PR #41, opened 2019-07-12
rewrite Data and add atom_pref label

njzjz:
- rewrite `Data` and `DataSystem`
- add `atom_pref` label


source/setup.py
```diff
@@ -34,7 +34,7 @@
     keywords='deepmd',
     install_requires=install_requires,        
     cmake_args=['-DTENSORFLOW_ROOT:STRING=%s' % tf_install_dir, 
-                '-DTF_GOOGLE_BIN:BOOL=FALSE', 
+                '-DTF_GOOGLE_BIN:BOOL=TRUE',
```
njzjz: revert to `FALSE`
```suggestion
                '-DTF_GOOGLE_BIN:BOOL=FALSE', 
```

PR #56, opened 2019-07-30
use setuptools_scm instead of setuptools-git-version

njzjz:
Then, we can use `python setup.py sdist` to package the source, and use `twine upload dist/*` to upload it to [pypi.org](https://pypi.org/simple). Users can use `pip install deepmd-kit` to install the python interface directly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,12 @@
 language: python
 python: "3.6"
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
+    - gcc-4.8
+    - g++-4.8
 env:
   matrix:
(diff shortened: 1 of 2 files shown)


njzjz:
why test fail...

njzjz:
为啥分了一台gcc=5.4的机器……

amcadmus:
Why does the test fail on a machine with gcc==5.4? Could you paste the error log?

amcadmus:
2019-07-30-205606_1854x590_scrot
I see it....

amcadmus:
The official tensorflow distribution was built with gcc==4.8 and the deepmd-kit should be built with the same gcc...

njzjz:
see tensorflow/tensorflow#27067

PR #68, opened 2019-09-15
add installation instructions for conda packages

njzjz:


njzjz:
BTW, users in China can use `https://mirrors.tuna.tsinghua.edu.cn/anaconda/cloud/conda-forge` instead of `conda-forge`.

PR #73, opened 2019-09-19
R0.13

denghuilu:
Add multi-GPU support unit

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,6 +513,100 @@ The option **`T`** specifies the temperature of the simulation, and the option *
 
 The **`atom_type`** set the type for the atoms in the system. The names of the atoms are those provided in the `conf_file` file. The **`atom_mass`** set the mass for the atoms. Again, the name of the atoms are those provided in the `conf_file`.
 
+# Multi-GPU support
+Recently, we updated code for multi-GPU support. The main changes are focused on the interface part of lammps and tensorflow,
+while the lammps package was also minorly modified. Here, we'll give you a brief overview of this upgrade and provide a simple test. First, you need to install the GPU support environment for deepmd-kit.
+## Install GPU support environments for deepmd-kit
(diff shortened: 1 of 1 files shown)


denghuilu:
> Does this work for a lower version of cuda, say, cuda-9.2?

It mainly depends on your TensorFlow version. If you use a lower version of TensorFlow, like TensorFlow-1.12, I think CUDA-9* is fine. Check [this](https://www.tensorflow.org/install/source#tested_build_configurations) for a closer look.

README.md
```diff
@@ -513,6 +513,131 @@ The option **`T`** specifies the temperature of the simulation, and the option *
...
+### Get the processor's node rank
+When working on platforms across nodes, we need to consider how to get the device rank mentioned in the previous section. If you're using Open MPI, it comes with a node-rank API, but if you use an Intel impi, you may need to use another method to specify node-rank. At present, we think it is a good way to be compatible with multi-platforms by dividing the MPI communicator based on the processor name. The code can be viewed [here](source/lmp/pair_nnp.cpp), focusing mainly on the get_node_rank function.
+### Cmake conditional compilation
+We introduced the USE-CUDA-TOOLKIT parameter as a control variable for whether to compile using the CUDA environment. If you want to build DeePMD-kit with CUDA-toolkit support, then execute cmake
+```bash
+cmake -DTF_GOOGLE_BIN=true -DUSE_CUDA_TOOLKIT=true -DTENSORFLOW_ROOT=$tensorflow_root \
```
njzjz: So, we don’t need to specify the CUDA’s PATH? Does it just use TensorFlow’s settings?

denghuilu: Cmake has a function, named "find_package(CUDA REQUIRED)", which can detect the CUDA environment by default. If you load the CUDA module or set the CUDA environment in your system, when you set "-DUSE_CUDA_TOOLKIT=true" , CMake will set the CUDA environment by default.

amcadmus: user can specify the cuda tool-kit path by CUDA_TOOLKIT_ROOT_DIR, please check
https://cmake.org/cmake/help/latest/module/FindCUDA.html

PR #78, opened 2019-09-25
simplify installation of TensorFlow

njzjz:
When compile tensorflow_cc, the dependencies of tensorflow have been downloaded already. So we don't need to download them again. Also, it seems that `libprotobuf.a` and `libnsync.a` are useless. I think they may be already a part of tensorflow. We only need to move headers to `include` folder.

It works well, but **please check it again**.


denghuilu:
It works well in my workstation

PR #79, opened 2019-09-26
implement PolarLoss by a more general TensorLoss. 

amcadmus:
More:
1, infer model_type and loss_type from fitting type


amcadmus:
2, enable console entry point

PR #85, opened 2019-10-01
support Tf2.0

njzjz:


njzjz:
@amcadmus what’s wrong with dpdata?

PR #121, opened 2019-11-12
add an ASE calculator interface

njzjz:
Example:
```python
from ase import Atoms
from deepmd.calculator import DP
water = Atoms('H2O',
              positions=[(0.7601, 1.9270, 1),
                         (1.9575, 1, 1),
                         (1., 1., 1.)],
              cell=[100, 100, 100],
              calculator=DP(model="frozen_model.pb", type_map=['O', 'H']))
print(water.get_potential_energy())
print(water.get_forces())
```
Optimization is also available:
```python
from ase.optimize import BFGS
dyn = BFGS(water)
dyn.run(fmax=1e-6)
print(water.get_positions())
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@
 	    - [Include deepmd in the pair style](#include-deepmd-in-the-pair-style)
 	    - [Long-range interaction](#long-range-interaction)
 	- [Run path-integral MD with i-PI](#run-path-integral-md-with-i-pi)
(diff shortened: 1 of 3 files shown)


amcadmus:
Please add a subsection like "use deep potential with ASE" to the "use deepmd-kit" section of readme

amcadmus:
The type_map is stored in the frozen model, and can be accessed by DeepPot.get_type_map()

PR #129, opened 2019-11-26
fix relative bug

njzjz:


source/lib/src/NNPInter.cc
```diff
@@ -1690,3 +1690,22 @@ compute_std_f (vector<VALUETYPE> &		std,
...
+  unsigned nloc = std.size();
+  for (unsigned ii = 0; ii < nloc; ++ii){
+      const VALUETYPE * tmp_avg = &(avg[ii*3]);
+      VALUETYPE vdiff[3];
+	  vdiff[0] = tmp_avg[0];
+      vdiff[1] = tmp_avg[1];
+      vdiff[2] = tmp_avg[2];
+      VALUETYPE f_norm = MathUtilities::dot(vdiff, vdiff);
+	  // relative std = std/(abs(f)+eps)
+      std[ii] /= f_norm + eps;
```
amcadmus: f_norm should be sqrt(dot(vdiff, vdiff))

njzjz: fixed

PR #145, opened 2020-01-06
fix bug of "-l-ldeepmd" error when manually set USE_CUDA_TOOLKIT=false

denghuilu:
when USE_CUDA_TOOLKIT=false, @DEEPMD_OP_CUDA@ is equal to none, which could cause a link error when compiling lammps.  So simply replicate deepmd_op


njzjz:
Is it necessary to add `-ldeepmd_op_cuda` when compile LAMMPS? I found [I didn't add it](https://github.com/deepmd-kit-recipes/lammps-dp-feedstock/blob/4991f6f3300d88cdd1c2d9faa267f476ae3acd6b/recipe/build.sh#L9) when building conda packages...

PR #148, opened 2020-01-09
fix bug of "unsupported architecture compute_75" when use CUDA-9.0

denghuilu:


source/op/cuda/CMakeLists.txt
```diff
@@ -14,14 +14,77 @@ SET(CMAKE_CXX_STANDARD 11)
...
+                        -gencode arch=compute_37,code=sm_37;
+                        -gencode arch=compute_50,code=sm_50;
+                        -gencode arch=compute_52,code=sm_52; # Tesla M40, Tesla M40, Quadro M6000...
+                        -gencode arch=compute_53,code=sm_53; 
+                        -gencode arch=compute_60,code=sm_60; # Pascal – GP100/Tesla P100 – DGX-1 (Generic Pascal)
+                        -gencode arch=compute_61,code=sm_61; # Pascal - GTX 1080, GTX 1070, GTX 1060, GTX 1050, GTX 1030, Titan Xp, Tesla P40, Tesla P4, Discrete GPU on the NVIDIA Drive PX2
+                        -O3; -Xcompiler -fPIC;
+        )
+elseif (${CUDA_VERSION_MAJOR} STREQUAL "7")
+    message(STATUS "CUDA major version is " ${CUDA_VERSION_MAJOR})
```
amcadmus: It would be better to move this line out of the conditions, because it is the same for all branches.

denghuilu: 👌

PR #180, opened 2020-02-13
user specified activation function for embedding and fitting nets

GeiduanLiu:
change activation function in input.json

### DIFF
--- a/source/train/common.py
+++ b/source/train/common.py
@@ -147,11 +147,6 @@ def j_must_have_d (jdata, key, deprecated_key) :
 def j_have (jdata, key) :
     return key in jdata.keys() 
   
-def j_whether_in_dict(activation_fn):
-    if activation_fn not in activation_fn_dict:
-        raise RuntimeError(activation_fn+" is not a valid activation function")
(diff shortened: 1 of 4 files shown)


source/train/common.py
```diff
@@ -1,8 +1,9 @@
 import warnings
 import numpy as np
+from deepmd.env import tf
 
 data_requirement = {}
-
+activation_fn_dict = {'tf.nn.relu':tf.nn.relu,'tf.nn.relu6':tf.nn.relu6,"tf.nn.softplus":tf.nn.softplus,"tf.sigmoid":tf.sigmoid,"tf.tanh":tf.tanh,"tf.nn.tanh":tf.nn.tanh}
```
njzjz: I'd prefer a simple activation function name:
```suggestion
activation_fn_dict = {
    "relu": tf.nn.relu,
    "relu6": tf.nn.relu6,
    "softplus": tf.nn.softplus,
    "sigmoid": tf.sigmoid,
    "tanh": tf.nn.tanh
}
```

GeiduanLiu: That's a nice suggestion. I will change it.

PR #194, opened 2020-02-22
user specified precision for embedding and fitting nets

GeiduanLiu:
support only tf.float16, tf.float32 and tf.float64. 
tf.float16 may cause nan problem


njzjz:
It’s not necessary to specify precision for custom op?

GeiduanLiu:
> It’s not necessary to specify precision for custom op?

We can specify precision for custom op while compiling.

amcadmus:
> It’s not necessary to specify precision for custom op?

The precision of the custom op was determined at compile time. What Geiduan implemented is to change the floating point precision only in the NNs.

njzjz:
> > It’s not necessary to specify precision for custom op?
> 
> The precision of the custom op was determined at compile time. What Geiduan implemented is to change the floating point precision only in the NNs.

I see. Is it possible to compile both precision? I think tensorflow must have done it, so we do not need to recompile tensorflow for different precision.

PR #276, opened 2020-10-12
fix bug of std:: string incompatible with TensorFlow 2.3

denghuilu:


njzjz (approved):
CI passed after patching https://github.com/deepmodeling/deepmd-kit/pull/276.diff

PR #290, opened 2020-11-09
Update Data.py

bwang-ecnu:
When condition “if coeff_atom_ener==1:” is established and one would like to ignore the energy_file, it seems to need to replace "coeff_atom_ener=1" with "coeff_ener=0".


njzjz (approved):

amcadmus (changes requested):
Shouldn't it be coeff_ener = 1? 
at line 478 the system energy is computed by reducing the atomic energies. thus we have the system energy label.

amcadmus (approved):

njzjz:
This PR was merged into master...

PR #350, opened 2021-02-02
model compression

denghuilu:
model compression through tabulating


deepmd/utils/tabulate.py
```diff
@@ -0,0 +1,185 @@
+import re
+import math
+import numpy as np
+import tensorflow.compat.v1 as tf
```
njzjz, line 4:
```suggestion
from deepmd.env import tf
```

njzjz (changes requested):

njzjz:
> ModuleNotFoundError: No module named 'tqdm'
You need to add it to `setup.py`.

deepmd/descriptor/se_a.py
```diff
@@ -226,6 +230,37 @@ def compute_input_stats (self,
...
+        distance
+                The nearest nbor distance between atoms
+        max_nbor_size
+                The max nbor size of atoms
+        env_mat_range
+                The output data range of the environment matrix
+        model_file
+                The original frozen model, that will be compressed
+        table_info
+                The configuration of the tabulation
```
amcadmus, line 254: Please explain do the four numbers in the list mean

deepmd/descriptor/se_a.py
```diff
@@ -226,6 +230,37 @@ def compute_input_stats (self,
...
+        """
+        Reveive the statisitcs (distance, max_nbor_size and env_mat_range) of the training data.
+        
+        Parameters
+        ----------
+        distance
+                The nearest nbor distance between atoms
+        max_nbor_size
+                The max nbor size of atoms
+        env_mat_range
```
amcadmus, line 249: What is the type of this variable ? List? what do the elements mean?

deepmd/descriptor/se_a.py
```diff
@@ -226,6 +230,37 @@ def compute_input_stats (self,
...
+                           env_mat_range,
+                           model_file = 'frozon_model.pb',
+                           table_config = [5, 0.01, 0.1, -1]
+    ) -> None:
+        """
+        Reveive the statisitcs (distance, max_nbor_size and env_mat_range) of the training data.
+        
+        Parameters
+        ----------
+        distance
```
amcadmus, line 245: change `distance`  to `min_nbor_dist` ?  Isn't it a variable that is never referred?

source/op/descrpt.cc
```diff
@@ -602,7 +602,7 @@ class DescrptOp : public OpKernel {
 
 #define REGISTER_CPU(T)                                                                 \
 REGISTER_KERNEL_BUILDER(                                                                \
-    Name("Descrpt").Device(DEVICE_CPU).TypeConstraint<T>("T"),                       \
+    Name("Descrpt").Device(DEVICE_CPU).TypeConstraint<T>("T"),                          \
```
amcadmus, line 605: Do not revise it!

source/lib/include/CustomeOperation.h
```diff
@@ -169,7 +169,7 @@ void compute_descriptor_se_a_cpu (
 }
 
 template<typename FPTYPE>
-void DescrptSeACPULauncher(const FPTYPE * coord, const int * type, const int * ilist, const int * jrange, const int * jlist, const FPTYPE * avg, const FPTYPE * std, FPTYPE * descrpt, FPTYPE * descrpt_deriv, FPTYPE * rij, int * nlist, const int nloc, const int nall, const int nnei, const int ntypes, const int ndescrpt, const float rcut_r, const float rcut_r_smth, const std::vector<int> sec_a, const bool fill_nei_a, const int max_nbor_size) {
+void DescrptSeACPULauncher(const FPTYPE * coord, const int * type, const int * ilist, const int * jrange, const int * jlist, const FPTYPE * avg, const FPTYPE * std, FPTYPE * descrpt, FPTYPE * descrpt_deriv, FPTYPE * rij, int * nlist, const int nloc, const int nall, const int nnei, const int ntypes, const int ndescrpt, const float rcut_r, const float rcut_r_smth, const std::vector<int> sec_a, const bool fill_nei_a, const int magic_number) {
```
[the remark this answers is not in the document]
denghuilu: Because the variable denotes the max nor size of atoms. The new name will help people to understand the code.

source/op/descrpt.cc
denghuilu: ok

denghuilu:
Compress a model, which including tabulating the embedding-net. 
The table is composed of fifth-order polynomial coefficients and is assembled from two sub-tables. The first table takes the stride(parameter) as it's uniform stride, while the second table takes 10 * stride as it's uniform stride
The range of the first table is automatically detected by deepmd-kit in class DeepTabulate, while the second table ranges from the first table's upper boundary(upper) to the extrapolate(parameter) * upper.
Finally, I added a check frequency parameter. It indicates how often the program checks for overflow(If the input environment matrix overflow the first or second table range) during inference.

deepmd/utils/env_mat_stat.py
```diff
@@ -0,0 +1,100 @@
...
+class EnvMatStat():
+    """
+    Class for getting training data information. 
+    It loads data from DeepmdData object, and measures the data info, including neareest nbor distance between atoms, max nbor size of atoms and the output data range of the environment matrix.
+    """
+    def __init__(self,
+                 ntypes : int,
+                 rcut,
+                 rcut_smth,
+                 sel) -> None:
```
amcadmus, line 19: Why do you need rcut_smth and sel?

deepmd/utils/env_mat_stat.py
```diff
@@ -0,0 +1,100 @@
+import math
+import numpy as np
+from tqdm import tqdm
+from deepmd.env import tf
+from typing import Tuple, List
+from deepmd.env import op_module
+from deepmd.env import default_tf_session_config
+from deepmd.RunOptions import global_np_float_precision
+
+class EnvMatStat():
```
amcadmus, line 10: I would suggest a name like `NeighborStat`

deepmd/utils/env_mat_stat.py
```diff
@@ -0,0 +1,100 @@
...
+                                         self.place_holders['type'],
+                                         self.place_holders['natoms_vec'],
+                                         self.place_holders['box'],
+                                         self.place_holders['default_mesh'],
+                                         sel = self.sel,
+                                         rcut = self.rcut,
+                                         rcut_smth = self.rcut_smth)
+        self.sub_sess = tf.Session(graph = sub_graph, config=default_tf_session_config)
+
+    def get_env_mat_stat(self,
```
amcadmus, line 57: `get_stat` would be better

deepmd/utils/tabulate.py
```diff
@@ -0,0 +1,252 @@
...
+class DeepTabulate():
+    """
+    Class for tabulation.
+    Compress a model, which including tabulating the embedding-net. 
+    The table is composed of fifth-order polynomial coefficients and is assembled from two sub-tables. The first table takes the stride(parameter) as it\'s uniform stride, while the second table takes 10 * stride as it\s uniform stride 
+    The range of the first table is automatically detected by deepmd-kit, while the second table ranges from the first table\'s upper boundary(upper) to the extrapolate(parameter) * upper.
+    """
+    def __init__(self,
+                 model_file,
+                 data_type,
```
amcadmus, line 20: `precision` would be consistent with other part of the code when referring to the precision of floating point numbers.

deepmd/utils/tabulate.py
```diff
@@ -0,0 +1,252 @@
...
+        self.last_layer_size = self.matrix["layer_" + str(self.layer_size)][0].shape[1]
+        # define tables
+        self.data = {}
+
+        # TODO: Need a check function to determine if the current model is properly
+
+    def build(self, 
+              min_nbor_dist,
+              rcut,
+              rcut_smth,
```
amcadmus, line 73: `rcut` is recorded in the model file. `rcut_smth` is not but we should added it to the model file.

amcadmus (changes requested):

deepmd/utils/tabulate.py
denghuilu: The precision would be consistent with the se_a. If we do model compression and precision conversion at the same time, we need this parameter.

denghuilu: Leave the precision parameter in the descriptor(se_a), the precision of class DeepTabulate is consistent with the input model precision

amcadmus (approved):

amcadmus (approved):

PR #351, opened 2021-02-03
migrate from travis-ci to github actions

njzjz:
it seems that travis-ci will be charged


njzjz:
https://github.com/njzjz/deepmd-kit/actions/runs/533026925

PR #352, opened 2021-02-03
add dosctings, pathlib, typing, uppercase constants etc. to few base files: WIP

marian-code:
I am pushing this PR as a follow up on dicussion #321 .

Some of the base modules where modified and also setup.py script

The list of changes:

* some basic code cleanup, formatting and refactoring where suitable with the aim to be more pythonic
* added detailed type hints
* ported parts of the code that handle paths from strings to `pathlib.Path` as this is the prefered way to interact with the filesystem in python 3.4+  and is a lot less error prone
* constants are now typed in uppercase to be easily recognizable as is the recommended naming convention in python
* added docstrings to all public methods and classes and to some private ones too
* removed old python 2 type formatting using `%` in favor of f-strings - this is the recommended way in python 3.6+ and again is a lot more readable and less error prone
* removed a few unused imports
* cleanup in setup.py and port to pathlib
* added some TODOs with ideas for improvement
* removed m2r from `setup.py` because it makes no sense to convert to `rst` format when `long_description_content_type=text/markdown`

I hope this is OK and if it is to your liking I can push other changes that help with the rewrite.


marian-code:
There seems to be weird error in tests.

pip install with skbuild runs OK but tests throw error:

```bash
tensorflow.python.framework.errors_impl.NotFoundError: /opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/deepmd/libop_abi.so: undefined symbol: _ZNK10tensorflow8OpKernel11TraceStringERKNS_15OpKernelContextEb
```
as if though `libop` was not compiled right. I am not very proficient with `C/C++` so I am a bit uncertain how to debug this.
I am able to run these test locally without problems on python=3.8, gcc=6.3 and tf=2.4.1 without any errors.

.github/workflows/test_python.yml
```diff
@@ -59,5 +59,6 @@ jobs:
     - run: |
         sudo apt update
         sudo apt install gcc-${{ matrix.gcc }} g++-${{ matrix.gcc }}
+    - run: pip install tensorflow
```
amcadmus, line 62: I feel this line triggered the failure in unittests.
@njzjz do you have any hint?

njzjz: let me check...

njzjz: Yes. And for original script, I forgot to add environment variables...

setup.py
```diff
@@ -1,95 +1,108 @@
...
+    if cmake_version in SpecifierSet("<3.7"):
+        setup_requires.append("cmake")
 
+Path("deepmd").mkdir(exist_ok=True)
 
 setup(
     name="deepmd-kit",
-    setup_requires=setup_requires,
-    use_scm_version={'write_to': 'deepmd/_version.py'},
+    setup_requires=INSTALL_REQUIRES,
```
njzjz, line 71: maybe broken here...
```suggestion
    setup_requires=setup_requires,
```

njzjz:
Ok, it turns green.

.github/workflows/test_python.yml
marian-code: seems like I was incorrectly trying to solve the lack of environment variables by installation of TF from pip

marian-code:
Is it OK if I push other changes? Any suggestion what I should work on?

marian-code (commented):
That was a stupid mistake, sorry you had to hunt that down.

njzjz:
> Is it OK if I push other changes? Any suggestion what I should work on?

Yes

amcadmus (changes requested):

deepmd/common.py
```diff
@@ -1,218 +1,453 @@
...
+    ----------
+    root_dir : Union[str, Path]
+        starting directory
+
+    Returns
+    -------
+    List[str]
+        list of string pointing to system directories
+    """
+    matches = [d for d in Path(root_dir).rglob("*") if (d / "type.raw").is_file()]
```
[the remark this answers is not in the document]
marian-code: It is fixed as requested, although I think using absolute paths is a bit safer.

amcadmus (approved):

PR #361, opened 2021-02-09
Rewoked logging and further rewrites

marian-code:
# Changes

* move local.py and slurm.py files to deepmd package, they do not need to be installed by Cmake
* renamed multiple files to conform with python `camel_case` style naming
* more constants converted to uppercase as is python convention
* completely reworked logging system.
  * built on python logging, works with both serial and MPI version.
  * It has much broader configuration capability by command line arguments.
  * For MPI it has three different modes for logging
    * only on master rank==0 as it was till now
    *  collect  log messages from all worker processes and print them on master
    * log in each process to separate file.
  * It also supports all the log levels of python logging module which can be easily set by command line arguments
  * the ability to have debug messages in code and then easily silence them by setting higher log level together with better MPI logging should hopefully enable much easier debugging
  * all this is under new `deepmd.loggers` module
* rewritten entry point file and added logger options to all parsers
* rewritten RunOptions
  * separated string constants from the class
  * removed logging facilities from the class
  * precision constants are now loaded from `run_config.ini` configuration file and are not defined in python module
* moved precision compile time settings that where written to RunOptions to  configuration ini file so the Cmake alters only the `run_config.ini` not the `RunOptions.py.in` which was inconvenient to work with because of this. Now the code and the constants are separated and `run_options.py` module loads the precision settings on import


marian-code:
I will resolve this later today by new commit

amcadmus (approved):

amcadmus:
Could you please firstly merge `deepmodeling:api` then push to ` marian-code:api`? Thanks!

amcadmus:
Please check https://github.com/marian-code/deepmd-kit/pull/2 for the fixing of the conflict.

njzjz (approved):

amcadmus (approved):

PR #362, opened 2021-02-11
send model file through mpi bcast

denghuilu:
### changes

- bcast the modle files to all MPI processors.
- fix a bug, change the default memory size of temporary tensor(used in descrpt_se_a.cu)to nloc * GPU_MAX_NBOR_SIZE(4096) to avoid a potential GPU memory access error.
- fix a bug, rm print in class NeighborStat.

### DIFF
--- a/source/lmp/pair_nnp.cpp
+++ b/source/lmp/pair_nnp.cpp
@@ -105,6 +105,54 @@ int PairNNP::get_node_rank() {
     return looprank;
 }
 
+std::string PairNNP::get_file_content(const std::string & model) {
+  int myrank = 0, root = 0;
+  MPI_Comm_rank(MPI_COMM_WORLD, &myrank);
+  unsigned nchar = 0;
+  std::string file_content;
+  if (myrank == root) {
+    checkStatus (ReadFileToString(Env::Default(), model, &file_content));
+    nchar = file_content.size();
+  }
+  MPI_Bcast(&nchar, 1, MPI_UNSIGNED, root, MPI_COMM_WORLD);  
+  char * buff = (char *)malloc(sizeof(char) * nchar);  
+  if (myrank == root) {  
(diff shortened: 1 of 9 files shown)


source/lmp/pair_nnp.cpp
```diff
@@ -105,6 +105,54 @@ int PairNNP::get_node_rank() {
...
+    char * buff = (char *)malloc(sizeof(char) * nchar);  
+    if (myrank == root) {  
+      memcpy(buff, file_contents[ii].c_str(), sizeof(char) * nchar);
+    }
+    MPI_Bcast(buff, nchar, MPI_CHAR, root, MPI_COMM_WORLD);
+    file_contents[ii].resize(nchar);
+    for (unsigned jj = 0; jj < nchar; ++jj) {
+      file_contents[ii][jj] = buff[jj];
+    }
+    free(buff);
```
amcadmus, line 151: This section of code should reuse the same piece in `std::string PairNNP::get_file_content(const std::string & model)`

amcadmus (changes requested):

source/lmp/pair_nnp.cpp
```diff
@@ -105,6 +105,37 @@ int PairNNP::get_node_rank() {
...
+  if (myrank == root) {  
+    memcpy(buff, file_content.c_str(), sizeof(char) * nchar);
+  }
+  MPI_Bcast(buff, nchar, MPI_CHAR, root, MPI_COMM_WORLD);
+  file_content.resize(nchar);
+  for (unsigned ii = 0; ii < nchar; ++ii) {
+    file_content[ii] = buff[ii];
+  }
+  return file_content;
+  free(buff);
```
amcadmus, line 128: This line is never reached

amcadmus (changes requested):

amcadmus (approved):

amcadmus (approved):

PR #368, opened 2021-02-18
reworked entrypoints, ops loading , deep potentials and more rewrites

marian-code:
# changes

* finished rewriting all entrypoint scripts
* the entry point scripts are now organized in separate submodule 
* reworked argument passing from command line parser to entrypoint scripts. The old system  of passing whole object was not transparent. One could not easily see which attributes the object contained. Now the parser unpacks the arguments from object to dict and sends all parameters to scripts as keywords.
* added setup.cfg to configure lintners. It provides the ability to globally silence unwanted errors
* added new suggestions about formatting of strings and whitespace
* reworked deep potentials. Now all potential classes have unified API - all inherit from `DeepTensor`
* custom ops are now installed into separate module
* reworked loading of the custom ops now they are loaded auto-magically on ops module import which removes the need for numerous manual imports of each op separately
* deep potential variants can now be initialized by factory
* small bugfixes mostly caused by previous changes

# Problems
* there are a lot of changes in transform script but it is not tested by unitest so there may be some bugs. I do not have suitable data at hand to test it and thus I hoped somebody could provide them.
* Also there is new test to check if the `DeepPotential` factory returns the right model but again I lack the appropriate models to test it. Because of this the test will fail! I am hoping somebody has these models at hand so there is no need to spend time to generate new ones.


amcadmus:
The reason why the unittest failed is that the `op/__init__.py` is not installed. 

One may solve this problem by adding `"deepmd/op"` to `packages` of `setup.py`

amcadmus:
Regarding your problems, I will add tests for the `infer` module asap.

amcadmus:
Regarding the unittests for infer classes, please see PR #373

amcadmus:
Regarding the unittests for transform, please see PR #374

marian-code:
Than you a lot, @amcadmus for help. I will try to resolve the failing tests.

marian-code:
* fixed failing transform test
* addressed incorrect handling of `None` for `set_loggers` `log_path` argument in `main.py`. This closes #375 
* fixed `op` module not found
*  finished implementation of infer classes factory test
All tests are passing now

deepmd/__init__.py
```diff
@@ -1,19 +1,12 @@
...
-from .infer.deep_polar import DeepPolar
-from .infer.deep_polar import DeepGlobalPolar
-from .infer.deep_wfc import DeepWFC
-from .infer.data_modifier import DipoleChargeModifier
+
+from . import cluster, descriptor, fit, loss, utils
 from .env import set_mkl
+from .infer import DeepPotential
+from .infer.data_modifier import DipoleChargeModifier
+from .infer.deep_eval import DeepEval
```
amcadmus, line 9: I am not sure if it is a good idea to provide DeepEval, because it is not supposed to be called by the users.

deepmd/infer/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+    RuntimeError
+        if model file does not correspond to any implementd potential
+    """
+    mf = Path(model_file)
+
+    model_type = DeepEval(
+        mf, load_prefix=load_prefix, default_tf_graph=default_tf_graph
+    ).model_type
+
+    print(mf, model_type)
```
amcadmus, line 59: log this line for debug?

amcadmus:
The command line training of this PR fails. There are two know issues.
- the `deepmd/entrypoints` is not installed
- the default of the command line option `--restart` is set to boolean, which causes a failure when using it as a `str`. 

To reproduce the second issue, 
```
cd /path/to/deepmd-kit/examples/water/train
dp train water_se_a.json
```
The error message is 
```
  File "/xxx/venvs/py3.6-tf1.8/bin/dp", line 11, in <module>
    load_entry_point('deepmd-kit==1.2.3.dev316+gece66cf.d20210224', 'console_scripts', 'dp')()
  File "/xxx/venvs/py3.6-tf1.8/lib/python3.6/site-packages/deepmd_kit-1.2.3.dev316+gece66cf.d20210224-py3.6-linux-x86_64.egg/deepmd/main.py", line 283, in main
    train(**dict_args)
  File "/xxx/venvs/py3.6-tf1.8/lib/python3.6/site-packages/deepmd_kit-1.2.3.dev316+gece66cf.d20210224-py3.6-linux-x86_64.egg/deepmd/entrypoints/train.py", line 185, in train
    try_distrib=jdata.get("with_distrib", False),
  File "/xxx/venvs/py3.6-tf1.8/lib/python3.6/site-packages/deepmd_kit-1.2.3.dev316+gece66cf.d20210224-py3.6-linux-x86_64.egg/deepmd/run_options.py", line 297, in __init__
    self.restart = os.path.abspath(restart)
  File "/xxx/venvs/py3.6-tf1.8/lib64/python3.6/posixpath.py", line 371, in abspath
    path = os.fspath(path)
TypeError: expected str, bytes or os.PathLike object, not bool
```

We strongly recommend you to test the commits by the examples before PR. Thank you.

deepmd/__init__.py
marian-code: Yes you are right, that was left there by accident

amcadmus:
> I will have a look at the problems. I only ran the unittests, I did not think of using examples as tests. Maybe it would be a good idea to make tests that use the data from examples? So one does not have to do this manually.

Unfortunately the entrypoints are not covered by unittest. Would it be a better idea to provide unittests for interpreting the command line options?

deepmd/infer/__init__.py
marian-code: I think this is not necessary, I used it just for debugging and forgotten to delete it.

marian-code:
I have fixed all the bugs that were discovered.

> Unfortunately the entrypoints are not covered by unittest. Would it be a better idea to provide unittests for interpreting the command line options?

I also added unit tests for argument parser which is good idea but I think we still need to test the "interaction" with entrypoint scripts. Because when some underlying code changes like the aforementioned `RunOptions` and one forgets to change the parser accordingly the current tests I have written will not catch the problem. I can have a look at it in some future commit.

amcadmus:
Please check this bug fixing:
https://github.com/marian-code/deepmd-kit/pull/3

amcadmus:
> I have fixed all the bugs that were discovered.
> 
> > Unfortunately the entrypoints are not covered by unittest. Would it be a better idea to provide unittests for interpreting the command line options?
> 
> I also added unit tests for argument parser which is good idea but I think we still need to test the "interaction" with entrypoint scripts. Because when some underlying code changes like the aforementioned `RunOptions` and one forgets to change the parser accordingly the current tests I have written will not catch the problem. I can have a look at it in some future commit.

Yes, that's necessary, thanks for consideration.

marian-code:
> > I have fixed all the bugs that were discovered.
> > > Unfortunately the entrypoints are not covered by unittest. Would it be a better idea to provide unittests for interpreting the command line options?
> > 
> > 
> > I also added unit tests for argument parser which is good idea but I think we still need to test the "interaction" with entrypoint scripts. Because when some underlying code changes like the aforementioned `RunOptions` and one forgets to change the parser accordingly the current tests I have written will not catch the problem. I can have a look at it in some future commit.
> 
> Yes, that's necessary, thanks for consideration.

That is probably better left for some other PR since this one is rather big already and adding more code will certainly not help with debugging.

amcadmus (approved):

marian-code:
Thanks for having patience with this as there were a lot of bugs.

PR #369, opened 2021-02-18
convert dp1.2 model to dp1.3 model

denghuilu:


amcadmus (changes requested):
The file name `convert.py` should be changed to e.g. `convet_to_13.py`

njzjz:
It is strange to push this commit to `r1.2`... Why not push to `r1.3` and call it `convert_from_12`?

amcadmus:
> It is strange to push this commit to `r1.2`... Why not push to `r1.3` and call it `convert_from_12`?

The 1.2 environment is needed to load the 1.2 model.

amcadmus (approved):

PR #379, opened 2021-02-26
optimize code structure of custom ops

denghuilu:
### changes

- set multi-device style(same as op prod_env_mat_xx) for prod_force_xx, prod_virial_xx, gelu and tabulate ops;
- optimize the code structure of cuda part;
- change the suffix gpu_nv to gpu_cuda;
- add custom_op.h in "api_cc/include" as the common header file for all custom ops;
- delete tabulate.cc and gelu.cc which were used in training specifically, now training and inference share the same op implementations.

### TODO

- add google unittest for multi_device ops' gpu implementations;
- use the gpu implementations of op ProdForceSeA and ProdVirialSeA for model training.

### problems

- should we change the file name, like, tabulate.cu to prod_tabulate.cu for the sake of name style unity?


source/lib/include/prod_virial.h
```diff
@@ -24,3 +24,28 @@ void prod_virial_r_cpu(
...
     const int nnei);
 
+#if GOOGLE_CUDA
+template<typename FPTYPE>
+void prod_virial_a_gpu_cuda(
+    FPTYPE * virial, 
+    FPTYPE * atom_virial, 
+    const FPTYPE * net_deriv, 
+    const FPTYPE * env_deriv, 
+    const FPTYPE * rij_deriv,
```
amcadmus, line 34: The meaning for `rij_deriv` is not clear. Here should be `rij`

source/lib/src/cuda/prod_env_mat.cu
```diff
@@ -0,0 +1,514 @@
...
+template<typename FPTYPE>
+__device__ inline FPTYPE dev_dot(
+    FPTYPE * arr1, 
+    FPTYPE * arr2) 
+{
+  return arr1[0] * arr2[0] + arr1[1] * arr2[1] + arr1[2] * arr2[2];
+}
+
+template<typename FPTYPE>
+__device__ inline void spline5_switch(
```
amcadmus, line 51: Place this function in `switcher.h`?

source/lib/src/cuda/tabulate.cu
```diff
@@ -14,466 +10,231 @@
...
+  }
 }
 
 template <typename FPTYPE>
-__forceinline__ 
-__device__ 
-FPTYPE dot(FPTYPE ll[4], FPTYPE rr[4]) {
-    return ll[0] * rr[0] + ll[1] * rr[1] + ll[2] * rr[2] + ll[3] * rr[3];
+__forceinline__ __device__ 
+FPTYPE dot(
```
amcadmus, line 44: Place all `dot` in `utilities.h` ?

source/op/prod_force_multi_device.cc
```diff
@@ -0,0 +1,190 @@
...
+    );
+    assert (nframes == force_shape.dim_size(0));
+    assert (nframes == net_deriv_tensor.shape().dim_size(0));
+    assert (nframes == in_deriv_tensor.shape().dim_size(0));
+    assert (nframes == nlist_tensor.shape().dim_size(0));
+    assert (nall * 3 == force_shape.dim_size(1));
+    assert (nloc * ndescrpt == net_deriv_tensor.shape().dim_size(1));
+    assert (nloc * ndescrpt * 3 == in_deriv_tensor.shape().dim_size(1));
+    assert (nloc * nnei == nlist_tensor.shape().dim_size(1));
+    assert (nnei * 4 == ndescrpt);
```
amcadmus, line 145: should be `nnei * 1 == ndescrpt`

source/api_cc/include/custom_op.h
```diff
@@ -0,0 +1,45 @@
...
+#ifdef HIGH_PREC
+typedef double VALUETYPE;
+typedef double ENERGYTYPE;
+#else 
+typedef float  VALUETYPE;
+typedef double ENERGYTYPE;
+#endif
+
+typedef double boxtensor_t ;
+typedef double compute_t;
```
amcadmus, line 27: It is better not adding the typedefs of `boxtensor_t` and `compute_t` in `custom_op.h`, because they are only used by old style op implementation.

amcadmus:
The unittests for the following implementations are missing
1. `gelu` both CPU and GPU
2. `tabulate` both CPU and GPU
3. `prod_env_mat_r` both CPU and GPU
4. `prod_force_a/r_gpu_cuda`
5. `prod_virial_a/r_gpu_cuda`

source/lib/src/cuda/prod_env_mat.cu
denghuilu: It's a device function, nvcc compiler is needed, so better to leave it here.

source/lib/src/cuda/tabulate.cu
denghuilu: It's a device function, nvcc compiler is needed, so better to leave it here.

denghuilu:
### update

- mv custom_op.h from api_cc/include to op
- rm HIGH_PREC as well as compute_t from custom_op.h
- import logging module into tabulate.py and neighbor_stat.py and baned tqdm module 
- resove conversions
- fix bug of libdeepmd_op_cuda.so's installation

### TODO
add unittests for custom ops

source/op/CMakeLists.txt
```diff
@@ -22,6 +22,9 @@ endif (BUILD_CPP_IF)
 if (BUILD_PY_IF)
   set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
   set(CMAKE_INSTALL_RPATH $ORIGIN)
+  set(LIB_BASE_DIR ${CMAKE_SOURCE_DIR})
+  include_directories(${LIB_BASE_DIR}/api_cc/include)
+  MESSAGE( STATUS ${LIB_BASE_DIR})
```
amcadmus, line 27: We should not need to include the headers in `api_cc`

source/lib/src/cuda/prod_env_mat.cu
amcadmus: Conditional compiling solve the problem

PR #383, opened 2021-03-01
test building gpu version C++ API

njzjz:


njzjz (commented):
needs to fix bugs in #382 and rebase

PR #396, opened 2021-03-03
fix bug in log levels settings parser

### LINKED ISSUE #381: bug in log level
I want to discuss the determination of the `log_level` by `argparse`, the code is in the following
https://github.com/deepmodeling/deepmd-kit/blob/18ba5b19d8d2d237e600c6478ba8c8ba59e48d7b/source/train/main.py#L42-L50

Surprisingly, the action of `"count"` is to ADD the number of flags to the default value of `log_level`. Then if one sets `-vvv`, he/she will have a `log_level` of `5` rather than `3`. To reproduce this behavoior, 
```
import argparse
parser = argparse.ArgumentParser()
parser.add_argument("-v", default=2, action="count", dest="level")
args = parser.parse_args(['-vvv'])
print(args.level)
```

marian-code:
Fixes #381

Also changed in github actions `ubuntu-latest` to `ubuntu-18.04` because github has updated the VMs and `latest==20.04` now. The `20.04` version does not have older gcc version like `4.8` for instance


marian-code:
Don't merge this yet please, I have found another small bug. Will fix it tomorrow.

marian-code:
Done

PR #418, opened 2021-03-16
Add support for model version

amcadmus:
- Model version in format `MODEL_VERSION_MAJOR.MODEL_VERSION_MINOR`
- Implemented in both c++ and python interfaced
- Disentangle `DeepPot` from `DeepTensor`. `DeepTensor` may become a base class of `DeepPot` when the `model` is better designed.
- Use `model_attr/output_dim` to determine the output dimension when inferring `DeepTensor`


amcadmus:
> * Disentangle `DeepPot` from `DeepTensor`. `DeepTensor` may become a base class of `DeepPot` when the `model` is better designed.

@marian-code please not the above change

marian-code:
> > * Disentangle `DeepPot` from `DeepTensor`. `DeepTensor` may become a base class of `DeepPot` when the `model` is better designed.
> 
> @marian-code please note the above change

Thanks for the heads up! I have been preoccupied with other things so I haven't been able to contribute anything for the last weeks.

PR #429, opened 2021-03-22
fix bugs of ase calculator and add tests

njzjz:


njzjz:
@marian-code Could you check the ase interface? There may be a bug there

marian-code:
Sorry, got to this only today, I see it is already fixed.

PR #495, opened 2021-04-09
API modification: support combined dipole training mode and global dipole test mode.

YWolfeee:
dp train / freeze / test check has finished, but tensorboard used in loss/tensor.py has not been tested yet.

Author: Haotian Ye (YWolfeee)
Date: 9, Apr, 2021.

### DIFF
--- a/deepmd/entrypoints/main.py
+++ b/deepmd/entrypoints/main.py
@@ -232,10 +232,10 @@ def parse_args(args: Optional[List[str]] = None):
     )
     parser_tst.add_argument(
         "-a",
-        "--atomic-energy",
+        "--atomic",
         action="store_true",
         default=False,
-        help="Test the accuracy of atomic energy",
+        help="Test the accuracy of atomic label, i.e. energy / dipole / polar",
     )
 
     # * compress model *****************************************************************
--- a/deepmd/entrypoints/test.py
+++ b/deepmd/entrypoints/test.py
@@ -580,6 +580,7 @@ def test_dipole(
 
     # do summation in atom dimension
     if has_atom_dipole == False:
+        dipole = np.reshape(dipole,(dipole.shape[0], -1, 3))
         atoms = dipole.shape[1]
         dipole = np.sum(dipole,axis=1)
 
--- a/deepmd/entrypoints/test.py
+++ b/deepmd/entrypoints/test.py
@@ -26,7 +26,7 @@ def test(
     rand_seed: Optional[int],
     shuffle_test: bool,
     detail_file: str,
-    atomic_energy: bool,
+    atomic: bool,
     **kwargs,
 ):
     """Test model predictions.
@@ -47,7 +47,7 @@ def test(
         whether to shuffle tests
     detail_file : Optional[str]
         file where test details will be output
-    atomic_energy : bool
+    atomic : bool
         whether per atom quantities should be computed
 
     Raises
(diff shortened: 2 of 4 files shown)


deepmd/entrypoints/main.py
```diff
@@ -232,10 +232,10 @@ def parse_args(args: Optional[List[str]] = None):
     )
     parser_tst.add_argument(
         "-a",
-        "--atomic-energy",
+        "--atomic",
```
njzjz, line 235: Please fix unit tests if you import a breaking change.

njzjz (changes requested):

deepmd/loss/tensor.py
```diff
@@ -12,24 +12,60 @@ class TensorLoss () :
...
             self.scale = 1.0
+
+        # YHT: added for global / local dipole combination
+        if self.atomic is True: # upper regulation, will control the lower behavior
+            self.local_weight,self.global_weight = 1.0,0.0
+        elif self.atomic is False: # upper regulation, will control the lower behavior
+            self.local_weight,self.global_weight = 0.0,1.0
+        else: # self.atomic is None, let the loss parameter decide which mode to use
+            if jdata is not None:
+                self.local_weight = jdata.get('atomic_' + self.tensor_name,None)
```
amcadmus, line 34: It seems that the name of the prefactors are `atomic_dipole` and `dipole`.   

I would suggest to change them to `pref_atomic_dipole` and `pref_dipole`.

deepmd/loss/tensor.py
```diff
@@ -12,24 +12,60 @@ class TensorLoss () :
...
+                    assert self.local_weight > 0.0, AssertionError('assign a zero weight to local dipole without setting a global weight')
+                    self.local_weight,self.global_weight = 1.0,0.0
+                    self.atomic = True
+                else:   # Both are not None
+                    self.atomic = True if self.local_weight != 0.0 else False
+                    assert (self.local_weight >0.0) or (self.global_weight>0.0), AssertionError('can not assian zero weight to both local and global mode')
+                    # normalize
+                    temp_sum = self.local_weight + self.global_weight
+                    self.local_weight   /=  temp_sum
+                    self.global_weight  /=  temp_sum
```
amcadmus, line 57: I suggest not doing this normalization. Leave the prefactors the same as those provided by the user.

deepmd/entrypoints/test.py
```diff
@@ -563,19 +564,31 @@ def test_dipole(
...
+    if has_atom_dipole == False:
+        dipole = np.reshape(dipole,(dipole.shape[0], -1, 3))
+        atoms = dipole.shape[1]
+        dipole = np.sum(dipole,axis=1)
+
     l2f = l2err(dipole - test_data["dipole"][:numb_test])
 
+    if has_atom_dipole == False:
+        l2f = l2f / atoms
+
```
amcadmus, line 591: The design here is not good.

We assume that `has_atom_dipole == False`, then by line 576, the label is ensured to be global. The atoms on line 584 is always 1.

When `has_atom_dipole == False`, the user wants to test the error in global dipole. The RMSE should be normalized by number of atoms. This information can be get from `test_data`.

amcadmus (changes requested):

deepmd/loss/tensor.py
YWolfeee: I will fix this

deepmd/loss/tensor.py
YWolfeee: I will fix this.

deepmd/entrypoints/test.py
YWolfeee: I don't think it's true. Suppose has_atom_dipole is false.
The frozen model will output local dipole. Therefore, it should be shaped as (nframs x sel_atom x 3).
However, the label is (nframes x 3) in this case.
Therefore, it's necessary to store natoms and do normalization

amcadmus: That's clear to me

YWolfeee:
Finish the above problem, including ut test and some advise from Han Wang. Tensorboard hasn't tested yet.

Author: Haotian Ye (YWolfeee)
Date: 12, Apr, 2021.

amcadmus (approved):

PR #517, opened 2021-04-15
Separate training and validation dataset 

ZiyaoLi:
Major:
- ideally-functioned validation supported.
- `sys_prob` evaluation style changed in `DeepmdDataSystem`.
- `argcheck` altered to support validation data.

Minor:
- `dargs` version corrected.
- `trainer` code structures optimized.
- some spelling errors corrected.
- varable names optimized.


setup.py
```diff
@@ -101,7 +101,7 @@
         f"-DTENSORFLOW_ROOT:STRING={tf_install_dir}",
         "-DBUILD_PY_IF:BOOL=TRUE",
         "-DBUILD_CPP_IF:BOOL=FALSE",
-        "-DFLOAT_PREC:STRING=high",
+        "-DFLOAT_PREC:STRING=low",
```
amcadmus, line 104: Could you please revert the default precision to "high"?

deepmd/train/trainer.py
```diff
@@ -443,15 +447,18 @@ def _init_sess_distrib(self):
         # ,
         # save_checkpoint_steps = self.save_freq)
 
-    def train (self, 
-               data) :
+    def train (self, train_data, valid_data=None) :
+
+        if valid_data is None:  # no validation set specified.
+            valid_data = train_data  # using training set as validation set.
```
amcadmus, line 453: The setting `valid_data = train_data` may not be a good option. It gives the user the wrong impression that their model is validated and the validate error looks almost the same as the training error. 

If `valid_data` is `None`, I would suggest not printing the column of valid error in `lcurve.out`.

deepmd/train/trainer.py
```diff
@@ -539,56 +532,81 @@ def train (self,
...
+            if 'find_' in kk:
+                feed_dict[self.place_holders[kk]] = batch[kk]
+            else:
+                feed_dict[self.place_holders[kk]] = np.reshape(batch[kk], [-1])
+        for ii in ['type']:
+            feed_dict[self.place_holders[ii]] = np.reshape(batch[ii], [-1])
+        for ii in ['natoms_vec', 'default_mesh']:
+            feed_dict[self.place_holders[ii]] = batch[ii]
+        feed_dict[self.place_holders['is_training']] = is_training
+        return feed_dict, batch['natoms_vec']
```
amcadmus, line 550: I do not understand why `batch['natoms_vec']` is returned. we can get the piece of information from `feed_dict['atoms_vec']`

deepmd/train/trainer.py
```diff
@@ -539,56 +532,81 @@ def train (self,
...
+        sum_results = None
+        for i in range(numb_batch):
+            feed_dict, natoms = self.get_feed_dict(data, is_training=False)
+            results = self.loss.eval(self.sess, feed_dict, natoms)
+            if sum_results is None:
+                sum_results = results
+            else:
+                for k, v in results:
+                    sum_results[k] += v
+        avg_results = {k: v / numb_batch for k, v in sum_results.items()}
```
amcadmus, line 611: Here the error should be calculate in the sense of root mean square, rather than simply mean.

PR #518, opened 2021-04-16
seperate impl of averging errors across systems

amcadmus:


ZiyaoLi (commented):
Would it be better to switch 'l2' to 'rmse' for unity?

amcadmus:
> Would it be better to switch 'l2' to 'rmse' for unity?

Sure we should use rmse

PR #524, opened 2021-04-17
Printing the training error with exactly the batch used to train 

amcadmus:
- printing the training error with exactly the batch used to train 
- remove the size of test data in `print_summary`
- fix bug of averaging over batches.


ZiyaoLi (approved):
all clear for me!

PR #535, opened 2021-04-19
Do not find protobuf for python

njzjz:
In conda-forge's recent released tensorflow, protobuf is external
from tensorflow and not inside the tensorflow directory.

### DIFF
--- a/source/cmake/Findtensorflow.cmake
+++ b/source/cmake/Findtensorflow.cmake
@@ -32,6 +32,7 @@ find_path(TensorFlow_INCLUDE_DIRS
   PATH_SUFFIXES "/include"
   NO_DEFAULT_PATH
   )
+if (BUILD_CPP_IF)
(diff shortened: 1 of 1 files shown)


amcadmus:
Does this change compatible with older versions of TF?

njzjz:
> Does this change compatible with older versions of TF?

This has nothing to do with TF versions. See https://github.com/conda-forge/tensorflow-feedstock/blob/40db5f43756877bea9e0dd2685cd5683b6527904/recipe/build.sh#L30, and here protobuf will be an external library and will not packaged into the TF package. (It's used to reduce the build time)

njzjz:
In fact, I think we don't need libprotobuf to build python package, is it correct?

amcadmus:
> 
> 
> In fact, I think we don't need libprotobuf headers to build python package, is it correct?

You are right

PR #552, opened 2021-04-22
add type embed

jxxiaoshaoye:


deepmd/fit/ener.py
```diff
@@ -26,6 +26,7 @@ def __init__ (self,
                   seed : int = 1,
                   atom_ener : List[float] = [],
                   activation_function : str = 'tanh',
+                  share_fitting: bool = False,
```
njzjz, line 29: Please edit the document below if you bring a new parameter.

PR #593, opened 2021-05-07
Merge Tensor Loss Training and Polar Normalization

YWolfeee:
This pull request mainly update the support of tensor training (dipole and polar);
(1) Combine global_polar and polar mode.
(2) Unify the process of this two mode.
(3) Support training with global label system and local label system simultaneously.
(4) Add a initial normalization to polar training model, which can speed up the training and decrease loss.
(5) change the input of loss parameters (examples and document will be upload later).

**Author: YWolfeee**
**Date: Apr, 7, 2021**


deepmd/loss/tensor.py
```diff
@@ -18,54 +18,31 @@ def __init__ (self, jdata, **kwarg) :
...
-                    #temp_sum = self.local_weight + self.global_weight
-                    #self.local_weight   /=  temp_sum
-                    #self.global_weight  /=  temp_sum
-                    
-            else: # Nothing been set, use default setting
-                self.local_weight,self.global_weight = 1.0,0.0
-                self.atomic = True
+        assert jdata is not None, "Please provide loss parameters!"
+        self.local_weight = jdata.get('pref_atomic_weight', None)
+        self.global_weight = jdata.get('pref_weight', None)
```
amcadmus, line 29: `pref` standing for prefactor and weight have similar meaning. How about using `pref` and `pref_atomic`?

deepmd/utils/data.py
```diff
@@ -461,11 +461,16 @@ def _load_data(self, set_name, key, nframes, ndof_, atomic = False, must = True,
...
-            data = np.reshape(data, [nframes, ndof])
+            try:    # YWolfeee: deal with data shape error
+                if atomic :
+                    data = data.reshape([nframes, natoms, -1])
+                    data = data[:,idx_map,:]
+                    data = data.reshape([nframes, -1])
+                data = np.reshape(data, [nframes, ndof])
+            except ValueError as err_message:
+                print(str(err_message))
+                print("This error may occur when your label mismatch it's name, i.e. you might store global tensor in `atomic_tensor.npy` or atomic tensor in `tensor.npy`.")
```
amcadmus, line 472: It is better to use `log.error` instead print.

deepmd/utils/data.py
```diff
@@ -461,11 +461,16 @@ def _load_data(self, set_name, key, nframes, ndof_, atomic = False, must = True,
...
+                exit()
```
amcadmus, line 473: It is better to raise exception, which may be catched by others, rather than die.

amcadmus (changes requested):
we also need to support `atomic_dipole` and `atomic_polarizability` in the `raw_to_set.sh`

PR #594, opened 2021-05-07
Add model deviation of virial

amcadmus:
- add model deviation and relative model deviation of virial
- refact the computation of std and relative std
- UT covering the case of relative std


njzjz (approved):

njzjz (approved):
LGTM

Issue #563, opened 2021-04-24
[BUG] Lammps compilation fails when using low-precision mode USER-DEEPMD


BaozCWJ:
**Summary**

The low-precision mode deepmd module cannot be compiled by LAMMPS

**Deepmd-kit version, installation way, input file, running commands, error log, etc.**

vserion: 2.0.0 devel branch
installation way: C++ interface
key commands: 
`cmake -DUSE_CUDA_TOOLKIT=true -DTENSORFLOW_ROOT=$tensorflow_root -DCMAKE_INSTALL_PREFIX=$deepmd_root -DFLOAT_PREC=low ..
`
After the compilation of low-precision deepmd,  there's an error when compile lammps.

**Steps to Reproduce**

Following the step after the above cmake command from [https://deepmd.readthedocs.io/en/master/install.html#install-the-c-interface](url), then the error will occur at the last step of **Install LAMMPS’s DeePMD-kit module**

**Further Information, Files, and Links**
<img width="970" alt="Screen Shot 2021-04-24 at 14 51 29" src="https://user-images.githubusercontent.com/37867949/115950537-66ef0e00-a50e-11eb-80b7-05a423744a9b.png">

njzjz:
So why did I build conda packages successfully? 😂
Build log: https://dev.azure.com/deepmd-kit-recipes/dfe7535b-bc98-4288-9a49-be8914966e7a/_apis/build/builds/401/logs/63

amcadmus:
If you change the precision of deepmd-kit package, the lammps module should be replace by doing the following in the lammps src dir
```
make no-user-deepmd
rm -fr USER-DEEPMD
cp -r /path/to/new/USER-DEEPMD .
make yes-user-deepmd
```

BaozCWJ:
> If you change the precision of deepmd-kit package, the lammps module should be replace by doing the following in the lammps src dir
> 
> ```
> make no-user-deepmd
> rm -fr USER-DEEPMD
> cp -r /path/to/new/USER-DEEPMD .
> make yes-user-deepmd
> ```

I tried these steps and  `make clean-all`, but it makes no difference.

amcadmus:
Same as @njzjz I cannot reproduce this issue.

njzjz:
@BaozCWJ you may post `Makefile.package` to let us see the compiling arguments.

BaozCWJ:
> @BaozCWJ you may post `Makefile.package` to let us see the compiling arguments.

[...]

PR #600, opened 2021-05-08
test lammps builds

njzjz:
in CI


njzjz:
So anyway, we cannot reproduce #563...

amcadmus (approved):

Issue #589, opened 2021-05-04
[BUG] source/train/Local.py support GPU id


matzmz:
It would be nice if we could get GPU by UUIDs or GPUs by indices.

if you use an UUID (e.g. export CUDA_VISIBLE_DEVICES="GPU-08041152-9141-4970-a445-5e8b634025b5" the file  'source/train/Local.py ' will exit with an error because the code try to cast the uuid to an integer value.

Could you please support also GPU device UUID ? Newest HPC schedulers use gpu uids instead of integer indexes.

denghuilu:
Now, the GPU UUID should work well with DeePMD-kit.

PR #620, opened 2021-05-12
add GPU UUID support for DP

denghuilu:


denghuilu:
#589 This should fix issue589.

PR #628, opened 2021-05-14
support MPI and other atom_styles for LAMMPS atomic keyword

njzjz:
fix problems left in #44

### DIFF
--- a/source/lmp/pair_deepmd.cpp
+++ b/source/lmp/pair_deepmd.cpp
@@ -606,33 +606,33 @@ void PairDeepMD::compute(int eflag, int vflag)
 	     << " " << setw(18) << all_f_avg;
 	     // << " " << setw(18) << avg_e
 	     // << " " << setw(18) << std_e_1 / all_nlocal
-  }
-  if (out_each == 1){
-    vector<double> std_f_all(all_nlocal);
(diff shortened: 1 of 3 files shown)


njzjz:
I tried to run in serial and in parallel, and the outputs are the same.

source/lmp/pair_deepmd.cpp
```diff
@@ -606,14 +606,33 @@ void PairDeepMD::compute(int eflag, int vflag)
...
+    if (rank == 0) {
+      for (int dd = 0; dd < all_nlocal; ++dd) {
+        std_f_all[tagrecv[dd]-1] = stdfrecv[dd];
+      }
+      for (int dd = 0; dd < all_nlocal; ++dd) {
+        fp << " " << setw(18) << std_f_all[dd];	
+      }
+    }
+  }
+  if (rank == 0) {
```
amcadmus, line 635: Could you plz indent this piece of code so it is compatible with the rest part?

amcadmus (changes requested):

source/lmp/pair_deepmd.cpp
njzjz: ok

amcadmus (approved):

PR #631, opened 2021-05-14
add support for converting models to 2.0 compatibility

amcadmus:


denghuilu:
These two interfaces work well on my local workstation. We need a documentation update

amcadmus:
> These two interfaces work well on my local workstation. We need a documentation update later.

I will take care

denghuilu (approved):

PR #640, opened 2021-05-18
append to out_file when LAMMPS restarts

### LINKED ISSUE #623: [Feature Request] Append to model_devi.out if restart
**Summary**

If LAMMPS restarts, `model_devi.out` should be appended to instead of creating a new file. 

**Detailed Description**

How to implement:
1. set `restartinfo` to 1 (default)
See: https://github.com/lammps/lammps/blob/9da49d9c6f521eed62ca0752297c9057d34d39db/src/pair.cpp#L61
2. add `read_restart` to `pair`, where open the file using `std::ofstream::app` mode
See:
https://github.com/lammps/lammps/blob/9da49d9c6f521eed62ca0752297c9057d34d39db/src/pair.cpp#L725-L729
http://www.cplusplus.com/reference/fstream/ofstream/open/
3. add a condition to skip opening a new file and adding headers if restart
See: https://github.com/deepmodeling/deepmd-kit/blob/f70437ffaca76b6047e9783433657ed93742825c/source/lmp/pair_deepmd.cpp#L838-L850

Potential problems: Since we don't save restart files every timestep, there may be some duplicated data before and after restart.

njzjz:
This ensures the out file will not be override when LAMMPS restarts.
This commit may be conflicted with #392. Commit
https://github.com/deepmodeling/deepmd-kit/commit/5597ea2b49f96e99a52a9779b04b6c12e5a79a04 should be dropped.

Resolves #623.

### DIFF
--- a/source/lmp/pair_deepmd.cpp
+++ b/source/lmp/pair_deepmd.cpp
@@ -237,6 +237,7 @@ PairDeepMD::PairDeepMD(LAMMPS *lmp)
   if (strcmp(update->unit_style,"metal") != 0) {
     error->all(FLERR,"Pair deepmd requires metal unit, please set it by \"units metal\"");
   }
+  restartinfo = 1;
   pppmflag = 1;
   respa_enable = 0;
   writedata = 0;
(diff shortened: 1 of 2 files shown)


amcadmus:
Could you patch this bug fixing to the master branch?

PR #641, opened 2021-05-18
append to out_file when LAMMPS restarts

njzjz:
This ensures the out file will not be override when LAMMPS restarts.
This commit may be conflicted with #392. Commit
@5597ea2b49f96e99a52a9779b04b6c12e5a79a04 should be dropped.


source/lmp/pair_nnp.cpp
```diff
@@ -800,6 +806,16 @@ void PairNNP::settings(int narg, char **arg)
   all_force.resize(numb_models);
 }
 
+void PairDeepMD::read_restart(FILE *)
```
njzjz, line 809: I didn't notice that the class name had changed.

PR #652, opened 2021-05-21
add an example of C++ inference to doc

njzjz:


tuoping (approved):
Should we list C++ interface in API.rst?

njzjz:
> Should we list C++ interface in API.rst?

There should be an automatic way to generate another document, say `API_cc.rst`.

amcadmus:
> 
> 
> > Should we list C++ interface in API.rst?
> 
> There should be an automatic way to generate another document, say `API_cc.rst`.

Yes via something like doxygen. But current the documentation of the C++ interface is lacking...

PR #656, opened 2021-05-21
Add ROCM support for DEEPMD

pkulzy:
1. modify CMakeLists.txt  located in the /source folder in order to be able to compile on the ROCM platform
2. Transfer all the .cu files in the /source/lib/src/cuda to .hip.cu in the  /source/lib/src/rocm
3. In *.cc *.c in /source folder, add modules that are appropriate for rocm
4. In UT files, add rocm modules


source/CMakeLists.txt
```diff
@@ -72,6 +74,29 @@ if (USE_CUDA_TOOLKIT)
...
+#define USE_ROCM_TOOLKIT
+if (DEFINED USE_ROCM_TOOLKIT)
+  if (USE_ROCM_TOOLKIT)
+    find_package(HIP REQUIRED)
+    add_compile_definitions(__HIP_PLATFORM_HCC__)
+  else()
+    message(STATUS "Will not build nv GPU support")
+  endif()
+else()
+  find_package(HIP QUIET)
```
njzjz, line 86: This is not available in the non-ROCM environment.

njzjz (changes requested):

source/lib/tests/CMakeLists.txt
```diff
@@ -40,6 +42,45 @@ if (USE_CUDA_TOOLKIT)
...
+  if(USE_ROCM_TOOLKIT)
+    link_directories(${HIP_ROOT_DIR}/lib)
+    add_definitions("-DTENSORFLOW_USE_ROCM")
+    find_package(HIP REQUIRED)
+    add_compile_definitions(__HIP_PLATFORM_HCC__)
+  else()
+    message(STATUS "Will not build rocm GPU support")
+  endif()
+else()
+  find_package(HIP REQUIRED)
```
njzjz, line 65: This is not available in the non-ROCM environment.

njzjz (changes requested):

njzjz (approved):

source/lib/CMakeLists.txt
```diff
@@ -2,6 +2,11 @@
 
 set (libname ${LIB_DEEPMD})
 
+# Devices that have both ROCM and CUDA are not currently supported
+if (USE_ROCM_TOOLKIT AND USE_CUDA_TOOLKIT)
+  message (FATAL_ERROR "Devices that have both ROCM and CUDA are not currently supported")
+endif()
+
```
amcadmus, line 9: We have already died in source/CMakeList.txt. Do we really need to die again here?

source/api_cc/src/DeepPot.cc
```diff
@@ -217,6 +230,20 @@ init (const std::string & model, const int & gpu_rank, const std::string & file_
...
+  hipGetDeviceCount(&gpu_num); // check current device environment
+  if (gpu_num > 0) {
+    options.config.set_allow_soft_placement(true);
+    options.config.mutable_gpu_options()->set_per_process_gpu_memory_fraction(0.9);
+    options.config.mutable_gpu_options()->set_allow_growth(true);
+    hipErrcheck(hipSetDevice(gpu_rank % gpu_num));
+    std::string str = "/gpu:";
+    str += std::to_string(gpu_rank % gpu_num);
+    graph::SetDefaultDevice(str, &graph_def);
+  }
```
amcadmus, line 244: Couldn't this part of code be merged with the cuda GPU config setting?

source/api_cc/src/DeepPot.cc
```diff
@@ -542,6 +573,16 @@ init (const std::vector<std::string> & models, const int & gpu_rank, const std::
...
   #endif // GOOGLE_CUDA
 
+
+  #if TENSORFLOW_USE_ROCM
+  if (gpu_num > 0) {
+      options.config.set_allow_soft_placement(true);
+      options.config.mutable_gpu_options()->set_per_process_gpu_memory_fraction(0.9);
+      options.config.mutable_gpu_options()->set_allow_growth(true);
+      hipErrcheck(hipSetDevice(gpu_rank % gpu_num));
+  }
```
amcadmus, line 583: Merged with cuda GPU setting?

source/api_cc/tests/CMakeLists.txt
```diff
@@ -76,6 +76,45 @@ if (USE_CUDA_TOOLKIT)
   add_subdirectory(${LIB_BASE_DIR}/src/cuda cuda_binary_dir)
 endif()
 
+# define USE_ROCM_TOOLKIT
+if(NOT DEFINED HIP_PATH)
+  if(NOT DEFINED ENV{HIP_PATH})
+    set(HIP_PATH "/opt/rocm/hip" CACHE PATH "Path to which HIP has been installed")
```
amcadmus, line 82: I do not understand why here we need hard coding.

source/api_cc/src/DeepPot.cc
pkulzy: this part of code, i use "hipErrcheck" function:240

source/api_cc/src/DeepPot.cc
pkulzy: if we want to merge this part of code with CUDA, maybe we should do sth to deal with hipErrcheck and cudaErrcheck

source/api_cc/tests/CMakeLists.txt
pkulzy: Because it's only when I give this absolute path that it can find the HIP package

source/lib/src/prod_env_mat.cc
```diff
@@ -320,3 +320,68 @@ void deepmd::env_mat_nbor_update(
...
 #endif // GOOGLE_CUDA
+
+#if TENSORFLOW_USE_ROCM
+void deepmd::env_mat_nbor_update(
+    InputNlist &inlist,
+    InputNlist &gpu_inlist,
+    int &max_nbor_size,
+    int* &nbor_list_dev,
+    const int * mesh, 
+    const int size)
```
amcadmus, line 331: Can this function merged with the cuda version?  I find it should be possible.

source/lib/src/rocm/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+# required cmake version
+cmake_minimum_required(VERSION 3.15)
+# project name
+project(deepmd_op_rocm)
+set(CMAKE_LINK_WHAT_YOU_USE TRUE)
+#set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "/opt/rocm/hip/cmake")
+if(NOT DEFINED HIP_PATH)
+  if(NOT DEFINED ENV{HIP_PATH})
+    set(HIP_PATH "/opt/rocm/hip" CACHE PATH "Path to which HIP has been installed")
```
amcadmus, line 9: Shouldn't this path be found by `find_package(HIP)` ?

source/lib/CMakeLists.txt
pkulzy: Because the TensorFlow version is one of rocm and cuda, so when faced with a choice between CUDA and ROCM, I threw out an error

source/api_cc/tests/CMakeLists.txt
pkulzy: image

pkulzy: If I comment out this code, it makes error  when I execute "make"

source/op/prod_env_mat_multi_device.cc
```diff
@@ -366,6 +440,44 @@ class ProdEnvMatAOp : public OpKernel {
...
+      array_int = int_temp.flat<int>().data(); 
+      array_longlong = uint64_temp.flat<unsigned long long>().data();
+
+      // launch the gpu(nv) compute function
+      deepmd::prod_env_mat_a_gpu_rocm(
+          em, em_deriv, rij, nlist, 
+          coord, type, gpu_inlist, array_int, array_longlong, max_nbor_size, avg, std, nloc, frame_nall, rcut_r, rcut_r_smth, sec_a);
+      if(b_nlist_map) _map_nlist_gpu_rocm(nlist, idx_mapping, nloc, nnei);
+      deepmd::delete_device_memory(firstneigh);
+      #endif //TENSORFLOW_USE_ROCM
```
amcadmus, line 480: Can this part merge with the cuda implementation? or introduce the condition only when the function name appears? or considering use template the unify the interface?

source/lib/src/prod_env_mat.cc
pkulzy: because merge this function with cuda, thus we will get the structure：
#if GOOGLE_CUDA
.......
#endif //GOOGLE_CUDA

#IF TENSORFLOW_USE_ROCM
........
#ENDIF//TENSORFLOW_USE_ROCM

#if TENSORFLOW_USE_ROCM || GOOGLE_USE_ROCM
this function
# endif //
I didn't use this structure when I migrated without wanting to change the code that I had written before

source/api_cc/src/DeepPot.cc
amcadmus: Shouldn't we unify hipErrCheck and cudaErrCheck?

source/op/prod_env_mat_multi_device.cc
pkulzy: introduce the condition only when the function name appears is OK,but I didn't adopt this when I transfer the code

source/api_cc/src/DeepPot.cc
pkulzy: yeah, maybe we should do this, but i don't know how to unify hipErrCheck and cudaErrCheck, they are two macros in /source/lib/include/gpu_cuda.h and /source/lib/include/gpu_rocm.h. Or your meaning is introducing the condition only when the function name appears

amcadmus: >  Or your meaning is introducing the condition only when the function name appears

Yes, because the TF config for both cuda and rocm are the same (they also share the same device name `"gpu"` in TF), so we do not want to maintain two separate pieces of codes.

amcadmus: > they are two macros in /source/lib/include/gpu_cuda.h and /source/lib/include/gpu_rocm.h

Are the definitions in the files the same as those in DeepPot.cc?

source/api_cc/tests/CMakeLists.txt
amcadmus: I mean the path to hip should be discovered by `find_package(HIP)`

pkulzy: Get, maybe I should find a macro in the FindHIP.cmake to replace absolutely path

source/api_cc/src/DeepPot.cc
pkulzy: how about three kinds of ErrCheck: CudaErrCheck  hipErrCheck  DeviceErrCheck, thus somewhere we need to merge the code with cuda, we use DeviceErrCheck

source/api_cc/src/DeepPot.cc
amcadmus: not really, use compile condition to switch between `hipErrCheck` and `cudaErrCheck`

```cpp
#if USE_ROCM_TOOLKIT || USE_CUDA_TOOLKIT
...
#ifdef USE_ROCM_TOOLKIT
  hipErrcheck()
#elif USE_CUDA_TOOLKIT
  cudaErrcheck()
#endif
...
#endif
```

source/api_cc/tests/CMakeLists.txt
amcadmus: We must do it. Otherwise we cannot support machines with hip installed to any other path.

source/op/prod_env_mat_multi_device.cc
amcadmus: Apparently you did not. I ask you to do so.

source/api_cc/src/DeepPot.cc
amcadmus: OK, since this PR is already very large, we leave it as is in this PR. We should consider using a unified interface for the error checks in the next PR.

njzjz:
Is it possible to add a test into GitHub Actions?

amcadmus:
> 
> 
> Is it possible to add a test into GitHub Actions?

same question, is it possible to run rocm code in a CPU environment?

pkulzy:
> > Is it possible to add a test into GitHub Actions?
> 
> same question, is it possible to run rocm code in a CPU environment?

I don't seem to understand your question, i think the coda for rocm is the same as for cuda.  When the CPU executes, it will run the CPU mode, not the ROCM mode.

pkulzy:
> > Is it possible to add a test into GitHub Actions?
> 
> same question, is it possible to run rocm code in a CPU environment?

I don't seem to understand your question, i think the code for rocm is the same as for cuda. When the CPU executes, it will run the CPU mode, not the ROCM mode.

source/api_cc/tests/CMakeLists.txt
pkulzy: image

pkulzy: I haven't found HIP-related macros or files on cmake 3.15，It is possible that cmake packages findHIP as a binary, thereby inheriting the older HIP

njzjz:
At least we can test compilation process

pkulzy:
> At least we can test compilation process
Do we have a similar test for CUDA code? If there is one, I think ROCM is also possible, but I am not sure about this

pkulzy:
> At least we can test compilation process
This is fine if it is only for the compilation process, but in this case we need to install the ROCM package

pkulzy:
> > > At least we can test compilation process
> > 
> > 
> > This is fine if it is only for the compilation process, but in this case we need to install the ROCM package
> 
> Is it hard to install the ROCM package in a virtual environment? For cudatoolkit, it's easy to install it using `apt-get`.
> https://github.com/deepmodeling/deepmd-kit/blob/5a8bc2c97f6c16cf783a2802ffa8da4ddd49c105/.github/workflows/build_cc.yml#L25
I tried installing ROCM under Ubuntu 18, but the resulting ROCM 4.2 installation lacked the necessary libraries . installation tutorial: https://rocmdocs.amd.com/en/latest/Installation_Guide/Installation-Guide.html#ubuntu

source/lib/src/rocm/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+SET(CMAKE_HIP_STANDARD 11)
+
+message(STATUS "HIP major version is " ${HIP_VERSION_MAJOR})
+
+set (HIP_HIPCC_FLAGS -hc; -fno-gpu-rdc; --amdgpu-target=gfx906; -fPIC; -O3; --std=c++11; -D__HIP_PLATFORM_HCC__)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -DCUB_IGNORE_DEPRECATED_CPP_DIALECT -DCUB_IGNORE_DEPRECATED_CPP_DIALECT")
+
+set (SOURCE_FILES
+    prod_env_mat.hip.cu prod_force.hip.cu prod_virial.hip.cu gelu.hip.cu tabulate.hip.cu coord.hip.cu neighbor_list.hip.cu prod_force_grad.hip.cu prod_virial_grad.hip.cu region.hip.cu
+)
```
amcadmus, line 24: why not using wildcard?

source/lib/src/rocm/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+# required cmake version
+cmake_minimum_required(VERSION 3.5)
+# project name
+project(deepmd_op_rocm)
+set(CMAKE_LINK_WHAT_YOU_USE TRUE)
+
+#find package ROCM
+list (APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/../../../cmake/)
```
amcadmus, line 8: `CMAKE_SOURCE_DIR` should be the top level directory of the CMake source tree. It seems that this command would not set the correct directory to the findROCM.

source/op/CMakeLists.txt
```diff
@@ -5,18 +5,33 @@ set(OP_LIB ${PROJECT_SOURCE_DIR}/lib/src/SimulationRegion.cpp ${PROJECT_SOURCE_D
...
 file(GLOB OP_SRC prod_force.cc prod_virial.cc descrpt.cc descrpt_se_a_ef.cc descrpt_se_a_ef.cc descrpt_se_a_ef_para.cc descrpt_se_a_ef_vert.cc pair_tab.cc prod_force_multi_device.cc prod_virial_multi_device.cc soft_min.cc soft_min_force.cc soft_min_virial.cc ewald_recp.cc gelu_multi_device.cc map_aparam.cc neighbor_stat.cc unaggregated_grad.cc tabulate_multi_device.cc prod_env_mat_multi_device.cc)
 file(GLOB OP_CUDA_SRC prod_force.cc prod_virial.cc descrpt.cc prod_env_mat_multi_device.cc pair_tab.cc prod_force_multi_device.cc prod_virial_multi_device.cc soft_min.cc soft_min_force.cc soft_min_virial.cc gelu_multi_device.cc tabulate_multi_device.cc)
+file(GLOB OP_ROCM_SRC prod_force.cc prod_virial.cc descrpt.cc prod_env_mat_multi_device.cc pair_tab.cc prod_force_multi_device.cc prod_virial_multi_device.cc soft_min.cc soft_min_force.cc soft_min_virial.cc gelu_multi_device.cc tabulate_multi_device.cc)
 file(GLOB OP_GRADS_SRC prod_force_grad.cc prod_force_grad_multi_device.cc prod_virial_grad.cc prod_virial_grad_multi_device.cc soft_min_force_grad.cc soft_min_virial_grad.cc )
 file(GLOB OP_PY *.py)
 
+list (APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/../cmake/)
```
amcadmus, line 12: See the same comment for `CMAKE_SOURCE_DIR`

source/CMakeLists.txt
```diff
@@ -51,26 +51,26 @@ file(READ ${PROJECT_SOURCE_DIR}/config/MODEL_VER MODEL_VERSION)
...
+  message (FATAL_ERROR "Devices that have both ROCM and CUDA are not currently supported")
 endif()
+
+# define USE_CUDA_TOOLKIT
 if (USE_CUDA_TOOLKIT)
+  find_package(CUDA REQUIRED)
   add_definitions("-D GOOGLE_CUDA")
-endif()
+  message(STATUS "Found CUDA in ${CUDA_TOOLKIT_ROOT_DIR}, build nv GPU support")
+endif(USE_CUDA_TOOLKIT)
```
amcadmus, line 64: else we should let the user know the user that we will not build with cuda support.

source/CMakeLists.txt
```diff
@@ -195,10 +195,14 @@ if (BUILD_CPP_IF)
...
-    set (LIB_DEEPMD_OP_CUDA		"deepmd_op")
-  endif()
+    set (LIB_DEEPMD_OP_DEVICE		"deepmd_op_cuda")
+  endif(USE_CUDA_TOOLKIT)
+  if (USE_ROCM_TOOLKIT)
+    set (LIB_DEEPMD_OP_DEVICE   "deepmd_op_rocm")
+  endif(USE_ROCM_TOOLKIT)
+  if ((NOT USE_ROCM_TOOLKIT) AND (NOT USE_CUDA_TOOLKIT)) 
+    set (LIB_DEEPMD_OP_DEVICE		"deepmd_op")
+  endif((NOT USE_ROCM_TOOLKIT) AND (NOT USE_CUDA_TOOLKIT))
```
amcadmus, line 205: use 
```cmake
if (AAA)
...
elif (BBB)
...
else
...
endif

```

source/CMakeLists.txt
```diff
@@ -51,26 +51,26 @@ file(READ ${PROJECT_SOURCE_DIR}/config/MODEL_VER MODEL_VERSION)
...
+  message(STATUS "Found CUDA in ${CUDA_TOOLKIT_ROOT_DIR}, build nv GPU support")
+endif(USE_CUDA_TOOLKIT)
+
+#define USE_ROCM_TOOLKIT
+if (USE_ROCM_TOOLKIT)
+  find_package(ROCM REQUIRED)
+  add_definitions("-D TENSORFLOW_USE_ROCM")
+  add_compile_definitions(__HIP_PLATFORM_HCC__)
+  message(STATUS "Found ROCM in ${ROCM_ROOT}, build AMD GPU support")
+endif (USE_ROCM_TOOLKIT)
```
amcadmus, line 72: else we should let the user know we will not build the source with rocm support

source/api_cc/tests/CMakeLists.txt
```diff
@@ -93,9 +92,14 @@ endif()
...
-else()
+endif(USE_CUDA_TOOLKIT)
+if(USE_ROCM_TOOLKIT)
+  target_link_libraries(runUnitTests gtest gtest_main ${libname} ${apiname} ${opname} pthread ${TensorFlow_LIBRARY} deepmd_op_rocm coverage_config)
+endif(USE_ROCM_TOOLKIT)
+if((NOT USE_ROCM_TOOLKIT) AND (NOT USE_CUDA_TOOLKIT))
   target_link_libraries(runUnitTests gtest gtest_main ${libname} ${apiname} ${opname} pthread ${TensorFlow_LIBRARY} coverage_config)
-endif()
+endif((NOT USE_ROCM_TOOLKIT) AND (NOT USE_CUDA_TOOLKIT))
+
```
amcadmus, line 102: again use 
```cmake
if (AAA)
...
elif (BBB)
...
else
...
endif
```

source/lib/src/rocm/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+add_compile_definitions(__HIP_PLATFORM_HCC__)
+
+# set c++ version c++11
+SET(CMAKE_CXX_STANDARD 11)
+SET(CMAKE_HIP_STANDARD 11)
+
+message(STATUS "HIP major version is " ${HIP_VERSION_MAJOR})
+
+set (HIP_HIPCC_FLAGS -hc; -fno-gpu-rdc; --amdgpu-target=gfx906; -fPIC; -O3; --std=c++11; -D__HIP_PLATFORM_HCC__)
```
amcadmus, line 19: haven't `__HIP_PLATFORM_HCC__` been already added to the definition??

source/op/CMakeLists.txt
```diff
@@ -5,18 +5,33 @@ set(OP_LIB ${PROJECT_SOURCE_DIR}/lib/src/SimulationRegion.cpp ${PROJECT_SOURCE_D
...
+  endif(USE_CUDA_TOOLKIT)
+  if(USE_ROCM_TOOLKIT)
+    add_library(${LIB_DEEPMD_OP} SHARED ${OP_ROCM_SRC})
+    find_package(ROCM REQUIRED)
+    include_directories(${ROCM_INCLUDE_DIRS})
+    target_link_libraries (${LIB_DEEPMD_OP} ${ROCM_LIBRARIES})
+  endif(USE_ROCM_TOOLKIT)
+  if((NOT USE_CUDA_TOOLKIT) AND (NOT USE_ROCM_TOOLKIT))
+    add_library(${LIB_DEEPMD_OP} SHARED ${OP_SRC})
+  endif()
```
amcadmus, line 34: The same issue

source/op/CMakeLists.txt
```diff
@@ -27,12 +42,22 @@ if (BUILD_PY_IF)
...
+    find_package(ROCM REQUIRED)
+    include_directories(${ROCM_INCLUDE_DIRS})
+    target_link_libraries (op_abi ${LIB_DEEPMD_OP_DEVICE})
+    target_link_libraries (op_grads ${LIB_DEEPMD_OP_DEVICE})
+  endif(USE_ROCM_TOOLKIT)
+  if((NOT USE_ROCM_TOOLKIT) AND (NOT USE_CUDA_TOOLKIT))
+    add_library(op_abi SHARED ${OP_SRC} ${OP_LIB})
+    add_library(op_grads SHARED ${OP_GRADS_SRC})
+  endif((NOT USE_ROCM_TOOLKIT) AND (NOT USE_CUDA_TOOLKIT))
+
```
amcadmus, line 60: same issue

amcadmus (changes requested):

pkulzy:
> > > At least we can test compilation process
> > 
> > 
> > This is fine if it is only for the compilation process, but in this case we need to install the ROCM package
> 
> Is it hard to install the ROCM package in a virtual environment? For cudatoolkit, it's easy to install it using `apt-get`.
> https://github.com/deepmodeling/deepmd-kit/blob/5a8bc2c97f6c16cf783a2802ffa8da4ddd49c105/.github/workflows/build_cc.yml#L25

The ROCM problem encountered earlier is due to the problem of my personal computer system. I used the WSL of windows to install, so some of the above problems have arisen, but the installation process is more complicated (estimated to be skilled in 2~5h) Not as well packaged as CUDA

source/lib/src/rocm/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+set(CMAKE_LINK_WHAT_YOU_USE TRUE)
+
+# set c++ version c++11
+SET(CMAKE_CXX_STANDARD 11)
+SET(CMAKE_HIP_STANDARD 11)
+
+message(STATUS "HIP major version is " ${HIP_VERSION_MAJOR})
+
+set (HIP_HIPCC_FLAGS -hc; -fno-gpu-rdc; --amdgpu-target=gfx906; -fPIC; -O3; --std=c++11)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -DCUB_IGNORE_DEPRECATED_CPP_DIALECT -DCUB_IGNORE_DEPRECATED_CPP_DIALECT")
```
amcadmus, line 14: Recommend using `add_definitions` rather than hacking `CMAKE_CXX_FLAGS` with `-DCUB_IGNORE_DEPRECATED_CPP_DIALECT -DCUB_IGNORE_DEPRECATED_CPP_DIALECT`

amcadmus (changes requested):

source/lib/src/rocm/CMakeLists.txt
pkulzy: Ok, I willly modify later, by the way why do you recommend using add_definitions? Is it because add_definition can affect C++ and C compilers?

amcadmus: Does `add_definiations` only affects the building tasks in source/lib/src/rocm? Could you please check if it has affects on building tasks outside rocm?

amcadmus:
Please resolve the conflict. Thanks!

Issue #668, opened 2021-05-26
[BUG] _GPU version gives wrong results_


dfz05:
**Summary**

GPU version gives wrong results. When I run a minimization task by using the GPU version (Deepmd-kit standalone 1.3.3), sometimes there is a sudden increase in the potential energy. However, if the CPU version was used, there is no such problem.

Step CPU PotEng TotEng Pxx Pyy Pzz Lx Ly Lz 
     999    97.233572   **-5404.4774   -5404.4774**    20.158298   -7.3528705    12.085191    34.105647    34.074605    34.071433 
    1000    98.504249   **-5261.9372   -5261.9372**   -4195.7038   -4834.5739   -3836.2866    34.105647    34.074605    34.071433 

The source files are too large for me to upload. Please contact me (dfzshiwo@163.com) for the input files.

denghuilu:
#680 check the pr page for more discussions

dfz05:
A similar bug was also found when running MC+MD by using the GPU version of deepmd-kit-2.0beta. The potential energy frequently shifts to unbelievable values. Nevertheless, this bug dose not cause any failure to the simulation process in this situation.

njzjz:
Fixed in #812.

PR #680, opened 2021-05-31
fix issue 668

denghuilu:
### About issue 668

- In issue 668, tensorflow may initialize op ProdEnvMatA twice, and the private member gpu_inlist of ProdEnvMatA did not get the proper initialization in the second time.
- Since step 2, the address of gpu_inlist maintains unchaged.

image

- The new implementation will automatically initialize gpu_inlist when the program detects that the gpu_inlist is empty.


denghuilu:
#668 This pr should fix the problem.

denghuilu:
Mark. This problem is more complicated than I expected, and there are still potential bugs that need to be fixed

PR #691, opened 2021-06-02
Fix uniform random seed

amcadmus:
If `seed` is set in descriptors or fitting nets, the different layers are initialized with the same random seed, this may slow down the training speed. This PR fixes the issue


njzjz (approved):

deepmd/descriptor/se_a.py
```diff
@@ -23,12 +23,13 @@ def __init__ (self,
                   axis_neuron: int = 8,
                   resnet_dt: bool = False,
                   trainable: bool = True,
-                  seed: int = 1,
+                  seed: int = None,
                   type_one_side: bool = True,
                   exclude_types: List[int] = [],
```
njzjz, line 28: It's not related to this PR but I just saw it -- the type of `exclude_types` should be `List[List[int]]` instead of `List[int]`

PR #704, opened 2021-06-05
Action to Sync gitee

tuoping:
Added mirror_gitee.yml in .github/workflow/
    An id with password-free access to both repo is needed. 
    Please refer to:
    "SSH_PRIVATE_KEY: Create a SSH key without a passphrase which has access to both repositories. On GitHub you can add the public key as a deploy key to the repository. "


.github/workflows/mirror_gitee.yml
```diff
@@ -0,0 +1,18 @@
...
+concurrency:
+  group: git-mirror
+
+jobs:
+  git-mirror:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: wearerequired/git-mirror-action@v1
+        env:
+          SSH_PRIVATE_KEY: ${{ secrets.SSH_PRIVATE_KEY }}
```
njzjz, line 15: Have you added secrets? I suggest add into **organization secrets**.

njzjz: Also, have you added the key into Gitee?

tuoping: I'm retracting this. I think there might be a better solution.

Issue #712, opened 2021-06-08
[BUG] `atom_ener` doesn't work in v2.0.0.b0


njzjz:
**Summary**

`atom_ener` doesn't work in v2.0.0.b0, even if I set both `model->fitting_net->atom_ener` and `model->descriptor->set_davg_zero`. The input can be seen in #658. However, it works if the model is trained and frozen in v1.3.3 and converted to v2.0 by #633.

**Deepmd-kit version, installation way, input file, running commands, error log, etc.**

The model frozen in v2.0.0.b0

**Steps to Reproduce**

The input is as the same as #658. After freezing the model, predict the energy of a single HW or OW atom using Python API. It's not zero.

```py
from deepmd.infer import DeepPot
import numpy as np
dp = DeepPot('graph.000.pb')
coord = np.array([[0,0,0]]).reshape([1, -1])
atype = [2]
e, f, v = dp.eval(coord, None, atype)
print(e)
```

Result for the model trained and frozen in v2.0.0.b0: `[[0.15377867]]` (not zero)

Result for the model trained and frozen in v1.3.3 and converted using #633: `[[-1.11022302e-15]]` (zero)

njzjz:
By the way, it's better to add a unittest for it (although I don't know how to do it).

PR #714, opened 2021-06-09
fix bug in ener fitting: zero net should share the parameters with the working net

amcadmus:
This fixes issue #712


njzjz (approved):
It works to me. The output is `[[0.]]`.

PR #715, opened 2021-06-09
Add subcommand to calculate model-deviation

Ericwang6:


amcadmus (changes requested):
1. please do NOT upload binary data files. use common.gen_data to generate a small system for test.
2. please do not upload binary model files (in total 6M!!). use test/infer/deeppot.pbtxt and test/infer/deeppot-1.pbtxt. Two models are enough for the test of model deviation. you can check test_deeppot_a.py for how to convert .pbtxt to .pb

amcadmus (changes requested):
I would suggest :
1. to add a python interface for the model deviation
2. build the command line model deviation calculation upon the python interface. 
The advantage is that one can either calculate the model deviation from his/her python script or from command line.

deepmd/entrypoints/model_devi.py
```diff
@@ -0,0 +1,156 @@
...
+    '''
+        fs : numpy.ndarray, size of `n_models x n_frames x n_atoms x 3`
+    '''
+    fs_mean = np.mean(fs, axis=0)
+    fs_err = np.sum((fs - fs_mean) ** 2, axis=-1)
+    fs_devi = np.mean(fs_err, axis=0) ** 0.5
+    max_devi_f = np.max(fs_devi, axis=1)
+    min_devi_f = np.min(fs_devi, axis=1)
+    avg_devi_f = np.mean(fs_devi, axis=1)
+    return max_devi_f, min_devi_f, avg_devi_f
```
njzjz, line 17: Isn't it the std for force **components**? But in the C++ API and LAMMPS, the std for **atomic** forces is calculated.

doc/getting-started.md
```diff
@@ -238,6 +239,48 @@ Model compression, with little loss of accuracy, can greatly speed up MD inferen
...
+                        The output file for results of model deviation
+                        (default: model_devi.out)
+  -f FREQUENCY, --frequency FREQUENCY
+                        The trajectory frequency of the system (default: 1)
+  -i ITEMS, --items ITEMS
+                        The physical quantities of which model deviation is
+                        calculated. (default: vf)
+```
+
+For more details respect to definition of model deviation and its application, please refer to `Yuzhi Zhang, Haidi Wang, Weijie Chen, Jinzhe Zeng, Linfeng Zhang, Han Wang, and Weinan E, DP-GEN: A concurrent learning platform for the generation of reliable deep learning based potential energy models, Computer Physics Communications, 2020, 107206.`
```
njzjz, line 282: Use `2020, 253, 107206` instead of `2020, 107206`.

Ericwang6: OK. I've changed that.

deepmd/entrypoints/model_devi.py
Ericwang6: Exactly. Thanks for pointing it out.

Ericwang6:
Several changes are made according to suggestions from @amcadmus and @njzjz .

deepmd/entrypoints/model_devi.py
Ericwang6: It is the std of atomic forces. I will use `np.linalg.norm` and `np.std` function to make it more clear.

doc/getting-started.md
```diff
@@ -238,6 +239,57 @@ Model compression, with little loss of accuracy, can greatly speed up MD inferen
...
+  -o OUTPUT, --output OUTPUT
+                        The output file for results of model deviation
+                        (default: model_devi.out)
+  -f FREQUENCY, --frequency FREQUENCY
+                        The trajectory frequency of the system (default: 1)
+```
+
+Furthermore, one can use the python interface to calulate model deviation.
+```python
+from deepmd.entrypoints.model_devi import calc_model_devi
```
njzjz, line 281: Maybe we can change the directory or give an alias. `entrypoints` isn't a good name for users to import. @amcadmus What do you think?

amcadmus: Agree, should be placed in `deepmd.infer`

doc/getting-started.md
```diff
@@ -9,14 +9,15 @@ In this text, we will call the deep neural network that is used to represent the
 3. [Freeze a model](#freeze-a-model)
 4. [Test a model](#test-a-model)
 5. [Compress a model](#compress-a-model)
-6. [Model inference](#model-inference)
+6. [Calculate model deviation](#calculate-model-deviation)
```
amcadmus, line 12: I would recommend 

1. merge the command line part with "test a model", because the model deviation is a method to test the quality of the model without labels. 
2. merge the python part with "Model inference/Python interface"

What do you think?

source/tests/test_model_devi.py
```diff
@@ -0,0 +1,35 @@
...
+        box = np.load(os.path.join(self.data_dir, "set.000/box.npy"))
+        atype = np.loadtxt(os.path.join(self.data_dir, "type.raw"))
+        model_devi = calc_model_devi(coord, box, atype, self.graphs)
+        for ii in range(1, 7):
+            self.assertAlmostEqual(model_devi[0][ii], self.expect[ii])
+    
+    def tearDown(self):
+        for pb in self.graph_dirs:
+            os.remove(pb)
+        shutil.rmtree(self.data_dir)
```
amcadmus, line 35: One may also use `common.del_data`

source/tests/test_model_devi.py
```diff
@@ -0,0 +1,35 @@
...
+        self.output = os.path.join(tests_path, "model_devi.out")
+        self.expect = np.array([0, 1.670048e-01, 4.182279e-04, 8.048649e-02, 5.095047e-01, 4.584241e-01, 4.819783e-01])
+    
+    def test_calc_model_devi(self):
+        coord = np.load(os.path.join(self.data_dir, "set.000/coord.npy"))
+        box = np.load(os.path.join(self.data_dir, "set.000/box.npy"))
+        atype = np.loadtxt(os.path.join(self.data_dir, "type.raw"))
+        model_devi = calc_model_devi(coord, box, atype, self.graphs)
+        for ii in range(1, 7):
+            self.assertAlmostEqual(model_devi[0][ii], self.expect[ii])
```
amcadmus, line 30: Could you please add a test case for  multiple frames? for example one can copy the frames in the "system" and check if model_devi[0] == model_devi[1]

amcadmus (changes requested):

doc/getting-started.md
Ericwang6: Agree

PR #725, opened 2021-06-10
support converting models generated in v1.3 to 2.0 compatibility

njzjz:
(Next plan: add v1.2 compatibility)

### DIFF
--- a/source/op/prod_env_mat_multi_device.cc
+++ b/source/op/prod_env_mat_multi_device.cc
@@ -25,6 +25,26 @@ REGISTER_OP("ProdEnvMatA")
     .Output("nlist: int32");
     // only sel_a and rcut_r uesd.
 
+// an alias of ProdEnvMatA -- Compatible with v1.3
+REGISTER_OP("DescrptSeA")
+    .Attr("T: {float, double}")
+    .Input("coord: T")
+    .Input("type: int32")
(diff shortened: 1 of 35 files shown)


njzjz:
@amcadmus I've changed the implementation: Now users can converting models directly using v2.0. There's no necessary to install the old version.

source/op/prod_env_mat_multi_device.cc
```diff
@@ -25,6 +25,26 @@ REGISTER_OP("ProdEnvMatA")
     .Output("nlist: int32");
     // only sel_a and rcut_r uesd.
 
+// an alias of ProdEnvMatA -- Compatible with v1.3
+REGISTER_OP("DescrptSeA")
```
amcadmus, line 29: Do we need to register `DescrptSeA`? the op name has changed to `ProdEnvMatA` by `convert`

njzjz: We need to load the old graph first when executing `convert`.

amcadmus: I see

amcadmus (approved):

denghuilu (approved):
A smart approach to register different ops with the same class ProdEnvMatA

Issue #753, opened 2021-06-16
Wiki page needs update


amcadmus:
**Summary**

It looks that the wiki page is not up-to-the date. Some update on the doc is not synced to the wiki. 

Do we really need the wiki pages, most of which are the same as the doc?

tuoping:
I don't think we need it.

njzjz:
I've removed duplicated pages. After the rest pages are processed, I'll close it.

njzjz:
I've closed wiki and also projects, since they are unused now.

PR #761, opened 2021-06-17
Typeembed develdoc

tuoping:
Added development guide for type-embedding (authered by jxxiaoshaoye)
This is copied from wiki.


tuoping:
#753

tuoping:
#753

PR #775, opened 2021-06-20
[CMake] Fix Python interpreter not defined.

shishaochen:
With CMake built-in API [FindPython](https://cmake.org/cmake/help/latest/module/FindPython.html), user can avoid explicity decaring variable `PYTHON_EXECUTABLE`.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -56,14 +56,15 @@ find_package(Threads)
 
 # auto op_cxx_abi
 if (NOT DEFINED OP_CXX_ABI)
-  if (BUILD_PY_IF) 
+  if (BUILD_PY_IF)
+    find_package(Python REQUIRED COMPONENTS Interpreter) 
     if (DEFINED TENSORFLOW_ROOT)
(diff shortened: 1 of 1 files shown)


njzjz:
This has been done by `scikit-build`: https://github.com/scikit-build/scikit-build/blob/c0741f9a1de9dd97e0a1e6f0b3b02fcd76221418/skbuild/cmaker.py#L195-L196

So, I don't find the reason why we need to manually find it. Could you please explain your situation?

shishaochen:
@njzjz It should be OK to install deepmd-kit through setup.py.
However, the official document [Download and install](https://deepmd.readthedocs.io/en/master/install.html) tell users to build & install using CMake directly. In this way, failure comes out as `PYTHON_EXECUTABLE` not defined.

njzjz:
When you build C++ module, `BUILD_PY_IF` should be `FALSE`, so I don't understand why it throws error.

shishaochen:
@njzjz Got it. I tried to build PY_IF and CPP_IF together through CMake.

Issue #546, opened 2021-04-20
[Feature Request] Auto convertion of model to DeePMD-kit v2.0 compatibility


amcadmus:
**Summary**

Automatic conversion of frozen model file from v1 compatible to v2 compatible.

**Detailed Description**

After the upgrading to DeePMD-kit v2, the frozen models trained with v1 cannot be used anymore. It would be useful to provide users a tool that convert the v1 frozen models to v2.

**Further Information, Files, and Links**

njzjz:
It is also valuable to export the model to an "universal" format, and import from this format. Thus, if this new format is stable, we only need to maintain the convert between the tensorflow model and this new format.

PR #633, opened 2021-05-14
support converting models to 2.0 compatibility

amcadmus:

PR #789, opened 2021-06-22
give a default value to T and convert models from v1.2

njzjz:
See also https://www.tensorflow.org/guide/create_op#backwards_compatibility.

Implement #546.

### DIFF
--- a/doc/troubleshooting/model-compatability.md
+++ b/doc/troubleshooting/model-compatability.md
@@ -6,16 +6,11 @@ DeePMD-kit guarantees that the codes with the same major and minor revisions are
 
 One can execuate `dp convert-from` to convert an old model to a new one.
 
-|   | Model v0.12 | Model v1.0 | Model v1.1 | Model v1.2 | Model v1.3 | Model v2.0 |
+| Model version | v0.12 | v1.0 | v1.1 | v1.2 | v1.3 | v2.0 |
 |:-:|:-----------:|:----------:|:----------:|:----------:|:----------:|:----------:|
-|DeePMD-kit v0.12 | 😄 | 😢 | 😢 | 😢 | 😢 | 😢 |
-|DeePMD-kit v1.0  | 😢 | 😄 | 😢 | 😢 | 😢 | 😢 |
-|DeePMD-kit v1.1  | 😢 | 😢 | 😄 | 😢 | 😢 | 😢 |
-|DeePMD-kit v1.2  | 😢 | 😢 | 😢 | 😄 | 😢 | 😢 |
-|DeePMD-kit v1.3  | 😢 | 😢 | 😢 | 😊 | 😄 | 😢 |
-|DeePMD-kit v2.0  | 😢 | 😢 | 😢 | 😊 | 😊 | 😄 |
+| Compatibility  | 😢 | 😢 | 😢 | 😊 | 😊 | 😄 |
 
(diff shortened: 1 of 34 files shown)


amcadmus (approved):

amcadmus:
How about adding a section in the [getting started ](https://github.com/deepmodeling/deepmd-kit/blob/devel/doc/getting-started.md) to introduce the users the conversions of model compatibility?

njzjz:
I add a table to troublesome and I am not sure if it should be moved to the getting started.

tuoping (changes requested):
How about add a link to doc/troubleshooting/model-compatibility in getting-started?

doc/troubleshooting/model-compatability.md
```diff
@@ -1,5 +1,21 @@
...
+
+One can execuate `dp convert-from` to convert an old model to a new one.
+
+|   | Model v0.12 | Model v1.0 | Model v1.1 | Model v1.2 | Model v1.3 | Model v2.0 |
+|:-:|:-----------:|:----------:|:----------:|:----------:|:----------:|:----------:|
+|DeePMD-kit v0.12 | 😄 | 😢 | 😢 | 😢 | 😢 | 😢 |
+|DeePMD-kit v1.0  | 😢 | 😄 | 😢 | 😢 | 😢 | 😢 |
+|DeePMD-kit v1.1  | 😢 | 😢 | 😄 | 😢 | 😢 | 😢 |
+|DeePMD-kit v1.2  | 😢 | 😢 | 😢 | 😄 | 😢 | 😢 |
+|DeePMD-kit v1.3  | 😢 | 😢 | 😢 | 😊 | 😄 | 😢 |
```
amcadmus, line 15: Do we have tools to convert v1.3 model to v1.2 compatibility?

denghuilu (approved):
works well in my local workstation

doc/troubleshooting/model-compatability.md
njzjz: No.

amcadmus: So why there is a smile? Is it a typo?

njzjz: The smile is where model version is v1.2 and deepmd-kit version is 1.3, not converse.

amcadmus: Can we use a v1.2 model in deepmd-kit v1.3?

amcadmus: So you mean we can convert a model v1.2 to v1.3 by using deepmd-kit v1.3.   I do not see this subcommand in v1.3

njzjz: We can convert a model v1.2 to v1.3 by using deepmd-kit v1.2. So this should be a 😢?

amcadmus: Yes, the manner of conversion is somewhat different... In deepmd-kit v1.2 and 1.3, it is `convert-to` (pending PR #631 and #633), but in v2.0 it is `convert-from`.

I would proposed to only keep the doc for v2.0, and write separate docs for v1.2 and 1.3, what do you think?

njzjz: Good idea.

amcadmus (approved):

Issue #742, opened 2021-06-14
Errors with DeepMD GPU version


Grace0987:
**Summary**

I installed DeepMD v 1.3.3 with the offline script "deepmd-kit-1.3.3-cuda10.0_gpu-Linux-x86_64.sh" .
Inorder to train a module, i used this command "dp train <json file>"  on  a GPU machine ,which started well with loading the 
corresponding CUDA and tensorflow libraries . But the Job did crash with the below error:

Grace0987:
Error :
 E tensorflow/stream_executor/cuda/cuda_blas.cc:428] failed to run cuBLAS routine: CUBLAS_STATUS_EXECUTION_FAILED
Traceback (most recent call last):
tensorflow.python.framework.errors_impl.InternalError: 2 root error(s) found.
  (0) Internal: Blas GEMM launch failed : a.shape=(29440, 25), b.shape=(25, 50), m=29440, n=50, k=25
         [[node filter_type_0/MatMul_1 (defined at /lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py:1751) ]]
         [[l2_force_test/_59]]
  (1) Internal: Blas GEMM launch failed : a.shape=(29440, 25), b.shape=(25, 50), m=29440, n=50, k=25
         [[node filter_type_0/MatMul_1 (defined at /lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py:1751) ]]
0 successful operations.
0 derived errors ignored.

Grace0987:
So, i thought to try the next releases, and i got the same error in a couple of next releases. My last try was to install the latest release "deepmd-kit-2.0.0.b0-cuda10.1_gpu-Linux-x86_64.sh". This time, when i trained the same model as previous, the cublas error did not appear. Now , it is complaining about few parameters to be changed in input file. 
name: A100-SXM4-40GB, pci bus id: 0000:47:00.0, compute capability: 8.0)
2021-05-27 03:06:07.971279: I tensorflow/compiler/jit/xla_gpu_device.cc:99] Not creating XLA devices, tf_xla_enable_xla_devices not set
DEEPMD INFO    ---Summary of DataSystem: training     -----------------------------------------------
DEEPMD INFO    found 3 system(s):
DEEPMD INFO                                        system  natoms  bch_sz   n_bch   prob  pbc
DEEPMD INFO                               ../data/data_0/     192       1      80  0.250    T
[...]

Issue #675, opened 2021-05-29
[BUG] _An error in `prod_env_mat.cu`_


Ericwang6:
**Summary**

An error in `/source/lib/src/cuda/prod_env_mat.cu`

**Deepmd-kit v2.0.0b0**

When training data for small organic molecules with `se_e2_a` descriptor, an error occurs: `cuda assert: DeePMD-kit:	illegal nbor list sorting /home/yingze/deepmd-kit/source/lib/src/cuda/prod_env_mat.cu 509`.

My input.json :
```json
{
    "model": {
        "type_map": [
            "C",
            "H",
            "N",
            "O"
        ],
        "descriptor": {
            "type": "se_e2_a",
            "sel": [
                48,
                40,
                48,
                48
            ],
            "rcut_smth": 0.5,
            "rcut": 6.0,
            "neuron": [
                20,
                40,
                80
            ],
            "resnet_dt": false,
            "axis_neuron": 8,
            "type_one_side": true,
            "seed": 1,
            "activation_function": "gelu"
        },
        "fitting_net": {
            "neuron": [
                240,
                240,
                240
            ],
            "resnet_dt": true,
            "seed": 1,
            "activation_function": "gelu"
        }
    },
    "learning_rate": {
        "type": "exp",
        "start_lr": 0.0001,
        "stop_lr": 5e-8,
        "decay_steps": 500
    },
    "loss": {
        "type": "ener",
        "start_pref_e": 0.02,
        "limit_pref_e": 10,
        "start_pref_f": 1000,
        "limit_pref_f": 1,
        "start_pref_v": 0,
        "limit_pref_v": 0
    },
    "training": {
        "numb_steps": 100000,
        "disp_file": "lcurve.out",
        "disp_freq": 1000,
        "numb_test": 1,
        "save_freq": 1000,
        "save_ckpt": "model.ckpt",
        "disp_training": true,
        "time_training": true,
        "training_data": {
            "batch_size": "auto",
            "systems": [
                "./C0H0N0O2",
                "./C1H3N1O1"
            ]
        }
    }
}
```
The program is run on pbs system.

**Steps to Reproduce**

An example of relevant data is attached here:
[issue.zip](https://github.com/deepmodeling/deepmd-kit/files/6564330/issue.zip)

njzjz:
#742 reported an same bug.

iProzd:
This bug occurs when:
1. using gelu as activation function (GPU environment);
[...]

PR #800, opened 2021-06-26
Fix bug of empty input in gelu.cu 

iProzd:
#675 and #742 maybe the same bugs when: 
1. using gelu as activation function (GPU environment);
2. there's one type (or types) of atom not showing in the system but in the type_map;
this causes the empty input in gelu.cu and thus breaks.


njzjz:
Is our gelu much faster than TensorFlow's (tensorflow/tensorflow#41178)? It seems that they implement gelu in only some python codes. See https://github.com/tensorflow/tensorflow/blob/v2.5.0/tensorflow/python/ops/nn_ops.py#L3505-L3548

njzjz (approved):

amcadmus:
> Is our gelu much faster than TensorFlow's ([tensorflow/tensorflow#41178](https://github.com/tensorflow/tensorflow/pull/41178))? It seems that they implement gelu only using some python codes. See https://github.com/tensorflow/tensorflow/blob/v2.5.0/tensorflow/python/ops/nn_ops.py#L3505-L3548

Yes, that's why we implement gelu by ourselves

Issue #755, opened 2021-06-16
[Feature Request] support `-std=c++17`


njzjz:
**Summary**

When add `-std=c++17` to `CXXFLAGS` (this is the default environment variables of conda compilers), CUDA files will throw the following errors (CPU version works well). It works after changing `-std=c++17` to `-std=c++11`.

It's not a bug, but I hope `-std=c++17` can be supported on some day.

UPDATE: I am not sure whether it's a bug. Maybe CMake should set the proper C++ standard option?

**Detailed Description**

```text
2021-06-15T23:36:35.3222435Z   [  1%] Building NVCC (Device) object lib/src/cuda/CMakeFiles/deepmd_op_cuda.dir/deepmd_op_cuda_generated_tabulate.cu.o
2021-06-15T23:36:35.3438702Z   nvcc warning : The 'compute_35', 'compute_37', 'compute_50', 'sm_35', 'sm_37' and 'sm_50' architectures are deprecated, and may be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
2021-06-15T23:36:36.4755310Z   nvcc warning : The 'compute_35', 'compute_37', 'compute_50', 'sm_35', 'sm_37' and 'sm_50' architectures are deprecated, and may be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
2021-06-15T23:36:37.9274712Z   /home/conda/feedstock_root/build_artifacts/deepmd-kit_1623798550400/_build_env/x86_64-conda-linux-gnu/include/c++/8.4.0/type_traits(165): error: "conjunction_v" is not a function or static data member
2021-06-15T23:36:37.9275291Z 
2021-06-15T23:36:37.9276049Z   /home/conda/feedstock_root/build_artifacts/deepmd-kit_1623798550400/_build_env/x86_64-conda-linux-gnu/include/c++/8.4.0/type_traits(165): error: "constexpr" is not valid here
2021-06-15T23:36:37.9276621Z 
2021-06-15T23:36:37.9277433Z   /home/conda/feedstock_root/build_artifacts/deepmd-kit_1623798550400/_build_env/x86_64-conda-linux-gnu/include/c++/8.4.0/type_traits(168): error: "disjunction_v" is not a function or static data member
2021-06-15T23:36:37.9277978Z 
[...]

PR #805, opened 2021-06-27
make deep tensor deriv again

y1xiaoc:
The main purpose of this PR is to implement the (component-wise) force and virial calculation for the global tensor predicted by deep tensor models. Since the outputs are tensors, the force and tensor are calculated on each separate component and later stacked together. Similar tasks has been implemented for dplr calculation in `DataModifier` class. The current one has the following difference:
- the derivative is calculated on the global tensor, instead of atomic ones like DataModifier.
- since global tensor contains few components, the derivative is implemented as  loop over components, which may be slower
- the component-wise derivative (i.e. jacobian) is given directly, and there's no need to pass a head gradient to the graph.

The design is chosen to fit the potential usage in plumed. If per-atom derivative is needed instead, consider using DataModifier or modify it.

In detail, the PR mainly do the following things:
- change the tensor model so that the graph includes new ops for calculating global tensor, force and (atomic) virial
- add a new function `eval_full` in `DeepTensor` python class, which resembles the `eval` function in `DeepPot` class, that will output  global tensor, force and virial (and atomic tensor and atomic virial if `atoimc=True`)
- add several overload to the `compute` function in `DeepTensor` c++ class, which again resemble the `compute` function in `DeepPot` class, and correspond to the python counter part

as well as several minor changes:
- add more unittests to support the new force and virial calculations, and a corner case for the order of the output tensor
- refine the old `eval` in the python class of `DeepTensor` to handle the pbc correctly
- fix a bug in the old c++ interface where the output tensor could be in wrong order (but old fix dplr was correct)
- change the output of tensor model to rank 2 (with shape [nframes, (natoms x) odim]) to be consistent with other part of the code (it also handles the suffix correctly now)

and some technical details:
- refine `.gitignore` to exclude some files generated during tests (but may not be all)
- force `std=c++11` even in cuda 11 (temporarily) to avoid compiling error (#755)

The implementation of this PR paid special attention to backward compatibility, so the old models should still be able to run on the new code without changing anything. However, to use the new force and virial feature, one needs to retrain (since I directly changed the procedure of building the graph) and refreeze a new model.


deepmd/.gitignore
```diff
@@ -0,0 +1,2 @@
+op/_*.py
```
njzjz, line 1: will this line also exclude `op/__init__.py`?

y1xiaoc: That's true. I added another line to exclude it.

source/api_cc/src/DeepTensor.cc
```diff
@@ -91,16 +109,203 @@ run_model (std::vector<VALUETYPE> &	d_tensor_,
...
+    return;
+  }
+
+  std::vector<Tensor> output_tensors;
+  deepmd::check_status (session->Run(input_tensors, 
+			    {name_prefix(name_scope) + "o_global_" + model_type, 
+			     name_prefix(name_scope) + "o_force", 
+			     name_prefix(name_scope) + "o_virial"},
+			    {}, 
+			    &output_tensors));
```
amcadmus, line 160: Please notice  #812  the two run models should eval the same output tensors.

source/api_cc/tests/test_utils.h
```diff
@@ -21,6 +21,26 @@ _fold_back(
   }
 }
```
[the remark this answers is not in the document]
y1xiaoc: Agree. Has changed it.

source/tests/common.py
```diff
@@ -359,3 +359,23 @@ def virial_dw_test (inter,
...
+    in_shape = x.shape
+    y0 = f(x)
+    out_shape = y0.shape
+    res = np.empty(out_shape+in_shape)
+    for idx in np.ndindex(*in_shape):
+        diff = np.zeros(in_shape)
+        diff[idx] += delta
+        y1 = f(x+diff)
+        res[(Ellipsis, *idx)] = (y1-y0) / delta
+    return res
```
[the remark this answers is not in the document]
y1xiaoc: Changed. Originally it was just a quick check.

source/api_cc/src/DeepTensor.cc
y1xiaoc: Solved. Now both would call the same run model.

amcadmus (approved):

PR #809, opened 2021-06-29
speedup ROCm kernel which uses atomicAdd

darelbeida:
This PR works on optimizing ROCm kernel which uses atomicAdd. 

By the experiments before, we already knew that atomicAdd on CUDA is much more efficient than on ROCm. A straightforward idea is to reduce the calling number of atomicAdd instruction.  By pre-summing the four values before doing atomicAdd, we gains nearly **4 times** speedup on `force_deriv_wrt_neighbors_a` and `virial_deriv_wrt_neighbors_a` ROCm kernels. 

Moreover, by switching the computing order inside and cross a block, we aim to reduce the crashing of writing same global memory address. In our benchmark, this switching gets **1.4 ~ 2.2 times** speedup.

We also benchmark the E2E performance on LAMMPS inference. On compressed copper model, this work gets **1.45x** speedup on pairs computing.

Baseline:
[2_compress.json.gz](https://github.com/deepmodeling/deepmd-kit/files/6730613/2_compress.json.gz)
screenshot-20210629-034744

After speedup:
[2_compress_optimized.json.gz](https://github.com/deepmodeling/deepmd-kit/files/6730617/2_compress_optimized.json.gz)
screenshot-20210629-034352

### DIFF
--- a/source/lib/src/rocm/prod_force.hip.cu
+++ b/source/lib/src/rocm/prod_force.hip.cu
@@ -51,12 +51,11 @@ __global__ void force_deriv_wrt_neighbors_a(
     const int nnei)
 {  
     // idy -> nnei
-    const unsigned int idx = blockIdx.x * blockDim.x + threadIdx.x;
-    const unsigned int idy = blockIdx.y;
+    const unsigned int idx = blockIdx.x;
+    const unsigned int idy = blockIdx.y * blockDim.x + threadIdx.x;
     const unsigned int idz = threadIdx.y;
-    const unsigned int idw = threadIdx.z;
     const int ndescrpt = nnei * 4;
-    if (idx >= nloc) {
+    if (idy >= nnei) {
         return;
     }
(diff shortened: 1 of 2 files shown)


denghuilu (approved):
This implementation could also be applied to the CUDA code

amcadmus:
We want to make sure that you have run the unittests in `source/lib/tests` under ROCm environment. 

The CI on github does not support UTs on GPUs.

darelbeida:
> We want to make sure that you have run the unittests in `source/lib/tests` under ROCm environment.
> 
> The CI on github does not support UTs on GPUs.

@amcadmus Unit tests in `source/lib/tests` and `source/api_cc/tests` paths are all passed under ROCm platform.

amcadmus (approved):

amcadmus:
@pkulzy  Could you please update the `_r` kernels accordingly? Please also see #811

Issue #824, opened 2021-07-02
[BUG] _Process killed when running lammps by using deepmd-kit-2.0.0.beta0 model on GPU and DCU_


Carrotkingdom:
**I am running a DP model with deepmd-kit-2.0.0.beta0 version, both original and compressed, on GPU and DCU. In all cases the process is killed after several steps, depending on the number of threads and environment.**

**Here are the model type, machine type, the number of threads and the corresponding step number before the process died.**

![image](https://user-images.githubusercontent.com/86839379/124285415-79c23680-db80-11eb-9df2-380c107a33f3.png)

**The lammps input file and the submit files on GPU and DCU are attached below.**

[MgZn.zip](https://github.com/deepmodeling/deepmd-kit/files/6754941/MgZn.zip)

pkulzy:
Will this problem occur when running on the CUDA platform?

denghuilu:
Have you tested these models on a single GPU or DCU?

pkulzy:
> Have you tested these models on a single GPU or DCU environment?

about 12300 on DCU

amcadmus:
I would suggest trying the latest commit on `devel`

denghuilu:
I will try it. > I would suggest trying the latest commit on `devel`

Carrotkingdom:
> Have you tested these models on a single GPU or DCU environment?

Same results on a single GPU. On CPU the process is normal and won't be killed.

denghuilu:
The latest GPU code has been tested on ehpc and there is no problem. Please close this   @amcadmus.

Carrotkingdom:
The latest DCU code on NSCC-ZZ is tested successfully @amcadmus .

PR #827, opened 2021-07-05
fix bug 824 and Synchronize updates to CUDA code 

pkulzy:
update the ROCM code to fix BUG #824 (because the array access is out of bounds)


pkulzy:
there is some wrong in code

PR #831, opened 2021-07-06
automatically determine the `sel` from the training data.

amcadmus:
Notice :
- The function `train` in the module `deepmd.entrypoints.train` is renamed to `train_dp` to avoid the conflict between the module and the function name, which causes trouble in the UT for the auto sel


denghuilu (approved):

njzjz (approved):

njzjz:
Document needs to be updated.

amcadmus:
@njzjz I miss used the auto merging. Could you please check my PR?

njzjz:
approve

PR #833, opened 2021-07-07
use cached Session to speed up py tests

njzjz:


njzjz:
Python test time reduce about 5 minutes.

amcadmus (approved):

PR #834, opened 2021-07-07
support to specify CUDA/ROCm root in python pkg building

darelbeida:
In some develop environment, CUDA/ROCm toolkit root cannot be detected automatically. This PR fixes it by allowing user to specify toolkit root as environment variable.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,9 +41,15 @@ if dp_variant == "cpu" or dp_variant == "":
     pass
 elif dp_variant == "cuda":
     cmake_args.append("-DUSE_CUDA_TOOLKIT:BOOL=TRUE")
+    cuda_root = os.environ.get("CUDA_TOOLKIT_ROOT_DIR")
+    if cuda_root:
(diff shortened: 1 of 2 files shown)


darelbeida:
need to confirm whether environment varialbe key should start with "DP_ " as other keys like "DP_VARIANT".

setup.py
```diff
@@ -41,9 +41,15 @@
     pass
 elif dp_variant == "cuda":
     cmake_args.append("-DUSE_CUDA_TOOLKIT:BOOL=TRUE")
+    cuda_root = os.environ.get("CUDA_TOOLKIT_ROOT_DIR")
+    if cuda_root:
+        cmake_args.append(f"-DCUDA_TOOLKIT_ROOT_DIR:STRING={cuda_root}")
 elif dp_variant == "rocm":
     cmake_args.append("-DUSE_ROCM_TOOLKIT:BOOL=TRUE")
-elif dp_variant != "":
```
njzjz, line 46: empty value should be allowed

njzjz (approved):

amcadmus (approved):

PR #840, opened 2021-07-12
try to use tf.data

njzjz:
Ref: https://www.tensorflow.org/guide/data
However, I don't notice a lot of performance improvement. Maybe it's not an important bottleneck...


shishaochen:
@njzjz Unless training & validation data is too huge to hold in host or device memory, there is no gain by involving `tf.data` which attempts to overlap input preprocessing and graph exectuion. Currently, it should be the fastest way that we use `numpy` to transform raw data and directly hold normalized samples in memory.

PR #530, opened 2021-04-19
Auto conversion of input script from v1 to v2 

ZiyaoLi:
compat v1 -> v2 implemented.

PR #844, opened 2021-07-13
also convert input v1 to v2 for dp compress

njzjz:
See also #530

### DIFF
--- a/deepmd/entrypoints/compress.py
+++ b/deepmd/entrypoints/compress.py
@@ -6,7 +6,7 @@ from typing import Optional
 
 from deepmd.common import j_loader
 from deepmd.utils.argcheck import normalize
-from deepmd.utils.compat import convert_input_v0_v1
+from deepmd.utils.compat import updata_deepmd_input
 
 from .freeze import freeze
 from .train import train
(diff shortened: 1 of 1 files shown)


deepmd/entrypoints/compress.py
```diff
@@ -64,7 +64,7 @@ def compress(
     """
     jdata = j_loader(INPUT)
     if "model" not in jdata.keys():
-        jdata = convert_input_v0_v1(jdata, warning=True, dump="input_v1_compat.json")
+        jdata = updata_deepmd_input(jdata, warning=True, dump="input_v1_compat.json")
```
ZiyaoLi, line 68: minor confusion: should default file name be `input_v2_compat.json`?

ZiyaoLi:
should the default file name be altered to `input_v2_compat`?
just in case of confusion.

deepmd/entrypoints/compress.py
njzjz: Yes...

ZiyaoLi (approved):

amcadmus (approved):

PR #866, opened 2021-07-17
remove cub include for CUDA>=11

njzjz:
CUDA 11 has included cub, so we don't need to include it any more.
cub can be upgraded along with the CUDA Toolkit to obtain bugfixes.
See https://github.com/NVIDIA/cub#releases.
See also https://github.com/NVIDIA/cub/blob/main/CHANGELOG.md.

P.S. It's still unclear to me that which version of C++ it supports.  (#755)


denghuilu (approved):

amcadmus (approved):

denghuilu:
> CUDA 11 has included cub, so we don't need to include it any more.
> cub can be upgraded along with the CUDA Toolkit to obtain bugfixes.
> See https://github.com/NVIDIA/cub#releases.
> See also https://github.com/NVIDIA/cub/blob/main/CHANGELOG.md.
> 
> P.S. It's still unclear to me that which version of C++ it supports. (#755)

Right now, the C++14 standard is adopted by default in cub. cub support cxx-11,14, and17 versions, which can be changed in cmake options. See: https://github.com/NVIDIA/cub/blob/main/CONTRIBUTING.md#cmake-options

PR #870, opened 2021-07-20
[Paddle] Fixed model save issues with Ener model

jim19930609:
Following issues fixed:
1. Removed @paddle.jit.to_static decorator. Model will be converted to
static graph at save time.

2. Manually set InputSpec for "Ener" model with "se_a" descriptor

3. Due to lack of support for "double" datatype at inference time,
default training precision was set to float (low precision)


JiabinYang (approved):
LGTM

PR #877, opened 2021-07-22
multi task training

jxxiaoshaoye:


deepmd/utils/learning_rate.py
```diff
@@ -14,7 +14,8 @@ def __init__ (self,
                   start_lr : float,
                   stop_lr : float = 5e-8,
                   decay_steps : int = 5000,
-                  decay_rate : float = 0.95
+                  decay_rate : float = 0.95,
+                  name: str = None
```
njzjz, line 18: There is no document for the new argument.

deepmd/entrypoints/train.py
```diff
@@ -216,6 +218,90 @@ def train(
...
+    ------
+    RuntimeError
+        if distributed training job nem is wrong
+    """
+    # load json database
+    jdata = j_loader(INPUT)
+
+    jdata = updata_deepmd_input(jdata, warning=True, dump="input_v2_compat.json")
+
+    sys_name = ['battery','AlCuMg','HfO2','water']
```
amcadmus, line 261: Hard coding should not appear!

deepmd/entrypoints/train.py
```diff
@@ -216,6 +218,90 @@ def train(
...
+    *,
+    INPUT: str,
+    init_model: Optional[str],
+    restart: Optional[str],
+    output: str,
+    mpi_log: str,
+    log_level: int,
+    log_path: Optional[str],
+    **kwargs,
+):
```
amcadmus, line 231: Basically `train_mt` is copied from `train`, this is NOT recommended. 
We do not want to maintain two copies of codes.

deepmd/model/ener.py
```diff
@@ -157,7 +160,7 @@ def build (self,
                                  box,
                                  mesh,
                                  input_dict,
-                                 suffix = suffix,
+                                 suffix = '',
```
amcadmus, line 163: What is the reason of this change? One cannot pass suffix to descriptor

deepmd/model/ener.py
```diff
@@ -146,7 +149,7 @@ def build (self,
             type_embedding = self.typeebd.build(
                 self.ntypes,
                 reuse = reuse,
-                suffix = suffix,
+                suffix = '',
```
amcadmus, line 152: What is the reason of this change? One cannot pass suffix to type_embedding

source/tests/test_model_se_t.py
```diff
@@ -62,8 +62,8 @@ def test_model(self):
         t_mesh             = tf.placeholder(tf.int32,   [None], name='i_mesh')
         is_training        = tf.placeholder(tf.bool)
         t_fparam = None
-
-        model_pred\
+        with tf.variable_scope('se_t'):
+            model_pred\
```
amcadmus, line 66: Shouldn't happen. May be related to the change of suffix in `model/ener.py`

amcadmus (changes requested):
Apart from the inline reviews, here are some issues should be resolved before merging: 

1. convention of the class name is CamelCase. the `DPTrainer_mt` should be `DPTrainerMultiTask`
2. Copied code should be avoided, for example `trainer_mt.py` is copied from `trainer.py` with little change. This should be avoided because we do not want to maintain two copies of codes. 
3. The example is 8k lines long!!! not one can understand such a long file. The new feature should be exemplified by a simple but complete example.
4. NO UT is provided with the new feature. 
5. NO doc is provided with the new feature.

shishaochen:
It is a really interesting topic that apply multi-task training into learning of physical properties. Based on first principles, if `energy` is accurate, other derived properties like `polar` and `dipole`,  should be accurate.

When descriptor `se_e2_a` is selected, there is no difference of `embedding_net` and `fitting_net` among targets `energy`, `polar` and `dipole`. If model parameters offer accurate prediction on `enery`, will the predictor graph of `polar` and `diplore` perform perfect as well?

deepmd/utils/data_system.py
```diff
@@ -508,6 +514,626 @@ def _prob_sys_size_ext(self, keywords):
...
+            nbatch_block = self.nbatches[block_stt[ii]:block_end[ii]]
+            tmp_prob = [float(i) for i in nbatch_block] / np.sum(nbatch_block)
+            sys_probs[block_stt[ii]:block_end[ii]] = tmp_prob * block_probs[ii]
+        return sys_probs
+
+class DeepmdDataDocker() :
+    """
+    Class for manipulating many data dockers. 
+    It is implemented with the help of DeepmdData
+    """
```
njzjz, line 1024: If most methods are copied from `DeepmdData`, why not inherit `DeepmdData`?

deepmd/utils/data_system.py
```diff
@@ -508,6 +514,626 @@ def _prob_sys_size_ext(self, keywords):
             sys_probs[block_stt[ii]:block_end[ii]] = tmp_prob * block_probs[ii]
         return sys_probs
 
+class DeepmdDataSystem_mt() :
+    """
+    Class for manipulating many data systems. 
+    It is implemented with the help of DeepmdData
+    """
```
njzjz, line 521: If many methods are copied from `DeepmdData`, why not inherit `DeepmdData`?

deepmd/utils/data_system.py
amcadmus: Either inherit from `DeepmdData` or implement with `DeepmdData` object as a private member.

amcadmus:
> It is a really interesting topic that apply multi-task training into learning of physical properties. Based on first principles or laws of nature, if `energy` is accurate, other derived properties like `polar` and `dipole`, should be accurate.
> 
> When descriptor `se_e2_a` is selected, there is no difference of `embedding_net` and `fitting_net` among targets `energy`, `polar` and `dipole`. The only difference is the type of `loss` and fields of `label`.
> 
> Here comes my question. If model parameters trained by `enery` label offer accurate prediction, will the prediction graph of `polar` and `diplore` perform perfect as well? More specific, models of `polar` and `dipole` reuse parameters of `energy` model.
> 
> If the answer is `yes`, multi-task training may be unnecessary. Otherwise, the key point is involving more types of labels to traing parameters better using limited count of frames.
> 
> Your thoughts are appreciated. @amcadmus @njzjz

The `energy`, `dipole` and `polar` are different physical quantities, and they are fit with different `fitting_net`s.  That's why we need multi-task training: different labels with different `fitting_net`s while sharing the same `descriptor`.

deepmd/utils/argcheck_mt.py
```diff
@@ -0,0 +1,685 @@
...
+        Argument("activation_function", str, optional = True, default = 'tanh', doc = doc_activation_function),
+        Argument("resnet_dt", bool, optional = True, default = False, doc = doc_resnet_dt),
+        Argument("precision", str, optional = True, default = "float64", doc = doc_precision),
+        Argument("trainable", bool, optional = True, default = True, doc = doc_trainable),
+        Argument("seed", [int,None], optional = True, doc = doc_seed),
+    ]        
+
+
+#  --- Descriptor configurations: --- #
+def descrpt_local_frame_args ():
```
njzjz, line 40: I don't think most of these methods are different from those in `argcheck.py`. It's silly to copy them again.

shishaochen:
> > It is a really interesting topic that apply multi-task training into learning of physical properties. Based on first principles or laws of nature, if `energy` is accurate, other derived properties like `polar` and `dipole`, should be accurate.
> > When descriptor `se_e2_a` is selected, there is no difference of `embedding_net` and `fitting_net` among targets `energy`, `polar` and `dipole`. The only difference is the type of `loss` and fields of `label`.
> > Here comes my question. If model parameters trained by `enery` label offer accurate prediction, will the prediction graph of `polar` and `diplore` perform perfect as well? More specific, models of `polar` and `dipole` reuse parameters of `energy` model.
> > If the answer is `yes`, multi-task training may be unnecessary. Otherwise, the key point is involving more types of labels to traing parameters better using limited count of frames.
> > Your thoughts are appreciated. @amcadmus @njzjz
> 
> The `energy`, `dipole` and `polar` are different physical quantities, and they are fit with different `fitting_net`s. That's why we need multi-task training: different labels with different `fitting_net`s while sharing the same `descriptor`.

In recommender system, multi-task training brings better performane on targets with less data. Because parameters of the shared bottom can be fully trained by other related targets' samples.

Thus, in multi-task training of DeePMD, what's the constitution of training data? Each sample has all target's labels, or each sample has partial targets' labels?

amcadmus:
> > > It is a really interesting topic that apply multi-task training into learning of physical properties. Based on first principles or laws of nature, if `energy` is accurate, other derived properties like `polar` and `dipole`, should be accurate.
> > > When descriptor `se_e2_a` is selected, there is no difference of `embedding_net` and `fitting_net` among targets `energy`, `polar` and `dipole`. The only difference is the type of `loss` and fields of `label`.
> > > Here comes my question. If model parameters trained by `enery` label offer accurate prediction, will the prediction graph of `polar` and `diplore` perform perfect as well? More specific, models of `polar` and `dipole` reuse parameters of `energy` model.
> > > If the answer is `yes`, multi-task training may be unnecessary. Otherwise, the key point is involving more types of labels to traing parameters better using limited count of frames.
> > > Your thoughts are appreciated. @amcadmus @njzjz
> > 
> > 
> > The `energy`, `dipole` and `polar` are different physical quantities, and they are fit with different `fitting_net`s. That's why we need multi-task training: different labels with different `fitting_net`s while sharing the same `descriptor`.
> 
> In recommender system, multi-task training brings better performane on targets with less data. Because parameters of the shared bottom can be fully trained by other related targets' samples.
> 
> Thus, in multi-task training of DeePMD, what's the constitution design of training data? Every sample has all target's labels, or we have multiple groups of samples while each group has partial targets' labels? @amcadmus

We want to provide full flexibility regarding the labels, i.e. we allow the use provide multiple data `System`s and each `System` may have different types of labels.

shishaochen:
@jxxiaoshaoye Are you still going to push this feature online? Otherwise, other community members could help.

examples/mix/input.json
```diff
@@ -0,0 +1,7957 @@
...
+			"start_pref_v":	0,
+			"limit_pref_v":	0,
+			"_comment":	" that's all"
+		}
+	
+    },
+
+    "training" : {
+	"training_data": {
+	    "systems":		{
```
shishaochen, line 128: We'd better provide sampled (or faked data) into repository. Then each example can be executed by user after this repository is cloned to local.

shishaochen (changes requested):

PR #882, opened 2021-07-23
Fix the empty neighbor distance array in neighbor_stat.py 

nicklin96:
atoms with no neighbors generate empty neighbor distance array causing np.min to crush.
add safe check and warning for atoms with no neighbors

### DIFF
--- a/deepmd/utils/neighbor_stat.py
+++ b/deepmd/utils/neighbor_stat.py
@@ -86,7 +86,7 @@ class NeighborStat():
                         dt = np.min(dt)              
                     else:
                         dt = self.rcut
-                        log.warn("Atoms with no neighbors found in %s. Please make sure it's what you expected."%jj)
+                        log.warning("Atoms with no neighbors found in %s. Please make sure it's what you expected."%jj)
                         
                     if dt < self.min_nbor_dist:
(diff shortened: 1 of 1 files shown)


deepmd/utils/neighbor_stat.py
```diff
@@ -82,7 +82,12 @@ def get_stat(self,
                                                 self.place_holders['box']: np.array(data_set['box'])[kk].reshape([-1, 9]),
                                                 self.place_holders['default_mesh']: np.array(data.default_mesh[ii]),
                                             })
-                    dt = np.min(dt)
+                    if dt.size != 0:
+                        dt = np.min(dt)              
+                    else:
+                        dt = self.rcut
+                        log.warn("Atoms with no neighbors found in %s. Please make sure it's what you expected."%jj)
```
njzjz, line 89: https://docs.python.org/3/library/logging.html#logging.Logger.warning

> As `warn` is deprecated, please do not use it - use `warning` instead.

nicklin96: Sorry! I have modified log.warn to log.warning

denghuilu (approved):

amcadmus (approved):

PR #888, opened 2021-07-26
fix build wheel: `policy.json` was renamed

njzjz:
`policy.json` was renamed to `manylinux-policy.json`. See pypa/auditwheel#311.


njzjz:
This commit has not been released... The error is due to pypa/manylinux#1143.

PR #911, opened 2021-08-01
fix 'NoneType' has no len() in `auto_sel`

njzjz:
The default value of `type_map` is `None`, so when you don't set `type_map`, you'll get this error.

https://github.com/deepmodeling/deepmd-kit/blob/043ac869bfcdc7f3a20aa24d04bb7c7b88abcc0b/deepmd/entrypoints/train.py#L225

### DIFF
--- a/deepmd/entrypoints/train.py
+++ b/deepmd/entrypoints/train.py
@@ -229,7 +229,7 @@ def get_sel(jdata, rcut):
     max_rcut = get_rcut(jdata)
     type_map = get_type_map(jdata)
 
-    if len(type_map) == 0:
+    if type_map and len(type_map) == 0:
         type_map = None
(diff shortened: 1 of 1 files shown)


njzjz:
By the way, can we let the auto-sel calculation become a default behavior, and throw a `RuntimeError` when it's more than those set by users? Usually we will get a bad model if `sel` is not enough for training data. It wastes time to train such a model.

amcadmus:
> By the way, can we let the auto-sel calculation become a default behavior, and throw a `RuntimeError` when it's more than those set by users? Usually we will get a bad model if `sel` is not enough for training data. It wastes time to train such a model.

We should allow the users setting a small `sel` and warn them that they are using a too small `sel` and they really know what they are doing.

amcadmus (approved):

PR #918, opened 2021-08-04
Passing error to TF instead of exit

### LINKED ISSUE #802: [BUG] dp as ASE calculator, no output until the whole script is finished
Hi, everyone, I am trying to use dp as ASE calculator to obtain energy and force for each input geometry. The dp force field has been trained under dp 1.2.0 in centos7 and performs well. 
Every time I submt the python script to put geometry into dp calculator iteratively, there is no output energy or force until all the geometries calculations are finished. However, normally, the whole cycle cannot be finished since the system memory soon ran out. Here is my code: 
def get_E(isomer):	
	isomer.set_cell(100 * np.identity(3))
	isomer.set_pbc((False, False, False))
	isomer.calc=DP(model="graph.pb")
	E=isomer.get_potential_energy()[0]
	return E

f=open('E_ML.txt','w')
for one_xyz in xyz:
	xyz_atoms_objs=ase.io.read(one_xyz,index=':')
	for isomer in xyz_atoms_objs:          
		E=get_E(isomer)
		f.write(str(E)+'\n')
f.close()
Any suggestions are sinerely welcomed!
Thanks.

njzjz:
This commit does three little things:
(1) create an exception called `deepmd::deepmd_exception` (based on `std::runtime_error`);
(2) throw this exception instead of `exit` or `std::runtime_error`;
(3) catch this exception in the op, and pass to TF using `OP_REQUIRES_OK`.
One more, the OOM error will raise ResourceExhausted, as the same as TF ops.

The benifit of doing so is that the TF side and Python side can processing other things, catch the error, and print the traceback.
This commit can also fix #802, where the Python didn't save the buffer to the file before exit.

The idea was inspired by TF: https://github.com/tensorflow/tensorflow/blob/5dcfc51118817f27fad5246812d83e5dccdc5f72/tensorflow/core/kernels/mkl/mkl_tfconv_op.h#L120-L126

### DIFF
--- a/source/api_cc/tests/CMakeLists.txt
+++ b/source/api_cc/tests/CMakeLists.txt
@@ -37,7 +37,7 @@ configure_file(
 set(opname "deepmd_op")
 set(OP_BASE_DIR ${CMAKE_SOURCE_DIR}/../../op)
 # file(GLOB OP_SRC ${OP_BASE_DIR}/*.cc)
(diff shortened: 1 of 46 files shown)


amcadmus (changes requested):
We should avoid copying the `try ... catch ...` in all OP implementations. 

Shall we define some function like 
```cc
#define void ComputeFunction(OpKernelContext* context);

void save_compute(OpKernelContext* context, const ComputeFunction &ff) {
  try{
    ff(context);
  } catch (...) {
    ....
  } catch (...) {
    ....
  }
}
```
where `ComputeFunction` is the implementation of the OP.

njzjz:
See https://github.com/deepmodeling/deepmd-kit/blob/9c8a0da4a91b8d060be312cafa82c379ccc8fc37/source/op/custom_op.cc#L4-L20 and https://github.com/deepmodeling/deepmd-kit/blob/9c8a0da4a91b8d060be312cafa82c379ccc8fc37/source/op/descrpt.cc#L52-L56

amcadmus:
> https://github.com/deepmodeling/deepmd-kit/blob/9c8a0da4a91b8d060be312cafa82c379ccc8fc37/source/op/custom_op.cc#L5-L5
> 

The name should have been `safe_compute`, sry my mistake

amcadmus (approved):

denghuilu (approved):

pkulzy (approved):
There is no problem on ROCm platform about UT and compiling

Issue #728, opened 2021-06-11
[Feature Request] Save the range of embedding matrix input in the DP model 


amcadmus:
**Summary**

Currently the model compression still need the training dataset to compute the range of the embedding matrix input, and to determine the range of tabulation. Saving the range of the embedding matrix input in the DP model will remove this dependency, so the user can still compress the model in the case that the original training data is not available.

njzjz:
We may raise RuntimeError if sel is enough during the training, if users use their settings, since the model will be bad. We may add a --force option if users insist on using the bad sel.

Issue #727, opened 2021-06-11
[Feature Request] Save the input script to DP model


amcadmus:
**Summary**

Better to save the setting provided in the training input script to the model, so a user can get all the training hyper parameter from the frozen model, i.e. ".pb" file. If it is implemented, the model compression will not need the original input script anymore. 

**Detailed Description**

One can serialize a dict to a string by using `pickle`, see [here](https://stackoverflow.com/questions/4342176/how-do-i-serialize-a-python-dictionary-into-a-string-and-then-back-to-a-diction). The string can be saved in the model.  

**Further Information, Files, and Links**

PR #921, opened 2021-08-04
Reconstruct the interface of model compression

denghuilu:
Before this pull request, the model compression module within the deepmd-kit would require the users to provide the training script as well as the training data to compress a given frozen model. Herein, this pull request brings a more useful and robust interface of model compression, which removes the dependences on training script and data from the model compression.

The main changes are:

- Save the training script #727 as well as the min_nbor_dist value #728  within the training process.
- Remove the dependences on training script and data from the model compression. Now the model compression command can be simply written as: ```dp compress -i original.pb -o compress.pb``` instead of ```dp compress input.json -i original.pb -o compress.pb```.
- Update the UTs for model compression.
- Update the docs of model compression.

The next pull request will add an item within the convert-from interface to allow the users to upgrade their existing frozen models to adapt the new interface of model compression.


deepmd/entrypoints/train.py
```diff
@@ -68,11 +71,15 @@ def train(
...
 
-    jdata = update_sel(jdata)
+    if is_compress == False:
+        jdata = update_sel(jdata)
 
     with open(output, "w") as fp:
         json.dump(jdata, fp, indent=4)
 
+    # save the training script into the graph
+    tf.constant(json.dumps(jdata), name='training_script', dtype=tf.string)
```
njzjz, line 81: Do we need to consider to use `zlib.compress` to save the compressed string? This can reduce the model size.

deepmd/common.py
```diff
@@ -483,3 +485,39 @@ def get_np_precision(precision: "_PRECISION") -> np.dtype:
...
+    GraphWithoutJdataOrMinNborDistError
+        If the training script or min_nbor_dist are within the frozen model
+    """
+    graph_def = tf.GraphDef()
+    with open(model_file, "rb") as f:
+        graph_def.ParseFromString(f.read())
+    with tf.Graph().as_default() as graph:
+        tf.import_graph_def(graph_def, name = "")
+        try:
+            jdata = graph.get_tensor_by_name('training_script:0')
```
shishaochen, line 515: Shall we collect these constant strings into an unified Python file like `constant.py`? Then:
- Name consistency between saver and reader can be ensured.
- Other guys may reference them in individual Python programs.

njzjz:
I have another question. What is the behavior if one uses another JSON to restart the training? For example, in active learning cycles, we can continue training from the last iteration.

denghuilu:
> I have another question. What is the behavior if one uses another JSON to restart the training? For example, in active learning cycles, we can continue training from the last iteration.

The model will save the latest json file.

deepmd/common.py
amcadmus: good idea

denghuilu: I see

amcadmus:
> I have another question. What is the behavior if one uses another JSON to restart the training? For example, in active learning cycles, we can continue training from the last iteration.

With this PR users do not need to provide training input script anymore.

amcadmus:
> I have another question. What is the behavior if one uses another JSON to restart the training? For example, in active learning cycles, we can continue training from the last iteration.

In the `init-model` mode, overwriting the existing training script should be the expected behavior.  

In the `restart` mode, the training script are expected to be the same.

deepmd/entrypoints/train.py
denghuilu: TensorFlow will automatically compress the tensors within the frozen model, so I don’t think we need to manually compress the json script.

deepmd/common.py
```diff
@@ -483,3 +485,39 @@ def get_np_precision(precision: "_PRECISION") -> np.dtype:
         return np.float64
     else:
         raise RuntimeError(f"{precision} is not a valid precision")
+
+def load_model_info(model_file: str) -> Tuple[dict, float]:
```
amcadmus, line 489: I would recommend split this function into two: one for loading json data with a given list of tensor names, and the other for loading any variable with a given list of tensor names. 

Both of them then can be reused in other cases where tensors with certain names are required to be evaluated.

deepmd/entrypoints/freeze.py
```diff
@@ -45,6 +45,8 @@ def _make_node_names(model_type: str, modifier_type: Optional[str] = None) -> Li
         "model_attr/tmap",
         "model_attr/model_type",
         "model_attr/model_version",
+        "min_nbor_dist",
+        "training_script",
```
amcadmus, line 49: I suggest the following changes:
`"min_nbor_dist"` -> `"train/min_nbor_dist"`
`"training_script"` -> `"train/training_script"`

deepmd/train/trainer.py
```diff
@@ -321,15 +329,22 @@ def _build_lr(self):
...
-            if data_dict[kk]['high_prec'] :
-                prec = GLOBAL_ENER_FLOAT_PRECISION
-            self.place_holders[kk] = tf.placeholder(prec, [None], name = 't_' + kk)
-            self.place_holders['find_'+kk] = tf.placeholder(tf.float32, name = 't_find_' + kk)
+        if self.is_compress :
+            for kk in ['coord', 'box']:
+                self.place_holders[kk] = tf.placeholder(GLOBAL_TF_FLOAT_PRECISION, [None], 't_' + kk)
+            for kk in ['energy', 'force', 'virial', 'atom_ener', 'atom_pref']:
+                self.place_holders[kk] = tf.placeholder(GLOBAL_TF_FLOAT_PRECISION, [None], 't_' + kk)
+                self.place_holders['find_' + kk] = tf.placeholder(tf.float32, [None], 't_' + kk)
```
amcadmus, line 337: `energy`, `force`, ... is not a complete list! You should check the `data_requirement` in `deepmd/common.py` for a complete list of required data. 

The place_holder builders for the standard and compress modes should be unified.

deepmd/utils/argcheck.py
```diff
@@ -332,14 +332,16 @@ def modifier_variant_type_args():
 
 #  --- model compression configurations: --- #
 def model_compression():
-    doc_compress = "The name of the frozen model file."
+    doc_compress = f"The name of the frozen model file."
     doc_model_file = f"The input model file, which will be compressed by the DeePMD-kit."
     doc_table_config = f"The arguments of model compression, including extrapolate(scale of model extrapolation), stride(uniform stride of tabulation's first and second table), and frequency(frequency of tabulation overflow check)."
+    doc_min_nbor_dist = f"The nearest distance between neighbor atoms saved in the frozen model."
```
amcadmus, line 338: Why asking users to provide this key?

deepmd/train/trainer.py
```diff
@@ -412,14 +427,16 @@ def _init_session(self):
...
 
         # if valid_data is None:  # no validation set specified.
         #     valid_data = train_data  # using training set as validation set.
 
         stop_batch = self.stop_batch
         self._init_session()
-
+        if self.is_compress:
+            self.saver.save (self.sess, os.getcwd() + "/" + self.save_ckpt)
+            return
```
amcadmus, line 439: `train` should not be called. 
we may provide a new method like `save_compressed`

deepmd/train/trainer.py
```diff
@@ -278,35 +281,40 @@ def _init_param(self, jdata):
...
-
     def build (self, 
-               data, 
+               data = None, 
                stop_batch = 0) :
         self.ntypes = self.model.get_ntypes()
-        # Usually, the type number of the model should be equal to that of the data
-        # However, nt_model > nt_data should be allowed, since users may only want to 
-        # train using a dataset that only have some of elements 
-        assert (self.ntypes >= data.get_ntypes()), "ntypes should match that found in data"
```
amcadmus, line 289: Why removing this part??

deepmd/train/trainer.py
```diff
@@ -278,35 +281,40 @@ def _init_param(self, jdata):
         else:
             self.valid_numb_batch = 1
 
-
```
amcadmus, line 281: Why removing this line?

amcadmus (changes requested):

deepmd/entrypoints/train.py
```diff
@@ -86,10 +93,10 @@ def train(
         log.info(message)
 
     run_opt.print_resource_summary()
-    _do_work(jdata, run_opt)
+    _do_work(jdata, run_opt, is_compress)
 
 
-def _do_work(jdata: Dict[str, Any], run_opt: RunOptions):
+def _do_work(jdata: Dict[str, Any], run_opt: RunOptions, is_compress: False):
```
njzjz, line 99:
```suggestion
def _do_work(jdata: Dict[str, Any], run_opt: RunOptions, is_compress: bool = False):
```

njzjz (changes requested):

deepmd/utils/argcheck.py
denghuilu: This key is not provided by the users, it is computed by the code

deepmd/train/trainer.py
```diff
@@ -321,15 +338,27 @@ def _build_lr(self):
...
+        else :
+            data_dict = data.get_data_dict()
+            for kk in data_dict.keys():
+                if kk == 'type':
+                    continue
+                prec = GLOBAL_TF_FLOAT_PRECISION
+                if data_dict[kk]['high_prec'] :
+                    prec = GLOBAL_ENER_FLOAT_PRECISION
+                self.place_holders[kk] = tf.placeholder(prec, [None], name = 't_' + kk)
+                self.place_holders['find_' + kk] = tf.placeholder(tf.float32, name = 't_find_' + kk)
```
amcadmus, line 358: DANGEROUS: copied code from line 345:352

deepmd/train/trainer.py
```diff
@@ -412,14 +442,12 @@ def _init_session(self):
...
 
-    def train (self, train_data, valid_data=None) :
+    def train (self, train_data = None, valid_data=None) :
 
         # if valid_data is None:  # no validation set specified.
         #     valid_data = train_data  # using training set as validation set.
 
         stop_batch = self.stop_batch
-        self._init_session()
-
```
amcadmus, line 422: Why moving `_init_session` to `_build_training`?

amcadmus (changes requested):

deepmd/train/trainer.py
denghuilu: We need self.saver within the _init_session to store the checkpoint, in the old implementation we must call trainer.train to init the self.save

deepmd/train/trainer.py
denghuilu: Any idea to get the place_holders without accessing to the training data?

deepmd/train/trainer.py
```diff
@@ -2,6 +2,7 @@
 import logging
 import os
 import time
+import json
```
njzjz, line 5: I don't see it's used in this file?

deepmd/train/trainer.py
```diff
@@ -617,3 +643,10 @@ def get_evaluation_results(self, batch_list):
                     sum_results[k] = sum_results.get(k, 0.) + v * results["natoms"]
         avg_results = {k: v / sum_natoms for k, v in sum_results.items() if not k == "natoms"}
         return avg_results
+    
+    def save_compressed(self):
+        """
+        Save the compressed graph
+        """
+        if self.is_compress:
+            self.saver.save (self.sess, os.getcwd() + "/" + self.save_ckpt)
```
njzjz, line 652: Use `os.path.json` instead of hard-code `/`, to prevent that someday we suddenly want to support windows.

deepmd/utils/argcheck.py
njzjz: I don't think it's a correct way to store `min_nbor_dist`... Here the Argument will also affect the document. Can we cache it in another place?

njzjz (commented):
Also, please fix lint warnings.

deepmd/train/trainer.py
```diff
@@ -280,33 +283,44 @@ def _init_param(self, jdata):
...
         else:
             log.info("training without frame parameter")
 
-        self.type_map = data.get_type_map()
+        # self.type_map = data.get_type_map()
+        if self.is_compress == False:
+            # Usually, the type number of the model should be equal to that of the data
+            # However, nt_model > nt_data should be allowed, since users may only want to 
+            # train using a dataset that only have some of elements 
+            assert (self.ntypes >= data.get_ntypes()), "ntypes should match that found in data"
```
njzjz, line 308: Isn't it duplicated with line 289-293?

deepmd/train/trainer.py
amcadmus: Write a method for generating place_holders

deepmd/train/trainer.py
amcadmus: It doesn't make sense to init session in build training. 

one may call it in _do_works.

denghuilu: I'll move it back

njzjz:
https://github.com/deepmodeling/deepmd-kit/blob/425a896abb66164139cd4334be34bf1df379e33f/deepmd/train/trainer.py#L288-L307

Duplicated lines

deepmd/utils/argcheck.py
```diff
@@ -332,14 +332,16 @@ def modifier_variant_type_args():
...
     
     return [
         Argument("compress", bool, optional = False, default = True, doc = doc_compress),
         Argument("model_file", str, optional = False, default = 'frozen_model.pb', doc = doc_model_file),
         Argument("table_config", list, optional = False, default = [5, 0.01, 0.1, -1], doc = doc_table_config),
+        Argument("min_nbor_dist", float, optional = False, default = 0.0, doc = doc_min_nbor_dist),
```
njzjz, line 344: In these four lines, why `optional` is set to `False` but has `default` value?

amcadmus (approved):

iProzd (approved):

njzjz (approved):

PR #924, opened 2021-08-05
set input DeepmdData.type_map to input type_map

### LINKED ISSUE #920: Type map issue in DeepmdData
**type map may be wrong in __init__ of Deepmddata**

In the __init__ function of Deepmddata class, it read type.raw and cast it from local type map (which is read from a system's type_map.raw) to the global type map which is defined in the input.json file. The code is below:
image

However, for example, in the Al Cu Mg alloy case, the global type map is [Mg, Al, Cu]. If a system only have one elements Al,  and the local type_map.raw is [Al]. After process of __init__, self.type_map would be [Mg], which is incorrect logically.

my deepmd-kit version is v2.0.0.b3, and I print out the type map before and after __init__ cast to identify want I'm trying to say.
image

amcadmus:
This fixes #920

### DIFF
--- a/deepmd/utils/data.py
+++ b/deepmd/utils/data.py
@@ -57,8 +57,7 @@ class DeepmdData() :
         if type_map is not None and self.type_map is not None:
             atom_type_ = [type_map.index(self.type_map[ii]) for ii in self.atom_type]
             self.atom_type = np.array(atom_type_, dtype = np.int32)
-            ntypes = len(self.type_map)
-            self.type_map = type_map[:ntypes]
+            self.type_map = type_map
         # make idx map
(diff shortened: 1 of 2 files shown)


njzjz (approved):

njzjz:
@tuoping: It seems that our document does not mention that the `type_map` in all systems should be the same.

amcadmus:
> @tuoping: It seems that our document does not mention that the `type_map` in all systems should be the same.

We do not need to set the `type_map` in all data systems the same.

njzjz:
Got it. I just understood the logic.

jxxiaoshaoye (approved):

PR #929, opened 2021-08-06
add multi task

yhliu918:


njzjz:
Please merge with devel.

njzjz:
err... We also need to update the target branch.

deepmd/entrypoints/main.py
```diff
@@ -162,6 +162,13 @@ def parse_args(args: Optional[List[str]] = None):
         default="out.json",
         help="The output file of the parameters used in training.",
     )
+    parser_train.add_argument(
+        "-mu",
```
shishaochen, line 166: Maybe short name `mt` is better?

deepmd/entrypoints/train.py
```diff
@@ -15,9 +15,11 @@
 from deepmd.infer.data_modifier import DipoleChargeModifier
 from deepmd.train.run_options import BUILD, CITATION, WELCOME, RunOptions
 from deepmd.train.trainer import DPTrainer
+from deepmd.train.trainer_mt import DPTrainer_mt
 from deepmd.utils.argcheck import normalize
+from deepmd.utils.argcheck_mt import normalize_mt
 from deepmd.utils.compat import updata_deepmd_input
-from deepmd.utils.data_system import DeepmdDataSystem
+from deepmd.utils.data_system import DeepmdDataSystem,DeepmdDataDocker
```
shishaochen, line 22: A space should be added between characters.

deepmd/entrypoints/train.py
```diff
@@ -112,9 +125,17 @@ def _do_work(jdata: Dict[str, Any], run_opt: RunOptions):
...
-    model = DPTrainer(jdata, run_opt=run_opt)
-    rcut = model.model.get_rcut()
-    type_map = model.model.get_type_map()
+    if not run_opt.multi_task:
+        model = DPTrainer(jdata, run_opt=run_opt)
+        rcut = model.model.get_rcut()
+        type_map = model.model.get_type_map()
+    else:
+        model = DPTrainer_mt(jdata, run_opt=run_opt)
+        for model_name in model.model_dict.keys():
```
shishaochen, line 134: Maybe we need some check on `rcut` and `type_map` among sub models. Otherwise, this code segment will only use those of the last sub mdoel.

deepmd/entrypoints/train.py
```diff
@@ -191,6 +220,36 @@ def get_data(jdata: Dict[str, Any], rcut, type_map, modifier):
 
     return data
 
+def get_data_mt(jdata: Dict[str, Any], rcut, type_map, modifier):
+    systems = j_must_have(jdata, "systems")
+    batch_size = j_must_have(jdata, "batch_size")
+    sys_probs = jdata.get("sys_probs", None)
+    auto_prob = jdata.get("auto_prob", "prob_sys_size")
+    total_data = []
+    for sub_sys in systems:
```
shishaochen, line 229: Shall we move the `for` iteration into the constructor of `DeepmdDataDocker`? Herein, it can has similar constructor interfaces with `DeepmdDataSystem` where `object-oriented programming` is obeyed.

deepmd/entrypoints/main.py
njzjz: I've never see the usage of `-` along with two letters. Does it really work? `-m` or `--mu` is more common.

deepmd/utils/data.py
```diff
@@ -153,11 +157,11 @@ def reduce(self,
         }
         return self
 
-    def get_data_dict(self) -> dict:
+    def get_data_dict(self):
         """
         Get the `data_dict`
         """
-        return self.data_dict
+        return self.data_dict,self.name
```
shishaochen, line 164: A space is needed.

deepmd/utils/data.py
```diff
@@ -153,11 +157,11 @@ def reduce(self,
         }
         return self
 
-    def get_data_dict(self) -> dict:
+    def get_data_dict(self):
```
shishaochen, line 160: Why remove type declaration of return values?

deepmd/train/trainer.py
```diff
@@ -321,7 +321,7 @@ def _build_lr(self):
 
     def _build_network(self, data):        
         self.place_holders = {}
-        data_dict = data.get_data_dict()
+        data_dict,data_name = data.get_data_dict()
```
shishaochen, line 324: A space is needed.

deepmd/entrypoints/main.py
shishaochen: Yes, it works but lacking some features of the single-character short option and not a good practice.

deepmd/train/trainer_mt.py
```diff
@@ -0,0 +1,713 @@
...
+
+# load grad of force module
+import deepmd.op
+
+from deepmd.common import j_must_have, ClassArg
+
+log = logging.getLogger(__name__)
+
+
+def _is_subdir(path, directory):
```
shishaochen, line 40: We'd better import it from `trainer.py` instead of writing a duplicated one.

deepmd/train/trainer_mt.py
```diff
@@ -0,0 +1,713 @@
...
+    path = os.path.realpath(path)
+    directory = os.path.realpath(directory)
+    if path == directory:
+        return False
+    relative = os.path.relpath(path, directory) + os.sep
+    return not relative.startswith(os.pardir + os.sep)
+
+def _generate_descrpt_from_param_dict(descrpt_param):
```
shishaochen, line 48: We'd better import `_generate_descrpt_from_param_dict` from `trainer.py` instead of writing a duplicated one.

deepmd/train/trainer_mt.py
```diff
@@ -0,0 +1,713 @@
...
+            sub_loss = self.loss_dict[loss_name]
+            optimizer = tf.train.AdamOptimizer(learning_rate = self.learning_rate_dict[loss_name])
+
+            if self.run_opt.is_distrib :
+                optimizer = tf.train.SyncReplicasOptimizer(
+                    optimizer,
+                    replicas_to_aggregate = self.run_opt.cluster_spec.num_tasks("worker"),
+                    total_num_replicas = self.run_opt.cluster_spec.num_tasks("worker"),
+                    name = "sync_replicas")
+                self.sync_replicas_hook = optimizer.make_session_run_hook(self.run_opt.is_chief)
```
shishaochen, line 434: Error is inevitable if distributed training is enabled. There should be only one `SyncReplicasOptimizer` and corresponding `SessionRunHook`.
Since we changed the parallel training implementation in the `devel` branch, you can actually remove related codes.

deepmd/train/trainer_mt.py
```diff
@@ -0,0 +1,713 @@
...
+
+            if self.run_opt.is_distrib :
+                optimizer = tf.train.SyncReplicasOptimizer(
+                    optimizer,
+                    replicas_to_aggregate = self.run_opt.cluster_spec.num_tasks("worker"),
+                    total_num_replicas = self.run_opt.cluster_spec.num_tasks("worker"),
+                    name = "sync_replicas")
+                self.sync_replicas_hook = optimizer.make_session_run_hook(self.run_opt.is_chief)   
+            self.optimizer_dict[loss_name] = optimizer  
+        grads = tf.gradients(self.l2_l_dict[data_name], trainable_variables)
```
shishaochen, line 436: The `data_name` comes from the first sampled batch and will be frozen into the TensorFlow graph.
Thus, each step it will only execute the sub graph of the loss minimizing defined on `data_name`.
In other words, either failures of session runs occur or only 1 target is trained on unexpected system frames.

deepmd/train/trainer_mt.py
jxxiaoshaoye: sure, I'll leave a not implemented runtimeError for distributed training option

deepmd/entrypoints/train.py
jxxiaoshaoye: okay, I'll get it fixed

deepmd/entrypoints/train.py
jxxiaoshaoye: type_map is consistent now, however rcut may vary between different descriptors, so I guess I'll put a list through building the data docker

deepmd/utils/data_system.py
```diff
@@ -509,6 +518,121 @@ def _prob_sys_size_ext(self, keywords):
...
+            self.pick_idx = method_idx
+        s_data = self.data_systems[self.pick_idx]
+        b_data = s_data.get_batch(sys_idx)
+        
+        return b_data
+
+    def get_data_system(self,name):
+        for i in range(self.nmethod):
+            if self.name_list[i] == name:
+                return self.data_systems[i]
```
njzjz, line 576: Here is a more Pythonic-style code:
```suggestion
        for iname, idata_system in zip(self.name_list, self.data_systems):
            if iname == name:
                return idata_system

```

deepmd/utils/argcheck_mt.py
```diff
@@ -0,0 +1,727 @@
...
+        Argument("activation_function", str, optional=True, default='tanh', doc=doc_activation_function),
+        Argument("resnet_dt", bool, optional=True, default=False, doc=doc_resnet_dt),
+        Argument("precision", str, optional=True, default="float64", doc=doc_precision),
+        Argument("trainable", bool, optional=True, default=True, doc=doc_trainable),
+        Argument("seed", [int, None], optional=True, doc=doc_seed),
+    ]
+
+
+#  --- Descriptor configurations: --- #
+def descrpt_local_frame_args():
```
njzjz, line 41: If many methods are as the same as `argcheck.py`, please import methods from argcheck instead of repeating writing them.
For example,
```py
from deepmd.utils.argcheck import descrpt_local_frame_args
```

deepmd/train/trainer_mt.py
```diff
@@ -0,0 +1,713 @@
...
+        descrpt = DescrptSeAEf(**descrpt_param)
+    elif descrpt_type == 'se_ar' :
+        descrpt = DescrptSeAR(descrpt_param)
+    else :
+        raise RuntimeError('unknow model type ' + descrpt_type)
+    return descrpt
+    
+
+
+class DPTrainer_mt (object):
```
njzjz, line 80: I think this class can inherit `DPTrainer` and remove duplicate methods.

deepmd/entrypoints/train.py
```diff
@@ -130,13 +151,22 @@ def _do_work(jdata: Dict[str, Any], run_opt: RunOptions):
...
+            valid_data = None
     else:
-        valid_data = None
+        train_data = get_data(jdata["training"]["training_data"], rcut, ipt_type_map, modifier)
+        train_data.print_summary("training")
+        if jdata["training"].get("validation_data", None) is not None:
+            valid_data = get_data(jdata["training"]["validation_data"], rcut, ipt_type_map, modifier)
+            valid_data.print_summary("validation")
+        else:
+            valid_data = None
```
njzjz, line 169: Can be written as below to avoid duplicated codes
```py
    if run_opt.multi_task:
        get_data_func = get_data_mt
    else:
        get_data_func = get_data
    train_data = get_data_func(jdata["training"]["training_data"], rcut, ipt_type_map, modifier)
    train_data.print_summary("training")
    if jdata["training"].get("validation_data", None) is not None:
        valid_data = get_data_func(jdata["training"]["validation_data"], rcut, ipt_type_map, modifier)
        valid_data.print_summary("validation")
    else:
        valid_data = None
```

deepmd/utils/argcheck_mt.py
njzjz: Can you clarify what's the difference between `argcheck.py` and `argcheck_mt.py`? I think descriptors should be the same. Do we need to create a new file?

deepmd/utils/data_system.py
```diff
@@ -29,7 +30,8 @@ def __init__ (self,
                   modifier = None,
                   trn_all_set = False,
                   sys_probs = None,
-                  auto_prob_style ="prob_sys_size") :
+                  auto_prob_style ="prob_sys_size",
+                  name : str = None,) :
```
njzjz, line 34: Document needs to be added.

deepmd/utils/data_system.py
```diff
@@ -509,6 +519,208 @@ def _prob_sys_size_ext(self, keywords):
...
+    It is implemented with the help of DeepmdData
+    """
+    def __init__ (self,
+                  data_systems,
+                  batch_size : int,
+                  rcut : int,
+                  type_map : List[str] = None,
+                  sys_probs = None,
+                  auto_prob_style ="prob_sys_size",
+                  modifier = None,
```
njzjz, line 534: these three parameters don't have documents

njzjz (commented):
Not all previous comments are resolved.

deepmd/utils/data_system.py
jxxiaoshaoye: When doing argcheck, all fitting or descriptor have another feature 'name', and the multi-task input.json have different structure, so we add a new argcheck_mt.py

jxxiaoshaoye: We clarify each component in the model and combining them together in multi-task process, so the input.json file is organized by this logic

njzjz: @jxxiaoshaoye Got it. I suggest to write something like
```py
from deepmd.utils.argcheck import descrpt_local_frame_args

def descrpt_local_frame_args_mt():
    return [
        Argument("name", str, optional=True, default='', doc="Descriptor name."),
        *descrpt_local_frame_args(),
     ]
```
Besides, if every descriptor should have a name and it's not related to a specific descriptor, I think you can put it here?
https://github.com/deepmodeling/deepmd-kit/blob/3c09f770fbe97318759d807c68842755ad187e6d/deepmd/utils/argcheck_mt.py#L691
```py
    base = Argument("base", dict, [
        Argument("name", str, optional=True, default='', doc="Descriptor name."),
    ], [descrpt_variant_type_args()], doc="")
```

njzjz: If you need help, I may send a PR in one or two days.

jxxiaoshaoye: Ok, I think I've got your points

deepmd/train/trainer_mt.py
njzjz: And I still keep this advice: this class can inherit `DPTrainer` and remove duplicate methods. Or you can consider to create a new class `DPTrainerBase` to merge the same methods.

amcadmus:
It is definitely not a good practice to write commit messages like "add multi-task v4". There are many instructions on how to write good commit message [for example](https://www.freecodecamp.org/news/writing-good-commit-messages-a-practical-guide/).

Please describe what you have done in the description of this PR. Good examples are #945 #921  Please follow them to update your PR. Thanks!

deepmd/utils/argcheck_mt.py
```diff
@@ -0,0 +1,273 @@
...
+    link_se_e3 = make_link('se_e3', 'model/descriptor[se_e3]')
+    link_se_a_tpe = make_link('se_a_tpe', 'model/descriptor[se_a_tpe]')
+    link_hybrid = make_link('hybrid', 'model/descriptor[hybrid]')
+    doc_descrpt_type = f'The type of the descritpor. See explanation below. \n\n\
+- `loc_frame`: Defines a local frame at each atom, and the compute the descriptor as local coordinates under this frame.\n\n\
+- `se_e2_a`: Used by the smooth edition of Deep Potential. The full relative coordinates are used to construct the descriptor.\n\n\
+- `se_e2_r`: Used by the smooth edition of Deep Potential. Only the distance between atoms is used to construct the descriptor.\n\n\
+- `se_e3`: Used by the smooth edition of Deep Potential. The full relative coordinates are used to construct the descriptor. Three-body embedding will be used by this descriptor.\n\n\
+- `se_a_tpe`: Used by the smooth edition of Deep Potential. The full relative coordinates are used to construct the descriptor. Type embedding will be used by this descriptor.\n\n\
+- `hybrid`: Concatenate of a list of descriptors as a new descriptor.'
```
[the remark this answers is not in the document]
jxxiaoshaoye: I think I can deal with the argcheck problem.
As for inheriting DPTrainer, I definitely think there's a must to do it. However, there are not much duplicate method we can remove, because each build and training process is modified and DPTrainer doesn't split some big function into small ones, which means I can't apply them, instead I need to go through like the whole _init_param to change some details

amcadmus (changes requested):
1. other `fit` and `loss` (deep tensor) should be added with attribute `name`
2. I am not sure if the data `source/tests/multi-task/data/` is necessary for your test cases, please check.

deepmd/train/run_options.py
```diff
@@ -90,7 +90,9 @@ def __init__(
         restart: Optional[str] = None,
         log_path: Optional[str] = None,
         log_level: int = 0,
-        mpi_log: str = "master"
+        mpi_log: str = "master",
+        try_distrib: bool = False,
```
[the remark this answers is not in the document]
jxxiaoshaoye: This is not added by me, I merged the devel (maybe) and there are some inconsistence here in run options

deepmd/model/ener.py
```diff
@@ -95,7 +101,7 @@ def _compute_input_stat (self, all_stat, protection = 1e-2) :
         self.descrpt.compute_input_stats(all_stat['coord'],
                                          all_stat['box'],
                                          all_stat['type'],
-                                         all_stat['natoms_vec'],
+                                         all_stat['natoms_vec'], # this should be the global one(dim 14) or local one(dim 11)?
```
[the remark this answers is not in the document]
jxxiaoshaoye: I'm thinking about how to organize data more reasonably, because now we are using global type map everywhere, and each descriptor deal with its own atom type

deepmd/train/run_options.py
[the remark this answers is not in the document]
amcadmus: I see that may because that you didn't solve the conflict correctly.

deepmd/entrypoints/train.py
```diff
@@ -112,9 +125,19 @@ def _do_work(jdata: Dict[str, Any], run_opt: RunOptions):
...
+    if not run_opt.multi_task:
+        model = DPTrainer(jdata, run_opt=run_opt)
+        rcut = model.model.get_rcut()
+        type_map = model.model.get_type_map()
+    else:
+        model = DPMultitaskTrainer(jdata, run_opt=run_opt)
+        for model_name in model.model_dict.keys():
+            sub_model = model.model_dict[model_name]
+            rcut_list.append(sub_model.get_rcut())
+            type_map = sub_model.get_type_map()
```
shishaochen, line 138: Thus, we only use the `type_map` of the last data system? Or it is ensured `type_map` among data systems are the same.

jxxiaoshaoye: they are the same for now

examples/mix/input.json
```diff
@@ -0,0 +1,170 @@
...
+    "model": {
+	"type_map":	["H","O","Hf"],
+	"type_embedding":{
+	    "neuron":		[2, 4, 8],
+	    "resnet_dt":	false,
+	    "seed":		1
+	},	    
+	"descriptor" :[
+		{
+	    	"type":		"se_e2_a",
```
shishaochen, line 12: Please take care of indents.

deepmd/train/trainer_mt.py
```diff
@@ -0,0 +1,488 @@
...
+        valid_results = self.get_evaluation_results(valid_batches,method)
+
+        cur_batch = self.cur_batch
+        current_lr = self.sess.run(self.learning_rate_dict[self.method_name_list[method]])
+        if print_header:
+            print_str = DPTrainer.print_header(fp, train_results, valid_results)
+            print_str += "  %6s" % 'method'
+            fp.write(print_str+'\n')
+            fp.flush()
+        print_str = DPTrainer.print_on_training(fp, train_results, valid_results, cur_batch, current_lr)
```
shishaochen, line 463: 不打印 HEAD 吗？不如，最后一列的 LR 会忽上忽下，看不出来属于谁。

deepmd/train/trainer.py
```diff
@@ -551,6 +575,7 @@ def get_global_step(self):
     #         fp.write(print_str)
     #         fp.close ()
 
+
     def valid_on_the_fly(self,
                          fp,
```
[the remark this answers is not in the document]
jxxiaoshaoye: I check out the devel branch, but fp is not deprecated?

jxxiaoshaoye:
Sorry, to call the multi task training, it is still using 'dp train -mt True input.json', however train and train_mt is separated

amcadmus:
> 
> 
> Sorry, to call the multi task training, it is still using 'dp train -mt True input.json', however train and train_mt is separated

Could you please remove the support of -mt option is multi-task is not supported in `dp train`?

deepmd/entrypoints/main.py
```diff
@@ -162,6 +163,14 @@ def parse_args(args: Optional[List[str]] = None):
         default="out.json",
         help="The output file of the parameters used in training.",
     )
+    parser_train.add_argument(
+        "-mt",
+        "--multi_task",
+        type=bool,
+        default=False,
```
njzjz, line 170: I suggest to use [`action="store_true"`](https://docs.python.org/3/library/argparse.html#action) instead.
```suggestion
        action="store_true",
```

deepmd/entrypoints/train.py
```diff
@@ -15,9 +15,12 @@
 from deepmd.infer.data_modifier import DipoleChargeModifier
 from deepmd.train.run_options import BUILD, CITATION, WELCOME, RunOptions
 from deepmd.train.trainer import DPTrainer
+from deepmd.train.trainer_mt import DPMultitaskTrainer
 from deepmd.utils.argcheck import normalize
+from deepmd.utils.argcheck_mt import normalize_mt
 from deepmd.utils.compat import updata_deepmd_input
 from deepmd.utils.data_system import DeepmdDataSystem
+from deepmd.utils.data_docker import DeepmdDataDocker
```
amcadmus, line 23: These imports should not be necessary

deepmd/entrypoints/train.py
```diff
@@ -210,28 +214,34 @@ def get_modifier(modi_data=None):
...
         for ii in descrpt_data['list']:
             rcut_list.append(ii['rcut'])
     else:
         rcut_list.append(descrpt_data['rcut'])
+    return rcut_list
+
+def get_rcut(jdata):
+    descrpt_data = jdata['model']['descriptor']
+    rcut_list = []
+    rcut_list.extend(parse_rcut(descrpt_data))
```
amcadmus, line 229: I do not think you need to revise the `get_rcut`.

deepmd/model/ener.py
```diff
@@ -95,7 +101,7 @@ def _compute_input_stat (self, all_stat, protection = 1e-2) :
         self.descrpt.compute_input_stats(all_stat['coord'],
                                          all_stat['box'],
                                          all_stat['type'],
-                                         all_stat['natoms_vec'],
+                                         all_stat['natoms_vec'],
```
amcadmus, line 104: plz avoid introduce such revision.

deepmd/utils/data_system.py
```diff
@@ -333,7 +341,7 @@ def get_batch(self, sys_idx : int = None):
         b_data = self.data_systems[self.pick_idx].get_batch(self.batch_size[self.pick_idx])
         b_data["natoms_vec"] = self.natoms_vec[self.pick_idx]
         b_data["default_mesh"] = self.default_mesh[self.pick_idx]
-        return b_data
+        return  b_data
```
amcadmus, line 344: a void such change

deepmd/utils/data_system.py
```diff
@@ -508,8 +519,6 @@ def _prob_sys_size_ext(self, keywords):
             sys_probs[block_stt[ii]:block_end[ii]] = tmp_prob * block_probs[ii]
         return sys_probs
 
-
-
```
amcadmus, line 512: avoid such change

amcadmus (changes requested):
plz revert all changes on adding/removing blank lines.

deepmd/entrypoints/train.py
yhliu918: if I don't split get_rcut into two function, I can't reuse it in the train_mt

deepmd/model/ener.py
```diff
@@ -111,10 +117,15 @@ def build (self,
                box, 
                mesh,
                input_dict,
-               suffix = '', 
+               suffix, # a dict of suffix, including type_embed, descrpt and fitting.
```
amcadmus, line 120: I think we should preserve the default value `''`

source/tests/test_multitask.py
```diff
@@ -0,0 +1,52 @@
...
+        assert(ret == 0), "DP train error!"
+        dd = np.loadtxt("lcurve.out",skiprows=1)[:,:9]
+        dd = dd.reshape([3,-1])
+        ref_loss = [24.8,19.6,
+                    3.79,1.81,
+                    2110.0,2110.0]
+
+        for ii in range(3):
+            for jj in range(2):
+                self.assertAlmostEqual(dd[ii][jj+1], ref_loss[ii*2+jj], places = 8)
```
amcadmus, line 52: why only compare the total loss?

amcadmus (changes requested):

amcadmus (approved):

PR #945, opened 2021-08-10
add LAMMPS plugin

njzjz:
- <del>Breaking changes:</del> update fix_dplr to support lammps/lammps#2560. <del>Old version may be not supported. No idea how to support both versions.</del> Both versions will be supported, if one provides either `LAMMPS_VERSION_NUMBER`, `LAMMPS_VERSION`, or `LAMMPS_SOURCE_ROOT`. If none is provided, default is 29Oct2020.
- Feature: support LAMMPS's new plugin package. See document for details.
- update document and examples for LAMMPS plugin. The old installation method will be still supported, although it's removed from documents.
- `OLD_LMP_PPPM` is replaced by `LAMMPS_VERSION_NUMBER`. (we can add back support for version<29Oct2020 removed in #302?)
- LAMMPS hasn't released a stable version, but I think it will be released in one or two months.


njzjz:
I have an idea to let one assign the LAMMPS version when using CMake if using the traditional version. (or automatically detect using new method) Then I can add the compatibility back.

njzjz:
Note: I've changed actions, so `Expected — Waiting for status to be reported` will appear util the setting changes.

amcadmus (approved):

doc/getting-started.md
```diff
@@ -414,6 +414,12 @@ Note that the model for MD simulations is required to be compatible with the Dee
 
 ### Run MD with LAMMPS
 
+Firstly, enable DeePMD-kit package in LAMMPS with `plugin` command:
+
+```
+plugin load path/to/deepmd/lib/libdeepmd_lmp.so
+```
+
```
amcadmus, line 422: It seems that the built-in mode is not supported by this PR. Then it is a breaking change, because all lammps input scripts have to be changed. 

It is ideal to support both built-in and plugin modes. We still need the built-in mode because it is easy to use, one do not need to load the shared lib in every lammps script.

amcadmus (changes requested):

doc/getting-started.md
njzjz: Ok. Note that I just removed built-in mode in the document, and I'll revert them.

source/install/build_cc.sh
```diff
@@ -20,7 +20,7 @@ NPROC=$(nproc --all)
 BUILD_TMP_DIR=${SCRIPT_PATH}/../build
 mkdir -p ${BUILD_TMP_DIR}
 cd ${BUILD_TMP_DIR}
-cmake -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX} -DINSTALL_TENSORFLOW=TRUE ${CUDA_ARGS} ..
+cmake -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX} -D LAMMPS_INSTALL_RPATH=ON -DINSTALL_TENSORFLOW=TRUE ${CUDA_ARGS} -DLAMMPS_VERSION=patch_30Jul2021 -DUSE_TTM=TRUE ..
```
amcadmus, line 23: `-D LAMMPS_INSTALL_RPATH=ON` dose not have any effect on the deepmd-kit code, am I getting it wrong?

njzjz: Err... I put it in the wrong place. It should be moved to LAMMPS cmake arguments.

njzjz:
```suggestion
cmake -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX} -DINSTALL_TENSORFLOW=TRUE ${CUDA_ARGS} -DLAMMPS_VERSION=patch_30Jul2021 -DUSE_TTM=TRUE ..
```

amcadmus (approved):

PR #966, opened 2021-08-12
fix lint errors and warnings

njzjz:


njzjz:
The import sort in VS Code is inconsistent with this lint...

PR #968, opened 2021-08-13
add code comment for InputNlist struct

denghuilu:


njzjz:
I forgot to mention.. Use the doxygen style introduced here https://wiki.rdkcentral.com/plugins/servlet/mobile?contentId=20807878#content/view/20807878

denghuilu:
@njzjz Any suggestion to solve this problem?
image

denghuilu:
> I forgot to mention.. Use the doxygen style introduced here https://wiki.rdkcentral.com/plugins/servlet/mobile?contentId=20807878#content/view/20807878

OK

njzjz:
https://github.com/deepmodeling/deepmd-kit/blob/ee0f6e6aa789e06a297d5e6f76000356b808474e/.github/workflows/test_python.yml#L57-L64

There should be a wrong cache of pypi index. However, GitHub provides no way to clean the cache.

njzjz:
See also https://pip.pypa.io/en/stable/topics/caching/#http-responses

njzjz (approved):

njzjz:
Finally, someone has the same cache issue: matplotlib/matplotlib#20834

PR #971, opened 2021-08-14
Remove pip cache as we need to flush

njzjz:
This reverts commit 6958e1a603894ea88a1bbb96e553c96356552097.


njzjz:
Not work, though.

PR #1003, opened 2021-08-21
fill docstring with templates

njzjz:
This commit uses `doq` to fill the docstring:
```sh
pip install doq
for ii in deepmd/*/*.py; do doq -f $ii -w --formatter=numpy; done
```

It just fill with the templates without any real documents. However, it
will be much easier to add documents to those that haven't had docstring
yet.


njzjz:
The automatic program has some issues which need to manually adjust...

PR #1009, opened 2021-08-21
add `InputNlist` into API doc

njzjz:
This struct is necessary for C++ APIs. Without document, no one can understand this struct...

@Yi-FanLi I believe you need it.

Preview below:

image

### DIFF
--- a/source/api_cc/include/common.h
+++ b/source/api_cc/include/common.h
@@ -32,9 +32,13 @@ typedef double ENERGYTYPE;
 
 struct NeighborListData 
 {
+  /// Number of core region atoms
   std::vector<int > ilist;
+  /// Array stores the core region atom's index
   std::vector<std::vector<int> > jlist;
+  /// Array stores the core region atom's neighbor atom number
   std::vector<int > numneigh;
+  /// Array stores the core region atom's neighbor index
   std::vector<int* > firstneigh;  
 public:
   void copy_from_nlist(const InputNlist & inlist);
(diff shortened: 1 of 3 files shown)


source/api_cc/include/common.h
```diff
@@ -32,9 +32,13 @@ typedef double ENERGYTYPE;
 
 struct NeighborListData 
 {
+  /// Number of core region atoms
   std::vector<int > ilist;
+  /// Array stores the core region atom's index
   std::vector<std::vector<int> > jlist;
+  /// Array stores the core region atom's neighbor atom number
   std::vector<int > numneigh;
+  /// Array stores the core region atom's neighbor index
```
amcadmus, line 41: this comment is for `numneigh`. Place it above `std::vector<int > numneigh;`? 

The dsame issue for `ilist` and `jlist`

njzjz: In #968, I see @denghuilu add the same comment to `InputNlist::firstneigh`? Isn't it the same thing as `NeighborListData::firstneigh`? Did I understand wrongly?

amcadmus: @denghuilu @njzjz  `jlist` stores the indexes of all the neighbors. `firstneigh[i]` stores the location of the first neighbor of `i`. So the neighbors of `i` are stored in `jlist[firstneigh[i] : firstneigh[i] + numneigh[i]`

@denghuilu Please check again you code you really use the neighbor list correctly....

denghuilu: @amcadmus I think you are right, so the NeighborListData is used to construct the InputNlist. From my point of view, firstneigh is a two-dimensional array that stores neighbor atom's index, even if it only refers to a series of jlist pointers. I'm using the InputNlist in the custom op, so I think it's fine.

amcadmus (approved):

denghuilu (approved):

PR #1017, opened 2021-08-23
add docstring to `DescrptSeA`

njzjz:
Contains other cleanups.

image

image

### DIFF
--- a/deepmd/descriptor/se_a.py
+++ b/deepmd/descriptor/se_a.py
@@ -16,23 +16,58 @@ from deepmd.utils.sess import run_sess
 from deepmd.utils.graph import load_graph_def, get_tensor_by_name_from_graph
 
 class DescrptSeA ():
-    """DeepPot-SE constructed from all information (both angular and radial) of
-    atomic configurations.
-    
-    The embedding takes the distance between atoms as input.
+    r"""DeepPot-SE constructed from all information (both angular and radial) of
+    atomic configurations. The embedding takes the distance between atoms as input.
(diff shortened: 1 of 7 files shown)


deepmd/descriptor/se_a.py
```diff
@@ -16,23 +16,58 @@
...
+    the relative coordinate and :math:`r_{ji}=\lVert \mathbf{R}_{ji} \lVert` is its norm.
+    The switching function :math:`s(r)` is defined as:
+
+    .. math::
+        s(r)=
+        \begin{cases}
+        \frac{1}{r}, & r<r_s \\
+        \frac{1}{r} \{ \frac{1}{2} \cos (\frac{\pi (r - r_s)}{ r_c - r_s} ) +\frac{1}{2} \}, & r_s \leq r<r_c \\
+        0, & r \geq r_c
+        \end{cases}
```
amcadmus, line 47: We are now using 5-th order polynomial as switcher. It is smooth upto 2nd order derivative. The cosine switcher is only smooth upto the first order derivative. 

https://github.com/njzjz/deepmd-kit/blob/24b433b20e51fc45c7ae587104d00cf732b0e92a/source/lib/include/switcher.h#L77-L92

njzjz: Oh, I don't know that.

njzjz: Updated equation:
image

amcadmus (approved):

PR #1019, opened 2021-08-23
Instantiate RunOptions first when training.

shishaochen:
PR #914 breaks down the parallel training with multiple GPU cards because `NeightStat` creates a `tf.Session` covering all GPUs before the default `tf.ConfigProto` is reset to only use CPU. Then, the trainer will fail when it binds to only 1 GPU card with a different device ID.

Since there is no GPU machines in GitHub CI, regression test cannot be activated to ensure new PRs won't effect the feature of parallel training. We have to test multi-GPU training at local.

### DIFF
--- a/source/tests/test_parallel_training.py
+++ b/source/tests/test_parallel_training.py
@@ -1,0 +1,35 @@
+import os
+import subprocess as sp
+import unittest
+
+from deepmd.cluster.local import get_gpus
+
+from common import tests_path
+
+
+class TestSingleMachine(unittest.TestCase):
+    def setUp(self):
+        try:
+            import horovod
+        except ImportError:
+            raise unittest.SkipTest("Package horovod is required for parallel-training tests.")
+        if len(get_gpus() or []) < 2:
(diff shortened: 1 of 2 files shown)


source/tests/test_parallel_training.py
```diff
@@ -0,0 +1,35 @@
...
+from common import tests_path
+
+
+class TestSingleMachine(unittest.TestCase):
+    def setUp(self):
+        try:
+            import horovod
+        except ImportError:
+            raise unittest.SkipTest("Package horovod is required for parallel-training tests.")
+        if len(get_gpus() or []) < 2:
```
njzjz, line 16: I think you also need to check whether tensorflow is built by CUDA? https://www.tensorflow.org/api_docs/python/tf/test/is_built_with_cuda

njzjz: Oh I think it should be not a problem. `get_gpus` should return empty if it's not built by CUDA.

amcadmus (approved):

njzjz (approved):

PR #1020, opened 2021-08-23
decouple activation function's type from model compression's process in SE_A, now tanh & gelu is both available.

liangadam:
I modified 3 pieces of code and wrote a brief unittest 'source/tests/test_tabulate.py' .

Now, when the 'activation_function' is designated as 'tanh' or 'gelu' in descriptor SE_A, model compression can produce a reasonable model. 

Moreover, cause the type of activation function is fully decoupled from the process of model compression, now we can add any other function type we need in a very easy way.

### DIFF
--- a/deepmd/utils/tabulate.py
+++ b/deepmd/utils/tabulate.py
@@ -34,7 +34,8 @@ class DPTabulate():
     def __init__(self,
                  model_file : str,
                  type_one_side : bool = False,
-                 exclude_types : List[List[int]] = []) -> None:
+                 exclude_types : List[List[int]] = [],
+                 activation_fn=tf.nn.tanh) -> None:
         """
         Constructor
         """
@@ -44,6 +45,15 @@ class DPTabulate():
         self.exclude_types = exclude_types
         if self.type_one_side and len(self.exclude_types) != 0:
             raise RunTimeError('"type_one_side" is not compatible with "exclude_types"')
+        
+        # functype
+        if activation_fn.__name__ == 'tf.nn.tanh' or activation_fn.__name__ == 'tanh':
+            self.functype = 1
+        elif activation_fn.__name__ == 'gelu':
+            self.functype = 2
+        else:
+            raise RunTimeError("Unknown actication function type!")
+        self.activation_fn = activation_fn
 
         self.graph, self.graph_def = load_graph_def(self.model_file)
         self.sess = tf.Session(graph = self.graph)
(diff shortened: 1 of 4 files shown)


njzjz (changes requested):
Also, please notice that tests failed.

deepmd/utils/tabulate.py
```diff
@@ -44,6 +45,15 @@ def __init__(self,
         self.exclude_types = exclude_types
         if self.type_one_side and len(self.exclude_types) != 0:
             raise RunTimeError('"type_one_side" is not compatible with "exclude_types"')
+        
+        # functype
+        if activation_fn.__name__ == 'tf.nn.tanh' or activation_fn.__name__ == 'tanh':
```
njzjz, line 50: I think you can directly check if the method is another method instead of checking its name
```suggestion
        if activation_fn is tf.nn.tanh:
```

denghuilu (commented):
The code indentation is inconsistent in multiple places, please check it.

deepmd/utils/tabulate.py
```diff
@@ -34,7 +34,8 @@ class DPTabulate():
     def __init__(self,
                  model_file : str,
                  type_one_side : bool = False,
-                 exclude_types : List[List[int]] = []) -> None:
+                 exclude_types : List[List[int]] = [],
+                 activation_fn=tf.nn.tanh) -> None:
```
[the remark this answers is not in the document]
liangadam: ok, thank you.

source/op/unaggregated_grad.cc
```diff
@@ -2,42 +2,90 @@
 #include "ComputeDescriptor.h"
 #include "neighbor_list.h"
 
+
+#define SQRT2_PI 0.7978845608028654
```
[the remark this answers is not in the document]
liangadam: alright.

deepmd/utils/tabulate.py
liangadam: OK, I will have a try. Thank you.

deepmd/utils/tabulate.py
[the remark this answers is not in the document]
njzjz: Use [`typing.Callable`](https://docs.python.org/3/library/typing.html#typing.Callable) as type hint here
```suggestion
                 activation_fn : Callable[[tf.Tensor], tf.Tensor] = tf.nn.tanh) -> None:
```
Note that you need to import `Callable` in line 5.

source/op/unaggregated_grad.cc
```diff
@@ -1,43 +1,90 @@
...
 
 template <typename FPTYPE>
 struct UnaggregatedDyDxSFunctor {
-    void operator()(const CPUDevice& d, const FPTYPE * y, const FPTYPE * w, const int length, const int width, FPTYPE * dy_dx) {
+    void operator()(const CPUDevice& d, const FPTYPE * y, const FPTYPE * w, const FPTYPE* xbar, const int length, const int width, FPTYPE * dy_dx, const int functype) {
         #pragma omp parallel for
         for (int ii = 0; ii < length; ii++) {
             for (int jj = 0; jj < width; jj++) {
-                dy_dx[ii * width + jj] = (1 - y[ii * width + jj] * y[ii * width + jj]) * w[jj];
+				dy_dx[ii * width + jj] = grad(xbar[ii * width + jj], y[ii * width + jj],functype)*w[jj];
```
njzjz, line 87: Please follow @denghuilu's suggestion and do not change the indentation here and somewhere else.

njzjz (changes requested):

source/op/unaggregated_grad.cc
liangadam: OK. Thank you.

njzjz:
Please revert changes from this commit to `tabulate.py`... We do not accept style format without actual change for [this reason](https://stackoverflow.com/a/39874565/9567349).

source/op/unaggregated_grad.cc
```diff
@@ -320,7 +423,7 @@ REGISTER_CPU(double);
 //     UnaggregatedDyDxSOp<GPUDevice, T>);                                                 \
 // REGISTER_KERNEL_BUILDER(                                                                \
 //     Name("UnaggregatedDyDx").Device(DEVICE_GPU).TypeConstraint<T>("T"),                 \
-//     UnaggregatedDyDxOp<GPUDevice, T>);                         
+//     UnaggregatedDyDxOp<GPUDevice, T>);
```
amcadmus, line 426: Plz revert all such format changes! thank you

amcadmus (changes requested):

PR #1056, opened 2021-08-29
fix device UT error

denghuilu:


njzjz (approved):

pkulzy (commented):
I tested it on ROCm platform. . The results showed that there are 23 skipped.
'''
Ran 265 tests in 974.050s
OK (skipped=23) 
'''
is that normal?

denghuilu:
> I tested it on ROCm platform. . The results showed that there are 23 skipped.
> 
> '''
> 
> Ran 265 tests in 974.050s
> 
> OK (skipped=23) 
> 
> '''
> 
> is that normal?

The same as cpu and gpu.

pkulzy (approved):

pkulzy:
> > I tested it on ROCm platform. . The results showed that there are 23 skipped.
> > '''
> > Ran 265 tests in 974.050s
> > OK (skipped=23)
> > '''
> > is that normal?
> 
> The same as cpu and gpu.

But why did the test skip

amcadmus:
> > > I tested it on ROCm platform. . The results showed that there are 23 skipped.
> > > '''
> > > Ran 265 tests in 974.050s
> > > OK (skipped=23)
> > > '''
> > > is that normal?
> > 
> > 
> > The same as cpu and gpu.
> 
> But why did the test skip

Skipped when the tf version is not satisfied with the UT.  see
https://github.com/deepmodeling/deepmd-kit/blob/1a25414dcd18b12316250baa330b690f3f5510fe/source/tests/test_deepdipole.py#L105

PR #1059, opened 2021-08-29
use `np.testing.assert_almost_equal` for array comparing

njzjz:
See my comment below.


source/tests/test_gen_stat_data.py
```diff
@@ -58,13 +55,13 @@ def test_merge_all_stat(self):
...
                                 1.0)
-        data1.add('force', 3, atomic = True, must = True)
+        data1.add('energy', 1, must = True)
         dp_random.seed(0)
         data2 = DeepmdDataSystem(['system_0', 'system_1'], 
                                 5, 
                                 10, 
                                 1.0)
-        data2.add('force', 3, atomic = True, must = True)
+        data2.add('energy', 1, must = True)
```
njzjz, line 64: The original `data1` and `data2` are not eqaul - but the original compare method didn't find it...

amcadmus (approved):

denghuilu (approved):

amcadmus (approved):

source/tests/test_gen_stat_data.py
amcadmus: This should be a bug in the UT...

PR #1066, opened 2021-08-30
upgrade os.system to subporcess.popen method

denghuilu:
Upgrade the os.system to subporcess.popen method. According to the [doc](https://docs.python.org/3/library/os.html#os.system) of os.system method:

> The subprocess module provides more powerful facilities for spawning new processes and retrieving their results; using that module is preferable to using this function. See the Replacing Older Functions with the subprocess Module section in the subprocess documentation for some helpful recipes.

In general, `subprocess.Popen()` is super-set of `os.system()`.

### DIFF
--- a/source/tests/test_model_compression.py
+++ b/source/tests/test_model_compression.py
@@ -1,6 +1,7 @@
 import os,sys,platform,shutil,dpdata,json
 import numpy as np
 import unittest
+import subprocess as sp
 
 from deepmd.infer import DeepPot
 from deepmd.env import MODEL_VERSION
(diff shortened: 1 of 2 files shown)


source/tests/test_model_compression.py
```diff
@@ -17,6 +18,16 @@ def _file_delete(file) :
     if os.path.exists(file):
         os.remove(file)
 
+def _subprocess_run(command):
+    popen = sp.Popen(command, shell=True, stdout=sp.PIPE, stderr=sp.STDOUT)
```
njzjz, line 22: I prefer setting `shell` to `False` and using `command.split()` instead of `command`. Using a shell [is not safe](https://docs.python.org/3/library/subprocess.html#security-considerations).

denghuilu: done

njzjz (approved):

amcadmus (approved):

PR #1071, opened 2021-08-31
fix Fail to build on macos: explicit instantiation of 'deepmd::Region' must occur in namespace 'deepmd'

amcadmus:


njzjz (approved):
I confirm it works.

PR #1077, opened 2021-08-31
fix typo: `lcueve.out`->`lcurve.out`

njzjz:


doc/train-input-auto.rst
```diff
@@ -898,7 +898,7 @@ model:
         .. _`model/fitting_net[polar]/scale`: 
 
         scale: 
-            | type: ``list`` | ``float``, optional, default: ``1.0``
+            | type: ``float`` | ``list``, optional, default: ``1.0``
```
njzjz, line 901: @y1xiaoc Is there some different behaviors to sort type?

y1xiaoc: dargs use a set to record dtypes. The order may change from 3.6 to 3.7 since the newer python guarantees that the iteration order is same as the add order.

amcadmus (approved):

PR #1081, opened 2021-09-01
add ABC for descriptors

njzjz:
I'm going to add abstract base classes for different object, where a list of methods and attributes is defined to normalize classes and their external call by other classes. It's also useful to develop and extend new classes.

The first one I did is the descriptor.


njzjz:
Documents can be read here: https://deepmodeling--1081.org.readthedocs.build/projects/deepmd/en/1081/api_py/deepmd.descriptor.html#deepmd-descriptor-descriptor-module

amcadmus (approved):

denghuilu (approved):

PR #1087, opened 2021-09-02
Fix a potential slice bug in se_t descriptor

denghuilu:
In the code review process, I found that `env_j` in se_t type always start the slice from position 0 while the slice size evaluates from `self.sel_a[type_j]` due to the setting of the `start_index_j`:

https://github.com/deepmodeling/deepmd-kit/blob/1a25414dcd18b12316250baa330b690f3f5510fe/deepmd/descriptor/se_t.py#L462-L498

And after the talk with Wanrun, we found that this may lead to repeated values ​​of `env_j` when there are multiple types of atoms. So I change the `start_index_j` value to avoid the replication of `env_j`. @amcadmus Please confirm the necessity of this change.

Note that this PR will not affect the single-atom type system.


deepmd/descriptor/se_t.py
```diff
@@ -477,7 +477,7 @@ def _filter(self,
                 env_i = tf.reshape(inputs_i, [-1, nei_type_i, 4])
                 # with natom x nei_type_i x 3
                 env_i = tf.slice(env_i, [0, 0, 1], [-1, -1, -1])
-                start_index_j = 0
+                start_index_j = start_index_i - self.sel_a[type_i]
```
amcadmus, line 480: How about moving the assignment of `start_index_j` before 472 the update of `start_index_i`, so we can write 
`start_index_j = start_index_i`

amcadmus (approved):

deepmd/descriptor/se_t.py
denghuilu: sure

amcadmus (approved):

PR #1091, opened 2021-09-03
fix(entrypoints/main): remove the argument -S

wangzyphysics:
the function `expand_sys_str(system)`  in test.py  has walked through all the sub directories in `system path`,
An error from utils/data.py will occur when  `set prefix` is delivered

if one has Multi-systems, the `system path` should be the parent path of all systems


amcadmus:
Could you plz provide detailed steps for reproducing the bug? Thanks!

wangzyphysics:
It turns out that I misunderstood the concept of the `-S`

PR #1098, opened 2021-09-05
add C API

njzjz:
currently it only contains two functions, but it can be a good example to write C APIs.


source/api_c/src/c_api.cc
```diff
@@ -0,0 +1,38 @@
...
+DP_ComputeResult DP_DeepPotCompute (DP_DeepPot* dp,
+                        const int natom,
+                        const VALUETYPE* coord,
+                        const int* atype,
+                        const VALUETYPE* cell) {
+    std::vector<VALUETYPE> coord_(coord, coord+natom*3);
+    std::vector<int> atype_(atype, atype+natom);
+    std::vector<VALUETYPE> cell_(cell, cell+9);
+    double e;
+    std::vector<VALUETYPE> f, v;
```
amcadmus, line 28: f and v will be destructed when `DP_DeepPotCompute` finishes. One may want to use new to allocate the memory.

doc/inference/cxx.md
```diff
@@ -22,4 +23,31 @@ gcc infer_water.cpp -D HIGH_PREC -L $deepmd_root/lib -L $tensorflow_root/lib -I
...
+
+int main(){
+  char* model = "graph.pb";
+  DP_DeepPot* dp = DP_NewDeepPot(model);
+  double coord[] = {1., 0., 0., 0., 0., 1.5, 1. ,0. ,3.};
+  double cell[] = {10., 0., 0., 0., 10., 0., 0., 0., 10.};
+  int atype[] = {1, 0, 1};
+  DP_ComputeResult result = DP_DeepPotCompute (dp, 3, &coord[0], &atype[0], &cell[0]);
+  const double e = result.energy, *f = (double*)result.force, *v = (double*)result.virial;
+}
```
amcadmus, line 41: dp, f and v should be freed before main ends.

amcadmus: Could you plz run with valgrind to check if there is any other memory leak?

amcadmus (changes requested):

njzjz:
I rewrite the interface. Tests are passed, and documents are added. I think there are no memory issues (not sure as I am not proficient in C).

doc/inference/cxx.md
```diff
@@ -23,3 +25,51 @@ and then run the program:
...
+  free(dp);
+}
+```
+
+where `e`, `f` and `v` are predicted energy, force and virial of the system, respectively.
+See {cpp:func}`DP_DeepPotCompute` for details.
+
+You can compile `infer_water.c` using `gcc`:
+```sh
+gcc infer_water.cpp -L $deepmd_root/lib -L $tensorflow_root/lib -I $deepmd_root/include -Wl,--no-as-needed -ldeepmd_c -Wl,-rpath=$deepmd_root/lib -Wl,-rpath=$tensorflow_root/lib -o infer_water
```
wanghan-iapcm, line 70:
```suggestion
gcc infer_water.c -L $deepmd_root/lib -L $tensorflow_root/lib -I $deepmd_root/include -Wl,--no-as-needed -ldeepmd_c -Wl,-rpath=$deepmd_root/lib -Wl,-rpath=$tensorflow_root/lib -o infer_water
```

source/api_c/tests/test_deeppot_a.cc
```diff
@@ -0,0 +1,116 @@
...
+
+  DP_DeepPotComputef (dp, natoms, coordf, atype, boxf, ener_, force_, virial_);
+
+  double ener = *ener_;
+  std::vector<float> force(force_, force_ + natoms * 3);
+  std::vector<float> virial(virial_, virial_ + 9);
+
+  EXPECT_LT(fabs(ener - expected_tot_e), 1e-6);
+  for(int ii = 0; ii < natoms*3; ++ii){
+    EXPECT_LT(fabs(force[ii] - expected_f[ii]), 1e-6);
```
wanghan-iapcm, line 111: Im surprised that the expected precision of the C interface is higher than that of the C++ interface (1e-4)

source/api_c/include/c_api.h
```diff
@@ -0,0 +1,72 @@
...
+  DP_DeepPot* dp,
+  const int natom,
+  const double* coord,
+  const int* atype,
+  const double* cell,
+  double* energy,
+  double* force,
+  double* virial
+  );
+
```
wanghan-iapcm, line 37: I would recommend that the compute method also returns atomic energy and virial.

source/api_c/tests/test_deeppot_a.cc
njzjz: The atomic energy does not have 1e-6 precision.

source/api_c/include/c_api.h
njzjz: Added.

wanghan-iapcm (approved):

amcadmus (approved):

PR #1102, opened 2021-09-06
Enable init-frz-model support for the original model

denghuilu:
This PR enables `init-frz-model` support for the original model within the `dp train` interface. Now both original and compress pb file(with `se_a` type descriptor) can be used to initialize the dp train interface.

The results in the example water benchmark system show that the `step 0` of the `lcurve.out` output of the original model(`original.out`) and the compressed model(`compress.out`) under the `dp train --init-frz-model` interface is consistent:

result of original.out:
```
root se_e2_a $ head original.out 
#  step      rmse_val    rmse_trn    rmse_e_val  rmse_e_trn    rmse_f_val  rmse_f_trn         lr
      0      2.58e+00    2.40e+00      2.81e-02    2.74e-02      8.17e-02    7.58e-02    1.0e-03
```

result of compress.out:
```
root se_e2_a $ head compress.out 
#  step      rmse_val    rmse_trn    rmse_e_val  rmse_e_trn    rmse_f_val  rmse_f_trn         lr
      0      2.58e+00    2.40e+00      2.81e-02    2.74e-02      8.17e-02    7.58e-02    1.0e-03
```

### DIFF
--- a/deepmd/utils/network.py
+++ b/deepmd/utils/network.py
@@ -92,7 +92,8 @@ def embedding_net(xx,
                   bavg = 0.0,
                   seed = None,
                   trainable = True, 
-                  uniform_seed = False):
+                  uniform_seed = False,
+                  initial_variables = None):
     r"""The embedding network.
 
     The embedding network function :math:`\mathcal{N}` is constructed by is the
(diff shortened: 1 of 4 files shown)


deepmd/utils/network.py
```diff
@@ -92,7 +92,8 @@ def embedding_net(xx,
                   bavg = 0.0,
                   seed = None,
                   trainable = True, 
-                  uniform_seed = False):
+                  uniform_seed = False,
+                  initial_variables = None):
```
njzjz, line 96: lack of documentation

njzjz (approved):

amcadmus (approved):

PR #1110, opened 2021-09-06
Fix bug of single precision model compression

denghuilu:
The model compression with float precision has problems in reading model parameters and parameter checking. This PR should be able to fix these problems. Now single-precision model compression can be used normally.

### DIFF
--- a/deepmd/entrypoints/compress.py
+++ b/deepmd/entrypoints/compress.py
@@ -5,8 +5,8 @@ import json
 import logging
 from typing import Optional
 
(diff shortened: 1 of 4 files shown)


njzjz (approved):

njzjz:
Does current unit tests also work for single precision?

denghuilu:
> Does current unit tests also work for single precision?

The current UT does not include specific single-precision tests. I'm thinking about adding the relative UT...

amcadmus (approved):

PR #1118, opened 2021-09-09
implement descriptor plugin; pass params to `Descriptor`

njzjz:
This PR adds plugin systems to descriptors. It's aimed to make deepmd-kit become a framework with any new or old descriptors.

This PR also adds support to pass parameters to `Descriptor` class, i.e. `descript = Descriptor(type="se_e2_a", rcut=6., rcut_smth=0.5, sel=[50])`, which is aimed to make it possible to build models with Python instead of command lines.

The next goal is to do the same thing to fittings, models, et al.

Also, I am going to write an example and maybe implement BPNN or other third-party descriptors.


deepmd/descriptor/se_ar.py
```diff
@@ -7,8 +7,9 @@
 from deepmd.env import op_module
 from .descriptor import Descriptor
 
+@Descriptor.register("se_ar")
 class DescrptSeAR (Descriptor):
-    def __init__ (self, jdata):
+    def __init__ (self, **jdata):
```
amcadmus, line 12: This change actually introduces a bug, `DescrptSeAR` because takes `dict` as input. 

`DescrptSeAR` is replace by `DescrptHybrid`. 

Now deepmd-kit does not accept the descriptor type `"se_ar"`, so we can safely remove it

njzjz: Ok, I've reverted changes to DescrptSeAR.

njzjz: Can we directly remove this file?

amcadmus: Yes we can. Remember also remove it UT.

amcadmus (approved):

denghuilu (approved):

PR #1125, opened 2021-09-10
fix "Call to method DeepTensor.__init__ with too many arguments"

njzjz:
As detected by LGTM, here is an error:
Call to method DeepTensor.__init__ with too many arguments; should be no more than 4.

I agree with the automatic code analysis tool.


amcadmus (approved):
Actually the class `DeepGlobalPolar` is not used by deepmd-kit anymore.

Issue #1109, opened 2021-09-06
[BUG] In the release 2.0.0 version, the lammps result is abnormal


gzq942560379:
I use the water data in the examples folder to train the model. When using this model for lammps MD inference, the results of lammps were abnormal. This problem did not occur in the previous bate version. I used the same model to get different results in the release version and the previous bate version. Below are my training input files, dp test output, and lammps output.

### training input.json
```json
{
    "_comment": " model parameters",
    "model": {
	"type_map":	["O", "H"],
	"descriptor" :{
	    "type":		"se_e2_a",
	    "sel":		[46, 92],
	    "rcut_smth":	0.50,
	    "rcut":		6.00,
	    "neuron":		[32, 64, 128],
	    "resnet_dt":	false,
	    "axis_neuron":	16,
	    "seed":		1,
	    "_comment":		" that's all"
	},
	"fitting_net" : {
	    "neuron":		[240, 240, 240],
	    "resnet_dt":	true,
	    "seed":		1,
	    "_comment":		" that's all"
	},
	"_comment":	" that's all"
    },

    "learning_rate" :{
	"type":		"exp",
	"decay_steps":	5000,
	"start_lr":	0.001,	
	"stop_lr":	3.51e-8,
	"_comment":	"that's all"
    },

    "loss" :{
	"type":		"ener",
	"start_pref_e":	0.02,
	"limit_pref_e":	1,
	"start_pref_f":	1000,
	"limit_pref_f":	1,
	"start_pref_v":	0,
	"limit_pref_v":	0,
	"_comment":	" that's all"
    },

    "training" : {
	"training_data": {
	    "systems":		["../data/data_0/", "../data/data_1/", "../data/data_2/"],
	    "batch_size":	"auto",
	    "_comment":		"that's all"
	},
	"validation_data":{
	    "systems":		["../data/data_3"],
	    "batch_size":	1,
	    "numb_btch":	3,
	    "_comment":		"that's all"
	},
	"numb_steps":	100000,
	"seed":		10,
	"disp_file":	"lcurve.out",
	"disp_freq":	100,
	"save_freq":	1000,
	"_comment":	"that's all"
    },    

    "_comment":		"that's all"
}
```

### test result
```bash
DEEPMD INFO    # number of test data : 1 
DEEPMD INFO    Energy RMSE        : 7.472860e-02 eV
DEEPMD INFO    Energy RMSE/Natoms : 3.892115e-04 eV
[...]

PR #1128, opened 2021-09-11
explicitly set neighbor request to full to fix bug #1109

amcadmus:
It seems that the default behavior of handling neigh_request is changed in Lammps `patch_14May2021`, which introduces a wrong stencil for the neighbor list. 

Before this fix the stencil is half
```
  (1) pair deepmd, perpetual
      attributes: , newton on
      pair build: full/bin/atomonly
      stencil: half/bin/3d/tri
      bin: standard
```
After this fix the stencil is full,
```
  (1) pair deepmd, perpetual
      attributes: full, newton on
      pair build: full/bin/atomonly
      stencil: full/bin/3d
      bin: standard
```


njzjz (approved):

njzjz:
works fine

PR #1144, opened 2021-09-13
support init-frz-model for se_r and se_t

njzjz:
I just copied several codes from se_a, which can also be applied to se_r and se_t.

### DIFF
--- a/deepmd/utils/graph.py
+++ b/deepmd/utils/graph.py
@@ -129,7 +129,7 @@ def get_embedding_net_nodes_from_graph_def(graph_def: tf.GraphDef, suffix: str =
         The embedding net nodes within the given tf.GraphDef object
     """
     embedding_net_nodes = {}
-    embedding_net_pattern = f"filter_type_\d+{suffix}/matrix_\d+_\d+|filter_type_\d+{suffix}/bias_\d+_\d+|filter_type_\d+{suffix}/idt_\d+_\d+|filter_type_all{suffix}/matrix_\d+_\d+|filter_type_all{suffix}/matrix_\d+_\d+_\d+|filter_type_all{suffix}/bias_\d+_\d+|filter_type_all{suffix}/bias_\d+_\d+_\d+|filter_type_all{suffix}/idt_\d+_\d"
+    embedding_net_pattern = f"filter_type_\d+{suffix}/matrix_\d+_\d+|filter_type_\d+{suffix}/bias_\d+_\d+|filter_type_\d+{suffix}/idt_\d+_\d+|filter_type_all{suffix}/matrix_\d+_\d+|filter_type_all{suffix}/matrix_\d+_\d+_\d+|filter_type_all{suffix}/bias_\d+_\d+|filter_type_all{suffix}/bias_\d+_\d+_\d+|filter_type_all{suffix}/idt_\d+_\d+"
     for node in graph_def.node:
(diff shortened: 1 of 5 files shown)


amcadmus (approved):

amcadmus:
> I just copied several codes from se_a, which can also be applied to se_r and se_t.

Should these methods be put in `Descriptor`?

njzjz:
> > I just copied several codes from se_a, which can also be applied to se_r and se_t.
> 
> Are they tested?

Yes and they work fine.

> Should these methods be put in `Descriptor`?

At least the hybrid descriptor needs specific methods so I prefer not putting in Descriptor.

However, both these descriptors has a same embedding network, so it is possible for them to share some methods (like those in this PR). So I would prefer adding another base class for them.

denghuilu (approved):

deepmd/utils/graph.py
```diff
@@ -129,7 +129,7 @@ def get_embedding_net_nodes_from_graph_def(graph_def: tf.GraphDef, suffix: str =
         The embedding net nodes within the given tf.GraphDef object
     """
     embedding_net_nodes = {}
-    embedding_net_pattern = f"filter_type_\d+{suffix}/matrix_\d+_\d+|filter_type_\d+{suffix}/bias_\d+_\d+|filter_type_\d+{suffix}/idt_\d+_\d+|filter_type_all{suffix}/matrix_\d+_\d+|filter_type_all{suffix}/bias_\d+_\d+|filter_type_all{suffix}/idt_\d+_\d"
+    embedding_net_pattern = f"filter_type_\d+{suffix}/matrix_\d+_\d+|filter_type_\d+{suffix}/bias_\d+_\d+|filter_type_\d+{suffix}/idt_\d+_\d+|filter_type_all{suffix}/matrix_\d+_\d+|filter_type_all{suffix}/matrix_\d+_\d+_\d+|filter_type_all{suffix}/bias_\d+_\d+|filter_type_all{suffix}/bias_\d+_\d+_\d+|filter_type_all{suffix}/idt_\d+_\d"
```
njzjz, line 132: I just added `filter_type_all{suffix}/matrix_\d+_\d+_\d+` and `filter_type_all{suffix}/bias_\d+_\d+_\d+` to support `se_e3`. @denghuilu I have two questions:
1. The pattern is too complex now. Is it possible to use something like `filter_type_*/*` to get all these things.
2. Is `filter_type_all{suffix}/idt_\d+_\d` a typo of `filter_type_all{suffix}/idt_\d+_\d+`?

denghuilu: > I just added `filter_type_all{suffix}/matrix_\d+_\d+_\d+` and `filter_type_all{suffix}/bias_\d+_\d+_\d+` to support `se_e3`. @denghuilu I have two questions:
> 
> 1. The pattern is too complex now. Is it possible to use something like `filter_type_*/*` to get all these things.
> 2. Is `filter_type_all{suffix}/idt_\d+_\d` a typo of `filter_type_all{suffix}/idt_\d+_\d+`?

1. I think the answer is no, we also have nodes like `filter_type_1/add_5,
filter_type_1/Tanh_3, filter_type_1/matrix_2_1/read` within the graph,

2. Yes, It's a typo

njzjz: Ok, I pushed a commit to fix this typo.

amcadmus (approved):

denghuilu (approved):

PR #1153, opened 2021-09-15
Add doc for model deviation

AnguseZhang:


doc/third-party/lammps-command.md
```diff
@@ -18,7 +18,10 @@ The DeePMD-kit package provides the pair_style `deepmd`
 pair_style deepmd models ... keyword value ...
 ```
 - deepmd = style of this pair_style
-- models = frozen model(s) to compute the interaction. If multiple models are provided, then the model deviation will be computed
+- models = frozen model(s) to compute the interaction. 
+If multiple models are provided, then only the first model serves to provide energy and force prediction for molecular dynamics, 
+and the model deviation will be computed among all models.
```
njzjz, line 23: The model deviation will be only computed every `out_freq` steps.

denghuilu (approved):

doc/third-party/lammps-command.md
denghuilu: Agree

amcadmus: @njzjz plz suggest a change.

doc/third-party/lammps-command.md
```diff
@@ -18,7 +18,10 @@ The DeePMD-kit package provides the pair_style `deepmd`
...
+
```
njzjz, line 24:
```suggestion
If multiple models are provided, then only the first model serves to provide energy and force prediction for each timestep of molecular dynamics, 
and the model deviation will be computed among all models every `out_freq` timesteps.
```

amcadmus (approved):

PR #1154, opened 2021-09-15
Bug fix of memory overflow when calculating model deviation

Ericwang6:
Calculate model deviation frame by frame to avoid memory overflow problems


deepmd/infer/model_devi.py
```diff
@@ -195,12 +195,13 @@ def make_model_devi(
...
     for data in data_sets:
-        coord = data["coord"]
-        box = data["box"]
-        atype = data["type"][0] 
-        devi = calc_model_devi(coord, box, atype, dp_models, nopbc=nopbc)
-        nframes_tot += coord.shape[0]
-        devis.append(devi)
+        coords = data["coord"]
+        boxs = data["box"]
+        atype = data["type"][0]
```
njzjz, line 200: Is there a reason to only use the first (?) type?

Ericwang6: I think atom types for all frames in a dataset is the same, aren't they?

njzjz: I don't think we can publicly ensure it if the returned array is the size of `nframes*natoms` instead of `natoms`

amcadmus: I agree with @njzjz , in this context we do not assume all frames of data["type"] has the same value.

amcadmus (changes requested):

amcadmus:
Next time you PR,  please checkout a new branch. 
your devel has too many commits that do not exist on the current devel branch.

amcadmus:
fixed by #1165

njzjz:
> fixed by #1165

I don't think so... `dp test` and `dp model-devi` are two different entry points.

amcadmus:
@njzjz It should be straightforward of using `AutoBatchSize` here, what do you think?

amcadmus (approved):

njzjz:
> @njzjz It should be straightforward of using `AutoBatchSize` here, what do you think?

Agree, and I've not decided whether we should use it on DeepPot -- the Python interface provided to users.

njzjz (approved):

amcadmus:
> > @njzjz It should be straightforward of using `AutoBatchSize` here, what do you think?
> 
> Agree, and I've not decided whether we should use it on DeepPot -- the Python interface provided to users.

Good idea!

PR #1155, opened 2021-09-15
Add plugin for gromacs

Ericwang6:
Add plugin for gromacs and relevant docs / examples


source/gmx/2020.2/CMakeLists.txt
```diff
@@ -0,0 +1,944 @@
...
+# If you want to redistribute modifications to GROMACS, please
+# consider that scientific software is very special. Version
+# control is crucial - bugs must be traceable. We will be happy to
+# consider code for inclusion in the official distribution, but
+# derived work must not be called official GROMACS. Details are found
+# in the README & COPYING files - if they are missing, get the
+# official version at http://www.gromacs.org.
+#
+# To help us fund GROMACS development, we humbly ask that you cite
+# the research papers on the package. Check out http://www.gromacs.org.
```
njzjz, line 33: This is the license information of GROMACS, which should be re-written.
Luckily I think the license of GROMACS is compatible with ours. Can you clarify how many codes of GROMACS are you using?

njzjz (changes requested):
Several problems:
1. New documentations are not added to sphinx's doctree. See [preview](https://deepmodeling--1155.org.readthedocs.build/projects/deepmd/en/1155/).
2. The current CMake system is not used. For example, `make gromacs` to generate the plugin like `make lammps`.
3. I don't think it's necessary to separate different versions. Preprocessor macros should be enough.
4. Examples can be smaller.
5. For `json.hpp`: I think there is already one in `third_party` directory.

There is also a very important comment below.

source/gmx/2020.2/CMakeLists.txt
njzjz: We should relicense with ours.

doc/install/install-gromacs.md
```diff
@@ -0,0 +1,33 @@
+# Install GROMACS with DeepMD
+## 1. Patch source code of GROMACS 
+Download source code of a supported gromacs version from https://manual.gromacs.org. Run the following command:
+```bash
+cd $deepmd_root/source/gmx && chmod +x patch.sh
```
njzjz, line 5: `git` allows you to `+x` before pushing.

njzjz (changes requested):

doc/install/install-gromacs.md
```diff
@@ -0,0 +1,33 @@
...
+./patch.sh -d $gromacs_root -v $version -p
+```
+where `deepmd_root` and `gromacs_root` refer to source code directory of deepmd-kit and gromacs respectively. And `version` represents the version of gromacs, **only support 2020.2 now**. You may patch another version of gromacs but still setting `version` to `2020.2`. However, we cannot ensure that it works.
+## 2. Install C++ api of deepmd-kit and tensorflow
+The C++ api of `deepmd-kit 2.x` and `tensorflow 2.x` are required.
++ Tips: C++ api of deepmd and tensorflow could be easily installed from the deepmd-kit offline packages. But before using tensorflow, you need to manually change the protobuf package to [version 3.9.2](https://github.com/protocolbuffers/protobuf/releases?after=v3.11.2) in `$deepmd_env_dir/include/google/protobuf` (the offline package will install a version of 3.14, which will cause incompability).
```
amcadmus, line 11: @Ericwang6 You said that the required protobuf version is 3.9.2 but provide a link to version <3.11.2, is there any reason?

I cannot find any definition for the variable `deepmd_env_dir`. Plz provide one.

amcadmus: @njzjz shall we downgrade protobuf in the offline package?

doc/install/install-gromacs.md
```diff
@@ -0,0 +1,33 @@
+# Install GROMACS with DeepMD
+## 1. Patch source code of GROMACS 
+Download source code of a supported gromacs version from https://manual.gromacs.org. Run the following command:
```
amcadmus, line 3: If you only support Gromacs 2020.2, then provide the download link to that specific version.

doc/third-party/gromacs.md
```diff
@@ -0,0 +1,120 @@
+# Running MD with GROMACS
+## NNP/MM Simulation
```
amcadmus, line 2: I would prefer `DP/MM` than `NNP/MM`

source/gmx/2020.2/CMakeLists.txt
amcadmus: I would strongly recommend you **patching** gromacs with the difference between your implementation and standard gromacs, rather than **overwriting** gromacs with your implementation.

amcadmus (changes requested):

doc/install/install-gromacs.md
njzjz: No, it's a dependency of a package built by Anaconda Inc. Is there a reason to mix the offline package and building from source?

Ericwang6:
> 5\. For `json.hpp`: I think there is already one in `third_party` directory.

The version here is 3.9.1, which is higher than that in `third_party` directory. And that old version is incompatible.

Ericwang6:
> 3\. I don't think it's necessary to separate different versions. Preprocessor macros should be enough.

@AnguseZhang and I both prefer to keep version-based patches as `plumed` does. Because there may be too may differences between versions in GROMACS... In this way can we ensure that users of different versions of GROMACS can use our plugin.

doc/third-party/gromacs.md
Ericwang6: OK

doc/install/install-gromacs.md
Ericwang6: @njzjz here is only a tip for users to quickly prepare a installed C++ api of `deepmd-kit` and `tensorflow`, since it is a really tough work to compile the C++ api of `tensorflow`

Ericwang6: > @Ericwang6 You said that the required protobuf version is 3.9.2 but provide a link to version <3.11.2, is there any reason?
> 
> I cannot find any definition for the variable `deepmd_env_dir`. Plz provide one.

Sure

Ericwang6:
> 4\. Examples can be smaller.

Well, do you mean reducing the number of atoms in our examples? I believe that methane in water is the smallest example to illustrate a DP/MM simulation.

Ericwang6:
> 2\. The current CMake system is not used. For example, `make gromacs` to generate the plugin like `make lammps`.

Here, the plugin is designed to provide scripts for users to manually modify the source code of gromacs. And  I don't think it's necessary to use the `cmake` system.

doc/install/index.rst
```diff
@@ -9,3 +9,4 @@ Installation
    install-lammps
    install-ipi
    build-conda
+   install-gromacs
```
njzjz, line 12: It should be located before `build-conda`.

doc/install/install-gromacs.md
njzjz: > @njzjz here is only a tip for users to quickly prepare a installed C++ api of `deepmd-kit` and `tensorflow`, since it is a really tough work to compile the C++ api of `tensorflow`

It should be officially recommended, though. We can directly provided a compiled GROMACS instead.

njzjz: Also, I think this section (Install C++ api...) is duplicate of other sections of the doc. We can assume one has already installed deepmd-kit, or just provide a link to the installation section.

njzjz: @amcadmus By the way, is it possible to avoid using any TensorFlow's API for our C++ API in our header? Then we will not need TensorFlow's header any more if we have already had C++ API library.

njzjz:
> > 5. For `json.hpp`: I think there is already one in `third_party` directory.
> 
> The version here is 3.9.1, which is higher than that in `third_party` directory. And that old version is incompatible.

Upgrading old version to new one?

njzjz:
> > 3. I don't think it's necessary to separate different versions. Preprocessor macros should be enough.
> 
> @AnguseZhang and I both prefer to keep version-based patches as `plumed` does. Because there may be too may differences between versions in GROMACS... In this way can we ensure that users of different versions of GROMACS can use our plugin.

I think `deepmd_plugin.cpp` and `deepmd_plugin.h` is not a version-based file, and all patch files can be merged into one file.

source/gmx/2020.2/CMakeLists.txt.patch
```diff
@@ -0,0 +1,29 @@
...
++add_definitions("-D HIGH_PREC")
++# ensure all dynamic libraries linked
++set (CMAKE_EXE_LINKER_FLAGS "-Wl,--no-as-needed")
++# add link libraries
++list (APPEND GMX_PUBLIC_LIBRARIES deepmd_op)
++list (APPEND GMX_PUBLIC_LIBRARIES deepmd_op_cuda)
++list (APPEND GMX_PUBLIC_LIBRARIES deepmd)
++list (APPEND GMX_PUBLIC_LIBRARIES deepmd_cc)
++list (APPEND GMX_PUBLIC_LIBRARIES tensorflow_cc)
++list (APPEND GMX_PUBLIC_LIBRARIES tensorflow_framework)
```
njzjz, line 24: `deepmd_cc` should be enough, as others are now a dependency of `deepmd_cc`. (use `ldd` to check it)

njzjz: Also note that `deepmd_op_cuda` is not available in cpu or hip mode.

doc/install/install-gromacs.md
```diff
@@ -0,0 +1,33 @@
...
+mkdir build
+cd build
+TENSORFLOW_ROOT="/path/to/tensorflow"
+DEEPMD_ROOT="/path/to/tensorflow"
+
+cmake3 .. -DCMAKE_CXX_STANDARD=14 \ # not required, but c++14 seems to be more compatible with higher version of tensorflow
+          -DGMX_TENSORFLOW_ROOT=${TENSORFLOW_ROOT} \
+          -DGMX_DEEPMD_ROOT=${DEEPMD_ROOT} \
+          -DGMX_MPI=ON -DGMX_GPU=CUDA \
+          -DCUDA_TOOLKIT_ROOT_DIR=/path/to/cuda \
```
njzjz, line 29: Does GROMACS support ROCm?

Ericwang6: Nope.

Ericwang6:
> > > 1. For `json.hpp`: I think there is already one in `third_party` directory.
> > 
> > 
> > The version here is 3.9.1, which is higher than that in `third_party` directory. And that old version is incompatible.
> 
> Upgrading old version to new one?

Sure. I'm not sure whether it will cause incompatibility for deepmd-kit. But I can test on it first.

doc/install/install-gromacs.md
Ericwang6: > Also, I think this section (Install C++ api...) is duplicate of other sections of the doc. We can assume one has already installed deepmd-kit, or just provide a link to the installation section.

Sure, I prefer to remove this section.

Ericwang6:
> > > 1. I don't think it's necessary to separate different versions. Preprocessor macros should be enough.
> > 
> > 
> > @AnguseZhang and I both prefer to keep version-based patches as `plumed` does. Because there may be too may differences between versions in GROMACS... In this way can we ensure that users of different versions of GROMACS can use our plugin.
> 
> I think `deepmd_plugin.cpp` and `deepmd_plugin.h` is not a version-based file, and all patch files can be merged into one file.

OK. Would you plz provide a reference on how to merge the patches into one file?

doc/install/index.rst
Ericwang6: Exactly...

njzjz:
[`git-format-patch`](https://git-scm.com/docs/git-format-patch)

source/gmx/2020.2/CMakeLists.txt.patch
```diff
@@ -0,0 +1,29 @@
...
+ 
++# DeepMD
++message(STATUS "Compling with DeepMD...")
++add_definitions(-w) # close warning
++include_directories(${GMX_DEEPMD_ROOT}/include)
++include_directories(${GMX_TENSORFLOW_ROOT}/include)
++link_directories(${GMX_DEEPMD_ROOT}/lib)
++link_directories(${GMX_TENSORFLOW_ROOT}/lib)
++# define high precision
++add_definitions("-D HIGH_PREC")
```
njzjz, line 15: question: this means it only supports models with double precision?

source/gmx/2020.2/CMakeLists.txt.patch
Ericwang6: > `deepmd_cc` should be enough, as others are now a dependency of `deepmd_cc`. (use `ldd` to check it)

Unfortunately, according to my test, all libs except `deepmd_op_cuda` should be linked, otherwise the program will crash...

source/gmx/2020.2/CMakeLists.txt.patch
Ericwang6: Yes. That's because we have not tested on whether single precision is enough for a gromacs simulation...And @denghuilu told me that the single-precision version is not quite stable yet due to lack of some unittests

Ericwang6:
@njzjz Uploading `json.hpp` to 3.9.1 will cause compilation errors in deepmd-kit

doc/install/install-gromacs.md
```diff
@@ -0,0 +1,33 @@
...
+Specify the installation path of tensorflow and deepmd using cmake options: `-DGMX_TENSORFLOW_ROOT` and `-DGMX_DEEPMD_ROOT`. Here is a sample compile scipt:
+```bash
+#!/bin/bash
+export CC=/usr/bin/gcc
+export CXX=/usr/bin/g++
+export CMAKE_PREFIX_PATH="/path/to/fftw-3.3.9" # fftw libraries
+mkdir build
+cd build
+TENSORFLOW_ROOT="/path/to/tensorflow"
+DEEPMD_ROOT="/path/to/tensorflow"
```
amcadmus, line 23: shouldn't it be "/path/to/deepmd"?

amcadmus:
> > 5. For `json.hpp`: I think there is already one in `third_party` directory.
> 
> The version here is 3.9.1, which is higher than that in `third_party` directory. And that old version is incompatible.

which part is incompatible? is it possible to upgrade the existing `json.hpp`?

doc/install/index.rst
njzjz: not resolved yet?

doc/install/install-gromacs.md
njzjz: If so, I think you should mention this thing.

source/gmx/2020.2/CMakeLists.txt.patch
njzjz: If so, I think you should mention this thing.

source/gmx/2020.2/CMakeLists.txt.patch
njzjz: Are you using the latest libraries?

doc/install/install-gromacs.md
Ericwang6: Oha, this is a mistake, I will fix it...

doc/install/install-gromacs.md
Ericwang6: OK.

source/gmx/2020.2/CMakeLists.txt.patch
Ericwang6: OK.

source/gmx/2020.2/CMakeLists.txt.patch
Ericwang6: This is resolved after using the latest version of deepmd-kit

Ericwang6:
Follow @njzjz 's advice, I've refactored the code in order to compile the gmx plugin along with deepmd-kit. Therefore, the number of changes and commits become too many to track for reviewers. So I'll close this pr first and open another pr for review.

PR #1160, opened 2021-09-19
Add plugin for GROMACS

Ericwang6:
1. Add plugin for gromacs, and relevant document and examples
2. Upgrade `source/3rdparty/json.hpp` to version 3.9.1


Ericwang6:
@amcadmus @njzjz I have refactored the gmx plugin code in order to compile it along with deepmd-kit. And also I have upgraded `json.hpp` to 3.9.1. I have closed the previous pr #1155 since it was too large to be tracked, and we may work on this pr instead. FYI @AnguseZhang @TablewareBox

njzjz (approved):
Looks good to me!

source/gmx/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+install (
+  FILES dp_gmx_patch
+  DESTINATION bin
+  PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
+)
+
+install (
+  DIRECTORY patches/
+  DESTINATION lib/patches
```
amcadmus, line 27: NOT a good name. Suppose that a user install deepmd in `/usr/`, then the patches will be installed to `/usr/lib/patches`, which may conflict with other software. I would suggest changing it to something like `lib/deepmd_patches`

doc/install/install-gromacs.md
```diff
@@ -0,0 +1,35 @@
...
+The C++ interface of `deepmd-kit 2.x` and `tensorflow 2.x` are required. And be aware that only deepmd-kit with **high precision** is supported now, since we cannot ensure single precision is enough for a GROMACS simulation. Specify the installation path of tensorflow and deepmd using cmake options: `-DGMX_TENSORFLOW_ROOT` and `-DGMX_DEEPMD_ROOT`. Here is a sample compile scipt:
+```bash
+#!/bin/bash
+export CC=/usr/bin/gcc
+export CXX=/usr/bin/g++
+export CMAKE_PREFIX_PATH="/path/to/fftw-3.3.9" # fftw libraries
+mkdir build
+cd build
+TENSORFLOW_ROOT="/path/to/tensorflow"
+DEEPMD_ROOT="/path/to/deepmd"
```
amcadmus, line 24: We should be able to compile `TENSORFLOW_ROOT` and `DEEPMD_ROOT` in `dp_gmx_patch`, so a user does not need to set them manually.

amcadmus (changes requested):

source/gmx/CMakeLists.txt
njzjz: `share` may be more proper than `lib`? See https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html
> The /usr/share hierarchy is for all read-only architecture independent data files. 
> This hierarchy is intended to be shareable among all architecture platforms of a given OS; thus, for example, a site with i386, Alpha, and PPC platforms might maintain a single /usr/share directory that is centrally-mounted. Note, however, that /usr/share is generally not intended to be shared by different OSes or by different releases of the same OS.
>Any program or package which contains or requires data that doesn't need to be modified should store that data in /usr/share (or /usr/local/share, if installed locally). It is recommended that a subdirectory be used in /usr/share for this purpose. Applications using a single file may use /usr/share/misc.
>Game data stored in /usr/share/games must be purely static data. Any modifiable files, such as score files, game play logs, and so forth, should be placed in /var/games.

source/CMakeLists.txt
```diff
@@ -182,6 +182,7 @@ if (BUILD_CPP_IF)
   if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 4.9)
     set (LIB_DEEPMD_NATIVE	"deepmd_native_md")
     set (LIB_DEEPMD_IPI		"deepmd_ipi")
+    set (LIB_DEEPMD_GROMACS "deepmd_gromacs")
   else ()
     message (STATUS "Your gcc/g++ version is ${CMAKE_CXX_COMPILER_VERSION}, so native MD and ipi are disabled. To enable them, use gcc/g++ >= 4.9.")
```
njzjz, line 187: This message should be updated.

njzjz (changes requested):

source/CMakeLists.txt
njzjz: wait, I think the new json.hpp supports GCC 4.8?

Ericwang6: Sure.

source/gmx/CMakeLists.txt
Ericwang6: Copy that

doc/install/install-gromacs.md
Ericwang6: That's a good idea.

njzjz:
4.9 needs to be changed to 4,8 in

https://deepmodeling--1160.org.readthedocs.build/projects/deepmd/en/1160/troubleshooting/installation.html
https://deepmodeling--1160.org.readthedocs.build/projects/deepmd/en/1160/install/install-from-source.html#install-the-tensorflow-s-c-interface

njzjz:
image

These two paragraphs are inconsistent

amcadmus (approved):

njzjz (approved):

Ericwang6:
@AnguseZhang @TablewareBox Give a quick review?

PR #1162, opened 2021-09-20
Add ROCm4.0.1 support for DEEPMD-KIT 

ZeyuLi0123:
ROCm's framework was tweaked after ROCm3.3, for example, some compile options and some dependency libraries.
The default version of ROCm is 2.9. If user wants to use ROCm-4.0.1, he/she should use "-DROCM_DTK=TRUE" to change the option.  

In addtion:
1. The GELU activation function comes with TF2.5, so I removed GELU compilation from ROCm4.0.1 compilation
2. I tested it on the cluster based on ROCm4.0.1 and TF2.5 and found it was still slower than the ROCm3.3 and TF2.2 we used for our previous training. For speed of inference, the two are about the same


njzjz:
> * The GELU activation function comes with TF2.5, so I removed GELU compilation from ROCm4.0.1 compilation

Does it really work? I think we use a different name from official TF.

ZeyuLi0123:
> > * The GELU activation function comes with TF2.5, so I removed GELU compilation from ROCm4.0.1 compilation
> 
> Does it really work? I think we use a different OP name from official TF.

I didn't test this activation function, but GELU is registered in ROCM-TF-2.5

njzjz:
> > > * The GELU activation function comes with TF2.5, so I removed GELU compilation from ROCm4.0.1 compilation
> > 
> > 
> > Does it really work? I think we use a different OP name from official TF.
> 
> I didn't test this activation function, but GELU is registered in ROCM-TF-2.5

Oh I think you are right. It's registered here: https://github.com/tensorflow/tensorflow/blob/919f693420e35d00c8d0a42100837ae3718f7927/tensorflow/python/ops/nn_ops.py#L3658

njzjz:
Did you add `-DROCM_DTK=TRUE` to `setup.py`? Also, this new variable should be referred in the document.

njzjz:
I think a problem here is that we still use custom GeLU OP in the Python API. Will it cause an error?

amcadmus:
> I didn't test this activation function, but GELU is registered in ROCM-TF-2.5

Please test it and show it works. otherwise we use our customized gelu.

denghuilu:
> I think a problem here is that we still use custom GeLU OP in the Python API. Will it cause an error?

It will cause an error in rocm dtk environment. And it's unclear in cuda environment now. I think we should address the gelu function in a later PR @amcadmus @njzjz

source/cmake/FindROCM.cmake
```diff
@@ -7,11 +7,21 @@
...
 
 # define the search path
-list(APPEND ROCM_search_PATHS ${ROCM_ROOT})
-list(APPEND ROCM_search_PATHS "/opt/rocm/")
+if(ROCM_ROOT)
+  list(APPEND ROCM_SEARCH_PATHS ${ROCM_ROOT})
+else()
+  if(ROCM_DTK)
+    list(APPEND ROCM_SEARCH_PATHS "/public/software/compiler/dtk/dtk-21.04/")
+    list(APPEND ROCM_SEARCH_PATHS "/public/software/compiler/rocm/dtk-21.04/")
```
njzjz, line 15: I think these two paths should be different in different machines?

ZeyuLi0123: yes, These two paths are the rocm_dtk paths that I have seen. I put them here as the default paths if user does not give the path of ROCm. And if the ROCm_dtk's path is not in the default path-lists, he/she can use `ROCM_ROOT` to attach to path

ZeyuLi0123: yes, These two paths are the rocm_dtk paths that I have seen. I put them here as the default paths if user does not give the path of ROCm. And if the ROCm_dtk's path is not in the default path-lists, user can use ROCM_ROOT to attach to path

njzjz: Is it the default installation path? It does not look like so.

ZeyuLi0123: As a matter of fact, it is the default installation path.
On the cluster, '/opt/rocm' maybe looks like the default path. But this is the path of the version 2.9

ZeyuLi0123:
There are some problems when uses tf-2.5 by ROCm-4.0.1, and I haven't located the problem yet. I turned PR off for a while

wankiwi:
Hello, developers. Can you tell which versions of ROCm and tensorflow are supported for Deepmd-kit now? For I try to install it on a cluster with dcu.

ZeyuLi0123:
> Hello, developers. Can you tell which versions of ROCm and tensorflow are supported for Deepmd-kit now? For I try to install it on a cluster with dcu.

Rocm-2.9, tensorflow base on Rocm2.9

wankiwi:
Thanks, I will try now.

PR #1175, opened 2021-09-28
Add model compression support for se_t type descriptor

denghuilu:
With the help of @liangadam , we have added the  model compression support for the `se_t` type descriptor. The main changes are:

- Reconstruct the DPTabulate class for the support the of 'se_t' type;
- Change the op name from `TabulateFusion` to `TabulateFusionSeA` for `se_a` type;
- Add op `TabulateFusionSeT` for 'se_t' type, and put the corresponding CPU and GPU implementations in tabulate.cc and tabulate.cu;
- Add enable_model_compression method for 'se_t.py';
- The default step of 'se_t' is 10 times larger than 'se_a'.

By using the default water example benchmark system, we found that the lammps output of the se_t compressed model is consistent. And with the embedding net size setting[25, 50, 100], [4, 8, 16], the compressed model gains speedup by a factor of 20(single V100 GPU) and 12(CPU) correspondingly.

TODO:

- [ ] Add unit test for 'se_t' model compression;
- [ ] Add doc support for 'se_t' model compression;
- [ ] Fix the bug of accuracy when the last_layer_size of embedding net is equal to 8.

### DIFF
--- a/deepmd/utils/tabulate.py
+++ b/deepmd/utils/tabulate.py
@@ -4,6 +4,7 @@ import logging
 import numpy as np
 from typing import Callable
 from typing import Tuple, List
+from scipy.special import comb
 from deepmd.env import tf
 from deepmd.env import op_module
 from deepmd.common import ACTIVATION_FN_DICT
@@ -25,6 +26,8 @@ class DPTabulate():
 
     Parameters
     ----------
+    descrpt_name
+            The descriptor name of the frozen model
     model_file
             The frozen model
     type_one_side
@@ -38,6 +41,7 @@ class DPTabulate():
             The suffix of the scope
     """
     def __init__(self,
+                 descrpt_name : str,
                  model_file : str,
                  type_one_side : bool = False,
                  exclude_types : List[List[int]] = [],
@@ -47,13 +51,11 @@ class DPTabulate():
         """
         Constructor
         """
-
+        self.descrpt_name = descrpt_name
         self.model_file = model_file
         self.type_one_side = type_one_side
         self.exclude_types = exclude_types
         self.suffix = suffix
-        if self.type_one_side and len(self.exclude_types) != 0:
-            raise RuntimeError('"type_one_side" is not compatible with "exclude_types"')
         
         # functype
         if activation_fn == ACTIVATION_FN_DICT["tanh"]:
@@ -87,27 +89,21 @@ class DPTabulate():
 
         self.embedding_net_nodes = get_embedding_net_nodes_from_graph_def(self.graph_def, suffix=self.suffix)
 
-        for tt in self.exclude_types:
-            if (tt[0] not in range(self.ntypes)) or (tt[1] not in range(self.ntypes)):
-                raise RuntimeError("exclude types" + str(tt) + " must within the number of atomic types " + str(self.ntypes) + "!")
-        if (self.ntypes * self.ntypes - len(self.exclude_types) == 0):
-            raise RuntimeError("empty embedding-net are not supported in model compression!")
-        
-        self.layer_size = len(self.embedding_net_nodes) // ((self.ntypes * self.ntypes - len(self.exclude_types)) * 2)
-        self.table_size = self.ntypes * self.ntypes
-        if type_one_side :
-            self.layer_size = len(self.embedding_net_nodes) // (self.ntypes * 2)
-            self.table_size = self.ntypes
-        # self.value_type = self.embedding_net_nodes["filter_type_0/matrix_1_0"].dtype #"filter_type_0/matrix_1_0" must exit~
-        # get trained variables
+        # move it to the descriptor class
+        # for tt in self.exclude_types:
+        #     if (tt[0] not in range(self.ntypes)) or (tt[1] not in range(self.ntypes)):
+        #         raise RuntimeError("exclude types" + str(tt) + " must within the number of atomic types " + str(self.ntypes) + "!")
+        # if (self.ntypes * self.ntypes - len(self.exclude_types) == 0):
+        #     raise RuntimeError("empty embedding-net are not supported in model compression!")
+        self.layer_size = self._get_layer_size()
+        self.table_size = self._get_table_size()
+
         self.bias = self._get_bias()
         self.matrix = self._get_matrix()
 
-        for item in self.matrix["layer_" + str(self.layer_size)]:
-            if len(item) != 0:
-                self.data_type = type(item[0][0])
-                self.last_layer_size = item.shape[1]
-        # define tables
+        self.data_type  = self._get_data_type()
+        self.last_layer_size = self._get_last_layer_size()
+
         self.data = {}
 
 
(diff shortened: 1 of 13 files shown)


source/op/_tabulate_grad.py
```diff
@@ -8,12 +8,22 @@
 from deepmd.env import tf
 # from deepmd.DescrptSeATabulate import last_layer_size
 
-@ops.RegisterGradient("TabulateFusion")
-def _tabulate_fusion_grad_cc (op, dy):    
-    dy_dx, dy_df = op_module.tabulate_fusion_grad(op.inputs[0], op.inputs[1], op.inputs[2], op.inputs[3], dy, op.outputs[0])
+@ops.RegisterGradient("TabulateFusionSeA")
```
njzjz, line 11: This rename will cause model compatibility issue, though.

denghuilu: The op `TabulateFusion` are still registered in the `tabulate_multi_device.cc` just like the `DescrptSeA` in `prod_env_mat_multi_device.cc`. So I think this change should not be able to affect the inference process. For the `init-frz-model` process, I will add the Decorators `@ops.RegisterGradient("TabulateFusion")`.

njzjz: Backward compatibility is fine but forward compatibility is broken. Do we need to bump model_version?

deepmd/utils/tabulate.py
```diff
@@ -254,21 +282,26 @@ def _layer_1(self, x, w, b):
...
+                if upper < ((1 / min_nbor_dist) * sw - self.davg[ii][0]) / self.dstd[ii][0]:
+                    upper = ((1 / min_nbor_dist) * sw - self.davg[ii][0]) / self.dstd[ii][0]
+        elif self.descrpt_name == "DescrptSeT":
+            for ii in range(self.ntypes):
+                for jj in range(1, 4):
+                    var = (sw * sw) / (min_nbor_dist * min_nbor_dist * self.dstd[ii][jj] * self.dstd[ii][jj])
+                    if lower > -var:
+                        lower = -var
+                    if upper < var:
+                        upper = var
```
njzjz, line 304:
```suggestion
            var = np.square(sw / (min_nbor_dist * self.dstd[:, 1:4]))
            lower = np.max(-var)
            upper = np.min(var)
```

deepmd/utils/tabulate.py
```diff
@@ -254,21 +282,26 @@ def _layer_1(self, x, w, b):
...
-            if lower > -self.davg[ii][0] / self.dstd[ii][0]:
-                lower = -self.davg[ii][0] / self.dstd[ii][0]
-            if upper < ((1 / min_nbor_dist) * sw - self.davg[ii][0]) / self.dstd[ii][0]:
-                upper = ((1 / min_nbor_dist) * sw - self.davg[ii][0]) / self.dstd[ii][0]
+        if self.descrpt_name == "DescrptSeA":
+            for ii in range(self.ntypes):
+                if lower > -self.davg[ii][0] / self.dstd[ii][0]:
+                    lower = -self.davg[ii][0] / self.dstd[ii][0]
+                if upper < ((1 / min_nbor_dist) * sw - self.davg[ii][0]) / self.dstd[ii][0]:
+                    upper = ((1 / min_nbor_dist) * sw - self.davg[ii][0]) / self.dstd[ii][0]
```
njzjz, line 296:
```suggestion
            lower = np.max(-self.davg[:, 0] / self.dstd[:, 0])
            upper = np.min(((1 / min_nbor_dist) * sw - self.davg[:, 0]) / self.dstd[:, 0])
```

deepmd/utils/tabulate.py
```diff
@@ -139,34 +135,57 @@ def build(self,
...
-                            tt = stride1
-                        hh = vv[jj + 1][kk] - vv[jj][kk]
-                        self.data[net][jj][kk * 6 + 0] = vv[jj][kk]
-                        self.data[net][jj][kk * 6 + 1] = dd[jj][kk]
-                        self.data[net][jj][kk * 6 + 2] = 0.5 * d2[jj][kk]
-                        self.data[net][jj][kk * 6 + 3] = (1 / (2 * tt * tt * tt)) * (20 * hh - (8 * dd[jj + 1][kk] + 12 * dd[jj][kk]) * tt - (3 * d2[jj][kk] - d2[jj + 1][kk]) * tt * tt)
-                        self.data[net][jj][kk * 6 + 4] = (1 / (2 * tt * tt * tt * tt)) * (-30 * hh + (14 * dd[jj + 1][kk] + 16 * dd[jj][kk]) * tt + (3 * d2[jj][kk] - 2 * d2[jj + 1][kk]) * tt * tt)
-                        self.data[net][jj][kk * 6 + 5] = (1 / (2 * tt * tt * tt * tt * tt)) * (12 * hh - 6 * (dd[jj + 1][kk] + dd[jj][kk]) * tt + (d2[jj + 1][kk] - d2[jj][kk]) * tt * tt)
+
+        if self.descrpt_name == "DescrptSeA":
```
njzjz, line 139: I prefer something like
```py
if isinstance(self.descrpt, deepmd.descriptors.DescrptSeA):
```

source/op/_tabulate_grad.py
denghuilu: I think you are right. How should I bump the model_version?

doc/freeze/compress.md
```diff
@@ -79,3 +79,7 @@ Model compression, with little loss of accuracy, can greatly speed up MD inferen
 **Acceptable original model version**
 
 The model compression interface requires the version of deepmd-kit used in original model generation should be `2.0.0-alpha.0` or above. If one has a frozen 1.2 or 1.3 model, one can upgrade it through the `dp convert-from` interface.(eg: ```dp convert-from 1.2/1.3 -i old_frozen_model.pb -o new_frozen_model.pb```) 
+
+**Acceptable descriptor type**
+
+Note only descriptors with `se_e3_a` or `se_t` type are supported by the model compression feature.
```
njzjz, line 85: `se_e3_a` -> `se_e2_a`

`hybrid` mixed with above descriptors is also supported.

denghuilu: Sure

source/op/_tabulate_grad.py
amcadmus: update the file source/config/MODEL_VER

denghuilu: got it

liangadam:
An error ‘incompatible model version’ occurred when using the produced model.pb to run lammps.

terminate called after throwing an instance of 'std::runtime_error'
  what():  incompatable model: version 1.1 in graph, but version 1.0  supported

denghuilu:
> An error ‘incompatible model version’ occurred when using the produced model.pb to run lammps.
> 
> terminate called after throwing an instance of 'std::runtime_error' what(): incompatable model: version 1.1 in graph, but version 1.0 supported

You should manually compile the deepmd-kit c++ interface from scratch to match the 1.1 model version.

PR #1176, opened 2021-09-29
bump default LAMMPS version to `stable_29Sep2021`

njzjz:
`stable_29Sep2021` will be released today.

- [x] retry actions after it is released


amcadmus (approved):
Checked the DPLR MD output of 20210929 is identical to 20201029

PR #1200, opened 2021-10-09
add `start_pref_pf` and `limit_pref_pf` to `loss` Argument

### LINKED ISSUE #1199: "start_pref_f" not support on DeepMD-kit v2.0.1 gpu cuda 11.3
### Discussed in https://github.com/deepmodeling/deepmd-kit/discussions/1197

<div type='discussions-op-text'>

<sup>Originally posted by **Franklalalala** October  8, 2021</sup>
I saw this feature in the code, loss function, which is exactly what i'm looking for. Since my data is not from md trajectory, most of the atoms have a extremely low force. Apply this "atom prefernce" feature may greatly improve the force prediction.
I've seen this. 
https://github.com/deepmodeling/deepmd-kit/discussions/358
And checked my data with this 
https://github.com/deepmodeling/deepmd-kit/discussions/709
I've tried this feature on my cpu version, the results looks acceptable. However, I can't run it with GPU version. BUG is: ArgumentKeyError: [at location `loss`] undefined key `start_pref_pf` is not allowed in strict mode.
What can i do to implement this feature with GPU? Lower my version? If so, which version? Thanks!</div>

njzjz:
fix #1199.

### DIFF
--- a/deepmd/utils/argcheck.py
+++ b/deepmd/utils/argcheck.py
@@ -478,6 +478,8 @@ def loss_ener():
     doc_limit_pref_v = limit_pref('virial')
     doc_start_pref_ae = start_pref('atom_ener')
     doc_limit_pref_ae = limit_pref('atom_ener')
+    doc_start_pref_pf = start_pref('atom_pref')
+    doc_limit_pref_pf = limit_pref('atom_pref')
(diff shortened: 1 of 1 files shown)


Franklalalala:
Wow, so efficient! Thanks a lot!

PR #1246, opened 2021-10-27
Fix bug in transfer.py

ZhengdQin:
fix the np.frombuffer in dp transfer


ZhengdQin:
fix the np.frombuffer in dp transfer, the point is explained as follows:
Interpret a buffer as a 1-dimensional array, which type is the given data_dtype, which means if the type is float32, the data in the buffer will be converted to float32, no matter what the original data type is. Let's see an example:
#===========================
temp1 = np.array([1.2, 4.5], dtype=np.float64)

temp1
Out[3]: array([1.2, 4.5])

temp2 = np.frombuffer(temp1, dtype=np.float32)

temp2
Out[5]: 
array([4.172325e-08, 1.900000e+00, 0.000000e+00, 2.281250e+00],
      dtype=float32)
#===========================
If we do not load the data with the old type, the loaded tensor values are wrong.

denghuilu (approved):

wanghan-iapcm (approved):

PR #1274, opened 2021-11-09
Accelerate model compression

denghuilu:
### Main changes:

- Due to the sorted environment matrix, during the execution of model compression process, we can use this rule to reuse GPU kernel registers. Thus accelerate the tabulate_fusion_grad_se_a kernel by a factor of 1.1.
- Abstract part of the code.

A rocm implementation will be added later.


njzjz:
#1275 should fix the compilation error.

wanghan-iapcm (approved):

PR #1283, opened 2021-11-11
Added all activation functions for model compression.

liangadam:
Now the model with activation functions such as  'relu', 'relu6', 'softplus', 'sigmoid'  in descriptor can be compressed, and the output results have been well tested.

Modified related document with activation function.

### DIFF
--- a/doc/train/train-input.rst
+++ b/doc/train/train-input.rst
@@ -5,7 +5,9 @@ Training Parameters
    
    
    **Available activation functions for descriptor:**
+   
    tanh
+   
    gelu
    relu
    relu6
--- a/doc/train/train-input.rst
+++ b/doc/train/train-input.rst
@@ -4,8 +4,8 @@ Training Parameters
    One can load, modify, and export the input file by using our effective web-based tool `DP-GUI <https://deepmodeling.org/dpgui/input/deepmd-kit-2.0>`_. All training parameters below can be set in DP-GUI. By clicking "SAVE JSON", one can download the input file for furthur training.
    
    
-   **Available activation functions for descriptor**:
-   - tanh
+   **Available activation functions for descriptor:**
+- tanh
    - gelu
    - relu
    - relu6
--- a/doc/train/train-input.rst
+++ b/doc/train/train-input.rst
@@ -5,12 +5,12 @@ Training Parameters
    
    
    **Available activation functions for descriptor:**
-      -tanh
-      -gelu
-      -relu
-      -relu6
-      -softplus
-      -sigmoid
+   -tanh
+   -gelu
+   -relu
+   -relu6
+   -softplus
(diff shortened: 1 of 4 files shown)


doc/train/train-input.rst
```diff
@@ -2,5 +2,18 @@ Training Parameters
 ======================================
 .. note::
    One can load, modify, and export the input file by using our effective web-based tool `DP-GUI <https://deepmodeling.org/dpgui/input/deepmd-kit-2.0>`_. All training parameters below can be set in DP-GUI. By clicking "SAVE JSON", one can download the input file for furthur training.
+   
+   
+   **Available activation functions for descriptor:**
```
njzjz, line 7: Why insert here?

wanghan-iapcm (commented):
How was this implementation tested?

doc/train/train-input.rst
liangadam: Because there is a lack of guidance on available activation functions, neither in DP-GUI nor for editing 'input.json' by oneself. 
You can tell me if there is a more suitable place to add.

liangadam:
> How was this implementation tested?

By using dp test to compare the results of models (trained with different activation functions) before and after compression. The test results show that all compressed models are very accurate.
Also，I deliberately add different types of errors to test the error reporting mechanism. The results are sensible.

Moreover, using lammps to run those models to make sure there are no abnormal result.

The following shows an test example: 
image

doc/train/train-input.rst
njzjz: I think it's mentioned, though.

image

wanghan-iapcm:
> > How was this implementation tested?
> 
> By using dp test to compare the results of models (trained with different activation functions) before and after compression. The test results show that all compressed models are very accurate. Also，I deliberately add different types of errors to test the error reporting mechanism. The results are sensible.
> 
> Moreover, using lammps to run those models to make sure there are no abnormal result.
> 
> The following shows an test example: image

It looks great. Have you tested all types of activation functions you implemented?

liangadam:
> > > How was this implementation tested?
> > 
> > 
> > By using dp test to compare the results of models (trained with different activation functions) before and after compression. The test results show that all compressed models are very accurate. Also，I deliberately add different types of errors to test the error reporting mechanism. The results are sensible.
> > Moreover, using lammps to run those models to make sure there are no abnormal result.
> > The following shows an test example: image
> 
> It looks great. Have you tested all types of activation functions you implemented?

Yes, All of them have been tested~

doc/train/train-input.rst
liangadam: Fine~ It's better to left document work to you because it‘s hard for me to think about it comprehensively.

njzjz (approved):

PR #1285, opened 2021-11-15
Enable mixed precision support for deepmd-kit

denghuilu:
This PR has enabled the mixed-precision training as well as the mixed precision inference process for deepmd-kit. Without any change of the input script, one can easily enable the mixed precision training by simply setting the environment variable `DP_ENABLE_MIXED_PREC` to `fp16`. 

Main changes:

- add `DP_ENABLE_MIXED_PREC` environmental variable for the control of mixed precision training. Note currently only `tf.float16` precision is enabled with the mixed precision setting.
- set the default embedding-net and fitting-net precision at `argcheck.py` according to the environment variable `DP_INTERFACE_PREC`.
- use dynamic loss scale for gradients update.
- add doc for mixed precision suppport.

According to our example water benchmark system, with `TF-2.6.0`, `CUDA-11.0` and `NVIDIA-V100 GPU` environment, the speed of the dp training process decreased slightly, while the inference process with 12288 atoms has gained a speedup by a factor of 3.

It is strongly recommended to enable the mixed precision settings with CUDA-11.0 or above CUDA-toolkit.


njzjz:
@wanghan-iapcm an import error is caught in the latest dpdata

deepmd/train/trainer.py
```diff
@@ -345,6 +348,9 @@ def _build_training(self):
             optimizer = self.run_opt._HVD.DistributedOptimizer(optimizer)
         else:
             optimizer = tf.train.AdamOptimizer(learning_rate = self.learning_rate)
+        if DP_ENABLE_MIXED_PRECISION:
+            # enable dynamic loss scale of the gradients
+            optimizer = tf.train.experimental.enable_mixed_precision_graph_rewrite(optimizer)
```
njzjz, line 353: This function has been moved to `tf.mixed_precision.enable_mixed_precision_graph_rewrite`. https://www.tensorflow.org/api_docs/python/tf/compat/v1/mixed_precision/enable_mixed_precision_graph_rewrite What TF version do you use? Do you know if it is supported in all TF versions?

amcadmus:
> @wanghan-iapcm an import error is caught in the latest dpdata

pymatgen... could you please help fix it? thanks!

njzjz:
> pymatgen... could you please help fix it? thanks!

See deepmodeling/dpdata#217.

deepmd/train/trainer.py
denghuilu: This function was found in Nvidia's official [documentation](https://docs.nvidia.com/deeplearning/performance/mixed-precision-training/index.html#tensorflow-amp). I have tested it with the TF-1.14.0 and TF-2.6.0 environment. Since it is a deprecated function, I will use the new `tf.mixed_precision.enable_mixed_precision_graph_rewrite` function.

njzjz: The method was available since v1.12 (https://github.com/tensorflow/tensorflow/commit/02730dcecf0e479c5be06adab16d5f6a0c03d11c) and then was renamed in v2.4 (https://github.com/tensorflow/tensorflow/commit/011228639301a8ed60d39c94ab096b074091bb4a#diff-38a59d213566539b30a6ce96962d40b68f18db2365e74e5e8bec220c43e92aec).

denghuilu:
There are some problems in the mixed precision training on the descriptors of se_r and se_t types, which are under investigation.

deepmd/train/trainer.py
denghuilu: Sure

denghuilu:
> There are some problems in the mixed precision training on the descriptors of se_r and se_t types, which are under investigation.

@amcadmus @njzjz There's still some errors when training mixed precision with the `se_r` or `se_t` types. So I suggest that we merge the `se_a` type first.

deepmd/train/trainer.py
```diff
@@ -345,6 +358,12 @@ def _build_training(self):
             optimizer = self.run_opt._HVD.DistributedOptimizer(optimizer)
         else:
             optimizer = tf.train.AdamOptimizer(learning_rate = self.learning_rate)
+        if self.mixed_prec is not None:
+            # check the TF_VERSION, when TF < 1.12, mixed precision is not allowed 
+            if TF_VERSION < "1.12":
```
njzjz, line 363:
```py
>>> "1.8"<"1.12"
False
```

njzjz (approved):

njzjz:
Can you also support hybrid?

denghuilu:
> Can you also support hybrid?

As we said, there's still some errors when using the `se_r` or `se_t` type descriptor. Hybrid is not yet ready for using.

njzjz:
It will be useful to `hybrid` mixed by two `se_a`.

deepmd/utils/network.py
```diff
@@ -44,6 +49,12 @@ def one_layer(inputs,
                             b_initializer, 
                             trainable = trainable)
         variable_summaries(b, 'bias')
+
+        if mixed_prec is not None and outputs_size != 1:
```
wanghan-iapcm, line 53: I do not like this idea. 
For dipole and polar, the size of output layer is not 1, but they are using fp16, which is not what we want.

deepmd/utils/network.py
```diff
@@ -65,6 +76,8 @@ def one_layer(inputs,
                 # return activation_fn(hidden_bn)
             else:
                 if use_timestep :
+                    if mixed_prec is not None and outputs_size != 1:
+                       idt = tf.cast(idt, get_precision(mixed_prec['compute_prec']))
```
wanghan-iapcm, line 80: Again `outputs_size != 1` may not be a good idea.

deepmd/descriptor/se_a.py
```diff
@@ -735,6 +749,8 @@ def _filter(
             name='linear', 
             reuse=None,
             trainable = True):
+        if self.mixed_prec is not None:
+            inputs = tf.cast(inputs, get_precision(self.mixed_prec['compute_prec']))
```
wanghan-iapcm, line 753: Do we need this line? the `inputs` are anyway cast to `compute_prec` in networks.one_layer or networks.embedding_net

denghuilu: 1. There's matrix multiplication outside the embedding net, we need to cast the inputs to match the dtype of the embedding net output.
2. Half precision slicing will be more efficient.

deepmd/utils/network.py
```diff
@@ -16,10 +17,14 @@ def one_layer(inputs,
               reuse=None,
               seed=None, 
               use_timestep = False, 
-              trainable = True,
+              trainable = False,
```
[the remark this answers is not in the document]
denghuilu: typo for debug, I'll fix it

deepmd/descriptor/descriptor.py
```diff
@@ -262,6 +262,24 @@ def enable_compression(self,
...
+                The mixed precision setting used in the embedding net
+        
+        Notes
+        -----
+        This method is called by others when the descriptor supported compression.
+        """
+        raise NotImplementedError(
+            "Descriptor %s doesn't support mixed precision training!" % type(self).__name__)
+
+
```
njzjz, line 282: lint errors appear here

deepmd/train/trainer.py
```diff
@@ -345,6 +358,15 @@ def _build_training(self):
             optimizer = self.run_opt._HVD.DistributedOptimizer(optimizer)
         else:
             optimizer = tf.train.AdamOptimizer(learning_rate = self.learning_rate)
+        if self.mixed_prec is not None:
+            TF_VERSION_LIST = [int(item) for item in TF_VERSION.split('.')]
```
njzjz, line 362: `int(item)` will cause an error if the version is a pre-release, e.g. `v2.6.0-rc1`. See https://github.com/tensorflow/tensorflow/blob/ff68385595088304cf772086b9a259a65b007622/tensorflow/core/public/version.h#L35-L37

njzjz: I suggest to use a third-party class [Specifiers](https://packaging.pypa.io/en/latest/specifiers.html)

deepmd/utils/argcheck.py
```diff
@@ -600,6 +600,24 @@ def validation_data_args():  # ! added by Ziyao: new specification style for dat
...
+    doc_output_prec  = 'The precision for mixed precision params. " \
+        "The trainable variables precision during the mixed precision training process, " \
+        "supported options are float32 only currently.'
+    doc_compute_prec  = 'The precision for mixed precision compute. " \
+        "The compute precision during the mixed precision training process, "" \
+        "supported options are float16 only currently.'
+
+    args = [
+        Argument("output_prec", str, optional=True, default="float32", doc=doc_output_prec),
+        Argument("compute_prec", str, optional=False, default="float16", doc=doc_compute_prec),
```
njzjz, line 613: The default behavior is to enable mixed precision?

denghuilu: The `mixed_precision` session is optional within the training session(see line 617), so it's false by default. However, when one have set the mixed_precision session, one must provide the `compute_prec` key.

wanghan-iapcm (approved):

PR #1291, opened 2021-11-18
update guidelines for the number of threads

### LINKED ISSUE #1284: [BUG] _Multi-GPU version is much slower than the single GPU
**Summary**
I am confused that the Multi-GPU version is much slower than the single GPU.

How to install： conda install deepmd-kit=2.0.3=*cpu libdeepmd=2.0.3=*cpu lammps-dp=2.0.0 horovod -c https://conda.deepmodeling.org
GPU: NVIDIA Tesla V100 16Gb
In my system, one core has four GPU

### Test results
In both cases, the same input file was used:
If with one GPU:
`dp train --mpi-log=master input.json   1>> train.log 2>> train.log`
the run speed is like this:
image

horovodrun -np 2  dp train --mpi-log=master input.json   1>> train.log 2>> train.log
the run speed is like this:
image

### Another problem
when training the model with version 2.0.2/3, It took me more than 1 hour to load the data, but it only took me a few mins if with version 2.0.0.b2.  why?

njzjz:
close #1284.

### DIFF
--- a/deepmd/env.py
+++ b/deepmd/env.py
@@ -126,6 +126,14 @@ def set_tf_default_nthreads():
     `TF_INTRA_OP_PARALLELISM_THREADS` and `TF_INTER_OP_PARALLELISM_THREADS`
     control TF configuration of multithreading.
     """
+    if "OMP_NUM_THREADS" not in os.environ or \
+       "TF_INTRA_OP_PARALLELISM_THREADS" not in os.environ or \
+       "TF_INTER_OP_PARALLELISM_THREADS" not in os.environ:
+        logging.warning(
+            "To get the best performance, it is recommended to adjust "
+            "the number of threads by setting the environment variables "
+            "OMP_NUM_THREADS, TF_INTRA_OP_PARALLELISM_THREADS, and "
+            "TF_INTER_OP_PARALLELISM_THREADS.")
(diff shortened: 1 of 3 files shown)


deepmd/env.py
```diff
@@ -126,6 +126,14 @@ def set_tf_default_nthreads():
...
     control TF configuration of multithreading.
     """
+    if "OMP_NUM_THREADS" not in os.environ or \
+       "TF_INTRA_OP_PARALLELISM_THREADS" not in os.environ or \
+       "TF_INTER_OP_PARALLELISM_THREADS" not in os.environ:
+        logging.warning(
+            "To get the best performance, it is recommended to adjust "
+            "the number of threads by setting the environment variables "
+            "OMP_NUM_THREADS, TF_INTRA_OP_PARALLELISM_THREADS, and "
+            "TF_INTER_OP_PARALLELISM_THREADS.")
```
wanghan-iapcm, line 136: Is it a good suggestion when the user train with GPU? 
I think a user only need to adjust them when he/she uses CPU.

njzjz: #1284 is confirmed to be caused by improper configuration of threads.

wanghan-iapcm (approved):

PR #1297, opened 2021-11-19
updata_deepmd_input when compress

njzjz:
This ensures the old model can be compressed by the new program.


denghuilu (approved):

wanghan-iapcm (approved):
Approved, but I find a typo: updata -> update

PR #1302, opened 2021-11-22
Fix bug with cudamemset's parameter

ZeyuLi0123:
cudaMemset and memset only accept integer type to fill the memory.
1.Alter template function: memset_device_memory
2.I notice there are many uses of memset or cudamemset use float type to fill the memory, but It is strange there isn't ant compile error....


wanghan-iapcm:
It is an interesting bug, I agree with your fix.

wanghan-iapcm (approved):

denghuilu (approved):

Issue #1299, opened 2021-11-19
[BUG] dp compress fail


neojie:
version: deepmd-kit 2.0.3, cpu version
description: dp compress -i pv.pb -o comp.pb -t input_v2_compat.json 
error:
```
2021-11-19 15:00:00.761448: I tensorflow/stream_executor/platform/default/dso_loader.cc:53] Successfully opened dynamic library libcudart.so.11.0
WARNING:tensorflow:From /u/home/j/jd848/project-lstixrud/dp2/dp2gpuconda/lib/python3.9/site-packages/tensorflow/python/compat/v2_compat.py:96: disable_resource_variables (from tensorflow.python.ops.variable_scope) is deprecated and will be removed in a future version.
Instructions for updating:
non-resource variables are not supported in the long term
DEEPMD INFO    stage 0: compute the min_nbor_dist
2021-11-19 15:00:14.806285: I tensorflow/core/platform/cpu_feature_guard.cc:142] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  SSE4.1 SSE4.2
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
[...]

PR #1306, opened 2021-11-22
Fix model compression bug when fparam or aparam is not zero

denghuilu:
See also #1299


wanghan-iapcm (approved):

wanghan-iapcm:
Please resolve the conflict

PR #1323, opened 2021-11-30
optimize `DPTabulate._build_lower` method

njzjz:
This commit optimizes the execution time of `DPTabulate._build_lower` from 1s to 0.1s.

I confirm that the returned value (`self.data[net]`) does not change.


denghuilu:
Why these changes matter to numpy's execution performance?

njzjz:
> Why these changes matter to numpy's execution performance?

https://numpy.org/devdocs/user/whatisnumpy.html

wanghan-iapcm (approved):

PR #1360, opened 2021-12-16
enable OpenMP for `prod_force` and `prod_virial`

njzjz:
About 1 ms can be saved in each training step.

### DIFF
--- a/source/lib/src/prod_force.cc
+++ b/source/lib/src/prod_force.cc
@@ -36,15 +36,17 @@ prod_force_a_cpu(
 
   memset(force, 0.0, sizeof(FPTYPE) * nall * 3);
   // compute force of a frame
+  #pragma omp parallel
   for (int i_idx = 0; i_idx < nloc; ++i_idx) {
     // deriv wrt center atom
+    #pragma omp single
     for (int aa = 0; aa < ndescrpt; ++aa) {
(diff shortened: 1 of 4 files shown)


source/lib/src/prod_force.cc
```diff
@@ -36,14 +36,17 @@ prod_force_a_cpu(
 
   memset(force, 0.0, sizeof(FPTYPE) * nall * 3);
   // compute force of a frame
+  #pragma omp parallel
```
wanghan-iapcm, line 39: Two threads with different i_idx may write on the same force `force[j_idx * 3 + xxx]`

njzjz: I realized that, so that `pragma omp for` is inside this loop and before another loop. The current version can pass the tests.

wanghan-iapcm: OK I understand you compute the neighbors of the same atom using multi-threading. 

What I do not understand is why you need `omp parallel` to generate threads here, but not at L49 where you really need multi-threading

njzjz: [image](https://www.openmp.org/wp-content/uploads/openmp-webinar-vanderPas-20210318.pdf)

wanghan-iapcm (approved):

denghuilu (approved):

PR #1361, opened 2021-12-16
Finish model compression for se_r descriptor!

liangadam:
Now model compression is available for se_r descriptor.

The compressed result has been checked with lammps , both on CPU and GPU. It's correct and the deviation is within 1e-8 in 1000 steps.

Acceleration performance: CPU 300%, GPU 120% (may be due to the small test system).

PR #1391, opened 2021-12-30
fix network precision under specific situation

njzjz:


deepmd/descriptor/se_a.py
```diff
@@ -736,7 +736,7 @@ def _filter_lower(
               if (not self.uniform_seed) and (self.seed is not None): self.seed += self.seed_shift
           else:
             # we can safely return the final xyz_scatter filled with zero directly
-            return tf.cast(tf.fill((natom, 4, outputs_size[-1]), 0.), GLOBAL_TF_FLOAT_PRECISION)
+            return tf.cast(tf.fill((natom, 4, outputs_size[-1]), 0.), self.filter_precision)
```
amcadmus, line 739: We cast the result back to global precision before return

njzjz: I don't see any `cast` in line 722 or line 747

njzjz: Is there something wrong?

wanghan-iapcm: > I don't see any `cast` in line 722 or line 747

747 should be cast back...

Why do we need to cast at L722?

njzjz: Instead of `_filter_lower`, I think we should cast back at line 839 before `_filter` returns.

wanghan-iapcm: I agree with you

wanghan-iapcm: Why not writing a decorator for doing that? It casts the inputs of `_filter` to `filter_precision` and casts back to global precision when `_filter` returns

njzjz: Good idea!

deepmd/common.py
```diff
@@ -487,3 +487,75 @@ def get_np_precision(precision: "_PRECISION") -> np.dtype:
...
+    cast it.
+    """
+    if tf.is_tensor(input) and input.dtype == from_precision:
+        return tf.cast(input, to_precision)
+    return input
+
+
+def cast_precision(func: Callable) -> Callable:
+    """A decorator that casts and casts back the input
+    and output tensor of a method.
```
wanghan-iapcm, line 522: Could you please write more on the logic behind `cast_precision`? 
- It casts i tensor from `GLOBAL_TF_FLOAT_PRECISION` to precision defined by property `precision`.
- It casts o tensor from `precision` to `GLOBAL_TF_FLOAT_PRECISION`.
- It checks the i/o list and only cast when an i or o is tensor and the tensor matches `GLOBAL_TF_FLOAT_PRECISION` or `precision`, respectively.

wanghan-iapcm (approved):

deepmd/descriptor/se_r.py
```diff
@@ -392,15 +392,15 @@ def _pass_filter(self,
                                      [ 0, start_index*      self.ndescrpt],
                                      [-1, natoms[2+type_i]* self.ndescrpt] )
                 inputs_i = tf.reshape(inputs_i, [-1, self.ndescrpt])
-                layer = self._filter_r(tf.cast(inputs_i, self.filter_precision), type_i, name='filter_type_'+str(type_i)+suffix, natoms=natoms, reuse=reuse, trainable = trainable, activation_fn = self.filter_activation_fn)
+                layer = self._filter_r(self.filter_precision, type_i, name='filter_type_'+str(type_i)+suffix, natoms=natoms, reuse=reuse, trainable = trainable, activation_fn = self.filter_activation_fn)
```
liangadam, line 395: The first variable should be "inputs_i" instead of “self.filter_precision”. It couldn't pass UT but it was merged into devel.

njzjz: In fact, this line is not covered by UT.

liangadam: Yes, it's covered by source/tests/test_model_compression_se_r.py, a UT just added in PR #1361.

Issue #1380, opened 2021-12-23
KeyError: "Registering two gradient with name 'Gelu'!


wanghan-iapcm:
### Discussed in https://github.com/deepmodeling/deepmd-kit/discussions/1371

<div type='discussions-op-text'>

<sup>Originally posted by **wankiwi** December 22, 2021</sup>
Hi, everyone.
I installed Deepmd-kit 2.0.3 successfully from the source code without any error message. But when I tried to train a model, I got the following error message.

`**KeyError: "Registering two gradient with name 'Gelu'! (Previous registration was in register /public/software/apps/DeepLearning/Tensorflow/tensorflow_2.3.3-rocm_3.9.1/lib/python3.6/site-packages/tensorflow/python/framework/registry.py:66)"**`

How to solve it? Thanks in advance.
PS: my Tensorflow version is tensorflow_rocm-2.3.3 on a machine with dcu 
[slurm-236155.zip](https://github.com/deepmodeling/deepmd-kit/files/7760756/slurm-236155.zip)

</div>

foxbrucecui:
Hello,Sincerely expect get your help.I also run Deepmd-kit in dcu.But I can only use 1 DCU when training models.
I found Horovod in the [install from source code.md](https://github.com/horovod/horovod/blob/master/docs/gpus.rst) to support multiple GPU acceleration,
I also found ROCM's compatibility for NCCL https://github.com/ROCmSoftwarePlatform/rccl
I follow it step by step,but it didn't work.
Do you have experience in using multiple DCU training?
Heartfelt thanks.

njzjz:
Fixed in #1406.

njzjz:
> Hello,Sincerely expect get your help.I also run Deepmd-kit in dcu.But I can only use 1 DCU when training models. I found Horovod in the [install from source code.md](https://github.com/horovod/horovod/blob/master/docs/gpus.rst) to support multiple GPU acceleration, I also found ROCM's compatibility for NCCL https://github.com/ROCmSoftwarePlatform/rccl I follow it step by step,but it didn't work. Do you have experience in using multiple DCU training? Heartfelt thanks.

@foxbrucecui The horovod should be built against MPI and NCCL.

Jack9959:
[...]

PR #1406, opened 2022-01-10
fix gelu grad multi definitions error

denghuilu:
#1380 should be fixed by this PR.

### DIFF
--- a/deepmd/common.py
+++ b/deepmd/common.py
@@ -2,6 +2,7 @@
 
 import json
 import warnings
+import tensorflow
 from functools import wraps
 from pathlib import Path
 from typing import (
@@ -64,7 +65,11 @@ def gelu(x: tf.Tensor) -> tf.Tensor:
     Original paper
     https://arxiv.org/abs/1606.08415
     """
-    return op_module.gelu(x)
+    try:
+        gelu = tensorflow.nn.gelu
+    except AttributeError:
+        gelu = op_module.gelu
+    return gelu(x)
 
 
 # TODO this is not a good way to do things. This is some global variable to which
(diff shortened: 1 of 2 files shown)


deepmd/common.py
```diff
@@ -64,7 +65,11 @@ def gelu(x: tf.Tensor) -> tf.Tensor:
     Original paper
     https://arxiv.org/abs/1606.08415
     """
-    return op_module.gelu(x)
+    try:
+        gelu = tensorflow.nn.gelu
+    except AttributeError:
+        gelu = op_module.gelu
```
wanghan-iapcm, line 71: `tf.nn.gelu` v.s. `op_module.gelu`, which is faster? 

If the latter is faster can we use it by default?

denghuilu: The later is indeed fast, but sometimes an error will be reported during a long MD process, we need to be more careful.

denghuilu: Maybe we could follow the implementations of tensorflow and do some more tests to figure out the reasons of this particular error.

wanghan-iapcm (approved):

njzjz:
Isn't it a breaking change from approximate to non-approximate? Also, the behavior will differ among different TensorFlow versions.

denghuilu:
> Isn't it a breaking change from approximate to non-approximate? Also, the behavior will differ among different TensorFlow versions.

Maybe we use the approximate

> Isn't it a breaking change from approximate to non-approximate? Also, the behavior will differ among different TensorFlow versions.

We use the approximate version of gelu, then we check the gelu errors later in another issue. Let's fix the multi definitions of gelu grad first.

njzjz (approved):

PR #1410, opened 2022-01-11
move `@docstring_parameter` to correct object

njzjz:


njzjz:
It's conflict with the equation....

PR #1416, opened 2022-01-13
only test/eval fitting properties

njzjz:
During profiling, I found that the virial will be calculated in the test process even if I don't provide virial data.
This is meaningless and the viral has never been outputted.
This commit removes such behavior.

(cherry picked from commit 9a55069b40dc2892adeed8b153c67ba2d407a62b)

image


ZiyaoLi:
looks fine to me.

ZiyaoLi (approved):
looks fine to me.

wanghan-iapcm (approved):

PR #1417, opened 2022-01-13
remove the dependency on `inputs` from `inputs_zero`

njzjz:
When profiling on multiple CPU threads, I notice that `inputs_zero` always runs after descriptor is computed. However, Some threads have nothing to do during `ProdEnvMatA`.

image

`inputs_zero` can be used to fill waiting time and runs with `ProdEnvMatA` at the same time...

image

(cherry picked from commit e704b4af1df7c23465615454bf7dfc0fc2f8d18f)

### DIFF
--- a/deepmd/fit/ener.py
+++ b/deepmd/fit/ener.py
@@ -332,6 +332,7 @@ class EnerFitting (Fitting):
     def build (self, 
                inputs : tf.Tensor,
                natoms : tf.Tensor,
+               nframes : tf.Tensor,
                input_dict : dict = {},
                reuse : bool = None,
                suffix : str = '', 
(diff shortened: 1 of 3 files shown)


wanghan-iapcm (commented):
It is not a good idea to change the **interface** of the `EnerFitting`  for the sake of little performance improvement.

njzjz:
> It is not a good idea to change the **interface** of the `EnerFitting` for the sake of little performance improvement.

I can add it to `input_dict` instead.

njzjz:
Done. It will also work when there's nothing in the `input_dict`.

wanghan-iapcm (approved):
I approve this PR, but I really do not like the `input_dict`, which makes the interfaces unclear...

PR #1418, opened 2022-01-14
enable TF remapper optimizer

njzjz:
TF supports a [remapper optimizer](https://www.tensorflow.org/guide/graph_optimization) which remaps subgraphs onto more efficient implementations by replacing commonly occuring subgraphs with optimized fused monolithic kernels. However, its support is limited: (1) `MatMul` + `BiasAdd` (not `Add`) + Activation; (2) Float32 (but not float64); (3) Activation is Tanh; (4) MKL is built and used.
This commit replaces `Add` by `BiasAdd` in the NN. The speed of a single op can be improved by about 20% when TF is using MKL and precision is set to float32. One can find `_MklNativeFusedMatMul` op in the profiler.

image
Original graph. Ops include MklMatMul, AddV2, and Tanh.

image
New graph. `_MklNativeFusedMatMul` is used here.

See also:
- https://www.tensorflow.org/guide/graph_optimization
- https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/grappler/optimizers/remapper.cc

(cherry picked from commit 8f2dc4404ddfc6308e3d42d29a35ee98f5d5cddd)

### DIFF
--- a/deepmd/utils/network.py
+++ b/deepmd/utils/network.py
@@ -56,7 +56,7 @@ def one_layer(inputs,
             w = tf.cast(w, get_precision(mixed_prec['compute_prec']))
             b = tf.cast(b, get_precision(mixed_prec['compute_prec']))
 
-        hidden = tf.matmul(inputs, w) + b
+        hidden = tf.nn.bias_add(tf.matmul(inputs, w), b)
         if activation_fn != None and use_timestep :
             idt_initializer = tf.random_normal_initializer(
                                     stddev=0.001,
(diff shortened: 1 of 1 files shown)


wanghan-iapcm (commented):
Does this change compatible to tf 1.x?

njzjz:
This PR only introduces [`tf.nn.bais_add`](https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/nn/bias_add), which is available in tf v1 and it is a recommended way to add the biases.

The support for tanh + MKL is introduced in https://github.com/tensorflow/tensorflow/pull/42173.

I don't know why only FP32 is supported.

njzjz:
I found that [Intel OneDNN](https://oneapi-src.github.io/oneDNN/dev_guide_data_types.html) 
does not support fp64 at all...

wanghan-iapcm (approved):

denghuilu (approved):

njzjz:
I have some other ideas: we can customize a remapper optimizer. TF has provided the [interface](https://github.com/tensorflow/tensorflow/blob/r2.7/tensorflow/core/grappler/optimizers/custom_graph_optimizer.h).

Issue #1469, opened 2022-02-10
[BUG] Mixed precision training


denghuilu:
**Summary**

We may have some troubles with the mixed precision training:
```
root water $ cd se_e2_a_mixed_prec
root se_e2_a_mixed_prec $ ls
input.json
root se_e2_a_mixed_prec $ dp train input.json
WARNING:tensorflow:From /root/denghui/tensorflow_venv/lib/python3.7/site-packages/tensorflow/python/compat/v2_compat.py:101: disable_resource_variables (from tensorflow.python.ops.variable_scope) is deprecated and will be removed in a future version.
Instructions for updating:
non-resource variables are not supported in the long term
WARNING:root:To get the best performance, it is recommended to adjust the number of threads by setting the environment variables OMP_NUM_THREADS, TF_INTRA_OP_PARALLELISM_THREADS, and TF_INTER_OP_PARALLELISM_THREADS.
WARNING:deepmd.train.run_options:Switch to serial execution due to lack of horovod module.
DEEPMD INFO    Calculate neighbor statistics... (add --skip-neighbor-stat to skip this step)
DEEPMD INFO    training data with min nbor dist: 0.8854385688525511
DEEPMD INFO    training data with max nbor size: [38, 72]
DEEPMD INFO     _____               _____   __  __  _____           _     _  _
DEEPMD INFO    |  __ \             |  __ \ |  \/  ||  __ \         | |   (_)| |
DEEPMD INFO    | |  | |  ___   ___ | |__) || \  / || |  | | ______ | | __ _ | |_
DEEPMD INFO    | |  | | / _ \ / _ \|  ___/ | |\/| || |  | ||______|| |/ /| || __|
DEEPMD INFO    | |__| ||  __/|  __/| |     | |  | || |__| |        |   < | || |_
DEEPMD INFO    |_____/  \___| \___||_|     |_|  |_||_____/         |_|\_\|_| \__|
DEEPMD INFO    Please read and cite:
DEEPMD INFO    Wang, Zhang, Han and E, Comput.Phys.Comm. 228, 178-184 (2018)
DEEPMD INFO    installed to:         /root/denghui/deepmd-kit/_skbuild/linux-x86_64-3.7/cmake-install
DEEPMD INFO    source :              v2.0.2-80-g0d8fe0a-dirty
DEEPMD INFO    source brach:         devel
DEEPMD INFO    source commit:        0d8fe0a
[...]

PR #1471, opened 2022-02-11
fix bug of mixed precision training

denghuilu:
#1469 should be fixed by this PR.

### DIFF
--- a/deepmd/descriptor/se_a.py
+++ b/deepmd/descriptor/se_a.py
@@ -738,7 +738,7 @@ class DescrptSeA (DescrptSe):
             # we can safely return the final xyz_scatter filled with zero directly
             return tf.cast(tf.fill((natom, 4, outputs_size[-1]), 0.), self.filter_precision)
           # natom x nei_type_i x out_size
-          xyz_scatter = tf.cast(tf.reshape(xyz_scatter, (-1, shape_i[1]//4, outputs_size[-1])), self.filter_precision)  
+          xyz_scatter = tf.reshape(xyz_scatter, (-1, shape_i[1]//4, outputs_size[-1]))  
           # When using tf.reshape(inputs_i, [-1, shape_i[1]//4, 4]) below
           # [588 24] -> [588 6 4] correct
           # but if sel is zero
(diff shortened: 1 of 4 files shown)


wanghan-iapcm (commented):
Do wen have similar issues in other descriptors?

denghuilu:
> Do wen have similar issues in other descriptors?

No, only `se_a` type support the mixed precision training.

deepmd/descriptor/se_a.py
```diff
@@ -738,7 +738,7 @@ def _filter_lower(
             # we can safely return the final xyz_scatter filled with zero directly
             return tf.cast(tf.fill((natom, 4, outputs_size[-1]), 0.), self.filter_precision)
           # natom x nei_type_i x out_size
-          xyz_scatter = tf.reshape(xyz_scatter, (-1, shape_i[1]//4, outputs_size[-1]))  
+          xyz_scatter = tf.cast(tf.reshape(xyz_scatter, (-1, shape_i[1]//4, outputs_size[-1])), self.filter_precision)
```
njzjz, line 741: If I understand correctly, it causes the error that [`deepmd.utils.network.embedding_net`](https://github.com/deepmodeling/deepmd-kit/blob/82c787de0ea98e3c032368a5fd4ee46a1fac6283/deepmd/utils/network.py#L99) accepts an input with `self.filter_precision` precision but returns one with `self.mixed_prec`. It is inconsistent. You should both cast and cast back inside the `embedding_net` or outside that method, but do not cast inside the `embedding_net` and cast back here.

njzjz (changes requested):

deepmd/utils/network.py
```diff
@@ -79,15 +79,13 @@ def one_layer(inputs,
...
-                    return tf.reshape(activation_fn(hidden), [-1, outputs_size]) * idt
+                    hidden = tf.reshape(activation_fn(hidden), [-1, outputs_size]) * idt
                 else :
-                    return tf.reshape(activation_fn(hidden), [-1, outputs_size])                    
-        else:
-            if useBN:
-                None
-                # return self._batch_norm(hidden, name=name+'_normalization', reuse=reuse)
-            else:
-                return hidden
```
wanghan-iapcm, line 90: why the case `activation_fn == None` is removed?

njzjz (approved):

PR #1476, opened 2022-02-14
calculate neighbor statistics from CLI

njzjz:
In many cases, we need to confirm neighbor statistics before training.
This command provides such CLI option.

### DIFF
--- a/source/tests/test_neighbor_stat.py
+++ b/source/tests/test_neighbor_stat.py
@@ -20,7 +20,7 @@ def gen_sys(nframes):
     return data
 
 
-class TestEnerShift(unittest.TestCase):
+class TestNeighborStat(unittest.TestCase):
     def setUp(self):
         data0 = gen_sys(1)
         sys0 = dpdata.LabeledSystem()
@@ -30,7 +30,7 @@ class TestEnerShift(unittest.TestCase):
     def tearDown(self):
         shutil.rmtree('system_0')
 
-    def test_ener_shift(self):
+    def test_neighbor_stat(self):
         # set rcut to 0. will cause a core dumped
         # TODO: check what is wrong
         for rcut in (3., 6., 11.):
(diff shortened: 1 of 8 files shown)


wanghan-iapcm (commented):
Could you please:
1. add to the document showing how it works. 
2. add a unittest for `neighbor_stat`
Thanks!

source/tests/test_neighbor_stat.py
```diff
@@ -0,0 +1,47 @@
...
+        data0 = gen_sys(1)
+        sys0 = dpdata.LabeledSystem()
+        sys0.data = data0
+        sys0.to_deepmd_npy('system_0', set_size = 10)
+        
+    def tearDown(self):
+        shutil.rmtree('system_0')
+
+    def test_neighbor_stat(self):
+        # set rcut to 0. will cause a core dumped
```
njzjz, line 34: @denghuilu setting `rcut` to `0.` will cause "core dumped". Any ideas?

source/tests/test_neighbor_stat.py
```diff
@@ -0,0 +1,47 @@
...
+        # TODO: check what is wrong
+        for rcut in (3., 6., 11.):
```
njzjz, line 36: @denghuilu setting `rcut` to `0.` will cause "core dumped". Any ideas?

njzjz: (Although, I don't think some one will set it to 0)

wanghan-iapcm (approved):
LGTM

source/tests/test_neighbor_stat.py
denghuilu: It might be caused by this line: https://github.com/deepmodeling/deepmd-kit/blob/159e45d248b0429844fb6a8cb3b3a201987c8d79/source/lib/src/neighbor_list.cc#L665

PR #1477, opened 2022-02-15
assign energy shift stats if atomic energies are assigned

njzjz:
Atomic energies stats may be incorrect. Here, we use assigned atomic energies instead.
I am training a model against multiple systems, where some atoms should not contribute energies. Before this commit, the loss of my model could not decrease. After this commit, it decreased quickly.

### DIFF
--- a/deepmd/fit/ener.py
+++ b/deepmd/fit/ener.py
@@ -129,6 +129,7 @@ class EnerFitting (Fitting):
             self.trainable = [self.trainable] * (len(self.n_neuron)+1)
         assert(len(self.trainable) == len(self.n_neuron) + 1), 'length of trainable should be that of n_neuron + 1'
         self.atom_ener = []
+        self.atom_ener_v = atom_ener
         for at, ae in enumerate(atom_ener):
             if ae is not None:
                 self.atom_ener.append(tf.constant(ae, self.fitting_precision, name = "atom_%d_ener" % at))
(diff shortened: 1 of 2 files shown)


wanghan-iapcm (commented):
Could you please add a UT for your update? 
For example you may provide 1 `atom_ener` for a three component system, and see if the computed `energy_shift` exactly use the assigned `atom_ener` and fits the other two element from the data.  
(I hope I understood your PR correctly..)

deepmd/fit/ener.py
```diff
@@ -197,8 +197,19 @@ def _compute_output_stats(self, all_stat, rcond = 1e-3):
             sys_tynatom = np.append(sys_tynatom, data[ss][0].astype(np.float64))
         sys_tynatom = np.reshape(sys_tynatom, [nsys,-1])
         sys_tynatom = sys_tynatom[:,2:]
+        if len(self.atom_ener) > 0:
+            # Atomic energies stats are incorrect if atomic energies are assigned.
+            # In this situation, we directly use these assigned energies instead of computing stats.
+            # This will make the loss decrease quickly
+            tot_assigned_ener = np.sum((ee for ee in self.atom_ener_v if ee is not None))
```
njzjz, line 204: I realize this is not correct

wanghan-iapcm (approved):

PR #1483, opened 2022-02-18
add an interface to eval descriptors

njzjz:
Fix #1393.


deepmd/infer/deep_pot.py
```diff
@@ -229,30 +233,29 @@ def eval(
...
-                e, f, v = self.auto_batch_size.execute_all(self._eval_inner, numb_test, natoms,
-                              coords, cells, atom_types, fparam = fparam, aparam = aparam, atomic = atomic, efield = efield)
-            else:
-                e, f, v = self._eval_inner(coords, cells, atom_types, fparam = fparam, aparam = aparam, atomic = atomic, efield = efield)
-            if self.modifier_type is not None:
-                me, mf, mv = self.dm.eval(coords, cells, atom_types)
-                e += me.reshape(e.shape)
-                f += mf.reshape(f.shape)
-                v += mv.reshape(v.shape)
-            return e, f, v
```
wanghan-iapcm, line 255: Any reason why `auto_batch_size` is removed?

wanghan-iapcm (commented):
I would recommend provide an separate method for inferring the descriptors, because strictly speaking, they are intermediate variable,  not the output of a model. The `eval` method is provided to infer the output of a model.

deepmd/infer/deep_pot.py
njzjz: Not removed. Instead, it's moved into a wrapper to make the code more clear.

deepmd/infer/deep_pot.py
```diff
@@ -231,30 +262,20 @@ def eval(
...
 
-    def _eval_inner(
+        if self.modifier_type is not None:
+            if atomic:
+                raise RuntimeError('modifier does not support atomic modification')
+            me, mf, mv = self.dm.eval(coords, cells, atom_types)
+            e, f, v = output[:3]
+            e += me.reshape(e.shape)
+            f += mf.reshape(f.shape)
+            v += mv.reshape(v.shape)
```
wanghan-iapcm, line 275: `output` is not updated with me, mf, mv before return.

wanghan-iapcm (changes requested):

deepmd/infer/deep_pot.py
njzjz: I don't think so. Here is a minimal example:
```py
>>> import numpy as np
>>> output = (np.ones(3), np.ones(3), np.ones(3), np.ones(3), np.ones(3))
>>> e,f,v = output[:3]
>>> e += 1
>>> f += 2
>>> v += 3
>>> output
(array([2., 2., 2.]), array([3., 3., 3.]), array([4., 4., 4.]), array([1., 1., 1.]), array([1., 1., 1.]))
```

See https://stackoverflow.com/a/35910888/9567349. In Python, `a+=b` is different from `a=a+b`. `a=a+b` creates a new object, but `a+=b` modifies the original object.

wanghan-iapcm: by ` e,f,v = output[:3]` e, f, v are copies of the list, not refs of the list element. You may do a test as 
```python
In [1]: a = [1,2,3]                                                             

In [2]: a[2]+=1                                                                 

In [3]: a                                                                       
Out[3]: [1, 2, 4]

In [4]: e,f,v=a[:3]                                                             

In [5]: e+=1                                                                    

In [6]: f+=1                                                                    

In [7]: v+=1                                                                    

In [8]: a                                                                       
Out[8]: [1, 2, 4]

In [9]: e, f, v                                                                 
Out[9]: (2, 3, 5)

```

wanghan-iapcm: I can reproduce your example. 
It seems that python does not copy the numpy arrays , but do copy numbers. 
Adding modification in this way confuses the readers of the code, I suggest you changing it to 
```python
output[0] += me.reshape(e.shape)
...
```

wanghan-iapcm (changes requested):

wanghan-iapcm (approved):

PR #1487, opened 2022-02-19
run test_python in the pre-built container

njzjz:


njzjz:
ok - I think it finally works

amcadmus (approved):

PR #1482, opened 2022-02-18
recover input stats from frozen models

njzjz:
Before, only NN parameters were recovered.

PR #1573, opened 2022-03-14
adjust sel of frozen models

njzjz:
This PR adds the folling features:

- save checkpoints when `numb_steps == 0`
- fix a typo in #1482
- restore `sel` in the graph
- support recovering when sel != original_sel (only when `set_davg_zero` is `true`)

So, when `set_davg_zero` is `true`, one can use `dp train input.json -f graph.pb` to get a new model, where `input.json` decreased or increased `sel` and set `numb_steps` to zero, and then freeze the model. Two models should get the same result (descriptors, energies, forces) when sel is enough for both models. The UT confirms it.
As discussed, the theory behind it is also clear.

Here we may have the following motivations to use this feature:
(1) One can train a model that sel fits the training data to get the best performance during training, and then increase the sel, to prevent sel is not enough;
(2) One can continue to train a small-sel model against the large-sel data, without retraining parameters;
(3) One can train a model for mixed systems with different sel, but run simulations for each system with small-sel model (even 0 if a system does not contain a element) to get best performance.

### DIFF
--- a/deepmd/descriptor/se_a.py
+++ b/deepmd/descriptor/se_a.py
@@ -13,7 +13,8 @@ from deepmd.utils.network import embedding_net, embedding_net_rand_seed_shift
 from deepmd.utils.tabulate import DPTabulate
 from deepmd.utils.type_embed import embed_atom_type
 from deepmd.utils.sess import run_sess
-from deepmd.utils.graph import load_graph_def, get_tensor_by_name_from_graph
+from deepmd.utils.graph import load_graph_def, get_tensor_by_name_from_graph, get_tensor_by_name
+from deepmd.utils.errors import GraphWithoutTensorError
 from .descriptor import Descriptor
 from .se import DescrptSe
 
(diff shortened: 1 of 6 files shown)


wanghan-iapcm (approved):
LGTM

wanghan-iapcm:
Could you add the introduction on how to use this feature into doc (e.g. doc/train/training-advanced.md)?

njzjz:
Sure.

wanghan-iapcm (approved):

deepmd/descriptor/se_a.py
```diff
@@ -827,7 +830,12 @@ def _filter(
           # inputs_reshape = tf.reshape(inputs, [-1, shape[1]//4, 4])
           # natom x 4 x outputs_size
           # xyz_scatter_1 = tf.matmul(inputs_reshape, xyz_scatter, transpose_a = True)
-          xyz_scatter_1 = xyz_scatter_1 * (4.0 / shape[1])
+          if self.original_sel is None:
+              # shape[1] = nnei * 4
+              nnei = shape[1] / 4
+          else:
+              nnei = np.sum(self.original_sel)
```
njzjz, line 837: I realized that adjusting `sel` again would lose this information.
We need to restore `nnei` forever.

PR #1574, opened 2022-03-16
restore original_sel in the graph

njzjz:
#1573 has an issue: when the frozen graph is recovered again, `original_sel` is not the initial one (but the new one).
In this commit, `original_sel` will be saved into the graph.


wanghan-iapcm (approved):
LGTM

PR #1575, opened 2022-03-16
fix model compression bug of nan output

### LINKED ISSUE #1444: [BUG] compressed model produces NaN forces
**Summary**

Here I have a system. Predicted forces by an uncompressed model:
image

Forces predicted by the compressed model:
image

We can see most of atoms have the same forces, but some of forces are NaN. (Python interface)

devel, f5b0abb7acc63aec7bd144ee914f0290bc690f05

**Steps to Reproduce**
I upload the model and data here: [1.zip](https://github.com/deepmodeling/deepmd-kit/files/7922038/1.zip).
The compress model is too big to upload, but it should be able to compress by `dp compress`.

dpdata's [`predict` method](https://github.com/deepmodeling/dpdata/blob/306172d51f08a63c7214efbbf40b2f25ba8c0484/dpdata/system.py#L1198) can be used here.

denghuilu:
This PR should also fix #1444 .

### DIFF
--- a/source/lib/src/cuda/tabulate.cu
+++ b/source/lib/src/cuda/tabulate.cu
@@ -164,7 +164,7 @@ __global__ void tabulate_fusion_se_a_fifth_order_polynomial(
   int breakpoint = nnei - 1;
 
   FPTYPE sum[MTILE] = {0.f};
-  int mark_table_idx = 0;
+  int mark_table_idx = -1;
   FPTYPE var[6];
   for (int ii = 0; ii < nnei; ii++) {
     FPTYPE xx = em_x[block_idx * nnei + ii];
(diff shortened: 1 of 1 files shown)


njzjz (approved):
I approve this PR. In addition, could you explain which situation will cause the bug, which is even not covered by the UT?

denghuilu:
> I approve this PR. In addition, could you explain which situation will cause the bug, which is even not covered by the UT?

This problem was caused by a previous optimization, which reuse the GPU registers to reduce the global memory access time. However, within that implementation, when the `table_idx` value obtained for the first time is 0, the initialization operation of the registers would be prevented. This is a logic error in programming, which will occur in rare cases.

denghuilu:
> > I approve this PR. In addition, could you explain which situation will cause the bug, which is even not covered by the UT?
> 
> This problem was caused by a previous optimization, which reuse the GPU registers to reduce the global memory access time. However, within that implementation, when the `table_idx` value obtained for the first time is 0, the initialization operation of the registers would be prevented. This is a logic error in programming, which will occur in rare cases.

see #1274 .

PR #1584, opened 2022-03-20
Devel

chtchelkatchev:
Hello. When working with deepmd, it is often necessary to perform thermodynamic integration, e.g., in order to build phase diagrams. There is an unpublished officially package dpti. But its documentation is incomplete and, moreover, the thermodynamic integration procedure there looks overly complicated compared to standard methods. Lampps has built-in thermodynamic integration routines. Are they compatible with deepmd? If yes, please provide at least one example with thermodynamic integration in the deepmd documentation. Thanks.


wanghan-iapcm:
Please raise it in the discussion panel of this repository https://github.com/deepmodeling/deepmd-kit/discussions 
I close this PR.

Issue #1583, opened 2022-03-20
[feature request] dp convert supports models trained with v0.12, v1.0 and v1.1 


wanghan-iapcm:
### Discussed in https://github.com/deepmodeling/deepmd-kit/discussions/1581

<div type='discussions-op-text'>

<sup>Originally posted by **zcb-code** March 19, 2022</sup>
Is that possible to added  v0.12, v1.0 and v1.1 with easy conda install? Because some models have already trained with these versions and supplied in supporting information. And it is difficult to install for a beginner.
</div>

The better way to solve the discussion is to support model conversion from v0.12, v1.0 and v1.1 to v2.

PR #1587, opened 2022-03-22
support `dp convert-from 1.1`

njzjz:
#1583
It looks like there's no breaking changes between v1.1 and v1.2. https://github.com/deepmodeling/deepmd-kit/compare/v1.1.5...v1.2.0


wanghan-iapcm:
For energy mode, yes. The portability and dipole fitting supported in 1.2 cannot be used in 1.1.

wanghan-iapcm (approved):

Issue #1589, opened 2022-03-23
[BUG] _"compute stress/atom" not working in Lammps_


DingChangjie:
### Bug summary

======= Description ===========
"compute stress/atom" command gives all-zero components, however "thermo" command still works and gives correct total stress of the system. 
======= Supporting information ==========
The Lammps binary file in this issue is distributed together with deepmd-kit, located in /bin/lmp.
To confirm this issue, an EAM empirical potential is also tested (see "Input Files, Running Commands, Error Log, etc.") as a comparison.

### DeePMD-kit Version

v2.0.3-1-g96e623d5

### TensorFlow Version

2.7.0

### How did you download the software?

Offline packages

### Input Files, Running Commands, Error Log, etc.

# Related files
All the related files and test results are in the Bugreport.zip file.
calc_stress.in : Lammps input file.
DP_test.pb : A DP potential for tungsten.
2013_Marinica_EAM2 : a well-known EAM potential.
relaxed.lmp : a configuration containing two interstitial loops, which should give non-zero atomic stress.
[Bugreport.zip](https://github.com/deepmodeling/deepmd-kit/files/8330170/Bugreport.zip)

### Steps to Reproduce

Simply modify and run calc_stress.in

wanghan-iapcm:
Please read the [link](https://github.com/deepmodeling/deepmd-kit/blob/master/doc/third-party/lammps-command.md#use-of-the-centroidstressatom-to-get-the-full-3x3-atomic-virial) to see if the `centroid/stress/atom` compute solves the issue.

DingChangjie:
> Please read the [link](https://github.com/deepmodeling/deepmd-kit/blob/master/doc/third-party/lammps-command.md#use-of-the-centroidstressatom-to-get-the-full-3x3-atomic-virial) to see if the `centroid/stress/atom` compute solves the issue.

It seems that the default Lammps binary in version 2.0.3 is not compiled with centroid/stress/atom support ... Here is the error message :

CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE

WARNING: No fixes defined, atoms won't move (../verlet.cpp:55)
ERROR: Pair style does not support compute centroid/stress/atom (../compute_centroid_stress_atom.cpp:131)
Last command: run 1

DingChangjie:
[...]

PR #1605, opened 2022-03-28
fix the atomic virial bug introduced by lammps PR #2481

amcadmus:
The compute style centroid/stress/atom does not work due to [lammps/lammps#2481](https://github.com/lammps/lammps/pull/2481) 

I change the constant of `centroidstressflag` according to the [definition](https://github.com/lammps/lammps/blob/e1835250c7351a2ba0145e448753216f8346168c/src/pair.h#L71-L74).

LAMMPS version : 24Mar2022

DeePMD-kit version: devel

This PR fixes the issue #1589


njzjz (commented):
Do we want to drop support for the old version? If not, using `LAMMPS_VERSION_NUMBER` may help

amcadmus:
Checked against LAMMPS 20201029, the outputs are consistent.

njzjz (approved):

PR #1610, opened 2022-03-29
refactor init_variable and support type embedding

### LINKED ISSUE #1578: refactor `Descriptor.init_variables`
### Summary

Making `Descriptor.init_variables` pass a GraphDef instead of a filename.

### Detailed Description

Now this method accepts the filename as input, so we need to open the graph every time. As discussed with @denghuilu , of course, we should pass the graph_def to improve performance.
https://github.com/deepmodeling/deepmd-kit/blob/1503fc3a10eb4b875431016d05285e51a5f22af9/deepmd/descriptor/descriptor.py#L353

_No response_

njzjz:
Fix #1578. This commit refactors init_variable for both Descriptor and Fitting, so they can accept graph and graph_def instead of the filename. Opening the same file multiple times is bad behavior.
Moves some codes in Trainer to Model. Indeed Trainer should not operate Descriptor and Fitting directly.
Add init_variable method to type_embedding.

@denghuilu - I think `enable_compression` may have some similar issues.

### DIFF
--- a/deepmd/fit/polar.py
+++ b/deepmd/fit/polar.py
@@ -6,7 +6,7 @@ from deepmd.env import tf
 from deepmd.common import add_data_requirement, cast_precision, get_activation_func, get_precision, ACTIVATION_FN_DICT, PRECISION_DICT, docstring_parameter
 from deepmd.utils.argcheck import list_to_doc
 from deepmd.utils.network import one_layer, one_layer_rand_seed_shift
-from deepmd.utils.graph import get_fitting_net_variables
+from deepmd.utils.graph import get_fitting_net_variables_from_graph_def
 from deepmd.descriptor import DescrptLocFrame
 from deepmd.descriptor import DescrptSeA
 from deepmd.fit.fitting import Fitting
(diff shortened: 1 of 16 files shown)


deepmd/fit/polar.py
```diff
@@ -511,17 +517,23 @@ def build (self,
...
-            The input frozen model file
+        graph : tf.Graph
+            The input frozen model graph
+        graph_def : tf.GraphDef
+            The input frozen model graph_def
+        suffix : str
+            suffix to name scope
         """
-        self.polar_fitting.init_variables(model_file)
+        self.polar_fitting.init_variables(graph=graph, graph_def=graph_def, suffix=suffix)
```
njzjz, line 536: It's unclear why this class does not inherit from `self.polar_fitting` directly...

wanghan-iapcm: The class `GlobalPolarFitting` is deprecated. `PolarFitting` can to all the things that the `GlobalPolarFitting` is designed to do.

wanghan-iapcm (approved):

denghuilu (approved):

denghuilu:
> 

Yes, enable_compress still passes the frozen_model to model compression process, which also can be replaced by the graph as well as graph_def.

PR #1618, opened 2022-04-01
add system names to model devi header

njzjz:
Otherwise, it's challenging to know which system the model deviation belongs to when multiple systems exist.

### DIFF
--- a/deepmd/infer/model_devi.py
+++ b/deepmd/infer/model_devi.py
@@ -44,7 +44,7 @@ def calc_model_devi_v(vs: np.ndarray):
     avg_devi_v = np.linalg.norm(vs_devi, axis=-1) / 3
     return max_devi_v, min_devi_v, avg_devi_v
 
-def write_model_devi_out(devi: np.ndarray, fname: str):
+def write_model_devi_out(devi: np.ndarray, fname: str, header: str=""):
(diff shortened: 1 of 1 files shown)


deepmd/infer/model_devi.py
```diff
@@ -44,17 +44,19 @@ def calc_model_devi_v(vs: np.ndarray):
...
     devi : numpy.ndarray
         the first column is the steps index
     fname : str
         the file name to dump
+    header : str, default=""
+        the header to dump
     '''
     assert devi.shape[1] == 7
-    header = "%10s" % "step"
+    header = "%s: %10s" % (header, "step")
```
wanghan-iapcm, line 59: Shall we write the header as a separated commented line? like 
```
# header ...
# step  max_devi.... 
```

njzjz: This can be implemented by adding a line breaking `\n`.

njzjz: (https://github.com/numpy/numpy/blob/b235f9e701e14ed6f6f6dcba885f7986a833743f/numpy/lib/npyio.py#L1422-L1423)

wanghan-iapcm (approved):

PR #1628, opened 2022-04-12
Fix compilation error while in a ROCm environment

denghuilu:


source/lib/src/rocm/tabulate.hip.cu
```diff
@@ -524,15 +524,6 @@ __global__ void tabulate_fusion_se_r_grad_fifth_order_polynomial(
...
   int lane_idx = threadIdx.x % 64;
 
-  bool unloop = false;
-  FPTYPE * iteratorA = (FPTYPE *)&_data[0]; // dy
-  for (int ii = 0; ii < MTILE; ii++) {
-    for (int jj = thread_idx; jj < last_layer_size; jj += blockDim.x) {
-      iteratorA[ii * last_layer_size + jj] = ;
-    }
-  }
-  __syncthreads();
```
njzjz, line 534: Are you sure that this line should also be removed?

denghuilu: Yes, there are no shared memory access operations

denghuilu:
@wanghan-iapcm After carefully checking, I found a variable declaration error within the test_tabulate_se_r.cc file. After fixing that error, all UTs have passed in the api_cc and lib folder.

source/lib/src/rocm/tabulate.hip.cu
njzjz: I am wondering - I do see this line in CUDA implementation 
https://github.com/denghuilu/deepmd-kit/blob/97c99e62d9f19285b68a834ee57dc7a0a084db03/source/lib/src/cuda/tabulate.cu#L557

wanghan-iapcm:
> @wanghan-iapcm After carefully checking, I found a variable declaration error within the test_tabulate_se_r.cc file. After fixing that error, all UTs have passed in the api_cc and lib folder within the ROCm environment.

That's great! Thanks!

source/lib/src/rocm/tabulate.hip.cu
denghuilu: > I am wondering - I do see this line in CUDA implementation
> 
> https://github.com/deepmodeling/deepmd-kit/blob/97c99e62d9f19285b68a834ee57dc7a0a084db03/source/lib/src/cuda/tabulate.cu#L557

The __syncthreads function is used to synchronise threads within the entire CUDA block, but these threads do not conflict with each other under the current implementation(There is a conflict in the previous implementation, and it is a waste to keep this __syncthreads), so this function is no longer needed.

njzjz (approved):

PR #1648, opened 2022-04-19
debug gtest

njzjz:


njzjz:
Looks like a newly updated conda package contains the whole google test namespace in its library, breaking the building...

PR #1665, opened 2022-04-28
use net-wise tabulate range

### LINKED ISSUE #1388: [Feature Request] reduce compressed graph size
**Summary**

The compressed graph size should be reduced.

**Detailed Description**

When there are many atomic types, the compressed graph size will be too large, causing it failed to freeze.

As discussed with @denghuilu, the table can be different for different types.

njzjz:
This PR uses different tabulate ranges for different NNs, in order to reduce the graph size.

Resolves #1388.

### DIFF
--- a/deepmd/utils/tabulate.py
+++ b/deepmd/utils/tabulate.py
@@ -414,8 +414,8 @@ class DPTabulate():
             upper = ((1 / min_nbor_dist) * sw - self.davg[:, 0]) / self.dstd[:, 0]
         elif isinstance(self.descrpt, deepmd.descriptor.DescrptSeT):
             var = np.square(sw / (min_nbor_dist * self.dstd[:, 1:4]))
-            lower = -var
-            upper = var
+            lower = np.min(-var, axis=1)
+            upper = np.max(var, axis=1)
         elif isinstance(self.descrpt, deepmd.descriptor.DescrptSeR):
             lower = -self.davg[:, 0] / self.dstd[:, 0]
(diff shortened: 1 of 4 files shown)


deepmd/utils/tabulate.py
```diff
@@ -161,81 +163,100 @@ def build(self,
...
+                xx = np.append(xx, np.arange(upper[ii], extrapolate * upper[ii], stride1, dtype = self.data_type))
+                xx = np.append(xx, np.array([extrapolate * upper[ii]], dtype = self.data_type))
+                xx_all.append(xx)
+            nspline = ((upper - lower) / stride0 + 2 * ((extrapolate * upper - upper) / stride1)).astype(int)
             idx = 0
             for ii in range(self.ntypes):
                 for jj in range(ii, self.ntypes):
                     net = "filter_" + str(ii) + "_net_" + str(jj)
-                    self._build_lower(net, xx, idx, upper, lower, stride0, stride1, extrapolate)
+                    self._build_lower(net, xx_all[ii], idx, upper[ii], lower[ii], stride0, stride1, extrapolate, nspline[ii])
```
denghuilu, line 205: Should it be changed to nspline instead of nspline[ii]?
```suggestion
                    self._build_lower(net, xx_all[ii], idx, upper[ii], lower[ii], stride0, stride1, extrapolate, nspline)
```

njzjz: Here `nspline` is an 1-D array (different from others).

denghuilu (approved):

PR #1668, opened 2022-04-28
rewrite data doc

njzjz:
Current data doc was wrttien in the inital version. It's out-of-date.


doc/data/system.md
```diff
@@ -0,0 +1,45 @@
+# System
+
+DeePMD-kit takes a **system** as data structure. A snapshot of a system is called a **frame**. A system may contain multiple frames with the same atom types and numbers, i.e. the same formula (like `H2O`). To contains data with different formula, one need to divide data into multiple systems.
+
+A system should contain system properties, input frame properties, and labeled frame properties. The system property contains the following property:
+
+ID       | Property                | Required/Optional    | Shape                    | Description
```
wanghan-iapcm, line 7: Also provide the name of the raw file?

wanghan-iapcm (commented):
Shall we also provide the raw file name for the frame properties and labels? I would recommend we do so.

wanghan-iapcm (approved):

PR #1669, opened 2022-04-29
migrate sphinx mathjax from jsdelivr to cdnjs

njzjz:
jsdelivr, which is the default mathjax path in the sphinx, has been blocked in China. See jsdelivr/jsdelivr#18392. Of course, we hope our rendered equations can be shown in China.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -251,3 +251,4 @@ html_css_files = ['css/custom.css']
 autodoc_default_flags = ['members']
 autosummary_generate = True
 master_doc = 'index'
+mathjax_path = 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/3.2.0/es5/tex-mml-chtml.min.js'


wanghan-iapcm:
How about the speed of rendering outside China?

njzjz:
It is amazing to know it's unblocked...

njzjz:
> Just out of interst, why not consider cdnjs mirrors available in China while having excellent global speeds, such as https://cdn.baomitu.com and https://cdn.bytedance.com

@hifocus baomitu was down last year for two weeks (https://www.v2ex.com/t/758137).

There is no evidence showing Bytedance CDN has excellent global speeds... It looks like they only have one server outside China.
image

hifocus:
> > Just out of interst, why not consider cdnjs mirrors available in China while having excellent global speeds, such as https://cdn.baomitu.com and https://cdn.bytedance.com
> 
> @hifocus baomitu was down last year for two weeks (https://www.v2ex.com/t/758137).
> 
> There is no evidence showing Bytedance CDN has excellent global speeds... It looks like they only have one server outside China. image

It was for ICP registration reasons.

For btyedance cdn, try `s{1-4}.pstatp.com`. If I remembered correctly, s1 was Aliyun global CDN, s2 was Wangsu, jsDelivr's China CDN provider in the past.

njzjz:
More than half of readers to our doc are outside the mainland China, so we cannot improve performance in China by sacrificing the speed in other regions.

hifocus:
> More than half of readers to our doc are outside the mainland China, so we cannot improve performance in China by sacrificing the speed in other regions.

Exactly, that's why my recommendations are globally optimised CDNs, including China and the world.

However I do respect your thoughts and ideas; therefore please only took my view as a friendly advise :)

PR #1697, opened 2022-05-10
fix grappler compilation error with TF 1.15 ~ 2.6

njzjz:
reported by @denghuilu


denghuilu (approved):
It works with TF-2.4.0 on my local workstation!

PR #1704, opened 2022-05-13
in model_devi, assumes nopbc if box is set to None

wanghan-iapcm:
The doc string says
```
Box to specify periodic boundary condition. If None, no pbc will be used
```
but it is not implemented as that. 

Fix the issue makes the behavior of model_devi the same as `DeepPot.eval`

### DIFF
--- a/deepmd/infer/model_devi.py
+++ b/deepmd/infer/model_devi.py
@@ -125,7 +125,7 @@ def calc_model_devi(coord,
     >>> graphs = [DP("graph.000.pb"), DP("graph.001.pb")]
     >>> model_devi = calc_model_devi(coord, cell, atype, graphs)
     '''
-    if not box:
+    if box is not None:
         nopbc = True
     else:
         nopbc = False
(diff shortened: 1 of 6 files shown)


deepmd/infer/model_devi.py
```diff
@@ -127,8 +125,10 @@ def calc_model_devi(coord,
     >>> graphs = [DP("graph.000.pb"), DP("graph.001.pb")]
     >>> model_devi = calc_model_devi(coord, cell, atype, graphs)
     '''
-    if nopbc:
-        box = None
+    if not box:
```
njzjz, line 128: This will not work when `box` is `np.ndarray`.

```py
>>> import numpy as np
>>> not np.ones(3)
  File "<stdin>", line 1, in <module>
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

So `is not None` should be used instead.
```suggestion
    if box is not None:
```

njzjz (changes requested):

njzjz (changes requested):
This line still uses old parameter:
https://github.com/deepmodeling/deepmd-kit/blob/9ccff11d2bcae61b98a5ee7885e6fdf4e9b50027/deepmd/infer/model_devi.py#L212

njzjz (approved):

PR #1706, opened 2022-05-13
nvnmd

LiuGroupHNU:
Our training procedure consists of not only the continuous neural network (CNN) training, but also the quantized neural network (QNN) training which uses the results of CNN as inputs,  to help users train machine-learning models that are compatible with the unique non von Neumann computer.


deepmd/descriptor/se_a.py
```diff
@@ -3,7 +3,7 @@
 from typing import Tuple, List, Dict, Any
 
 from deepmd.env import tf
-from deepmd.common import get_activation_func, get_precision, ACTIVATION_FN_DICT, PRECISION_DICT, docstring_parameter, cast_precision
+from deepmd.common import get_activation_func, get_precision, ACTIVATION_FN_DICT, PRECISION_DICT, docstring_parameter
```
njzjz, line 6: Your submission reverts a lot of recent commits. You might commit based on the wrong parent commit.

njzjz (changes requested):

wanghan-iapcm:
I close this PR because it tries to revert recent comments on deepmd-kit.

PR #1715, opened 2022-05-17
debug setuptools-scm

njzjz:


njzjz:
related issue: https://github.com/actions/checkout/issues/760

PR #1729, opened 2022-05-25
add JIT support to gelu

njzjz:
With JIT, float32 perforamce is improved, but float64 performance is decreased.

Usage:
```sh
export DP_JIT_LEVEL=1
```

### DIFF
--- a/deepmd/common.py
+++ b/deepmd/common.py
@@ -20,7 +20,7 @@ from typing import (
 import numpy as np
 import yaml
 
-from deepmd.env import op_module, tf
+from deepmd.env import op_module, tf, jit_wrapper
 from tensorflow.python.framework import tensor_util
 from deepmd.env import GLOBAL_TF_FLOAT_PRECISION, GLOBAL_NP_FLOAT_PRECISION
 from deepmd.utils.sess import run_sess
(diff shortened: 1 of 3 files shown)


wanghan-iapcm (approved):

deepmd/common.py
```diff
@@ -67,9 +67,13 @@ def gelu(x: tf.Tensor) -> tf.Tensor:
     """
     def gelu_wrapper(x):
         try:
-            return tensorflow.nn.gelu(x, approximate=True)
+            tensorflow.nn.gelu
```
wanghan-iapcm, line 70: What is the influence of using approximate=False instead of the approximate=True ? 
How does it get the env `DP_JIT_LEVEL`?

njzjz: This line is to check if TF supports gelu... The actual line is 74-76.

> How does it get the env DP_JIT_LEVEL?
line 76 does such things.

njzjz:
I found a better way to enable JIT... I will submit another PR.

deepmd/common.py
wanghan-iapcm: I see Thanks!

PR #1730, opened 2022-05-25
add JIT option

njzjz:
Here are my results (2080 Super, TF 2.9, CUDA 11.6):

(unit: ms/step)
  | no JIT | JIT
-- | -- | --
water tanh fp32 | 8.45 | 7.78
water tanh fp64 | 33.37 | 34.09
water gelu fp32 | 19.69 | 14.23
water gelu fp64 | 48.39 | 49.19

Conclusion:
1. TF JIT only works with FP32 and slows down FP64 (or my card only has good FP32 performance?).
2. JIT has a great performance on TF's GeLU.
3. JIT only works for static shapes, so training with different natoms (or different atom types) needs a lot of compilation time (not sure if it's worth it).
4. JIT only works for a GPU.

Usage
```sh
export DP_JIT=1
```
(It sets `TF_XLA_FLAGS=--tf_xla_auto_jit=2` but `DP_JIT=1` is easier to remember)


njzjz:
@wanghan-iapcm, any reviews for this PR?

wanghan-iapcm (approved):

PR #1737, opened 2022-06-01
using int64 within the memory allocation operations

denghuilu:
We use int32 to store the size or shape of tensors in the graph. And the multiplication of two large integers is often used as a memory allocation parameter, which is very dangerous as the number of atoms grows. Currently, a single GPU with 32 GiB or more device memory can process up to 240,000 atoms and may cause a potential waste of device resources.

To solve this problem, I have modified the code as follows:

1. Replace some `int` with `int64` within the `*_multi_device.cc` files to avoid memory allocation failure.
2. Change the related CUDA and ROCm implementations.
3. Avoid possible out-of-bounds reshape operations within the `deepmd/fit/ener.py` file.
4. The CPU part remains unchanged, simply because a single CPU core cannot handle such a large system.

After these changes, I tested using the 80 GiB A100 GPU with the example water system. 

Double precision, compressed model, max number of atoms: 835,584
```
Step PotEng KinEng TotEng Temp Press Volume 
       0 -1.3031679e+08    35642.501 -1.3028114e+08          330    4889.2972    8387686.3 
      10 -1.3031618e+08    35032.193 -1.3028114e+08    324.34939   -750.61864    8387686.3 
Loop time of 9.2185 on 1 procs for 10 steps with 835584 atoms

Performance: 0.047 ns/day, 512.139 hours/ns, 1.085 timesteps/s
100.5% CPU use with 1 MPI tasks x no OpenMP threads
```

Single precision, compressed model, max number of atoms: 1,951,488
```
Step PotEng KinEng TotEng Temp Press Volume 
       0 -3.0430224e+08    83242.333 -3.0421899e+08          330    21683.542     19589256 
      10 -3.043003e+08      81313.6 -3.0421899e+08    322.35387    16440.116     19589256 
Loop time of 11.1804 on 1 procs for 10 steps with 1951488 atoms

Performance: 0.039 ns/day, 621.135 hours/ns, 0.894 timesteps/s
97.8% CPU use with 1 MPI tasks x no OpenMP threads
```

### DIFF
--- a/deepmd/fit/ener.py
+++ b/deepmd/fit/ener.py
@@ -277,7 +277,7 @@ class EnerFitting (Fitting):
         # cut-out inputs
         inputs_i = tf.slice (inputs,
                              [ 0, start_index, 0],
-                             [-1, natoms, self.dim_descrpt] )
+                             [-1, natoms, -1] )
         inputs_i = tf.reshape(inputs_i, [-1, self.dim_descrpt])
         layer = inputs_i
         if fparam is not None:
--- a/deepmd/fit/ener.py
+++ b/deepmd/fit/ener.py
@@ -276,8 +276,8 @@ class EnerFitting (Fitting):
     ):
         # cut-out inputs
         inputs_i = tf.slice (inputs,
-                             [ 0, start_index*      self.dim_descrpt],
-                             [-1, natoms* self.dim_descrpt] )
+                             [ 0, start_index, 0],
+                             [-1, natoms, self.dim_descrpt] )
         inputs_i = tf.reshape(inputs_i, [-1, self.dim_descrpt])
         layer = inputs_i
         if fparam is not None:
@@ -419,13 +419,13 @@ class EnerFitting (Fitting):
                                                 trainable = False,
                                                 initializer = tf.constant_initializer(self.aparam_inv_std))
             
-        inputs = tf.reshape(inputs, [-1, self.dim_descrpt * natoms[0]])
+        inputs = tf.reshape(inputs, [-1, natoms[0], self.dim_descrpt])
         if len(self.atom_ener):
             # only for atom_ener
             nframes = input_dict.get('nframes')
             if nframes is not None:
                 # like inputs, but we don't want to add a dependency on inputs
-                inputs_zero = tf.zeros((nframes, self.dim_descrpt * natoms[0]), dtype=self.fitting_precision)
+                inputs_zero = tf.zeros((nframes, natoms[0], self.dim_descrpt), dtype=self.fitting_precision)
             else:
                 inputs_zero = tf.zeros_like(inputs, dtype=self.fitting_precision)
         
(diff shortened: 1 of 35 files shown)


deepmd/fit/ener.py
```diff
@@ -276,8 +276,8 @@ def _build_lower(
     ):
         # cut-out inputs
         inputs_i = tf.slice (inputs,
-                             [ 0, start_index*      self.dim_descrpt],
-                             [-1, natoms* self.dim_descrpt] )
+                             [ 0, start_index, 0],
+                             [-1, natoms, self.dim_descrpt] )
```
njzjz, line 280: dipole and polar need the same change
https://github.com/deepmodeling/deepmd-kit/blob/10fdc930d25673ac979f78c8f43148c3fad4d47f/deepmd/fit/dipole.py#L133-L141
https://github.com/deepmodeling/deepmd-kit/blob/10fdc930d25673ac979f78c8f43148c3fad4d47f/deepmd/fit/polar.py#L309-L324

njzjz, line 280: Looks like descriptors have the same issue, for example
https://github.com/deepmodeling/deepmd-kit/blob/b534355be64f2e819ba38f9793558cee7e5eb912/deepmd/descriptor/se_a.py#L520-L521
https://github.com/denghuilu/deepmd-kit/blob/b534355be64f2e819ba38f9793558cee7e5eb912/deepmd/descriptor/se_a.py#L556-L557
https://github.com/deepmodeling/deepmd-kit/blob/b534355be64f2e819ba38f9793558cee7e5eb912/deepmd/descriptor/se_t.py#L417-L418
https://github.com/deepmodeling/deepmd-kit/blob/b534355be64f2e819ba38f9793558cee7e5eb912/deepmd/descriptor/se_t.py#L445-L446
https://github.com/deepmodeling/deepmd-kit/blob/b534355be64f2e819ba38f9793558cee7e5eb912/deepmd/descriptor/se_t.py#L452-L453
https://github.com/deepmodeling/deepmd-kit/blob/10fdc930d25673ac979f78c8f43148c3fad4d47f/deepmd/descriptor/hybrid.py#L192

I guess `sel` in your model is not big, so you won't catch an exception

njzjz, line 280: I think casting `natoms` from `int32` to `int64` will also work if you don't want to change dims.

njzjz: 2,147,483,647/4/1,951,488=275.1
You may test a bigger sum of `sel` like 300

wanghan-iapcm:
Have you run the UTs for `source/lib`, `source/api_cc` on GPU and DCU?

njzjz (approved):

denghuilu:
> > Have you run the UTs for `source/lib`, `source/api_cc` on GPU and DCU?
> 
> All UTs have passed on a GPU environment. The DCU tests will be performed tomorrow.

All UTs have passed on a DCU environment with TensorFlow 1.14 except the following TF error:
```
C++ exception with description "DeePMD-kit Error: TensorFlow Error: Not found: Op type not registered 'MatrixDiagV3' in binary running on e10r4n00. Make sure the Op and Kernel are registered in the binary running in this process. Note that if you are loading a saved graph which used ops from tf.contrib, accessing (e.g.) `tf.contrib.resampler` should be done before importing the graph, as contrib ops are lazily registered when the module is first accessed." thrown in SetUp().
[  FAILED  ] TestInferDeepPolarNew.cpu_lmp_nlist (387 ms)
```

PR #1738, opened 2022-06-02
Automatically label new pull requests based on the paths of files being changed

njzjz:
https://github.com/actions/labeler


wanghan-iapcm (approved):
Looks great!

PR #1741, opened 2022-06-04
replace GPU 1./sqrt with rsqrt

njzjz:
Per NVIDIA doc:
> 11.1.3. Reciprocal Square Root
> **The reciprocal square root should always be invoked explicitly as rsqrtf() for single precision and rsqrt() for double precision.** The compiler optimizes 1.0f/sqrtf(x) into rsqrtf() only when this does not violate IEEE-754 semantics.

See https://docs.nvidia.com/cuda/cuda-c-best-practices-guide/index.html#reciprocal-square-root


njzjz:
The unit test with CUDA passed as expected.

denghuilu (approved):
The unit test with ROCm also passed as expected.

PR #1787, opened 2022-06-26
Merge devel into master

amcadmus:

PR #1788, opened 2022-06-26
Merge master into devel

amcadmus:


njzjz:
Can we merge the commit with the new version tag into the devel branch after #1787 is merged? Every time the devel branch is one version behind as the new tag is not merged...

njzjz (approved):

PR #1794, opened 2022-06-28
Update index.rst

ChuLiangFu:
fix the link?


njzjz (changes requested):
The new link does not work.

Screenshot from 2022-06-28 15-16-54

PR #1798, opened 2022-06-29
DP Mask Model

dingye18:
I implement a DP Mask Model with self-defined `se_a_mask` op and `EnerForcesMaskLoss`.
The main contribution comes from two points.

1. Add the argument for `mask_matrix.npy` (shape [num_frames, num_tot_atoms]), used in se_a_mask.  User can specific on which atom is real or virtual. So that we can concatenate the frames which have different elements and atoms into one numpy file. 
2. Add the argument `mask_matrix4force.npy` (shape [ num_frames, num_tot_atoms * 3 ]). Actually, this is a coefficient matrix which used in the module `EnerForcesMaskLoss`. The forces losses on each atom will be multiplied by this matrix. User can specify the forces on which atom should be included in the training process. This would be benefit to the DP/MM hybrid potential.

Besides, I did implement `DeepPotMaskModel` for running of `se_a_mask` op in the C API interface. Integration of this model with OpenMM will be done in future.
`atom_num4element.npy` and `atom_num4frame.npy` are needed for loss normalization on each element and each frame in printing.


deepmd/model/ener.py
```diff
@@ -231,8 +231,13 @@ def build (self,
         energy_raw = tf.reshape(energy_raw, [-1, natoms[0]], name = 'o_atom_energy'+suffix)
         energy = tf.reduce_sum(global_cvt_2_ener_float(energy_raw), axis=1, name='o_energy'+suffix)
 
-        force, virial, atom_virial \
-            = self.descrpt.prod_force_virial (atom_ener, natoms)
+        
+        if hasattr(self.descrpt, 'descrpt_type') and self.descrpt.descrpt_type == "se_a_mask":
```
njzjz, line 235:
```py
isinstance(self.descrpt, DescrptSeAMask)
```

source/lib/src/fmt_nlist.cc
```diff
@@ -8,7 +8,7 @@
 
 using namespace deepmd;
 
-template<typename FPTYPE> 
+/*template<typename FPTYPE>
```
njzjz, line 11: Comments should not be used to remove codes

deepmd/descriptor/se_a_mask.py
```diff
@@ -0,0 +1,722 @@
...
+from deepmd.utils.network import embedding_net, embedding_net_rand_seed_shift
+from deepmd.utils.tabulate import DPTabulate
+from deepmd.utils.type_embed import embed_atom_type
+from deepmd.utils.sess import run_sess
+from deepmd.utils.graph import load_graph_def, get_tensor_by_name_from_graph
+from .descriptor import Descriptor
+from .se import DescrptSe
+
+
+@Descriptor.register("se_a_mask")
```
njzjz, line 21: We may call `se_e2_a_mask` as `se_a` has been renamed

deepmd/utils/argcheck.py
```diff
@@ -522,6 +552,7 @@ def loss_variant_type_args():
     return Variant("type", 
                    [Argument("ener", dict, loss_ener()),
                     Argument("tensor", dict, loss_tensor()),
+                    Argument("masked_energy_forces", dict, loss_ener())
```
njzjz, line 555: Document should be added into this `Argument`, otherwise it's hard to understand what `masked_energy_forces` is.

deepmd/loss/ener.py
```diff
@@ -384,4 +384,166 @@ def print_on_training(self,
         return print_str   
 
 
-
+class EnerForcesMaskLoss():
```
njzjz, line 387: Please provide document into this class

njzjz (changes requested):

source/lib/include/fmt_nlist.h
```diff
@@ -6,6 +6,30 @@
...
+      {
+      }
+  bool operator < (const NeighborInfo & b) const 
+      {
+	return (type < b.type || 
+		(type == b.type && 
+		 (dist < b.dist || 
+		  (dist == b.dist && index < b.index) ) ) );
+      }
+};
```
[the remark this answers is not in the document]
dingye18: Thanks for point out. I will fix this in next commit.

deepmd/loss/ener.py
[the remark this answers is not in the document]
dingye18: Got it. It will be fixed in next commit.

deepmd/descriptor/se_a_mask.py
```diff
@@ -0,0 +1,722 @@
...
+class DescrptSeAMask (Descriptor):
```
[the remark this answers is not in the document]
dingye18: Yes, it can be inherit `DescrptSeA`. The duplicated methods will be deleted in next commit.

source/api_cc/include/DeepPot.h
```diff
@@ -427,6 +427,92 @@ class DeepPotModelDevi
   // function used for nborlist copy
   std::vector<std::vector<int> > get_sel() const;
 };
+
+class DeepPotMaskModel: public DeepPot{
+public:
+  /**
+  * @brief DPMaskModel constructor without initialization.
```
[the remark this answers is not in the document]
dingye18: This is a C API interface class for the usage of `DescrptSeAMask`. But it's an under developed class. I haven't test it on a trained DP Mask graph.

source/lib/include/fmt_nlist.h
```diff
@@ -6,6 +6,30 @@
...
+struct NeighborInfo 
+{
+  int type;
+  FPTYPE dist;
+  int index;
+  NeighborInfo () 
+      : type (0), dist(0), index(0) 
+      {
+      }
+  NeighborInfo (int tt, double dd, int ii)
```
njzjz, line 19: `FPTYPE`

deepmd/descriptor/se_a_mask.py
```diff
@@ -0,0 +1,429 @@
...
+                            ]
+                
+                davg = np.tile(davgunit, self.ndescrpt // 4)
+                dstd = np.tile(dstdunit, self.ndescrpt // 4)
+                all_davg.append(davg)
+                all_dstd.append(dstd)
+
+        if not self.set_davg_zero:
+            self.davg = np.array(all_davg)
+        self.dstd = np.array(all_dstd)
```
njzjz, line 228: Correct me if I am wrong, but it looks like `descrpt_se_a_mask` never uses `avg` and `std`? Is it an expected behavior or something unimplemented?
The computing of `avg` and `std` looks useless.

deepmd/loss/ener.py
```diff
@@ -384,4 +384,169 @@ def print_on_training(self,
...
 
+    Parameters
+    ----------
+
+    """
+    def __init__(self,
+                 starter_learning_rate : float, 
+                  start_pref_e : float = 0.02,
+                  limit_pref_e : float = 1.00,
+                  start_pref_f : float = 1000,
```
njzjz, line 401: `1000` is indeed an `int` but not `float`.

```python
>>> type(1000)
int
```

deepmd/descriptor/se_a_mask.py
```diff
@@ -0,0 +1,429 @@
...
+
+        avg_zero = np.zeros([self.ntypes,self.ndescrpt]).astype(GLOBAL_NP_FLOAT_PRECISION)
+        std_ones = np.ones ([self.ntypes,self.ndescrpt]).astype(GLOBAL_NP_FLOAT_PRECISION)
+        sub_graph = tf.Graph()
+        with sub_graph.as_default():
+            name_pfx = 'd_sea_mask_'
+            for ii in ['coord']:
+                self.place_holders[ii] = tf.placeholder(GLOBAL_NP_FLOAT_PRECISION, [None, None], name = name_pfx+'t_'+ii)
+            self.place_holders['type'] = tf.placeholder(tf.int32, [None, None], name=name_pfx+'t_type')
+            self.place_holders['mask_matrix'] = tf.placeholder(tf.int32, [None, None], name=name_pfx+'t_mask')
```
njzjz, line 153: From the docs I see `mask_matrix` can be either `0` or `1`, so why not use `tf.bool`?

deepmd/loss/ener.py
```diff
@@ -384,4 +384,169 @@ def print_on_training(self,
...
+        self.has_e = (self.start_pref_e != 0.0 or self.limit_pref_e != 0.0)
+        self.has_f = (self.start_pref_f != 0.0 or self.limit_pref_f != 0.0)
+    
+        self.debug_op = []
+        
+        # data required
+        add_data_requirement('energy', 1, atomic=False, must=False, high_prec=True)
+        add_data_requirement('force',  3, atomic=True,  must=False, high_prec=False)
+        add_data_requirement('mask_matrix', 1, atomic=True, must=True, high_prec=False)
+        add_data_requirement('mask_matrix4force', 3, atomic=True, must=False, high_prec=False)
```
njzjz, line 424: The default "default" value is zero. For `mask_matrix4force`, should the default value set to one?

njzjz, line 424: Maybe a new argument is required here to assign type. Is it np.float32 by default?

source/api_cc/include/DeepPot.h
```diff
@@ -427,6 +427,92 @@ class DeepPotModelDevi
...
+  **/
+  DeepPotMaskModel(const std::string & model, const int & gpu_rank = 0, const std::string & file_content = "");
+  /**
+  * @brief Initialize the DP.
+  * @param[in] model The name of the frozen model file.
+  * @param[in] gpu_rank The GPU rank. Default is 0.
+  * @param[in] file_content The content of the model file. If it is not empty, DP will read from the string instead of the file.
+  **/
+  void init (const std::string & model, const int & gpu_rank = 0, const std::string & file_content = "");
+public:
```
njzjz, line 452: duplicate?

source/op/descrpt_se_a_mask.cc
```diff
@@ -0,0 +1,337 @@
...
+                    descrpt_deriv_atom[jj * 12 + 4] = 2. * x * y * inr4;        // on y.
+                    descrpt_deriv_atom[jj * 12 + 5] = 2. * x * z * inr4;        // on z.
+                    // derive of the component y/r2
+                    descrpt_deriv_atom[jj * 12 + 6] = 2. * y * x * inr4;        // on x.
+                    descrpt_deriv_atom[jj * 12 + 7] = 2. * y * y * inr4 - inr2; // on y.
+                    descrpt_deriv_atom[jj * 12 + 8] = 2. * y * z * inr4;        // on z.
+                    // derive of the component z/r2
+                    descrpt_deriv_atom[jj * 12 + 9] = 2. * z * x * inr4;         // on x.
+                    descrpt_deriv_atom[jj * 12 + 10] = 2. * z * y * inr4;        // on y.
+                    descrpt_deriv_atom[jj * 12 + 11] = 2. * z * z * inr4 - inr2; // on z.
```
njzjz, line 270: `1.` and `2.` is `float64` but it should be `compute_t` before it multiplies by another `compute_t`.

source/op/descrpt_se_a_mask.cc
```diff
@@ -0,0 +1,337 @@
...
+
+typedef double boxtensor_t;
+typedef double compute_t;
+
+REGISTER_OP("DescrptSeAMask")
+    .Attr("T: {float, double} = DT_DOUBLE")
+    .Input("coord: T")
+    .Input("type: int32")
+    .Input("mask: int32")
+    .Attr("total_atom_num: int")
```
njzjz, line 17: What's the difference between this attr and the length of `type`?

source/op/descrpt_se_a_mask.cc
```diff
@@ -0,0 +1,337 @@
...
+        OP_REQUIRES(context, (nsamples == type_tensor.shape().dim_size(0)), errors::InvalidArgument("number of samples should match"));
+        OP_REQUIRES(context, (nsamples == mask_matrix_tensor.shape().dim_size(0)), errors::InvalidArgument("number of samples should match"));
+
+        // Set n_descrpt for each atom. Include 1/rr, cos(theta), cos(phi), sin(phi)
+        int n_descrpt = 4;
+
+        // Create an output tensor
+        TensorShape descrpt_shape;
+        descrpt_shape.AddDim(nsamples);
+        descrpt_shape.AddDim(total_atom_num * total_atom_num * n_descrpt);
```
njzjz, line 63: This may need to be cast to `int64` for huge systems.

deepmd/descriptor/se_a_mask.py
njzjz: Note that compression relies on the correct avg and std. If not used, please set to 0 and 1.

https://github.com/deepmodeling/deepmd-kit/blob/e8a57caa127d3a6c7d262c3e2567a634a0c17259/deepmd/utils/tabulate.py#L397-L399

dingye18: Yes, for this version, it never uses `avg` and `std`. But I don't know that if any other functions, such as the methods in `se_a` or outside functions, will call this `avg` or `std`. This implementation is used for the compatibility of `DescrptSeAMask` with other ops.

deepmd/loss/ener.py
dingye18: OK, I will fix this.

deepmd/descriptor/se_a_mask.py
dingye18: That's a legacy code. `tf.bool` might be ok. But I haven't test it. I will try to change this into `tf.bool` in next commit.

deepmd/loss/ener.py
dingye18: Yes, I will make the default value to be 1.0 in next commit.

dingye18: You mean the type of `mask_matrix.....`? It looks like that `add_data_requirement` did provide the type limitation setting.

source/api_cc/include/DeepPot.h
dingye18: Sorry. I found that `DeepPotMaskModel` is under developed. That's not finished for inference. I will fix this later.

source/op/descrpt_se_a_mask.cc
dingye18: In principle, there is no different between the length of `type` and `total_atom_num`. The attr here is used for C API inference in `DeepPotMaskModel` and input data dimension checking.

source/op/descrpt_se_a_mask.cc
dingye18: In my thought, mask model will be applied small system, it can't afford a >200 particles system. Thus, int32 might be enough for `total_atom_num * total_atom_num * n_descrpt`.

deepmd/descriptor/se_a_mask.py
dingye18: > Note that compression relies on the correct avg and std. If not used, please set to 0 and 1.
> 
> https://github.com/deepmodeling/deepmd-kit/blob/e8a57caa127d3a6c7d262c3e2567a634a0c17259/deepmd/utils/tabulate.py#L397-L399

OK, it will be fixed.

deepmd/loss/ener.py
njzjz: Yes, I can see it does only support `np.float32` or `np.float64`..

https://github.com/deepmodeling/deepmd-kit/blob/03c874291586c139e47d56f1e029e0782332187a/deepmd/utils/data.py#L175-L178

deepmd/descriptor/se_a_mask.py
```diff
@@ -0,0 +1,429 @@
...
+                                         dstd.shape, 
+                                         dtype = GLOBAL_TF_FLOAT_PRECISION,
+                                         trainable = False,
+                                         initializer = tf.constant_initializer(dstd))
+            '''
+            
+        coord = tf.reshape (coord_, [-1, natoms[1] * 3])
+        box   = tf.reshape (box_, [-1, 9])
+        atype = tf.reshape (atype_, [-1, natoms[1]])
+        mask_matrix = input_dict["mask_matrix"]
```
njzjz, line 308: I am not quite understanding here - where does `input_dict["mask_matrix"]` come from? I don't see any related codes

dingye18: That's load from `mask_matrix.npy` in set directory as I used `add_data_requirement('mask_matrix', 1, atomic=True, must=True, high_prec=False)`. As I used this in `EnerForcesMaskLoss` class, the data system will load `mask_matrix.npy` into `input_dict` as the value of key `mask_matrix` automatically

njzjz: `add_data_requirement` is only in `EnerForcesMaskLoss`. Does it mean it must be used along with `EnerForcesMaskLoss`? Or may you also add `add_data_requirement` to `DescrptSeAMask`? (`data_requirement` is a dict so it will not be added twice)

njzjz:
Could you provide an unittest to ensure it can work?

deepmd/descriptor/se_a_mask.py
dingye18: OK. I will do this later.

dingye18:
> Could you provide an unittest to ensure it can work?

Of course. But it may take a few days. Since I'm a little busy today......

dingye18:
Many works are required for the consistent of dp_mask and DP 2.x version. Refactor of dp_mask implementation is needed. I will post a new PR and unit test cases later.

njzjz:
> Refactor of dp_mask implementation is needed.

@dingye18 @wanghan-iapcm @iProzd We can schedule a short meeting before refactoring.

wanghan-iapcm:
> 

Sure, @njzjz please help arranging the meeting, thanks!

PR #1800, opened 2022-06-30
add core api docs

njzjz:
Although, nothing in `libdeemd.so` is documented...


njzjz:
Available at https://deepmodeling--1800.org.readthedocs.build/projects/deepmd/en/1800/api_core/api_core.html

wanghan-iapcm (approved):

PR #1804, opened 2022-07-01
add op docs

njzjz:
Also add op docs to ProdEnvMatA Op to show an example of how to write Op docs.

### DIFF
--- a/source/op/prod_env_mat_multi_device.cc
+++ b/source/op/prod_env_mat_multi_device.cc
@@ -10,7 +10,7 @@ REGISTER_OP("ProdEnvMatA")
     .Attr("T: {float, double} = DT_DOUBLE")
     .Input("coord: T")          //atomic coordinates
     .Input("type: int32")       //atomic type
-    .Input("natoms: int32")     //local atomic number; each type atomic number; daizheyingxiangqude atomic numbers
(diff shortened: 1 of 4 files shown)


source/op/prod_env_mat_multi_device.cc
```diff
@@ -10,7 +10,7 @@ REGISTER_OP("ProdEnvMatA")
     .Attr("T: {float, double} = DT_DOUBLE")
     .Input("coord: T")          //atomic coordinates
     .Input("type: int32")       //atomic type
-    .Input("natoms: int32")     //local atomic number; each type atomic number; daizheyingxiangqude atomic numbers
```
njzjz, line 13: @denghuilu could you clarify what is "daizheyingxiangqude"? Is it 带着影像去的?

njzjz:
Preview at https://deepmodeling--1804.org.readthedocs.build/projects/deepmd/en/1804/api_op.html

source/op/prod_env_mat_multi_device.cc
denghuilu: It's a typo...

njzjz: 😂

wanghan-iapcm (approved):

PR #1862, opened 2022-08-17
revert prod_force OMP in #1360

njzjz:
Sometimes when the box is quite small (i.e. box size < 2 * rcut), the same atom may repeat to appear in the neighbor list built by DP. This cause inaccurate results when using OMP.


njzjz:
I also fix the failure of GitHub Actions due to the breaking change of setuptools 64.

denghuilu (approved):

wanghan-iapcm (approved):

PR #1866, opened 2022-08-19
Upload attention based model

iProzd:
Here I upload the implementation of attention based model (DPA-1), see details in paper: https://arxiv.org/abs/2208.08236.
The changes can mainly be listed as following:

1. Add a new descriptor '**se_atten**', using attention schemes and type embeddings (deepmd/descriptor/se_atten.py);

2. Add a new custom op '**ProdEnvMatAMix**', mixing atoms with different types in the sorting, which is only suitable for descriptor using type embeddings (source/op/prod_env_mat_multi_device.cc and other sub files, here I try the best to reuse the previous codes);  

3. Other changes to match the newly added descriptor, such as 'sel' supporting 'int' type as input (deepmd/entrypoints/train.py ) and type embedding with padding (deepmd/utils/type_embed.py);

4. **Mixed_type**: a new format of DeepmdData  to put different systems sharing the same 'nloc' (instead of same fingerprints) together. (deepmd/utils/data.py & data_system.py) Note that this is only the training code on this new format of data. The generation code of this new format will soon be uploaded in _dpdata_.

5. An example of DPA-1 on water system (examples/water/se_atten/input.json);

6. The unittests on both C++&CUDA&ROCM and python interface of newly added apis(source/lib/tests & source/tests).

Unittests passed individually for these newly added features. Thanks for reviewing and disccusions are welcome.


iProzd:
@amcadmus @njzjz

source/lib/src/prod_env_mat.cc
```diff
@@ -175,6 +175,111 @@ prod_env_mat_r_cpu(
...
+    const int nall, 
+    const float rcut, 
+    const float rcut_smth, 
+    const std::vector<int> sec) 
+{
+  const int nnei = sec.back();
+  const int nem = nnei * 4;
+  memset(em, 0.0, sizeof(FPTYPE) * nloc * nem);
+  memset(em_deriv, 0.0, sizeof(FPTYPE) * nloc * nem * 3);
+  memset(rij, 0.0, sizeof(FPTYPE) * nloc * nnei * 3);
```
njzjz, line 203: The second parameter of `memset` is passed as an `int`. See https://cplusplus.com/reference/cstring/memset/

deepmd/entrypoints/train.py
```diff
@@ -90,7 +90,10 @@ def train(
 
     jdata = normalize(jdata)
 
-    if not is_compress and not skip_neighbor_stat:
+    if jdata['model']['descriptor']['type'] in ['se_atten'] and isinstance(jdata['model']['descriptor']['sel'], list):
+        jdata['model']['descriptor']['sel'] = sum(jdata['model']['descriptor']['sel'])
+
+    if not is_compress and not skip_neighbor_stat and jdata['model']['descriptor']['type'] not in ['se_atten']:
         jdata = update_sel(jdata)
```
njzjz, line 97: I see se_atten is excluded, but why not first update_sel and then make the sum?

source/lib/src/neighbor_list.cc
```diff
@@ -820,6 +820,56 @@ build_nlist_cpu(
...
+    if(b_nlist_map){
+    for (int ii = 0; ii < nloc; ++ii){
+      for (int jj = 0; jj < nnei; ++jj){
+        int nlist_idx = ii*nnei+jj;
+        int record = nlist[nlist_idx];
+        if (record >= 0){	
+          int temp = nlist_map[record];
+          nlist[nlist_idx] = temp;	  
+          ntype[nlist_idx]=type[temp];
+          nmask[nlist_idx]=1.;
```
njzjz, line 846:
```
(FPTYPE)1.
```

njzjz, line 846: If only 1. and 0. are supported, why not use `bool`? It should use less memory.

examples/water/se_atten/input.json
```diff
@@ -0,0 +1,70 @@
+{
```
njzjz, line 1: Please add this file to [test_examples.py](https://github.com/deepmodeling/deepmd-kit/blob/c31b2835f677094431b3a630cd93c889e2d41f64/source/tests/test_examples.py).

deepmd/descriptor/se_atten.py
```diff
@@ -0,0 +1,766 @@
...
+            The excluded pairs of types which have no interaction with each other.
+            For example, `[[0, 1]]` means no interaction between type 0 and type 1.
+    set_davg_zero
+            Set the shift of embedding net input to zero.
+    activation_function
+            The activation function in the embedding net. Supported options are |ACTIVATION_FN|
+    precision
+            The precision of the embedding net parameters. Supported options are |PRECISION|
+    uniform_seed
+            Only for the purpose of backward compatibility, retrieves the old behavior of using the random seed
```
njzjz, line 55: Some parameters are missing.

deepmd/fit/ener.py
```diff
@@ -168,7 +168,8 @@ def get_numb_aparam(self) -> int:
         return self.numb_fparam
 
     def compute_output_stats(self, 
-                             all_stat: dict
+                             all_stat: dict,
+                             large_batch_mode: bool = False
```
njzjz, line 172: The new parameter was not added to the docstring.

deepmd/utils/data.py
```diff
@@ -524,6 +534,12 @@ def _check_pbc(self, sys_path: DPPath):
             pbc = False
         return pbc
 
+    def _check_mode(self, sys_path):
+        large_batch_mode = False
+        if os.path.isfile(os.path.join(sys_path, 'set.000', 'real_atom_types.npy')):
+            large_batch_mode = True
+        return large_batch_mode
```
njzjz, line 541: These four lines can be written as one line:
```py
return os.path.isfile(os.path.join(sys_path, 'set.000', 'real_atom_types.npy'))
```

njzjz (changes requested):
This is a preliminary review.

Besides, there is no documentation to describe it.

deepmd/utils/data.py
njzjz: Besides, use DPPath class

source/lib/src/prod_env_mat.cc
```diff
@@ -175,6 +175,111 @@ prod_env_mat_r_cpu(
   }
 }
 
+template<typename FPTYPE>
+void
+deepmd::
+prod_env_mat_a_mix_cpu(
```
wanghan-iapcm, line 181: We should be able to implement it within `prod_env_mat_a_cpu`. 

Adding one parameter `const int * f_type` with default value `type` to `prod_env_mat_a_cpu` would work.

wanghan-iapcm: The same comment for the cuda and rocm variances  of `prod_env_mat`

source/op/prod_env_mat_multi_device.cc
```diff
@@ -896,7 +956,314 @@ class ProdEnvMatROp : public OpKernel {
...
+	  context, &coord, coord_cpy, &f_type, type_cpy, idx_mapping, 
+	  inlist, ilist, numneigh, firstneigh, jlist,
+	  frame_nall, mem_cpy, mem_nnei, max_nbor_size,
+	  box, mesh_tensor.flat<int>().data(), nloc, nei_mode, rcut_r, max_cpy_trial, max_nnei_trial);
+      // launch the cpu compute function
+      deepmd::prod_env_mat_a_mix_cpu(
+	  em, em_deriv, rij, nlist, 
+	  coord, f_type, type, inlist, max_nbor_size, avg, std, nloc, frame_nall, rcut_r, rcut_r_smth, sec_a);
+      // do nlist mapping if coords were copied
+    _map_nei_info_cpu(nlist, ntype, nmask, type, &idx_mapping[0], nloc, nnei, ntypes, b_nlist_map);
```
wanghan-iapcm, line 1243: have you mapped the neighbor index when the coords are copied?

wanghan-iapcm (changes requested):
- The organization and the data format of the DPA model is different from the standard DP model. We should explain in the doc what is the difference and how to prepare data.

doc/model/train-se-atten.md
```diff
@@ -0,0 +1,109 @@
...
+   "Mg",
+   "Al",
+   "Cu"
+  ]
+```
+which should include all the elements in the dataset you want to train on. The detail of data format can be found in [here](data/data-conv.md).
+Note that this data format requires that, only those frames with the same fingerprint(i.e. the number of atoms of different element) can be put together as a unit system.
+This may lead to sparse frame number in those rare systems. 
+
+An ideal way is to put systems with same total number of atoms together, which is the way we trained DPA-1 on OC2M. This API will be uploaded on dpdata soon for a more convenient experience.
```
wanghan-iapcm, line 99: You should explain, in detail, how one prepares the training data for DPA-1. Someone may want to manually prepare data, rather than using dpdata.

wanghan-iapcm (changes requested):

doc/model/train-se-atten.md
```diff
@@ -0,0 +1,109 @@
+# DPA-1: Pretraining of Attention-based Deep Potential Model for Molecular Simulation 
+
+![ALT](../images/model_se_atten.png "model_se_atten")
+
+Here we propose DPA-1, a Deep Potential model with a novel attention mechanism, which is highly effective for representing the conformation and chemical spaces of atomic systems and learning the PES.
+
+See [this paper](https://arxiv.org/abs/2208.08236) for more information. DPA-1 is implemented as a new descriptor `"se_atten"` for model training, which can be used after simply editing the input.json.
+
+# Installation 
+DPA-1 will be merged into DeePMD-kit official repo: [github](https://github.com/deepmodeling/deepmd-kit), and before that, for early adopters, you can refer to the following steps for a quick start:
```
njzjz, line 10: This should not be merged

deepmd/fit/ener.py
```diff
@@ -495,6 +516,7 @@ def build (self,
                 [tf.reshape(inputs,[-1,self.dim_descrpt]),atype_embed],
                 axis=1
             )
+            self.tmp_input = inputs
```
njzjz, line 519: What is it used for?

njzjz:
Depth of TOC is wrong:
image

deepmd/utils/network.py
```diff
@@ -204,7 +204,10 @@ def embedding_net(xx,
             xx = tf.cast(xx, get_precision(mixed_prec['compute_prec']))
             w  = tf.cast(w,  get_precision(mixed_prec['compute_prec']))
             b  = tf.cast(b,  get_precision(mixed_prec['compute_prec']))
-        hidden = tf.reshape(activation_fn(tf.nn.bias_add(tf.matmul(xx, w), b)), [-1, outputs_size[ii]])
+        if activation_fn != None:
```
njzjz, line 207: Please use `is not None`. See https://peps.python.org/pep-0008/#programming-recommendations

source/op/prod_env_mat_multi_device.cc
```diff
@@ -147,6 +147,27 @@ REGISTER_OP("DescrptSeR")
...
+    .Attr("rcut_r: float")
+    .Attr("rcut_r_smth: float")
+    .Attr("sel_a: list(int)")
+    .Attr("sel_r: list(int)")   //all zero
+    .Output("descrpt: T")
+    .Output("descrpt_deriv: T")
+    .Output("rij: T")
+    .Output("nlist: int32")
+    .Output("ntype: int32")
+    .Output("nmask: T");;
```
njzjz, line 169: Maybe you could add a summary to the new OP, especially the difference from the old OP.

```cpp
.Doc(R"(Something

something something
something.)");
```

deepmd/utils/argcheck.py
```diff
@@ -239,19 +239,63 @@ def descrpt_hybrid_args():
...
+    - `int`. The maximum number of neighbor atoms to be considered. We recommend it to be less than 200. \n\n\
+    - `List[int]`. The length of the list should be the same as the number of atom types in the system. `sel[i]` gives the selected number of type-i neighbors. Only the summation of `sel[i]` matters, and it is recommended to be less than 200.'
+    doc_rcut = 'The cut-off radius.'
+    doc_rcut_smth = 'Where to start smoothing. For example the 1/r term is smoothed from `rcut` to `rcut_smth`'
+    doc_neuron = 'Number of neurons in each hidden layers of the embedding net. When two layers are of the same size or one layer is twice as large as the previous layer, a skip connection is built.'
+    doc_axis_neuron = 'Size of the submatrix of G (embedding matrix).'
+    doc_activation_function = f'The activation function in the embedding net. Supported activation functions are {list_to_doc(ACTIVATION_FN_DICT.keys())}. Note that "gelu" denotes the custom operator version, and "gelu_tf" denotes the TF standard version.'
```
njzjz, line 251: Duplicated period.
image

deepmd/utils/type_embed.py
```diff
@@ -90,10 +93,14 @@ def __init__(
         self.seed = seed
         self.filter_resnet_dt = resnet_dt
         self.filter_precision = get_precision(precision)
-        self.filter_activation_fn = get_activation_func(activation_function)
+        if activation_function is not None:
+            self.filter_activation_fn = get_activation_func(activation_function)
```
wanghan-iapcm, line 97: revise `get_activation_function` to allow the input being "none" and None?

deepmd/train/trainer.py
```diff
@@ -112,14 +115,25 @@ def _init_param(self, jdata):
...
                 activation_function=typeebd_param['activation_function'],
                 precision=typeebd_param['precision'],
                 trainable=typeebd_param['trainable'],
-                seed=typeebd_param['seed']
+                seed=typeebd_param['seed'],
+                padding=padding
+            )
+        elif descrpt_type == 'se_atten':
+            self.typeebd = TypeEmbedNet(
+                neuron=[8],
```
wanghan-iapcm, line 133: Why the `neuron` is hard coded?

doc/model/train-se-atten.md
```diff
@@ -0,0 +1,122 @@
...
+
+ID         | Property                         | File                | Required/Optional    | Shape                    | Description
+---------- | -------------------------------- | ------------------- | -------------------- | -----------------------  | -----------
+fake_type  | Atom type indexes (place holder) | type.raw            | Required             | Natoms                   | All zeros to fake the type input
+type_map   | Atom type names                  | type_map.raw        | Required             | Ntypes                   | Atom names that map to atom type contained in all the frames, which is unnecessart to be contained in the periodic table
+real_type  | Atom type indexes of each frame  | real_atom_types.npy | Required             | Nframes \* Natoms        | Integers that describe atom types in each frame, corresponding to indexes in type_map
+type_number| Number of atoms in each type     | real_atom_numbs.npy | Required             | Nframes \* Ntypes        | Integers that describe the totoal number of atoms with the same type in each frame
```
wanghan-iapcm, line 108: I do not see where do you implement the IDs `fake_type`, `real_type` and `type_number`. The ID means the keys in the data_dict of DeepmdData.

deepmd/utils/data.py
```diff
@@ -453,6 +452,17 @@ def _load_set(self, set_name: DPPath) :
                 tmp_in = data[k_in].astype(GLOBAL_ENER_FLOAT_PRECISION)
                 data[kk] = np.sum(np.reshape(tmp_in, [nframes, self.natoms, ndof]), axis = 1)
 
+        if self.mixed_type:
+            type_path = set_name / "real_atom_types.npy"
+            data['type'] = type_path.load_numpy().astype(np.int32).reshape([nframes, -1])
+            natoms = data['type'].shape[1]
+            vec_path = set_name / "real_atom_numbs.npy"
+            tmp = vec_path.load_numpy().astype(np.int32).reshape([nframes, -1])
+            data['real_natoms_vec'] = np.concatenate((np.tile(np.array([natoms, natoms], dtype=np.int32), (nframes, 1)),
```
wanghan-iapcm, line 461: Should not the `real_natom_vec` be calculated from `real_atom_types`?

deepmd/utils/data_system.py
```diff
@@ -84,6 +84,12 @@ def __init__ (self,
                     modifier = modifier, 
                     trn_all_set = trn_all_set
                 ))
+        if self.data_systems[0].mixed_type:
+            for data_sys in self.data_systems[1:]:
+                assert data_sys.mixed_type, "all systems must have the same format for mixed_type!"
```
wanghan-iapcm, line 89: The error message is not clear... 
"if one of the system is of mixed_type, then all of the systems should be of mixed_type"

But, the check is only triggered when the data_system[0] is mixed_type. Suppose the case data_system[0] is normal system and data_system[1] is mixed_type, the current check cannot catch the case. 

We should raise error if **ANY** of the data_system is mixed_type and at least one of them is not.

source/tests/common.py
```diff
@@ -24,31 +24,64 @@ def j_loader(filename):
...
+        gen_data_mixed_type(nframes)
+
+
 class Data():
-    def __init__ (self, 
-                  rand_pert = 0.1, 
-                  seed = 1, 
+    def __init__ (self,
+                  rand_pert = 0.1,
+                  seed = 1,
```
wanghan-iapcm, line 81: Please do not introduce format editing in this PR.

wanghan-iapcm (changes requested):

deepmd/common.py
```diff
@@ -424,6 +424,8 @@ def get_activation_func(
     RuntimeError
         if unknown activation function is specified
     """
+    if activation_fn is None or activation_fn in ['none', 'None']:
+        return None
```
njzjz, line 428: You can return `lambda x: x` for None, and this will not be a special case but have the same behavior as other activation functions.

deepmd/entrypoints/train.py
njzjz: This is not resolved

deepmd/utils/argcheck.py
```diff
@@ -28,17 +28,17 @@ def type_embedding_args():
...
-    doc_activation_function = f'The activation function in the embedding net. Supported activation functions are {list_to_doc(ACTIVATION_FN_DICT.keys())}. Note that "gelu" denotes the custom operator version, and "gelu_tf" denotes the TF standard version.'
+    doc_activation_function = f'The activation function in the embedding net. Supported activation functions are {list_to_doc(ACTIVATION_FN_DICT.keys())} Note that "gelu" denotes the custom operator version, and "gelu_tf" denotes the TF standard version.'
     doc_precision = f'The precision of the embedding net parameters, supported options are {list_to_doc(PRECISION_DICT.keys())} Default follows the interface precision.'
     doc_trainable = 'If the parameters in the embedding net are trainable'
     
     return [
-        Argument("neuron", list, optional = True, default = [2, 4, 8], doc = doc_neuron),
-        Argument("activation_function", str, optional = True, default = 'tanh', doc = doc_activation_function),
+        Argument("neuron", list, optional = True, default = [8], doc = doc_neuron),
+        Argument("activation_function", str, optional = True, default = None, doc = doc_activation_function),
```
njzjz, line 37: This is a breaking change! We should not introduce it.

njzjz (approved):

wanghan-iapcm (approved):

PR #1870, opened 2022-08-21
remove white space from train_attr/training_script

njzjz:
#921 discussed that the tensors are compressed in the graph file. But it looks no... So at least we remove white space.

When any graph file is opened, it is clear that `train_attr/training_script` is stored as raw text.
image


denghuilu (approved):
LGTM

wanghan-iapcm (approved):

PR #1891, opened 2022-09-03
Fix bugs when init_frz_model using tebd.

iProzd:
Fix bugs when init_frz_model using tebd.
Add init_variables for attention variables.
Fix precision of self.t_bias_atom_e in se_atten.py.
Change the default values in argcheck for se_atten (typos).
Improve the name scope in se_atten.py.
Delete some redundant variables in se_atten.py.
Move tf.constant from se_a.__init__() into build.
Add UTs for init_frz_model.


source/tests/init_frz_model/input.json
```diff
@@ -0,0 +1,66 @@
...
+		"batch_size":	"auto",
+		"_comment":		"that's all"
+},
+"validation_data":{
+		"systems":		["init_frz_model/data"],
+		"batch_size":	1,
+		"numb_btch":	3,
+		"_comment":		"that's all"
+},
+"numb_steps":	100,
```
wanghan-iapcm, line 56: `numb_steps ==  1` would work if you also set `save_freq` to 1.

deepmd/fit/ener.py
```diff
@@ -570,6 +571,10 @@ def init_variables(self,
         if self.numb_aparam > 0:
             self.aparam_avg = get_tensor_by_name_from_graph(graph, 'fitting_attr%s/t_aparam_avg' % suffix)
             self.aparam_inv_std = get_tensor_by_name_from_graph(graph, 'fitting_attr%s/t_aparam_istd' % suffix)
+        try:
+            self.bias_atom_e = get_tensor_by_name_from_graph(graph, 'fitting_attr%s/t_bias_atom_e' % suffix)
+        except:
```
njzjz, line 576:
```py
except GraphWithoutTensorError:
```

source/tests/test_init_frz_model_se_atten.py
```diff
@@ -0,0 +1,158 @@
...
+        frozen_model = str(tests_path / "dp-frozen.pb")
+        ckpt_name = "init_frz_se_atten.ckpt"
+        ckpt = str(tests_path / ckpt_name)
+        run_opt_ckpt = RunOptions(init_model=ckpt, log_level=20)
+        run_opt_frz = RunOptions(init_frz_model=frozen_model, log_level=20)
+        INPUT = str(tests_path / "input.json")
+        jdata = j_loader(str(tests_path / os.path.join("init_frz_model", "input.json")))
+        jdata["training"]["training_data"]["systems"] = data_file
+        jdata["training"]["validation_data"]["systems"] = data_file
+        jdata["training"]["save_ckpt"] = ckpt_name
```
njzjz, line 54:
```suggestion
        jdata["training"]["save_ckpt"] = ckpt
```

njzjz: Note that `ckpt_name` and `ckpt` is not the same thing when `tests_path` is not the current working directory.

njzjz (approved):

deepmd/descriptor/se_a.py
```diff
@@ -164,10 +164,9 @@ def __init__ (self,
         self.embedding_net_variables = None
         self.mixed_prec = None
         self.place_holders = {}
-        nei_type = np.array([])
+        self.nei_type = np.array([])
         for ii in range(self.ntypes):
-            nei_type = np.append(nei_type, ii * np.ones(self.sel_a[ii])) # like a mask 
-        self.nei_type = tf.constant(nei_type, dtype = tf.int32)
+            self.nei_type = np.append(self.nei_type, ii * np.ones(self.sel_a[ii])) # like a mask
```
njzjz, line 169: It looks like it should be an int32. However, the default type for `np.ones` is float32.

denghuilu (approved):

deepmd/descriptor/se_a.py
njzjz: btw, it may be written as
```py
np.repeat(np.arange(self.ntypes), self.sel_a)
```

njzjz:
#1901 should fix the failed tests.

PR #1896, opened 2022-09-06
fix DeprecationWarning of imp

njzjz:
> deepmd/env.py:8: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses

See also https://docs.python.org/3/library/imp.html#imp.reload

### DIFF
--- a/deepmd/env.py
+++ b/deepmd/env.py
@@ -5,7 +5,7 @@ import os
 import re
 import platform
 from configparser import ConfigParser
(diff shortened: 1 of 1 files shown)


njzjz:
Python 3.12 will be released in October 2023.

wanghan-iapcm (approved):

PR #1898, opened 2022-09-06
bump to C++ 17 for TF 2.10

njzjz:
Also, (1) replace `GREATER_EQUAL` by `VERSION_GREATER_EQUAL`. `GREATER_EQUAL` does not work with `2.10`; (2) upgrade manylinux to 2_28 to have a newer compiler that supports C++ 17.


njzjz:
GCC 5 does not support C++ 17. Shall we remove `Test Python / Test Python (3.10, 5)` before bumping TF version to 2.10?

wanghan-iapcm (approved):

PR #1916, opened 2022-09-15
Update installation document and add support for previous versions of TensorFlow.

Angel-Jia:


doc/install/install-from-source.md
```diff
@@ -29,6 +29,12 @@ source $tensorflow_venv/bin/activate
 pip install --upgrade pip
 pip install --upgrade tensorflow
 ```
+If you prefer to use previous versions of TensorFlow, all packages related to TensorFlow must share the same version, e.g.,
+```bash
+source $tensorflow_venv/bin/activate
+pip install --upgrade pip
+pip install tensorflow==2.9.* tensorflow-estimator==2.9.* keras==2.9.*
```
njzjz, line 36: I don't know why it's required. TensorFlow's dependencies will be installed automatically.
See https://gist.github.com/njzjz/73b528ae9a995ae5be1beba0fbd3140f.
image

Angel-Jia: Tensorflow-2.9 is not a good example, you can try on 2.6 version. Keras-2.10.0 and tensorflow-estimator-2.10.0 will be installed and there will be some errors in training attention model due to the wrong version of keras. I also encounter other problems about tensorflow-estimator.

njzjz: Failed to reproduce: https://gist.github.com/njzjz/018f92a422f9f1672d134dc885a99cb6

Angel-Jia: I just found that's because I was using conda:
image
How about moving it to `Troubleshooting`?

njzjz: This is just a bug in TF 2.6.0. See https://github.com/tensorflow/tensorflow/pull/52929 and https://github.com/tensorflow/tensorflow/pull/52927.
You should install the latest patch version i.e. 2.6.5.

PR #1921, opened 2022-09-18
highlight LAMMPS codes in doc

njzjz:
See also https://github.com/lammps/pygments-lammps


njzjz:
image
image
image

wanghan-iapcm (approved):
Great!

PR #1923, opened 2022-09-18
enforce type_map change when mixed_type

iProzd:
enforce type_map change when mixed_type.
add out_of_bound check for 'real_atom_types.npy'.

### DIFF
--- a/deepmd/utils/data.py
+++ b/deepmd/utils/data.py
@@ -52,6 +52,9 @@ class DeepmdData() :
         # load atom type
         self.atom_type = self._load_type(root)
         self.natoms = len(self.atom_type)
+        if self.mixed_type:
+            # nframes x natoms
+            self.atom_type_mix = self._load_type_mix(self.dirs[0])
         # load atom type map
         self.type_map = self._load_type_map(root)
         if self.type_map is not None:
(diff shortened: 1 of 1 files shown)


iProzd:
@njzjz @wanghan-iapcm

deepmd/utils/data.py
```diff
@@ -456,13 +468,17 @@ def _load_set(self, set_name: DPPath) :
...
 
         if self.mixed_type:
-            type_path = set_name / "real_atom_types.npy"
-            real_type = type_path.load_numpy().astype(np.int32).reshape([nframes, -1])
+            real_type = self.atom_type_mix.reshape([nframes, self.natoms])
             data['type'] = real_type
             natoms = data['type'].shape[1]
+            # nframes x ntypes
+            atom_type_nums = np.array([(real_type == i).sum(axis=-1) for i in range(self.get_ntypes())],
+                                      dtype=np.int32).T
```
njzjz, line 476: For a more elegant code I think you can try [`numpy.bincount`](https://numpy.org/doc/stable/reference/generated/numpy.bincount.html):
```py
atom_type_nums = np.bincount(real_type, minlength=self.get_ntypes()).T
```

deepmd/utils/data.py
```diff
@@ -63,6 +66,15 @@ def __init__ (self,
             if not self.mixed_type:
                 atom_type_ = [type_map.index(self.type_map[ii]) for ii in self.atom_type]
                 self.atom_type = np.array(atom_type_, dtype = np.int32)
+            else:
+                type_idx_map = [type_map.index(ii) for ii in self.type_map]
+                atom_type_mix_ = -np.ones_like(self.atom_type_mix).astype(np.int32)
+                for ii in range(len(type_idx_map)):
+                    atom_type_mix_[self.atom_type_mix == ii] = type_idx_map[ii]
```
njzjz, line 73: This is equivalent to
```python
atom_type_mix_ = np.array(type_idx_map)[self.atom_type_mix]
```

deepmd/utils/data.py
```diff
@@ -63,6 +66,15 @@ def __init__ (self,
             if not self.mixed_type:
                 atom_type_ = [type_map.index(self.type_map[ii]) for ii in self.atom_type]
                 self.atom_type = np.array(atom_type_, dtype = np.int32)
+            else:
+                type_idx_map = [type_map.index(ii) for ii in self.type_map]
```
njzjz, line 70: Considering `self.type_map` is a subset of `type_map`, it can be written as
```py
type_idx_map = np.searchsorted(type_map, self.type_map)
```

deepmd/utils/data.py
iProzd: I tried np.bincount, but it does not support 2-dim array here (nframes x natoms), unless you split in the nframes dim, which is more ugly than 'for' in ntypes dim here.

njzjz (approved):

iProzd:
@wanghan-iapcm

PR #1963, opened 2022-10-03
fix typo in the doc of `loc_frame`

### LINKED ISSUE #1962: axis_rule of local_frame
Hi dper,
    I hope use the local_frame descriptor, but i do not understand and know how to set this part
image
I look for help. It will be nice if there is a more accurate explanation and an example.

amcadmus:
fixes #1962


njzjz (approved):

njzjz:
Changes should be synced here:
https://github.com/deepmodeling/deepmd-kit/blob/923b98ca918985dfea9147fcd6d4c0eb06fc09e5/deepmd/utils/argcheck.py#L106-L112

PR #1994, opened 2022-10-12
fix library suffix for macos

njzjz:


njzjz:
Not required, as macos packages contain all library files.

PR #2013, opened 2022-10-19
fix OP_CXX_ABI for TF>=2.9

njzjz:
TF 2.9 removes the tf_cxx11_abi_flag function. The previous implementation in #1736 was wrong as `_GLIBCXX_USE_CXX11_ABI` is set by the compiler but not the TF header...
Here I propose a new way - try to compile against a TF function that has `std::string` with both two flags.
This commit also merges some duplicated codes.


njzjz:
I've tested with both `_GLIBCXX_USE_CXX11_ABI=0` and `_GLIBCXX_USE_CXX11_ABI=1` TF libraries. It works fine.

wanghan-iapcm (approved):

Issue #1948, opened 2022-09-23
[Feature Request] Merge high and low api_cc into one library


njzjz:
### Summary

Currently high and low precision for api_cc are in two different libraries. The issue proposals some ways to merge them into one library.

### Detailed Description

As we hope the change will not break the current C++ API (using double precision of `DeepPot` without assigning dtype) for third-party packages, there are three ways to do it.

1. Automatically detect the dtype of `o_energy` to expose methods of both precision. This is the best way but may be a bit hard to implement. 
2. Add a new template class `template <typename dtype> class DeepPotDtype`, then `using DeepPot = DeepPotDtype<double>`.
3. C++ 17 allows setting a default value to the template `template <typename dtype=double> class DeepPot`. Then `DeepPot` can be simply called without specify dtype.

PR #2031, opened 2022-10-28
uncouple model precision and interface precision

njzjz:
This commit allows loading float models using the double interface, or the other way around.
The model precision is perceived from the type of `descrpt_attr/rcut`. All `run_model` and `session_input_tensors` are rewritten as templates.
This is an essential step of #1948. Next, functions in the whole module can be migrated to templates.


source/api_cc/src/DataModifier.cc
```diff
@@ -179,8 +203,13 @@ compute (std::vector<VALUETYPE> &		dfcorr_,
   nlist_data.make_inlist(nlist);
   // make input tensors
   std::vector<std::pair<std::string, Tensor>> input_tensors;
-  int ret = session_input_tensors (input_tensors, dcoord_real, ntypes, datype_real, dbox, nlist, std::vector<VALUETYPE>(), std::vector<VALUETYPE>(), atommap, nghost_real, 0, name_scope);
-  assert (nloc_real == ret);
+  if (dtype == tensorflow::DT_DOUBLE) {
+    int ret = session_input_tensors<double> (input_tensors, dcoord_real, ntypes, datype_real, dbox, nlist, std::vector<VALUETYPE>(), std::vector<VALUETYPE>(), atommap, nghost_real, 0, name_scope);
+    assert (nloc_real == ret);
```
wanghan-iapcm, line 208: The assertion may be moved out of the if-else.

source/api_cc/src/DeepPot.cc
```diff
@@ -750,14 +785,21 @@ compute (std::vector<ENERGYTYPE> &		all_energy,
...
 
     all_energy.resize (numb_models);
     all_force.resize (numb_models);
     all_virial.resize (numb_models);
-    assert (nloc == ret);
+
     for (unsigned ii = 0; ii < numb_models; ++ii) {
-        run_model (all_energy[ii], all_force[ii], all_virial[ii], sessions[ii], input_tensors, atommap, nghost);
+      if (dtype == tensorflow::DT_DOUBLE) {
+        int ret = session_input_tensors <double> (input_tensors, dcoord_, ntypes, datype_, dbox, nlist, fparam, aparam, atommap, nghost, ago);
```
wanghan-iapcm, line 795: The `session_input_tensors` should be placed out of the for loop, because all models share the same inputs, and we only need to make the input tensors once.

source/api_cc/src/DeepPot.cc
```diff
@@ -792,16 +834,23 @@ compute (std::vector<ENERGYTYPE> &		all_energy,
...
     all_force .resize (numb_models);
     all_virial.resize (numb_models);
     all_atom_energy.resize (numb_models);
     all_atom_virial.resize (numb_models); 
-    assert (nloc == ret);
+
     for (unsigned ii = 0; ii < numb_models; ++ii) {
-        run_model (all_energy[ii], all_force[ii], all_virial[ii], all_atom_energy[ii], all_atom_virial[ii], sessions[ii], input_tensors, atommap, nghost);
+      if (dtype == tensorflow::DT_DOUBLE) {
+        int ret = session_input_tensors <double> (input_tensors, dcoord_, ntypes, datype_, dbox, nlist, fparam, aparam, atommap, nghost, ago);
```
wanghan-iapcm, line 846: The same issue here.

wanghan-iapcm (changes requested):
This PR is great. 

I have a question regarding the coverage of the UTs. 
Ideally we should test all the four cases:
case 0: model_type == double  interface_type == double
case 1:  model_type == double  interface_type == float
case 2:  model_type == float  interface_type == double
case 3:  model_type == float  interface_type == float
Do we currently only cover case 0? In this pr we may at least provide tests for the case 1

njzjz:
I added case 1. I found the accuracy of the float interface is not very high.

wanghan-iapcm (approved):

wanghan-iapcm:
> I added case 1. I found the accuracy of the float interface is not very high.

The precision (an epsilon of 1e-4 ) is expected.

PR #2038, opened 2022-10-31
add dp test help doc

likefallwind:
add dp test help doc

### DIFF
--- a/deepmd/entrypoints/main.py
+++ b/deepmd/entrypoints/main.py
@@ -281,7 +281,7 @@ def main_parser() -> argparse.ArgumentParser:
         "--detail-file",
         type=str,
         default=None,
-        help="Files where details of energy, force and virial accuracy/accuracy per atom will be written",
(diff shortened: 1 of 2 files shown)


deepmd/entrypoints/main.py
```diff
@@ -281,7 +281,7 @@ def main_parser() -> argparse.ArgumentParser:
         "--detail-file",
         type=str,
         default=None,
-        help="File where details of energy force and virial accuracy will be written",
+        help="Files where details of energy, force and virial accuracy/accuracy per atom will be written",
```
njzjz, line 284: Actually the argument assigns **the prefix of** files.

likefallwind: So maybe the proper description is "The prefix to files where details of energy, force and virial accuracy/accuracy per atom will be written"?

njzjz: Agree

likefallwind: ok I have changed the file

njzjz (approved):

Issue #2045, opened 2022-11-02
Cannot use hdf5 dataset as described in document


Angel-Jia:
### Summary

Converting the dataset into hdf5 format as described in https://github.com/deepmodeling/deepmd-kit/blob/master/doc/data/data-conv.md will get this error:
DEEPMD CRITICAL dir ./hdf5_water#O64H128 is not a valid dir

### DeePMD-kit Version

DeePMD-kit v2.1.6.dev0+g6e3d4a6.d20221102

### TensorFlow Version

2.9

### Details

Take `examples/water` dataset as example, hdf5 file is obtained by:
```
cd examples/water && mkdir hdf5 && cd hdf5
ss = dpdata.MultiSystems(type_map=['O', 'H'])
ss.from_deepmd_npy('../data/', labeled=True)
ss.to(fmt='deepmd/hdf5',directory='hdf5_water')

f = h5py.File('hdf5_water', 'r')
print(f.keys())    # output is: <KeysViewHDF5 ['O64H128']>
```
Then add this file into `systems` in `input.json`:
```
"systems":		["./hdf5_water#O64H128"],
```
I got this error during training:
DEEPMD CRITICAL dir ./hdf5_water#O64H128 is not a valid dir
Traceback (most recent call last):
  File "/home/yaosk/anaconda3/envs/dp2.1.5/bin/dp", line 8, in <module>
    sys.exit(main())
  File "/home/yaosk/anaconda3/envs/dp2.1.5/lib/python3.8/site-packages/deepmd/entrypoints/main.py", line 562, in main
    train_dp(**dict_args)
  File "/home/yaosk/anaconda3/envs/dp2.1.5/lib/python3.8/site-packages/deepmd/entrypoints/train.py", line 94, in train
    jdata = update_sel(jdata)
  File "/home/yaosk/anaconda3/envs/dp2.1.5/lib/python3.8/site-packages/deepmd/entrypoints/train.py", line 354, in update_sel
    descrpt_data = update_one_sel(jdata, descrpt_data)
  File "/home/yaosk/anaconda3/envs/dp2.1.5/lib/python3.8/site-packages/deepmd/entrypoints/train.py", line 323, in update_one_sel
    tmp_sel = get_sel(jdata, rcut, one_type=descriptor['type'] in ('se_atten',))
  File "/home/yaosk/anaconda3/envs/dp2.1.5/lib/python3.8/site-packages/deepmd/entrypoints/train.py", line 284, in get_sel
[...]

PR #2046, opened 2022-11-02
fix bug in hdf5 path

Angel-Jia:
fix bug in hdf5 path: https://github.com/deepmodeling/deepmd-kit/issues/2045
Removing `/` at the begining of the key


njzjz (changes requested):
Please check HDF5 documentation. Every path starts with `/`.

https://confluence.hdfgroup.org/plugins/servlet/mobile#content/view/50078525
https://docs.h5py.org/en/stable/high/group.html

Instead, our documentation may have a typo.

PR #2072, opened 2022-11-08
build, package, and release C library

### LINKED ISSUE #2058: [Feature Request] Release C library binary
### Summary

Nowadays deepmd-kit C++ library has several downstream projects. As tensorflow is hard to build, it's also a pain to build deepmd-kit.

However, as C API is added in #1098 (although it still needs to improve), we can release a C library binary. It doesn't require using the same compiler and C++ standard as what the C++ library requires, so people who want to manually compile the downstream projects can directly download it and used to compile it. 

We also consider improving C API, and even provide a pure header based on C API to provide the same C++ classes as that in C++ API (like `json.hpp`). So downstream projects do not need to change functions but just change the header to use the C library.

### Detailed Description

The packaged C library should only contain:
- `lib/libdeepmd_c.so`
- the linked libraries of `libdeepmd_c.so` except those in `/lib64` such as `libc.so.6`, `libm.so.6`
- `include/deepmd/c_api.h` and potentially `include/deepmd/cxx_api.h`
- licenses

The minimum glibc requirement should be 2.17.

Currently the packages from conda can meet the requirement, so a CI script can be automatically executed to reuse them.

Some similar packages:
- TensorFlow C library download: `https://www.tensorflow.org/install/lang_c`

njzjz:
Fix #2058.

This patch builds DeePMD-kit C library against TensorFlow 2.9.2 C++ library with CUDA support on CentOS 7 and then packages DeePMD-kit C library along with TensorFlow libraries. Then it is tested with GCC 4.9.0 and GLIBC 2.19. Finally, the package is uploaded and released. GitHub Actions and Docker drive all these things.
The minimum GLIBC requirement is 2.17, and the minimum GLIBCXX requirement is 3.4.19. It should work with any version of GCC.
Note that one still needs to install CUDA Toolkit 11 and cuDNN 8 manually for GPUs, otherwise TF falls back to CPUs.

### DIFF
--- a/.github/workflows/package_c.yml
+++ b/.github/workflows/package_c.yml
@@ -1,0 +1,26 @@
+name: Build C library
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build_c:
+    name: Build C library
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Package C library
+        run: ./source/install/docker_package_c.sh
+      - name: Test C library
+        run: ./source/install/docker_test_package_c.sh
+      # for download and debug
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
+        with:
+          path: ./libdeepmd_c.tar.gz
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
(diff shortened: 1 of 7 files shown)


njzjz:
The library can be downloaded from https://github.com/deepmodeling/deepmd-kit/actions/runs/3418532003

wanghan-iapcm (approved):

PR #2079, opened 2022-11-09
add a header-only C++ DeepPot class based on C library

njzjz:
This patch adds a header-only C++ `deepmd::hpp::DeepPot` class which ONLY relies on the DeePMD-kit C library, so one can use any compiler that supports C++ 11 to build against it.

image

Not supported yet compared to C++ library: fparam, aparam, nlist as input, assign gpu_rank, print_summary


wanghan-iapcm (approved):
Interesting PR

PR #2082, opened 2022-11-10
dynamically load CUDA runtime library

njzjz:
This patch dynamically loads the CUDA runtime library (`libcudart.so`) instead of static linking. I referred to how TF does it and copied some open-source codes from TF. The advantage is that a CUDA package can be distributed that works on CPUs without any CUDA Toolkit libraries installed or packaged. (Currently, the packaged C library tarball contains `libcudart.so`)
Also fixes some minor issues.


njzjz:
btw, all tests pass with `USE_CUDA_TOOLKIT=TRUE`.

wanghan-iapcm (approved):

PR #2093, opened 2022-11-14
Upload the multi-task training api

iProzd:
Upload the multi-task training api.
Add tebd support for dipole&polar fitting nets.


iProzd:
A hidden bug need to be fixed:
`data_requirement` in deepmd/common.py may have different values with the same key. (For example, multiple dipole fitting nets with different `type_sel` in key 'atomic_dipole') 
`data_requirement` is a global parameter and I'm not yet sure how to modify this to suit multiple fitting nets. 
Disscusions needed.

deepmd/utils/argcheck.py
```diff
@@ -767,19 +789,143 @@ def normalize_hybrid_list(hy_list):
...
+
+
+def normalize_fitting_net_dict(fitting_net_dict):
+    new_dict = {}
+    base = Argument("base", dict, [], [fitting_variant_type_args()], doc="")
+    for fitting_key_item in fitting_net_dict:
+        data = base.normalize_value(fitting_net_dict[fitting_key_item], trim_pattern="_*")
+        base.check_value(data, strict=True)
+        new_dict[fitting_key_item] = data
+    return new_dict
```
njzjz, line 837: This could be something implemented in dargs. Currently, we have `repeat=True` for `list`, but it is possible to make it also available for `dict` (`dict.values()`). ping @y1xiaoc

deepmd/env.py
```diff
@@ -120,6 +121,29 @@
...
+    "o_rmat_{}": "o_rmat",
+    "o_rmat_deriv_{}": "o_rmat_deriv",
+    "o_nlist_{}": "o_nlist",
+    "o_rij_{}": "o_rij",
+    "o_dm_force_{}": "o_dm_force",
+    "o_dm_virial_{}": "o_dm_virial",
+    "o_dm_av_{}": "o_dm_av",
+    "o_wfc_{}": "o_wfc",
+}
+
```
njzjz, line 146: Please fix the lint warning here.

deepmd/entrypoints/train.py
```diff
@@ -270,18 +291,41 @@ def get_nbor_stat(jdata, rcut, one_type: bool = False):
...
+            tmp_data.get_batch()
+            assert tmp_data.get_type_map(), \
+                "In multi-task mode, 'type_map.raw' must be defined in data systems {}! ".format(systems)
+            train_data_list.append(tmp_data)
+        ntypes = len(type_map)
+        neistat = NeighborStat(ntypes, rcut, one_type=one_type)
+        min_nbor_dist = []
+        max_nbor_size = []
+        for train_data in train_data_list:
+            tmp_min_nbor_dist, tmp_max_nbor_size = neistat.get_stat(train_data)
```
njzjz, line 322: Why not just merge `train_data` into one and pass it to `get_stat`?

deepmd/fit/polar.py
```diff
@@ -303,72 +307,168 @@ def build (self,
...
+                # (nframes x natoms) x naxis x 3(coord)
+                final_layer = tf.matmul(final_layer, rot_mat_i)
+                # (nframes x natoms) x 3(coord) x 3(coord)
+                final_layer = tf.matmul(rot_mat_i, final_layer, transpose_a = True)
+                # nframes x natoms x 3 x 3
+                final_layer = tf.reshape(final_layer, [tf.shape(inputs)[0], natoms[2+type_i], 3, 3])
+                # shift and scale
+                sel_type_idx = self.sel_type.index(type_i)
+                final_layer = final_layer * self.scale[sel_type_idx]
+                final_layer = final_layer + self.constant_matrix[sel_type_idx] * tf.eye(3, batch_shape=[tf.shape(inputs)[0], natoms[2+type_i]], dtype = GLOBAL_TF_FLOAT_PRECISION)
```
njzjz, line 395: I wonder `dtype` is `self.fitting_precision` instead of `GLOBAL_TF_FLOAT_PRECISION`.

deepmd/model/multi.py
```diff
@@ -0,0 +1,366 @@
...
+                    model_dict[fitting_key][gname] = global_out
+                    model_dict[fitting_key]["force"] = force
+                    model_dict[fitting_key]["virial"] = virial
+                    model_dict[fitting_key]["atom_virial"] = atom_virial
+
+        return model_dict
+
+    def _import_graph_def_from_frz_model(self, frz_model, feed_dict, return_elements):
+        graph, graph_def = load_graph_def(frz_model)
+        return tf.import_graph_def(graph_def, input_map=feed_dict, return_elements=return_elements, name="")
```
njzjz, line 366: I don't understand how it works considering the model containing multiple outputs cannot be frozen.

deepmd/env.py
```diff
@@ -120,6 +121,29 @@
         r"model_attr/t_tab_data|"
 )
 
+REMOVE_SUFFIX_DICT = {
```
njzjz, line 124: `fitting_attr`?

deepmd/model/multi.py
```diff
@@ -0,0 +1,366 @@
...
+        if self.typeebd is not None:
+            type_embedding = self.typeebd.build(
+                self.ntypes,
+                reuse=reuse,
+                suffix=suffix,
+            )
+            input_dict['type_embedding'] = type_embedding
+            input_dict['atype'] = atype_
+
+        if frz_model == None:
```
njzjz, line 196: `is None`

njzjz:
`data_requirement` can be a class property for `Trainer`, `Model`, `Loss`, etc. The ideal code is
```py
self.data_requirement += self.loss.data_requirement
```

njzjz (changes requested):

deepmd/utils/argcheck.py
y1xiaoc: I'm not sure if that is desirable. `repeat` is here to deal with the case where one may have indefinitely many homogeneous dicts in that list. For dict, to make the type checking work, one need to specify the keys in advance. Therefore, there should be a fixed set of keys although the corresponding values are the same type. In that case, one should just write a loop to generate the Argument. If the use case is that no matter what the key is, the values follows the same dict type, maybe one should reconsider reform it into a list of dict with the original key being one of the attributes.

deepmd/fit/dipole.py
```diff
@@ -120,46 +122,103 @@ def build (self,
         dipole
                 The atomic dipole.
         """
+        if input_dict is None:
+            input_dict = {}
+        type_embedding = input_dict.get('type_embedding', None)
+        atype = input_dict.get('atype', None)
+        nframes = input_dict.get('nframes')
```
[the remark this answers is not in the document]
iProzd: Only fitting net with tebd has valid `atype`. It's a parameter together with `input_dict['type_embedding']`.

deepmd/model/multi.py
iProzd: Typo.

deepmd/env.py
iProzd: `fitting_attr_{}/` is contained in situation "_{}/" --> "/".

deepmd/utils/argcheck.py
iProzd: After discussion with @wanghan-iapcm , maybe it's not a suitable choice to make fitting_net_dict into a `list`: 
1. users may have to manully pair the fitting_net and its training datasystems;
2. or we must add an additional key in fitting_net to identify itself, which may have more conflicts than `dict`.

y1xiaoc: It's seems that I didn't get the full context. As I mentioned if there are finite choices of keys then this can be directly written as a loop to generate the Argument. The only use case is there are indefinite choices of keys. Is that the case? Or maybe we can schedule a video call on this.

deepmd/utils/argcheck.py
```diff
@@ -586,6 +590,15 @@ def loss_args():
     return ca
 
 
+def loss_dict_args():
+    doc_loss_dict = 'The dictionary of definitions of multiple loss functions in multi-task mode. ' \
+                    'Each loss_dict[fitting_key], with user-defined name `fitting_key` in `model/fitting_net_dict`, is the single definition of loss function, whose type should be set to `tensor`, `ener` or left unset.\n\.'
```
njzjz, line 595: `\.` might be a typo and should be removed. See also #2122.

deepmd/utils/argcheck.py
```diff
@@ -754,10 +779,19 @@ def gen_json(**kwargs):
...
         nvnmd_args(),
     ), cls=ArgumentEncoder)
 
+def gen_args(**kwargs):
+    return [nvnmd_args(),
+            training_args(),
+            loss_dict_args(),
+            loss_args(),
+            learning_rate_args(),
+            model_args()]  # show in reverse order
```
njzjz, line 793: I may fix it in dargs and a separated PR if it is a bug..

njzjz (approved):
Approve in principle

deepmd/utils/argcheck.py
njzjz: I fix the bug in https://github.com/deepmodeling/dargs/pull/22. May I release a new version of dargs?

njzjz: I release dargs v0.3.4. It's better to bump the dargs version for docs

https://github.com/deepmodeling/deepmd-kit/blob/060311eb972f25ee62ccd08efc4c880d2cc6afc2/setup.py#L83

deepmd/model/multi.py
```diff
@@ -0,0 +1,352 @@
...
+from deepmd.env import tf
+from deepmd.utils.pair_tab import PairTab
+from deepmd.utils.graph import load_graph_def
+from deepmd.env import global_cvt_2_ener_float, MODEL_VERSION, GLOBAL_TF_FLOAT_PRECISION
+from deepmd.env import op_module
+from .model import Model
+from .model_stat import make_stat_input, merge_sys_stat
+
+
+class MultiModel(Model):
```
wanghan-iapcm, line 13: It is nice to introduce the `MultiModel`. Is it possible to write a function that can either a dict of model, as that the current `MultiModel` does, or a single model.  

I find the model building code in `MultiModel` is a duplication of the model making code in modules `model.**`

deepmd/train/trainer.py
```diff
@@ -655,60 +805,94 @@ def valid_on_the_fly(self,
...
-            feed_dict = self.get_feed_dict(batch, is_training=False)
-            results = self.loss.eval(self.sess, feed_dict, natoms)
-
-            for k, v in results.items():
-                if k == "natoms":
-                    sum_natoms += v
-                else:
-                    sum_results[k] = sum_results.get(k, 0.) + v * results["natoms"]
-        avg_results = {k: v / sum_natoms for k, v in sum_results.items() if not k == "natoms"}
+        def eval_single_list(single_batch_list, loss, sess, get_feed_dict_func, prefix=''):
```
wanghan-iapcm, line 868: I am not sure if we need to introduce the `eval_single_list` in the scope `_build_training`.

wanghan-iapcm (changes requested):

deepmd/utils/argcheck.py
iProzd: Done in this PR.

deepmd/model/multi.py
iProzd: I'll improve this in next PR.

wanghan-iapcm (approved):

PR #2097, opened 2022-11-16
fix codecov

njzjz:
Codecov python package is deprecated does not work any more. This patch migrates it to the GitHub Actions version.


njzjz:
It works here. The key point is to set `gcov` to `true`.
```yaml
    - uses: codecov/codecov-action@v3
      with:
        gcov: true
```

wanghan-iapcm (approved):

Issue #2105, opened 2022-11-18
[BUG] TestInferDeepPotA.cpu_lmp_nlist_type_sel failed when requesting atomic energies and atomic virial


njzjz:
### Bug summary

At present, `TestInferDeepPotA.cpu_lmp_nlist_type_sel` passes. But when using the `compute` method that requests atomic energies and atomic virials, it throws `Segmentation fault`.

### DeePMD-kit Version

devel

### TensorFlow Version

2.9.0

### How did you download the software?

Built from source

### Input Files, Running Commands, Error Log, etc.

The current codes:
https://github.com/deepmodeling/deepmd-kit/blob/37c69bc545d9ef4eb25021eed90a1dab1dd431f0/source/api_cc/tests/test_deeppot_a.cc#L370

If we change it to 
```cpp
  std::vector<VALUETYPE> force_(nall*3, 0.0), virial(9, 0.0), atomic_energy(nall*3, 0.0), atomic_virial(nall*9, 0.0);
  dp.compute(ener, force_, virial, atomic_energy, atomic_virial, coord_cpy, atype_cpy, box, nall-nloc, inlist, 0);
```

We will get
```
4074706 Segmentation fault      (core dumped)
```

This means this API has a potential bug.

### Steps to Reproduce

After changing that codes, run `source/install/test_cc.sh` or `source/install/test_cc_local.sh`.

wanghan-iapcm:
I cannot reproduce the bug, or at least the Seg fault should not happen at `dp.compute`. Could you please provide your test_deeppot_a.cc?

njzjz:
https://github.com/deepmodeling/deepmd-kit/commit/b831685e9429beb1594b23db29611a3ed1fcf70e

I can reproduce the segmentation fault on the GitHub: https://github.com/njzjz/deepmd-kit/commit/b831685e9429beb1594b23db29611a3ed1fcf70e/checks?check_suite_id=9399836409

njzjz:
c1e4a3702a30e1ee8905c897d9a3075c1352babf

Reproduced on the GitHub Actions:
https://github.com/njzjz/deepmd-kit/commit/c1e4a3702a30e1ee8905c897d9a3075c1352babf/checks?check_suite_id=9399906381

wanghan-iapcm:
The failure is expected. One has to fold back the ghost atoms to local atoms like

[...]

PR #2106, opened 2022-11-18
c: support neighbor list input in the C API

### LINKED ISSUE #2086: [Feature Request] Support neighbor list input in the C API
### Summary

This issue asks for neighbor list support in the C API. 

Currently the C API uses deepmd-kit native neighbor list support, which should be fine for small systems. However, if one wants to do large scale simulations or wants to port deepmd to an MD package, he/she needs an API supports neighbor lists. This is supported by the C++ API but not the C API. 

### Detailed Description

In detail the C API is asked to support the follow C++ method

https://github.com/deepmodeling/deepmd-kit/blob/91ad80777dd94df7d3b90508c94382667c17e6da/source/api_cc/include/DeepPot.h#L144-L157

_No response_

njzjz:
Fix #2086.

The test `cpu_lmp_nlist_type_sel` is skipped until #2105 is fixed.

### DIFF
--- a/source/api_cc/src/DeepPot.cc
+++ b/source/api_cc/src/DeepPot.cc
@@ -697,18 +697,16 @@ compute (ENERGYTYPE &			dener,
   int nloc_real = nall_real - nghost_real;
   dcoord.resize(nall_real * 3);
   datype.resize(nall_real);
-  datom_energy.resize(nall_real);
   // fwd map
   select_map<VALUETYPE>(dcoord, dcoord_, fwd_map, 3);
   select_map<int>(datype, datype_, fwd_map, 1);
-  select_map<VALUETYPE>(datom_energy, datom_energy_, fwd_map, 1);
   // aparam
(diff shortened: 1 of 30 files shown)


wanghan-iapcm (approved):

wanghan-iapcm:
Also add an example to the doc?

wanghan-iapcm:
I keep this PR open until #2105 is fixed.

source/api_cc/src/DeepPot.cc
```diff
@@ -712,6 +712,7 @@ compute (ENERGYTYPE &			dener,
     assert (nloc_real == atommap.get_type().size());
 
         nlist_data.copy_from_nlist(lmp_list);
+        nlist_data.shuffle_exclude_empty(fwd_map);
```
njzjz, line 715: #2111 should be merged before this PR so that the change to this line will disappear in this PR.

PR #2114, opened 2022-11-21
docs: fix spellings and grammars

njzjz:


wanghan-iapcm (approved):
Nice work. Did you do it by some tool?

PR #2119, opened 2022-11-22
c: fix `ComputeNlist` -> `ComputeNList` in the C API

njzjz:
I don't know why gcc feels it ok (so the unit test passed), but gfortran complains about it.


wanghan-iapcm (approved):
Interesting bug...

PR #2155, opened 2022-12-04
avoid unnecessary std::endl

njzjz:
`std::endl` = `\n` + `flush`, and in these situations, it is not required to flush every line:
1. multiple lines into stdout and stderr. Only the last line is required to flush;
2. frequently writing to a file. In the DeePMD-kit, model_devi.out was flushed every line, but usually, one does not need to monitor it.

See also:
https://en.cppreference.com/w/cpp/io/manip/endl
https://stackoverflow.com/a/17468264/9567349


njzjz:
Not clear how to flush it...

PR #1423, opened 2022-01-16
support `type_one_side` along with `exclude_types`

njzjz:
This PR supports `type_one_side` along with `exclude_types`, by reusing the same NN parameters.

PR #2166, opened 2022-12-06
support DPRc + type embedding

njzjz:
This patch supports DPRc + type embedding, including
1. refactor `exclude_types` + `type_one_side`. Considering `exclude_types` is equivalent to $\forall i,j \in \text{exclude types}, \frac{\partial{E}}{\partial{r_{ij}}}=0$, a mask is created to apply to the environmental matrix, which is faster than the previous implementation in #1423.
2. implement `atom_ener` + type embedding.

### DIFF
--- a/deepmd/descriptor/se_r.py
+++ b/deepmd/descriptor/se_r.py
@@ -472,31 +472,14 @@ class DescrptSeR (DescrptSe):
             inputs_i = tf.reshape(inputs_i, [-1, self.ndescrpt])
             type_i = -1
             if len(self.exclude_types):
-                # generate a mask
-                type_mask = np.array([
-                    [1 if (tt_i, tt_j) not in self.exclude_types else 0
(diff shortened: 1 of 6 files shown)


deepmd/descriptor/se_r.py
```diff
@@ -475,6 +471,33 @@ def _pass_filter(self,
             inputs_i = inputs
             inputs_i = tf.reshape(inputs_i, [-1, self.ndescrpt])
             type_i = -1
+            if len(self.exclude_types):
```
wanghan-iapcm, line 474: This branch is repeating with `se_a`, except some shape difference. Write a method in the base class descriptor and reuse the method in both `se_r` and `se_a`?

njzjz: Merged.

wanghan-iapcm (approved):

PR #2208, opened 2022-12-30
support the custom training times for each frame

njzjz:
Before, `sys_probs` only specifies the probability of each system. This patch allows `numb_copy.npy` (`int`) to specify each frame is copied by the "numb_copy" (int) times. The default is 1 for all frames, which keeps the same behavior as before.

### DIFF
--- a/deepmd/common.py
+++ b/deepmd/common.py
@@ -122,6 +122,7 @@ def add_data_requirement(
     type_sel: Optional[bool] = None,
     repeat: int = 1,
     default: float = 0.,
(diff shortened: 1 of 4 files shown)


wanghan-iapcm (commented):
I suggest not naming it as "probability", because a probability should be summed to 1. 

"numb_copy" may be a better name, i.e. each frame is copied by the "numb_copy" (int) times. This is the behavior of your PR.

wanghan-iapcm (approved):

PR #2235, opened 2023-01-10
Support dp test with txt

HuangJiameng:
Support using a text file(E.G. `list.txt`) to specify the list of test sets

list.txt
```
data_dir/sys.000
data_dir/sys.001
data_dir/sys.002
...
```
`/data_dir/sys.000` should be a directory with "type.raw".
Then one can use `-l list.txt` to replace `-s data_dir`.


deepmd/entrypoints/test.py
```diff
@@ -57,9 +60,13 @@ def test(
     RuntimeError
         if no valid system was found
     """
-    all_sys = expand_sys_str(system)
-    if len(all_sys) == 0:
-        raise RuntimeError("Did not find valid system")
+    if(datafile is not None):
```
iProzd, line 63: Bracket is not necessary

deepmd/entrypoints/test.py
```diff
@@ -57,9 +60,13 @@ def test(
...
+        all_sys = open(datafile, 'r')
+        all_sys = [i.strip() for i in all_sys.readlines()]
+    else:
+        all_sys = expand_sys_str(system)
+        if len(all_sys) == 0:
```
iProzd, line 68: Move this condition out of ‘if’ scope above.

deepmd/entrypoints/main.py
```diff
@@ -271,6 +271,13 @@ def main_parser() -> argparse.ArgumentParser:
         type=str,
         help="The system dir. Recursively detect systems in this directory",
     )
+    parser_tst.add_argument(
+        "-f",
```
iProzd, line 275: It seems that -f and -s can not exist at the same time. 
It's better to use 
'''
parser_tst_subgroup = parser_tst.add_mutually_exclusive_group()
parser_tst_subgroup.add_argument("-s", ... 
parser_tst_subgroup.add_argument("-f", ...
'''

iProzd (changes requested):

deepmd/entrypoints/test.py
```diff
@@ -57,9 +60,13 @@ def test(
     RuntimeError
         if no valid system was found
     """
-    all_sys = expand_sys_str(system)
-    if len(all_sys) == 0:
-        raise RuntimeError("Did not find valid system")
+    if(datafile is not None):
+        all_sys = open(datafile, 'r')
```
njzjz, line 64: This file is not closed. See also https://stackoverflow.com/a/7396043/9567349

deepmd/entrypoints/test.py
```diff
@@ -57,9 +60,13 @@ def test(
...
+        all_sys = [i.strip() for i in all_sys.readlines()]
```
njzjz, line 65: Simpler:
```py
f.read().splitlines()
```

deepmd/entrypoints/test.py
```diff
@@ -39,6 +40,8 @@ def test(
         path where model is stored
     system : str
         system directory
+    datafile: str
```
njzjz, line 43: Numpydoc format: 
> The colon must be preceded by a space

```suggestion
    datafile : str
```

See https://numpydoc.readthedocs.io/en/latest/format.html

deepmd/entrypoints/main.py
```diff
@@ -271,6 +271,13 @@ def main_parser() -> argparse.ArgumentParser:
...
+        "--datafile",
+        default=None,
+        type=str,
+        help="The list of systems to test.",
```
njzjz, line 279: This help does not indicate that the input should be a filename.

njzjz (changes requested):
Also, it is encouraged to add a test.

iProzd (approved):

deepmd/entrypoints/test.py
```diff
@@ -57,7 +60,13 @@ def test(
     RuntimeError
         if no valid system was found
     """
-    all_sys = expand_sys_str(system)
+    if datafile is not None:
+        all_sys = open(datafile, 'r')
+        all_sys = all_sys.read().splitlines()
+        datafile.close()
```
njzjz, line 66: `datafile` is `str`.

njzjz:
```suggestion
        with open(datafile, 'r') as f:
            all_sys = f.read().splitlines()
```

njzjz (changes requested):

deepmd/entrypoints/test.py
njzjz: By the way, you should manually test your commit if there is no unit test.

HuangJiameng: done

njzjz (approved):

PR #2254, opened 2023-01-15
update frame parameter according to LAMMPS compute command

mingzhong15:
here i add two functions in `pair_deepmd.h` and `pair_deepmd.cpp`, so that we can update the frame parameter `fparam` according to LAMMPS compute command or ttm_fix command during DPMD simulations

- `make_compute_fparam`, we can use `compute` command in the LAMMPS to update the frame parameter at each simulation step. for example, the temperature of the whole system.

example:
```cpp
pair_style     deepmd cp.pb compute TEMP
compute        TEMP all temp
```

- `make_ttm_fparam`, similar to `make_ttm_aparam`, just a version for fparam-dependent DP model.

(In ultrafast laser irradiated metal foil, if the thickness is less than the mean free path of excited electrons, the electron-subsystem can be considered as uniformly heated due to ballistic transport of electrons. So we can assume a uniform distribution of electron temperature, so frame parameter woule be suitable for describing such two-temperature state.)
example:
```cpp
pair_style      deepmd cp.pb ttm LASER
pair_coeff      * *

fix             LASER all ttm/mod 1213 parameter.txt ${Nx_grid} 1 1 TTM_init.dat ${TE_DUMP} TTM_out.dat
```


njzjz (changes requested):
1. Please fix the compilation errors.
2. Please update documentation for your new parameters.

wanghan-iapcm (commented):
I would suggest changing `compute` to `fparam_from_compute`. You use a compute style to assign the value of `fparam`, the name `fparam_from_compute` is more self-explanatory.

wanghan-iapcm:
Also the user interface
```
pair_style     deepmd cp.pb fparam_from_compute TEMP
compute        TEMP all temp
```
should be better.

wanghan-iapcm:
Please update the doc https://github.com/deepmodeling/deepmd-kit/blob/master/doc/third-party/lammps-command.md

mingzhong15:
Thank you for your advice, i had 
1. changed the function name from `compute` to `fparam_from_compute` for clarity
2. changed the user interface from `compute` to `fparam_from_compute`
3. updated the documentation `lammps-command.md` for `fparam_from_compute` parameters

njzjz:
@mingzhong15 It looks that you force-push the branch and causes the PR closed. Could you recover 6ac0bff1a4171f49900d72b6b88568b27ae248b2 and reopen the PR?
```
git checkout devel
git reset --hard 6ac0bff1a4171f49900d72b6b88568b27ae248b2
git push -f
```
Opening a new branch and a new PR is also acceptable.

mingzhong15:
@njzjz sorry, i am working on it

njzjz (approved):

wanghan-iapcm (approved):

Issue #2236, opened 2023-01-10
[Feature Request] support C library for LAMMPS


njzjz:
### Summary

Support C library for LAMMPS.

### Detailed Description

There are some APIs not implemented yet:

- [x] fparam, aparam
  - [x] `DeepPot` #2415
  - [x] `DeepPotModelDevi` #2486
  - [x] add tests for them #2351
- [x] `deepmd::select_by_type` #2491
  - [x] add tests #2337
- [x] `deepmd::select_map` #2491
  - [x] add tests #2338
- [x] `deepmd::read_file_to_string` #2412
  - [x] add tests #2338
- [x] `deepmd::deepmd_exception` #2344
  - [x] add tests #2338
- [x] `deep_pot_model_devi` stats
  - [x] `deep_pot_model_devi.compute_avg` (no energy) #2493
  - [x] `deep_pot_model_devi.compute_std` #2493
  - [x] <del>`deep_pot_model_devi.compute_std_e`</del> removed in #2492
  - [x] `deep_pot_model_devi.compute_std_f` #2493
  - [x] `deep_pot_model_devi.compute_relative_std` #2493
  - [x] `deep_pot_model_devi.compute_relative_std_f` #2493

After resolving all the above, we can support the C API for LAMMPS.
- [ ] support C API for LAMMPS
  - [x] add tests for `compute deeptensor/atom` #2507
  - [ ] add tests for `kspace_style pppm/dplr`

PR #2309, opened 2023-02-10
support multiple frames DeepPot in C/hpp API

njzjz:
* Add DeepPot C API v2 to support multiple frames. Also, preserve the arguments for fparam and aparam for future use (planned in #2236). V1 is kept for API and ABI compatibility.
* Support multiple frames for DeepPot hpp API. The API compatibility is kept. Add tests for new situation.

### DIFF
--- a/source/install/test_cc.sh
+++ b/source/install/test_cc.sh
@@ -21,6 +21,6 @@ make install
 mkdir -p ${BUILD_TMP_DIR}/exec_tests
 cd ${BUILD_TMP_DIR}/exec_tests
 
-${INSTALL_PREFIX}/bin/runUnitTests_lib
-${INSTALL_PREFIX}/bin/runUnitTests_cc
+#${INSTALL_PREFIX}/bin/runUnitTests_lib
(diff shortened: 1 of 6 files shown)


source/install/test_cc.sh
```diff
@@ -21,6 +21,6 @@ make install
 mkdir -p ${BUILD_TMP_DIR}/exec_tests
 cd ${BUILD_TMP_DIR}/exec_tests
 
-${INSTALL_PREFIX}/bin/runUnitTests_lib
-${INSTALL_PREFIX}/bin/runUnitTests_cc
+#${INSTALL_PREFIX}/bin/runUnitTests_lib
+#${INSTALL_PREFIX}/bin/runUnitTests_cc
```
wanghan-iapcm, line 25: Why are the test for lib and cc commented out ?

njzjz: Typo - I want to make local tests faster but don't want to push it.

wanghan-iapcm (approved):

PR #2313, opened 2023-02-11
Implementation of se_a_mask op

dingye18:
* Implement se_a_mask op. Training and inference (c++ interface) are tested. Unit test files are required.

* Fix bug in atom pref setting for forces.

* Add the unit test for descrpt se_a_mask

* Remove the unit test for dp mask model

* Update dp mask system path for unit test


deepmd/descriptor/se_a_mask.py
```diff
@@ -0,0 +1,833 @@
...
+from .descriptor import (
+    Descriptor,
+)
+from .se import (
+    DescrptSe,
+)
+
+
+@Descriptor.register("se_a_mask")
+class DescrptSeAMask(Descriptor):
```
njzjz, line 54: To make `dp compress` work, you have to make `DescrptSeAMask` inherit from `DescrptSeA`. Is every method in `DescrptSeAMask` different from `DescrptSeA`?

source/op/descrpt_se_a_mask.cc
```diff
@@ -0,0 +1,342 @@
...
+
+REGISTER_OP("DescrptSeAMask")
+    .Attr("T: {float, double} = DT_DOUBLE")
+    .Input("coord: T")
+    .Input("type: int32")
+    .Input("mask: int32")
+    .Input("box: T")         // Not used in practice
+    .Input("natoms: int32")  // Not used in practice
+    .Input("mesh: int32")    // Not used in practice
+    .Attr("total_atom_num: int")
```
njzjz, line 21: It is difficult to understand the difference between `natoms` and `total_atom_num`. Is there a better name?

deepmd/fit/ener.py
```diff
@@ -549,16 +554,20 @@ def build(
...
-        if self.numb_aparam > 0:
-            aparam = input_dict["aparam"]
-            aparam = tf.reshape(aparam, [-1, self.numb_aparam])
-            aparam = (aparam - t_aparam_avg) * t_aparam_istd
-            aparam = tf.reshape(aparam, [-1, self.numb_aparam * natoms[0]])
+
+        if self.use_aparam_as_mask:
+            aparam = None
+        else:
+            aparam = None
```
njzjz, line 565: It seems that `aparam = None` can be put outside the `if...else`.

njzjz (changes requested):

source/op/descrpt_se_a_mask.cc
njzjz: Or at least add a doc to it.

deepmd/descriptor/se_a_mask.py
dingye18: `comput_input_stats` is not supported in se_a_mask. 
Thus, `merge_input_stats` and `enable_compression` may not supported either.

deepmd/utils/argcheck.py
```diff
@@ -374,6 +374,57 @@ def descrpt_se_atten_args():
     ]
 
 
+@descrpt_args_plugin.register("se_a_mask")
+def descrpt_se_a_mask_args():
+    doc_sel = 'This parameter set the number of selected neighbors for each type of atom. It can be:\n\n\
```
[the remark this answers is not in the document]
dingye18: Fixed in the latest commit.

deepmd/descriptor/se_a_mask.py
dingye18: In the latest commit. DescrptSeAMask is inherit from DescrptSeA.
Methods `get_rcut`, `compute_input_stats`, `build`, `prod_force_virial` are overloaded. 
However, I can't verify if `dp compress` works with `se_a_mask`.

source/op/descrpt_se_a_mask.cc
dingye18: It's removed for now. Variable `total_atom_num` can also be fetched from `natoms`.

deepmd/fit/ener.py
dingye18: Fixed in latest commit.

dingye18: Fixed in latest commit.

deepmd/utils/argcheck.py
```diff
@@ -374,6 +374,57 @@ def descrpt_se_atten_args():
...
+    doc_resnet_dt = 'Whether to use a "Timestep" in the skip connection'
+    doc_type_one_side = r"If true, the embedding network parameters vary by types of neighbor atoms only, so there will be $N_\text{types}$ sets of embedding network parameters. Otherwise, the embedding network parameters vary by types of centric atoms and types of neighbor atoms, so there will be $N_\text{types}^2$ sets of embedding network parameters."
+    doc_exclude_types = "The excluded pairs of types which have no interaction with each other. For example, `[[0, 1]]` means no interaction between type 0 and type 1."
+    doc_precision = f"The precision of the embedding net parameters, supported options are {list_to_doc(PRECISION_DICT.keys())} Default follows the interface precision."
+    doc_trainable = "If the parameters in the embedding net is trainable"
+    doc_seed = "Random seed for parameter initialization"
+    doc_set_davg_zero = "Set the normalization average to zero. This option should be set when `atom_ener` in the energy fitting is used"
```
njzjz, line 392: It looks that the normalization average is always zero, and thus, this option does not take any effect as the document describes.

deepmd/descriptor/se_a_mask.py
```diff
@@ -0,0 +1,436 @@
...
+            )
+        self.sub_sess = tf.Session(graph=sub_graph, config=default_tf_session_config)
+        self.original_sel = None
+
+    def get_rcut(self) -> float:
+        """
+        Returns: the cutoff radius:
+        """
+        warnings.warn("The cutoff radius is not used for this descriptor")
+        return -1.0
```
njzjz, line 241: I remember that `rcut` will be saved into the model and loaded during the MD simulations. Will `-1` work? I think it should be `999.0` or `float("inf")` (not sure if inf works) here.

dingye18: I did test the c++ interface for frozen model with this op. It works.

deepmd/utils/argcheck.py
dingye18: This option is deleted in the latest commit.

njzjz (approved):

wanghan-iapcm (changes requested):
- please add to the doc how to use the se_a_mask and provide an example
- I do not see the necessity of the changes in the deepmd/loss/ener.py.

dingye18:
> * please add to the doc how to use the se_a_mask and provide an example
> * I do not see the necessity of the changes in the deepmd/loss/ener.py.

* The doc for se_a_mask is added. Example input and data system are put into `examples/zinc_protein` directory.
* The changes in `deepmd/loss/ener.py` is necessary. Since the return `virial` tensor is set to be None in `se_a_mask` and `energy.npy` is not provided by default. The previous version will  throw an error when doing `tf.reshape(virial, [-1])` and trying to find `energy.npy` for `l2_ener_loss` calculation.

wanghan-iapcm (approved):

PR #2328, opened 2023-02-17
Delete dp4mask.pb....

dingye18:
I just found that the test binary graph dp4mask.pb was put in the source directory of deepmd-kit.....


njzjz:
We should add `*.pb` to `.gitignore`.

njzjz (approved):

wanghan-iapcm (approved):

PR #2343, opened 2023-02-24
Support multiple sets in one system for mixed_type format

iProzd:


deepmd/utils/data.py
```diff
@@ -61,15 +61,22 @@ def __init__(
...
+            "then all of the sets in this system should be of mixed_type format!"
+        )
+        if self._check_mode(self.dirs[0]):
+            for set_item in self.dirs[1:]:
+                assert self._check_mode(set_item), error_format_msg
+            self.mixed_type = True
+        else:
+            for set_item in self.dirs[1:]:
+                assert not self._check_mode(set_item), error_format_msg
+            self.mixed_type = False
```
njzjz, line 76: Here two conditions can be merged
```suggestion
        self.mixed_type = self._check_mode(self.dirs[0]):
        for set_item in self.dirs[1:]:
            assert self._check_mode(set_item) == self.mixed_type, error_format_msg
```

deepmd/utils/data.py
```diff
@@ -480,7 +478,19 @@ def _load_set(self, set_name: DPPath):
                 )
 
         if self.mixed_type:
-            real_type = self.atom_type_mix.reshape([nframes, self.natoms])
+            # nframes x natoms
+            atom_type_mix = self._load_type_mix(set_name)
+            if self.enforce_type_map:
+                try:
+                    atom_type_mix_ = self.type_idx_map[atom_type_mix].astype(np.int32)
+                except RuntimeError as e:
```
njzjz, line 486: Isn't it `IndexError`?

njzjz (changes requested):

njzjz:
Please remove "Note that we put all the frames in one set set.000" from the documentation.

wanghan-iapcm (approved):

njzjz (approved):

PR #2348, opened 2023-02-26
Support different learning rate settings for each fitting net in multi-task mode

HuangJiameng:
It is a breaking change to the user interface in multi-task mode. Please review carefully.

Replace "learning_rate" with "learning_rate_dict" in multi-task mode to support different learning rate settings for each fitting net, which is familiar to "loss_dict". The format of `lcurve.out` will be like the following:
```
#  step   sys1_rmse_val sys1_rmse_trn   sys1_rmse_e_val sys1_rmse_e_trn   sys1_rmse_f_val sys1_rmse_f_trn   sys1_lr  
 sys2_rmse_val sys2_rmse_trn   sys2_rmse_e_val sys2_rmse_e_trn   sys2_rmse_f_val sys2_rmse_f_trn   sys2_lr
...
```
The learning rate of each system will be printed separately.


HuangJiameng:
@iProzd Please review this PR.

njzjz:
> It is a breaking change to the user interface in multi-task mode.

The breaking change is not necessary

HuangJiameng:
> > It is a breaking change to the user interface in multi-task mode.
> 
> The breaking change is not necessary. `learning_rate_dict` can be optional

supp. What do you think? @iProzd

deepmd/utils/argcheck.py
```diff
@@ -1239,6 +1251,7 @@ def gen_doc(*, make_anchor=True, make_link=True, **kwargs):
         make_anchor = True
     ma = model_args()
     lra = learning_rate_args()
+    lrda = learning_rate_dict_args()
```
njzjz, line 1254: Please also add to `gen_args` and `gen_json`.

deepmd/train/trainer.py
```diff
@@ -269,21 +269,48 @@ def fitting_net_init(fitting_type_, descrpt_type_, params):
...
-        elif scale_by_worker == "sqrt":
-            self.scale_lr_coef = np.sqrt(self.run_opt.world_size).real
-        else:
-            self.scale_lr_coef = 1.0
-        lr_type = lr_param.get("type", "exp")
-        if lr_type == "exp":
-            self.lr = LearningRateExp(
-                lr_param["start_lr"], lr_param["stop_lr"], lr_param["decay_steps"]
-            )
+        if not self.multi_task_mode:
```
njzjz, line 272: Here a lot of codes are duplicated. I suggest creating a new method that returns lr and coef to merge duplicated codes.

deepmd/train/trainer.py
```diff
@@ -671,6 +709,45 @@ def _build_training(self):
         else:
             self.train_op = {}
             for fitting_key in self.fitting_type_dict:
+                if self.run_opt.is_distrib:
```
njzjz, line 712: These codes can also be merged to a method that returns optimizer.

njzjz (changes requested):

iProzd:
> 

I agree that `learning_rate_dict` should be optional, which is not the same as the necessary of `loss_dict`. In multi-task mode, user must define each loss for each dataset, since the labels might be different, while user can provide one single learning rate under most situations. We shall allow both the `learning_rate` and `learning_rate_dict` in `multi-task` mode: maybe we can process the jdata in argcheck, if user use one single `learning_rate`, we can extend it to multiple `learning_rate_dict`?

iProzd:
> > > It is a breaking change to the user interface in multi-task mode.
> > 
> > 
> > The breaking change is not necessary. `learning_rate_dict` can be optional
> 
> supp. What do you think? @iProzd

I agree that `learning_rate_dict` should be optional, which is not the same as the necessary of `loss_dict`. In multi-task mode, user must define each loss for each dataset, since the labels might be different, while user can provide one single learning rate under most situations. We shall allow both the `learning_rate` and `learning_rate_dict` in `multi-task` mode: maybe we can process the jdata in argcheck, if user use one single `learning_rate`, we can extend it to multiple `learning_rate_dict`? @HuangJiameng @njzjz

deepmd/train/trainer.py
```diff
@@ -597,69 +654,43 @@ def _build_network(self, data, suffix=""):
...
+                    optimizer = (
+                        tf.train.experimental.enable_mixed_precision_graph_rewrite(
+                            optimizer
+                        )
+                    )
+                else:
+                    optimizer = tf.mixed_precision.enable_mixed_precision_graph_rewrite(
+                        optimizer
+                    )
+
```
njzjz, line 693: This is actually what `_build_optimizer` is. Can you extract it to a new method?

source/tests/test_init_frz_model_multi.py
```diff
@@ -61,6 +61,7 @@ def _init_models():
     jdata = j_loader(str(tests_path / os.path.join("init_frz_model", "input.json")))
     fitting_config = jdata["model"].pop("fitting_net")
     loss_config = jdata.pop("loss")
+    learning_rate_config = jdata.pop("learning_rate")
```
njzjz, line 64: Is there still a breaking change? Old tests should be passed without modification.

deepmd/train/trainer.py
```diff
@@ -829,22 +913,27 @@ def train(self, train_data=None, valid_data=None):
                         train_batches = {}
                         valid_batches = {}
                         # valid_numb_batch_dict
-                        for fitting_key in train_data:
-                            train_batches[fitting_key] = [
-                                train_data[fitting_key].get_batch()
+                        for fitting_key_ii in train_data:
```
wanghan-iapcm, line 916: Is it really necessary to change the name of iterator?

deepmd/train/trainer.py
```diff
@@ -597,69 +654,43 @@ def _build_network(self, data, suffix=""):
...
-                )
-            elif _TF_VERSION < Version("2.4.0"):
-                optimizer = tf.train.experimental.enable_mixed_precision_graph_rewrite(
-                    optimizer
-                )
-            else:
-                optimizer = tf.mixed_precision.enable_mixed_precision_graph_rewrite(
-                    optimizer
-                )
         if not self.multi_task_mode:
```
wanghan-iapcm, line 663: Do not write large if-else condition. try to pack optimizer-making codes in functions and call the functions in conditions.

deepmd/train/trainer.py
```diff
@@ -1087,7 +1213,7 @@ def print_on_training(
                     prop_fmt = "   %11.2e"
                     for k in train_results[fitting_key].keys():
                         print_str += prop_fmt % (train_results[fitting_key][k])
-        print_str += "   %8.1e\n" % cur_lr
```
wanghan-iapcm, line 1090: lr in standard mode would be missing?

wanghan-iapcm (changes requested):

source/tests/test_init_frz_model_multi.py
HuangJiameng: There is no breaking change now. I test it with water demo that using "learning_rate" in multi-task mode.

deepmd/train/trainer.py
HuangJiameng: Yes, it is necessary. Otherwise, it will cover the `fitting_key` for `valid_on_the_fly`.

source/tests/test_init_frz_model_multi.py
njzjz, line 64: Could you keep the old test and add a new test for your PR?

wanghan-iapcm (approved):

HuangJiameng:
> Could you keep the old test and add a new test for your PR?

When checking this, I find that the problem is whether the new config inherit the learning rate of the system from the init model? Currently, most pre-trained models do not use a `learning_rate_dict`, and in order to maintain backward compatibility, new "input.json" needs to manually set the learning rate of each system in the pre-trained model. This is expected to cause some trouble... @iProzd  Do you have any better ideas?

HuangJiameng:
Perhaps the settings should be prioritized in the following order: User-provided > Pre-trained model > Default.

njzjz (approved):

PR #2351, opened 2023-02-28
fix C++ inference of multiple frames with fparam/aparam

njzjz:
Fix C++ inference of multiple frames with fparam/aparam. The input fparam/aparam can be in a single or multiple frames. Add Python and C++ tests for fparam/aparam.


wanghan-iapcm (commented):
may not be a good idea to place the tiling of the fparam/aparam into the `validate_fparam_aparam`. 
Write an new private method to do the tiling?

wanghan-iapcm (approved):

PR #830, opened 2021-07-06
speedup tabulate cuda kernel by reducing shm using

darelbeida:
This PR works on tabulate fusion CUDA kernel, including:

* speedup `tabulate_fusion_fifth_order_polynomial` kernel 1.3 times by reducing shm using (and thread syncing). 
* fix misusage between `KTILE` and `MTILE` in `tabulate_fusion_grad_fifth_order_polynomial` kernel.
* format the `warp_idx` variable used in `tabulate_fusion_grad_fifth_order_polynomial` kernel.

PR #2357, opened 2023-02-28
fix ROCm tabulate_fusion_se_a_grad_fifth_order_polynomial

### LINKED ISSUE #2303: [BUG] Run MD with LAMMPS has a temperature instability problem on the AMD ROCM platform
### Bug summary

When executing the calculation example, the temperature is unstable when LAMMPS runs the force field trained by deepmd

### DeePMD-kit Version

master branch

### TensorFlow Version

tensorflow2.7.0

Offline packages

in file to add the command line：
plugin load libdeepmd_lmp.so

The output result of lammps operation:
 the temperature value is unstable. When temp=2000k, more than 10,000 temp values will appear during the operation.

### Steps to Reproduce

1. install deepmd-kit
2. install lammps
3. run
4. Draw a graph of temperature change

_No response_

njzjz:
Follow #830 to fix #2303.

### DIFF
--- a/source/lib/src/rocm/tabulate.hip.cu
+++ b/source/lib/src/rocm/tabulate.hip.cu
@@ -162,7 +162,7 @@ __global__ void tabulate_fusion_se_a_grad_fifth_order_polynomial(
   }
   __syncthreads();
   FPTYPE ago = __shfl(em_x[block_idx * nnei + nnei - 1], 0);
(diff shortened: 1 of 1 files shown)


denghuilu (approved):

wanghan-iapcm:
@denghuilu @njzjz do we have similar issue in the cuda implementation?

PR #2358, opened 2023-03-01
fix: add declaration of strlen to deepmd.hpp

njzjz:
Fix `error: ‘strlen’ was not declared in this scope`


njzjz:
#2359 should fix the CI...

wanghan-iapcm (approved):

PR #2363, opened 2023-03-02
improve cxx11 abi flags and messages for Red Hat Developer Toolset

njzjz:
Since the Red Hat Developer Toolset does not support the custom `_GLIBCXX_USE_CXX11_ABI` flag, this PR improves the flag and the message.
- if both `_GLIBCXX_USE_CXX11_ABI=0` and `_GLIBCXX_USE_CXX11_ABI=1` work, throw a warning and set `_GLIBCXX_USE_CXX11_ABI=1` (this flag should not take actual effect).
- if both flags do not work, throw an error message about the compiler issue.

### DIFF
--- a/source/cmake/Findtensorflow.cmake
+++ b/source/cmake/Findtensorflow.cmake
@@ -119,10 +119,10 @@ if(BUILD_CPP_IF)
       )
     endif()
   endforeach()
-else(BUILD_CPP_IF AND NOT USE_TF_PYTHON_LIBS)
+else(BUILD_CPP_IF)
   message(
     STATUS "Disabled cpp interface build, looking for tensorflow_framework")
-endif(BUILD_CPP_IF AND NOT USE_TF_PYTHON_LIBS)
+endif(BUILD_CPP_IF)
 
 # tensorflow_framework
 if(NOT TensorFlowFramework_FIND_COMPONENTS)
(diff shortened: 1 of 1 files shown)


Yi-FanLi (approved):
I tested this change with GCC 12.1.1 20220628 (Red Hat 12.1.1-3). Before applying this PR, the error message is:
CMake Error at cmake/Findtensorflow.cmake:345 (message):                                                                                                                                                            
  Failed to detect OP_CXX_ABI, please set it manually                                                                                                                                                               
Call Stack (most recent call first):                                                                                                                                                                                
  CMakeLists.txt:134 (find_package)                                                                                                                                                                                 
                                                                                                                                                                                                                    
                                                                                                                                                                                                                    
-- Configuring incomplete, errors occurred!      

After applying this change, the error message becomes:
CMake Error at cmake/Findtensorflow.cmake:352 (message):
  Both _GLIBCXX_USE_CXX11_ABI=0 and 1 do not work.  The reason may be that
  your C++ compiler (e.g.  Red Hat Developer Toolset) does not support the
  custom cxx11 abi flag.
Call Stack (most recent call first):
  CMakeLists.txt:134 (find_package)


-- Configuring incomplete, errors occurred!

Thus I think this PR behaves as expected.

wanghan-iapcm (approved):

PR #2366, opened 2023-03-04
support mixed natoms training

njzjz:


wanghan-iapcm (commented):
Is it possible to on-the-fly pad virtual atom for a batch of training data? This would not need the user to rewrite their training data.

wanghan-iapcm (approved):

iProzd (approved):

PR #2371, opened 2023-03-07
Adapt DeePMD in Ascend platform (NPU)

ZhengdQin:
1. Add transfer-to-ascend module, one can use command "dp transfer-to-ascend mix_precision -i model.pb -o Ascend_transfer.pb" to transfer the "model.pb" to mix-precision "Ascend_transfer.pb", the transfered graph can be excuted on Ascend platform.
2. Adapt dp test module on Ascend platform.
3. Adapt Lammps +deepMD on Ascend platform.


njzjz:
Close, as this PR seems to be dead.

PR #2372, opened 2023-03-08
support mixed systems training

njzjz:
With the new option, a batch of data could come from all systems with different numbers of atoms and is generated and padded into a mixed system on the fly.

This PR is independent of #2366, which provides a manual way to use mixed systems. In theory #2366 should have better performance and this PR is more convenient.

### DIFF
--- a/deepmd/utils/data_system.py
+++ b/deepmd/utils/data_system.py
@@ -393,31 +393,60 @@ class DeepmdDataSystem:
         if not hasattr(self, "default_mesh"):
             self._make_default_mesh()
         if not self.mixed_systems:
-            if sys_idx is not None:
-                self.pick_idx = sys_idx
-            else:
-                # prob = self._get_sys_probs(sys_probs, auto_prob_style)
-                self.pick_idx = dp_random.choice(
-                    np.arange(self.nsystems), p=self.sys_probs
-                )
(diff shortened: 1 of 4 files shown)


iProzd (approved):

deepmd/utils/data_system.py
```diff
@@ -375,19 +383,88 @@ def get_batch(self, sys_idx: Optional[int] = None):
...
+            b_data["natoms_vec"] = self.natoms_vec[self.pick_idx]
+            b_data["default_mesh"] = self.default_mesh[self.pick_idx]
         else:
-            # prob = self._get_sys_probs(sys_probs, auto_prob_style)
-            self.pick_idx = dp_random.choice(np.arange(self.nsystems), p=self.sys_probs)
-        b_data = self.data_systems[self.pick_idx].get_batch(
-            self.batch_size[self.pick_idx]
-        )
-        b_data["natoms_vec"] = self.natoms_vec[self.pick_idx]
-        b_data["default_mesh"] = self.default_mesh[self.pick_idx]
```
wanghan-iapcm, line 390: better to write two functions for get_batch (standard) and get_batch (mixed) and switch between the functions in the if-else condition.

wanghan-iapcm (approved):

PR #2415, opened 2023-03-20
C API: support fparam and aparam for DeepPot

njzjz:


njzjz:
Some extra effort is still required for `DeepPotModelDevi`.

wanghan-iapcm (approved):

PR #2461, opened 2023-04-17
fix nopbc in finetune, DeepTensor test, and DipoleChargeModifier

njzjz:


wanghan-iapcm (changes requested):
Actually the dplr data modifier does not work with open system (nopbc), because we use the Ewald algorithm, which assumes PBC, to compute the Coulomb interaction. 

We shall through an error rather than taking box to `None`

iProzd (approved):

wanghan-iapcm (approved):

PR #2433, opened 2023-03-29
Fix model-devi with mixed_type format

iProzd:
Fix model-devi with mixed_type format;
Add UTs;
Add detection of whether models support mixed_type inference.

PR #2481, opened 2023-04-24
set specific mesh shapes for mixed type

njzjz:
This PR sets specific mesh shapes (7 for pbc and 1 for no pbc) for mixed type.

So when an operator does not throw mixed type, it will throw errors. It will be more accurate than the implementation in #2433 to check if a model supports DPRc.

`make_default_mesh` is updated and out-of-date behaviors are removed.

### DIFF
--- a/deepmd/common.py
+++ b/deepmd/common.py
@@ -208,6 +208,12 @@ def select_idx_map(atom_types: np.ndarray, select_types: np.ndarray) -> np.ndarr
 def make_default_mesh(pbc: bool, mixed_type: bool) -> np.ndarray:
     """Make mesh.
 
+    Only the size of mesh matters, not the values:
+    * 6 for PBC, no mixed types
+    * 0 for no PBC, no mixed types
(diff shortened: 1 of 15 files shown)


deepmd/common.py
```diff
@@ -205,28 +205,23 @@ def select_idx_map(atom_types: np.ndarray, select_types: np.ndarray) -> np.ndarr
     return np.concatenate(idx_map)
 
 
-# TODO not really sure if the docstring is right the purpose of this is a bit unclear
-def make_default_mesh(test_box: np.ndarray, cell_size: float = 3.0) -> np.ndarray:
-    """Get number of cells of size=`cell_size` fit into average box.
+def make_default_mesh(pbc: bool, mixed_type: bool) -> np.ndarray:
+    """Make mesh.
```
wanghan-iapcm, line 209: provide an explanation for the hack?

iProzd (approved):

Issue #2499, opened 2023-05-01
[BUG] 'use_srtab' tag is not compatible with ’DP_INTERFACE_PREC=low.‘


ixbozai:
### Bug summary

Everything works just fine when using 'DP_INTERFACE_PREC=high', but when using ’DP_INTERFACE_PREC=low‘, the training crashed at the beginning. The log of the crash is like:
DEEPMD INFO    start training at lr 1.00e-03 (== 1.00e-03), decay_step 10000, decay_rate 0.944061, final lr will be 1.00e-08
2023-05-01 08:39:05.804580: F tensorflow/core/framework/tensor.cc:719] Check failed: dtype() == expected_dtype (2 vs. 1) float expected, got double
/var/spool/slurm/d/job31244/slurm_script: line 14: 204085 Aborted                 (core dumped) dp train input.json

### DeePMD-kit Version

2.2.1

### TensorFlow Version

2.0?

### How did you download the software?

conda

### Input Files, Running Commands, Error Log, etc.

dp train input.json

### Steps to Reproduce

export DP_INTERFACE_PREC=low, and using following tags in the input.json
"use_srtab":  "/home/dimer.txt",
"smin_alpha": 0.1,
"sw_rmin": 0.7,
"sw_rmax": 0.8

dimer.txt just like this
0.30	1682.51752	407.02759	80.67498	
0.40	979.46505	230.92072	46.80655	
0.50	598.45385	133.44833	29.13083	
0.60	373.45964	76.03379	18.85596	
0.70	229.20296	42.39376	12.57054	
0.80	136.66362	23.24933	8.55577	
0.90	77.60532	12.36206	5.82518	
1.00	40.14691	6.17995	3.92981	
1.10	17.48767	2.71482	2.63077	
1.20	4.58016	0.79020	1.74970	
1.30	-2.19558	-0.28164	1.15680	
1.40	-5.45479	-0.87593	0.75982	
1.50	-6.95876	-1.19893	0.49499	
1.60	-7.63337	-1.36857	0.31941	
1.70	-7.88454	-1.46196	0.20365	
1.80	-7.90061	-1.52263	0.12778	
1.90	-7.77277	-1.57175	0.07835	
2.00	-7.54904	-1.61792	0.04637

ixbozai:
Thanks for the fast response. But the problem still there after I followed the fix in #2500

njzjz:
> But the problem still there

I can't reproduce this behavior.

[...]

PR #2500, opened 2023-05-01
fix dtype in PairTabOp

njzjz:
Fix https://github.com/deepmodeling/deepmd-kit/issues/2499.

### DIFF
--- a/deepmd/utils/pair_tab.py
+++ b/deepmd/utils/pair_tab.py
@@ -9,9 +9,15 @@ from scipy.interpolate import (
     CubicSpline,
 )
 
+from deepmd.env import (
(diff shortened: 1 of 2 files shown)


wanghan-iapcm (approved):

ixbozai:
Hi, after the new release of 2.2.2, the 'use_srtab' is compatible with float precision, but new problem comes. After 20000 steps of  training, all the RMSEs become NaN, and the final force field is just unusable (crush immediately when running MD). Again, this happened only for training with float precision.

njzjz:
There are probably some problems in data stat when using pairwise tabulated potentials. The initial loss value is too large.

PR #2519, opened 2023-05-10
lmp/ipi: remove float precision

njzjz:
For the following reason:
(1) api_cc supports both interfaces of the model and will convert between double and float;
(2) LAMMPS and i-Pi only support double.

Thus, it's useless to have float precision for the LAMMPS plugin or i-Pi driver. This PR removes it to make codes more clear.

Also, fix a compilation error when setting `DP_USING_C_API=OFF`.


wanghan-iapcm (approved):

Entropy-Enthalpy:
> LAMMPS and i-Pi only support double.

Is there any source that states "LAMMPS only supports double"?

wanghan-iapcm:
source code

PR #2532, opened 2023-05-16
Enable model compression for se_atten

nahso:


njzjz:
pre-commit.ci run

deepmd/train/trainer.py
```diff
@@ -180,6 +180,9 @@ def fitting_net_init(fitting_type_, descrpt_type_, params):
             self.fitting_type = fitting_type
             fitting_param.pop("type", None)
             fitting_param["descrpt"] = self.descrpt
+            fitting_param["compress"] = self.is_compress
+            if self.is_compress and fitting_type != "ener":
+                raise RuntimeError("compression can only be performed in energy model, but " + fitting_type + " is given")
```
njzjz, line 185: I don't think there is any reason to limit the energy model, which breaks the existing compression for the tensor models.

deepmd/fit/ener.py
```diff
@@ -722,6 +729,9 @@ def init_variables(
         self.fitting_net_variables = get_fitting_net_variables_from_graph_def(
             graph_def, suffix=suffix
         )
+        if isinstance(self.descrpt, deepmd.descriptor.DescrptSeAtten):
+            type_embedding = get_tensor_by_name_from_graph(graph, 't_typeebd')
+            self.compress_type_embedding = tf.convert_to_tensor(type_embedding)
```
njzjz, line 734: It's more straightforward to do it directly in the type embedding class (`TypeEmbed.init_variables`) than doing it in the fitting network.
Actually, it's not directly related to the fitting network.

deepmd/utils/tabulate.py
```diff
@@ -175,7 +175,24 @@ def build(
         """
         # tabulate range [lower, upper] with stride0 'stride0'
         lower, upper = self._get_env_mat_range(min_nbor_dist)
-        if isinstance(self.descrpt, deepmd.descriptor.DescrptSeA):
+        if isinstance(self.descrpt, deepmd.descriptor.DescrptSeAtten):
```
njzjz, line 178: I would also suggest supporting the situation for se_a + type_embedding. They can share the same OP.

deepmd/descriptor/se_atten.py
```diff
@@ -283,6 +296,165 @@ def compute_input_stats(
...
+        two_side_type_embedding = np.concatenate([type_embedding_nei, type_embedding_center], -1) # ntypes * ntypes * (Y+Y)
+        two_side_type_embedding = np.reshape(two_side_type_embedding, [-1, two_side_type_embedding.shape[-1]])
+        return two_side_type_embedding
+
+
+    def _get_two_side_embedding_net_variable(self, graph_def, varialbe_name):
+        ret = {}
+        for i in range(1, self.layer_size + 1):
+            node = get_pattern_nodes_from_graph_def(graph_def,
+                                                    f'filter_type_all/{varialbe_name}_{i}_two_side_ebd') \
```
njzjz, line 409: `filter_type_all` should be `filter_type_all{suffix}`

deepmd/descriptor/se_atten.py
```diff
@@ -283,6 +296,165 @@ def compute_input_stats(
...
+            [f'filter_type_all/{varialbe_name}_{i}_two_side_ebd']
```
njzjz, line 410: `filter_type_all` -> `filter_type_all{suffix}`

deepmd/descriptor/se_atten.py
```diff
@@ -960,6 +1130,35 @@ def init_variables(
             The suffix of the scope
         """
         super().init_variables(graph=graph, graph_def=graph_def, suffix=suffix)
+        self.two_side_embeeding_net_variables = {}
+
+        for i in range(1, self.layer_size + 1):
+            node = get_pattern_nodes_from_graph_def(graph_def,
+                                                    f'filter_type_all/matrix_{i}_two_side_ebd') \
```
njzjz, line 1137: `filter_type_all` -> `filter_type_all{suffix}`

deepmd/descriptor/se_atten.py
```diff
@@ -960,6 +1130,35 @@ def init_variables(
...
+            [f'filter_type_all/matrix_{i}_two_side_ebd']
```
njzjz, line 1138: `filter_type_all` -> `filter_type_all{suffix}`

deepmd/descriptor/se_atten.py
```diff
@@ -960,6 +1130,35 @@ def init_variables(
...
+            dtype = tf.as_dtype(node.dtype).as_numpy_dtype
+            tensor_shape = tf.TensorShape(node.tensor_shape).as_list()
+            if (len(tensor_shape) != 1) or (tensor_shape[0] != 1):
+                tensor_value = np.frombuffer(
+                    node.tensor_content, dtype=tf.as_dtype(node.dtype).as_numpy_dtype
+                )
+            else:
+                tensor_value = get_tensor_by_type(node, dtype)
+            self.two_side_embeeding_net_variables[f'filter_type_all/matrix_{i}_two_side_ebd'] = np.reshape(tensor_value, tensor_shape)
```
njzjz, line 1147: `filter_type_all` -> `filter_type_all{suffix}`

deepmd/descriptor/se_atten.py
```diff
@@ -960,6 +1130,35 @@ def init_variables(
...
+
+            node = get_pattern_nodes_from_graph_def(graph_def,
+                                                    f'filter_type_all/bias_{i}_two_side_ebd') \
```
njzjz, line 1150: `filter_type_all` -> `filter_type_all{suffix}`

deepmd/descriptor/se_atten.py
```diff
@@ -960,6 +1130,35 @@ def init_variables(
...
+            [f'filter_type_all/bias_{i}_two_side_ebd']
```
njzjz, line 1151: `filter_type_all` -> `filter_type_all{suffix}`

deepmd/descriptor/se_atten.py
```diff
@@ -960,6 +1130,35 @@ def init_variables(
...
+            dtype = tf.as_dtype(node.dtype).as_numpy_dtype
+            tensor_shape = tf.TensorShape(node.tensor_shape).as_list()
+            if (len(tensor_shape) != 1) or (tensor_shape[0] != 1):
+                tensor_value = np.frombuffer(
+                    node.tensor_content, dtype=tf.as_dtype(node.dtype).as_numpy_dtype
+                )
+            else:
+                tensor_value = get_tensor_by_type(node, dtype)
+            self.two_side_embeeding_net_variables[f'filter_type_all/bias_{i}_two_side_ebd'] = np.reshape(tensor_value, tensor_shape)
```
njzjz, line 1160: `filter_type_all` -> `filter_type_all{suffix}`

deepmd/fit/ener.py
```diff
@@ -470,7 +474,10 @@ def build(
         if input_dict is None:
             input_dict = {}
         bias_atom_e = self.bias_atom_e
-        type_embedding = input_dict.get("type_embedding", None)
+        if self.compress and isinstance(self.descrpt, deepmd.descriptor.DescrptSeAtten):
+            type_embedding = input_dict.get("type_embedding_from_graph", None)
```
njzjz, line 478: I suggest that `TypeEmbedNet.build` directly returns `type_embedding_from_graph`, and we don't need to hack `Fitting`

deepmd/utils/tabulate.py
```diff
@@ -403,7 +420,12 @@ def _get_bias(self):
         bias = {}
         for layer in range(1, self.layer_size + 1):
             bias["layer_" + str(layer)] = []
-            if isinstance(self.descrpt, deepmd.descriptor.DescrptSeA):
+            if isinstance(self.descrpt, deepmd.descriptor.DescrptSeAtten):
+                node = self.embedding_net_nodes[
+                    f"filter_type_all/bias_{layer}"
```
njzjz, line 425: self.suffix

deepmd/utils/tabulate.py
```diff
@@ -462,7 +484,12 @@ def _get_matrix(self):
         matrix = {}
         for layer in range(1, self.layer_size + 1):
             matrix["layer_" + str(layer)] = []
-            if isinstance(self.descrpt, deepmd.descriptor.DescrptSeA):
+            if isinstance(self.descrpt, deepmd.descriptor.DescrptSeAtten):
+                node = self.embedding_net_nodes[
+                    f"filter_type_all/matrix_{layer}"
```
njzjz, line 489: self.suffix

deepmd/utils/argcheck.py
```diff
@@ -334,6 +334,7 @@ def descrpt_hybrid_args():
 
 @descrpt_args_plugin.register("se_atten")
 def descrpt_se_atten_args():
+    doc_compressible = "Whether to build a compressible se_atten."
```
njzjz, line 337:
```suggestion
    doc_compressible = "Whether to build a compressible se_atten that is supported  by `dp compress`."
```

deepmd/utils/type_embed.py
```diff
@@ -127,6 +129,8 @@ def build(
         embedded_types
             The computational graph for embedded types
         """
+        if self.type_embedding_net_variables is not None:
+            return self.type_embedding_from_graph
```
njzjz, line 133: There will be three situations here when `self.type_embedding_net_variables is not None`:
- compress
- restart from an original model
- restart from a compressed model

The embedding network should be trainable for the second situation.

njzjz (changes requested):

source/lib/src/cuda/tabulate.cu
```diff
@@ -318,6 +318,219 @@ __global__ void tabulate_fusion_se_a_grad_grad_fifth_order_polynomial(
   }
 }
 
+template <typename FPTYPE, int MTILE, int KTILE>
+__global__ void tabulate_fusion_se_atten_fifth_order_polynomial(
```
wanghan-iapcm, line 322: In principle the function for `se_atten` can reuse those for `se_e2_a`.

wanghan-iapcm (changes requested):

deepmd/descriptor/se_atten.py
njzjz, line 1: Could you additionally do two tests to make sure the code is correct?
1. test the compress works in all of the following precision:
- [x] type embedding FP64, se_atten FP64
- [ ] type embedding FP64, se_atten FP32
- [ ] type embedding FP32, se_atten FP64
- [ ] type embedding FP32, se_atten FP32

njzjz, line 1: 2. test the hybrid descriptor of se_atten works with compression

njzjz (changes requested):

deepmd/utils/tabulate.py
njzjz: I closed this conversation as I think it can be supported in a separate PR. It is not resolved.

deepmd/utils/type_embed.py
```diff
@@ -158,6 +163,7 @@ def init_variables(
         graph: tf.Graph,
         graph_def: tf.GraphDef,
         suffix="",
+        model_type="original_model",
```
njzjz, line 166: Please also pass `model_type` here:
https://github.com/deepmodeling/deepmd-kit/blob/bf4521a94c7372b33d629021efe63e96af016aa6/deepmd/model/ener.py#L360

njzjz (changes requested):

njzjz (approved):
LGTM

njzjz (changes requested):

njzjz (approved):

wanghan-iapcm:
@nahso is your pr ready for review?

nahso:
> @nahso is your pr ready for review?

yes, it's ready

deepmd/utils/argcheck.py
```diff
@@ -356,6 +359,9 @@ def descrpt_se_atten_args():
     doc_attn_mask = "Whether to do mask on the diagonal in the attention matrix"
 
     return [
+        Argument(
+            "compressible", bool, optional=True, default=False, doc=doc_compressible
```
wanghan-iapcm, line 363: Shall we change the default to `True`? and write down in the document that setting it to `False` will fall back to the old behavior of `se_atten`

deepmd/utils/argcheck.py
```diff
@@ -334,9 +334,7 @@ def descrpt_hybrid_args():
 
 @descrpt_args_plugin.register("se_atten")
 def descrpt_se_atten_args():
-    doc_compressible = (
-        "Whether to build a compressible se_atten that is supported by `dp compress`."
-    )
+    doc_compressible = "Whether to build a compressible `se_atten` that is supported by `dp compress`. Attention: the compressible `se_atten` disables attention layers. Setting it to `False` will fallback to the previous behavior of `se_atten`."
```
wanghan-iapcm, line 337: In principle the model compression happens to the embedding network and is decoupled from the attention layers. I do not understand why one cannot use attention layers when he/she set the this key to True

wanghan-iapcm: I agree that when using attention layers the acceleration of the model compression may not be as significant as that without attention layers.

njzjz:
I have questions about the new key and the default value changing the behavior. When using the fine-tuning feature, the parameters are loaded from the old model, and the new default value will cause the new behavior and break the old model. @iProzd, please correct me if I am wrong.

source/op/tabulate_multi_device.cc
```diff
@@ -38,6 +39,7 @@ REGISTER_OP("TabulateFusionSeA")
     .Input("table_info: T")
     .Input("em_x: T")
     .Input("em: T")
+    .Input("two_embed: T")
```
njzjz, line 42: This is a breaking change. Old models will fail to run.

Please avoid revise old OPs. Instead, add a new OP (maybe called xxV2) that share the same kernel with the old one.

njzjz (changes requested):

source/op/tabulate_multi_device.cc
nahso: Why old models will fail to run? Passing a null pointer to this op will let it fallback to the previous op behaviour.

njzjz: That's how TensorFlow graphs work. When loading the graph from a file, the code to define the graph will never be visited again.

wanghan-iapcm (approved):

deepmd/utils/argcheck.py
```diff
@@ -334,6 +334,7 @@ def descrpt_hybrid_args():
 
 @descrpt_args_plugin.register("se_atten")
 def descrpt_se_atten_args():
+    doc_stripped_type_embedding = "Whether to strip the type embedding into a seperate embedding net. Setting it to `False` will fallback to the previous version of `se_atten` which is non-compressible. "
```
njzjz, line 337:
```suggestion
    doc_stripped_type_embedding = "Whether to strip the type embedding into a separated embedding network. Setting it to `False` will fall back to the previous version of `se_atten` which is non-compressible."
```

source/lib/src/tabulate.cc
```diff
@@ -118,6 +120,12 @@ void deepmd::tabulate_fusion_se_a_cpu(FPTYPE* out,
         FPTYPE a5 = table[table_idx * last_layer_size * 6 + 6 * kk + 5];
         FPTYPE var =
             a0 + (a1 + (a2 + (a3 + (a4 + a5 * xx) * xx) * xx) * xx) * xx;
+        if (enable_se_atten) {
+          FPTYPE t = two_embed[ii * nnei * last_layer_size +
+                               jj * last_layer_size + kk];
+          var = var * t + t;
+        }
```
njzjz, line 127: Please check here - Codecov said it is not covered.

image

njzjz (changes requested):
It can be merged after these two comments are addressed.

source/lib/src/tabulate.cc
nahso: Actually it has been tested in source/lib/tests/test_tabulate_se_a.cc:701.
image

njzjz: I see. The issue might occur on the coverage side.

njzjz (approved):

PR #2549, opened 2023-05-21
refactor: uncouple Descriptor and Fitting from Trainer

njzjz:


deepmd/train/trainer.py
```diff
@@ -470,12 +211,9 @@ def build(self, data=None, stop_batch=0, origin_type_map=None, suffix=""):
...
             if not self.is_compress and data.mixed_type:
-                assert self.descrpt_type in [
-                    "se_atten"
-                ], "Data in mixed_type format must use attention descriptor!"
-                assert self.fitting_type in [
-                    "ener"
-                ], "Data in mixed_type format must use ener fitting!"
+                assert isinstance(
+                    self.fitting, EnerFitting
+                ), "Data in mixed_type format must use ener fitting!"
```
iProzd, line 216: Why delete 'se_atten' descriptor judgement for mixed_type?

deepmd/model/ener.py
```diff
@@ -454,3 +465,38 @@ def natoms_not_match(self, force, natoms, atype):
...
+                    and do least sqaure on the errors to obtain the target shift as bias.
+            'statistic' : directly use the statistic energy bias in the target dataset.
+        """
+        self.fitting.change_energy_bias(
+            data,
+            frozen_model,
+            origin_type_map,
+            full_type_map,
+            bias_shift,
+            self.data_stat_nbatch,
```
iProzd, line 501: While `data_stat_nbatch` is suitable for data stating, maybe it's sometimes too large for `change_energy_bias`?(since this will triger GPU computation)

deepmd/train/trainer.py
njzjz: This PR uncouples the descriptor and the fitting from the trainer, as a trainer may not contain only one descriptor.
Theoretically, we should check it directly in `descriptor.build`. However, I think #2481 is a better idea to judge `mixed_type`. We don't need it if #2481 is merged.

deepmd/model/ener.py
njzjz: I see. It's a typo.

iProzd (approved):

PR #2561, opened 2023-05-24
reduce model size for dplr unittest

Yi-FanLi:
This PR uses small models to test dplr.


source/lmp/tests/lrmodel.pbtxt
njzjz, line 1: Why does the model have more lines when you reduce the size?
image

njzjz: > while the previous one is 20 M

??

njzjz: Can you reduce 1.3MB to <1.0MB?

Yi-FanLi: Yes, now it is reduced to 950 K.

njzjz (approved):

PR #2564, opened 2023-05-24
fix large files checking

njzjz:
It seems to me that the previous implementation does not really work in the CI... This PR will prevent any new large files (>1MB).


.pre-commit-config.yaml
```diff
@@ -11,6 +11,17 @@ repos:
     -   id: check-yaml
     #-   id: check-json
     -   id: check-added-large-files
+        args: ['--maxkb=1024', '--enforce-all']
+        # TODO: remove the following after resolved
+        exclude: |
+            (?x)^(
+                source/lmp/tests/lrmodel.pbtxt|
```
njzjz, line 18: Remove this line if #2561 is merged first. Otherwise, remove this line in #2561.

wanghan-iapcm (approved):

PR #2580, opened 2023-05-31
fix dplr: correct type check in get_valid_pairs

Yi-FanLi:
The binary_search method in C++ is only valid for sorted vectors. 
The following case in "type_associate" of "fix dplr" will raise error: 4 5 3 6. 4 and 3 are real atom types and stored in sel_type. The vector sel_type is sorted. 5 and 6 are associated with 4 and 3, respectively. The resulting sel_type vector is 3 4, and the resulting dpl_type vector is 6 5. If a pair 4 5 is searched, then this line

https://github.com/deepmodeling/deepmd-kit/blob/e7d68c165c20aea04af686f8a81b179154b1d679/source/lmp/fix_dplr.cpp#L195-L196

will do binary search for 4 in 3 4, which results in True; and will search for 5 in 6 5, which results in False. 

Therefore, I am copying dpl_type vector to the dpl_type_sorted vector and sort it. The binary search should be used on the sorted vector.

## The above solution is abandoned.
In stead, do not use binary_search in the type check of get_valid_pairs. 
The current method is: first find the type of the real atom, and check if the type of its bonded atom is consistent with that defined by type_associate.


source/lmp/fix_dplr.cpp
```diff
@@ -115,6 +115,8 @@ FixDPLR::FixDPLR(LAMMPS *lmp, int narg, char **arg)
   for (int ii = 0; ii < sel_type.size(); ++ii) {
     dpl_type.push_back(type_asso[sel_type[ii]]);
   }
+  dpl_type_sorted = dpl_type;
```
njzjz, line 118: Do we need to store two vectors? It seems to me that `dpl_type` is never used other than this line.

njzjz (changes requested):

source/lmp/fix_dplr.cpp
Yi-FanLi: Solved by commit [9cbc513](https://github.com/deepmodeling/deepmd-kit/pull/2580/commits/9cbc51338b9d9f4412b100d6faa85ffc5a84269c).

njzjz (approved):

source/lmp/fix_dplr.cpp
```diff
@@ -190,21 +188,46 @@ void FixDPLR::get_valid_pairs(vector<pair<int, int> > &pairs) {
...
-                      dtype[bondlist[ii][1]])) {
-      idx0 = bondlist[ii][0];
-      idx1 = bondlist[ii][1];
-    } else if (binary_search(sel_type.begin(), sel_type.end(),
-                             dtype[bondlist[ii][1]]) &&
-               binary_search(dpl_type.begin(), dpl_type.end(),
-                             dtype[bondlist[ii][0]])) {
-      idx0 = bondlist[ii][1];
-      idx1 = bondlist[ii][0];
+    it = find(sel_type.begin(), sel_type.end(), dtype[bondlist[ii][0]]);
```
wanghan-iapcm, line 191: declare `it` here rather than declaring it as a member of the class?

Yi-FanLi: OK

Yi-FanLi: Done by commit [a7639dd](https://github.com/deepmodeling/deepmd-kit/pull/2580/commits/a7639dd2f13f5dddf3e25f4faab65bd7005313f0).

wanghan-iapcm (approved):

PR #2584, opened 2023-06-02
Support minimization in dplr

Yi-FanLi:
The corresponding functions (min_pre_force and min_post_force) in fix dplr are added so that the positions and forces of Wannier centroids can be correctly assigned during minimization.

### DIFF
--- a/source/lmp/fix_dplr.cpp
+++ b/source/lmp/fix_dplr.cpp
@@ -134,6 +134,8 @@ int FixDPLR::setmask() {
   mask |= POST_INTEGRATE;
   mask |= PRE_FORCE;
   mask |= POST_FORCE;
+  mask |= MIN_PRE_FORCE;
+  mask |= MIN_POST_FORCE;
   return mask;
 }
 
@@ -166,6 +168,14 @@ void FixDPLR::setup(int vflag) {
     evflag = 0;
   }
 }
+
+/* ---------------------------------------------------------------------- */
+
+void FixDPLR::min_setup(int vflag) {
+  setup(vflag);
+}
+
+/* ---------------------------------------------------------------------- */
 
 void FixDPLR::get_valid_pairs(vector<pair<int, int> > &pairs) {
   pairs.clear();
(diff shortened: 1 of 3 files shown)


source/lmp/fix_dplr.cpp
```diff
@@ -207,9 +215,13 @@ void FixDPLR::get_valid_pairs(vector<pair<int, int> > &pairs) {
                  "find a bonded pair the types of which are not associated");
     }
     if (!(idx0 < nlocal && idx1 < nlocal)) {
-      error->all(FLERR,
-                 "find a bonded pair that is not on the same processor, "
-                 "something should not happen");
+      int idx0_ = atom->map(atom->tag[idx0]);
+      int idx1_ = atom->map(atom->tag[idx1]);
+      if (!(idx0_ > 0 && idx0_ < nlocal && idx1_ > 0 && idx1_ < nlocal)) {
```
njzjz, line 220: Should `>0` be `>=0`?

source/lmp/fix_dplr.cpp
```diff
@@ -515,6 +529,16 @@ void FixDPLR::post_force(int vflag) {
   }
 }
 
+/* ---------------------------------------------------------------------- */
+
+void FixDPLR::min_pre_force(int vflag) { pre_force(vflag); }
```
njzjz, line 534: Can you add a test to cover this line?

njzjz (changes requested):

source/lmp/fix_dplr.cpp
Yi-FanLi: Yes!

source/lmp/fix_dplr.cpp
Yi-FanLi: Solved by commit [c303b00](https://github.com/deepmodeling/deepmd-kit/pull/2584/commits/c303b00f448620c9a6df48e46df477e6665ef6bd).

njzjz (approved):

wanghan-iapcm (approved):

PR #2588, opened 2023-06-05
Reduce dp mask pb size and fix bug in  dim_fparam/dim_aparam fetching

dingye18:


source/api_c/src/c_api.cc
```diff
@@ -26,8 +26,8 @@ DP_Nlist* DP_NewNlist(int inum_,
 
 DP_DeepPot::DP_DeepPot() {}
 DP_DeepPot::DP_DeepPot(deepmd::DeepPot& dp) : dp(dp) {
-  dfparam = dp.dim_aparam();
-  daparam = dp.dim_fparam();
+  dfparam = dp.dim_fparam();
+  daparam = dp.dim_aparam();
```
njzjz, line 30: Lines 65-66 should also be fixed.

njzjz (approved):

PR #2600, opened 2023-06-10
prod_force: support multiple frames in parallel

njzjz:
The previous `prod_force` did not support multiple frames in parallel, which was slow when the batch size was large.

This PR adds support so that prod_force can be parallelized in the dimension of the samples.

When the batch size is about 70, the `prod_force` op is 10x faster than before on GPU cards.

### DIFF
--- a/source/lib/src/prod_force.cc
+++ b/source/lib/src/prod_force.cc
@@ -28,12 +28,14 @@ void deepmd::prod_force_a_cpu(FPTYPE* force,
                               const int nall,
                               const int nnei,
                               const int nframes,
+                              const int nloc_nloc,
                               const int start_index) {
   const int ndescrpt = 4 * nnei;
 
   memset(force, 0, sizeof(FPTYPE) * nframes * nall * 3);
   // compute force of a frame
-  for (int i_idx = start_index; i_idx < start_index + nframes * nloc; ++i_idx) {
+  for (int i_idx = start_index; i_idx < start_index + nframes * nloc_nloc;
+       ++i_idx) {
     int kk = i_idx / nloc;  // frame index
     int ll = i_idx % nloc;  // atom index
     int i_idx_nall = kk * nall + ll;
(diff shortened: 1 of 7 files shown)


denghuilu (approved):
LGTM

source/lib/src/prod_force.cc
```diff
@@ -27,20 +27,26 @@ void deepmd::prod_force_a_cpu(FPTYPE* force,
...
 
-  memset(force, 0, sizeof(FPTYPE) * nall * 3);
+  memset(force, 0, sizeof(FPTYPE) * nframes * nall * 3);
   // compute force of a frame
-  for (int i_idx = start_index; i_idx < start_index + nloc; ++i_idx) {
+  for (int i_idx = nframes * start_index;
+       i_idx < nframes * (start_index + nloc_nloc); ++i_idx) {
+    int kk = i_idx / nloc;  // frame index
+    int ll = i_idx % nloc;  // atom index
+    int i_idx_nall = kk * nall + ll;
```
wanghan-iapcm, line 41: The index may not be right if start_index != 0

njzjz: This is right. `start_index` only decides what indexes are computed (in this op). `i_idx` should not be affected by `start_index`.

The above codes are equivalent to
```py
  for (int i_idx = 0;
       i_idx < nframes * nloc; ++i_idx) {
    if (i_idx < nframes * start_index || i_idx >= nframes * (start_index + nloc_nloc))
      // skip in this task; should be compute in another task
      continue;
    int kk = i_idx / nloc;  // frame index
    int ll = i_idx % nloc;  // atom index
    int i_idx_nall = kk * nall + ll;
```

wanghan-iapcm: OK I see the way how you divide the task.  
Further comments:
1. start_index and nloc_nloc are very bad names, because one cannot easily see what are they referring to by their names. 
2. possible to provide default value for nloc_nloc?
3. unit test against the case of start_index != 0 is necessary

njzjz: To be clear, this is an old feature in https://github.com/deepmodeling/deepmd-kit/pull/1429. A unit test has been added in that PR to test the OP.

njzjz: I did the change in b7787ee13c1f76f78a86829c2995f12899181ceb (the commit time is not accurate):
1. change the name to `thread_nloc` and `thread_start_index`;
2. add an overload to support default value;
3. add docs.

PR #2613, opened 2023-06-13
Add zbl example

Chengqian-Zhang:
I add the example of using zbl in training process.


examples/water/zbl/input.json
njzjz, line 1: This file should be added to `source/tests/test_examples.py` to ensure the parameters are always correct.

njzjz (changes requested):

examples/water/zbl/input.json
Chengqian-Zhang: I have done it. Please check.

wanghan-iapcm (approved):

njzjz (approved):

PR #2625, opened 2023-06-20
Fix DPLR: Support time-dependent efield 

Yi-FanLi:
Supports the uses to define time-dependent efield through LAMMPS's variable feature.


source/lmp/fix_dplr.h
```diff
@@ -54,6 +54,8 @@ class FixDPLR : public Fix {
   double compute_scalar(void) override;
   double compute_vector(int) override;
 
+  enum { NONE, CONSTANT, EQUAL };
```
njzjz, line 57: Is there any reason to make it public?

source/lmp/fix_dplr.cpp
```diff
@@ -157,6 +192,48 @@ void FixDPLR::init() {
...
+    zvar = input->variable->find(zstr);
+    if (zvar < 0)
+      error->all(FLERR, "Variable {} for z-field in fix {} does not exist",
+                 zstr, style);
+    if (input->variable->equalstyle(zvar))
+      zstyle = EQUAL;
+    else
+      error->all(FLERR, "Variable {} for z-field in fix {} is invalid style",
+                 zstr, style);
+  }
```
njzjz, line 231: Is it possible to write a loop? Here x, y, z are exactly the same.

source/lmp/fix_dplr.h
Yi-FanLi: I made it public by doing something similar to https://github.com/lammps/lammps/blob/c5d9f901d9e8336602dee9558ffd6a5aab30e4d1/src/fix_efield.h#LL44C20-L44C20. LAMMPS did this because its FixEfield class has a friend class FixQEqReaxFF. In our case, there is no need to make it public. I have changed it to private.

source/lmp/fix_dplr.cpp
Yi-FanLi: It is possible, but I do not recommend doing that. This part is almost the same as that in LAMMPS's fix_efield.cpp. Keeping them the same makes it easier to maintain.

njzjz: I see.

njzjz (approved):

source/lmp/fix_dplr.cpp
```diff
@@ -157,6 +192,48 @@ void FixDPLR::init() {
   // 	 << vv[ii][2] << " "
   // 	 << endl;
   // }
+  // check variables
+
+  if (xstr) {
+    xvar = input->variable->find(xstr);
+    if (xvar < 0)
+      error->all(FLERR, "Variable {} for x-field in fix {} does not exist",
+                 xstr, style);
```
wanghan-iapcm, line 201: please add `{` and `}` around the line.
The same issue for all single line `if` statements.

wanghan-iapcm (changes requested):
- Could you please add to the document how to use the TD efield?
- UT is needed.

source/lmp/fix_dplr.cpp
njzjz: @wanghan-iapcm Do we need to add this option: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces

wanghan-iapcm: > @wanghan-iapcm Do we need to add this option: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces

Yes please.

njzjz: I submit #2629.

Yi-FanLi: UT has been added now.

Yi-FanLi: > please add `{` and `}` around the line. The same issue for all single line `if` statements.

Braces added.

Yi-FanLi: Document also added.

source/lmp/fix_dplr.h
```diff
@@ -71,6 +71,12 @@ class FixDPLR : public Fix {
   std::vector<double> efield_fsum, efield_fsum_all;
   int efield_force_flag;
   void get_valid_pairs(std::vector<std::pair<int, int> > &pairs);
+  int varflag;
+  char *xstr, *ystr, *zstr, *estr, *pstr;
```
wanghan-iapcm, line 75: it seems that estr and pstr are not used.

wanghan-iapcm (approved):

source/lmp/fix_dplr.h
Yi-FanLi: Yes. I deleted them.

PR #2628, opened 2023-06-21
add compressible and non-compressible se_atten examples

nahso:


examples/water/se_atten/input.json
```diff
@@ -7,7 +7,7 @@
     ],
     "descriptor": {
       "type": "se_atten",
-      "stripped_type_embedding": true,
+      "stripped_type_embedding": false,
```
wanghan-iapcm, line 10: new behavior `true` is preferred. @njzjz plz also check

njzjz:
> new behavior `true` is preferred. @njzjz plz also check 

Agree.

examples/water/se_atten/input.json
njzjz: > new behavior `true` is preferred. @njzjz plz also check

Agree.

examples/water/se_atten_compressible/input.json
njzjz, line 1: This new file should be added to source/tests/test_examples.py to ensure the parameters are always correct.

njzjz (changes requested):

examples/water/se_atten_non_compressible/input.json
```diff
@@ -0,0 +1,87 @@
+{
+  "_comment": " model parameters",
+  "model": {
+    "type_map": [
+      "O",
+      "H"
+    ],
+    "descriptor": {
+      "type": "se_atten",
+      "stripped_type_embedding": false,
```
njzjz, line 10: I think @wanghan-iapcm means setting this option to true with 2 attention layers.

njzjz (approved):

wanghan-iapcm:
close due to #2633

PR #2634, opened 2023-06-25
cmake: migrate from `FindCUDA` to CUDA language

njzjz:
`FindCUDA` will be removed in CMake 3.27, per https://cmake.org/cmake/help/latest/release/3.27.html#deprecated-and-removed-features.


source/install/package_c.sh
```diff
@@ -20,7 +20,7 @@ cmake -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX} \
 	-DPACKAGE_C=TRUE \
 	-DUSE_TF_PYTHON_LIBS=TRUE \
 	..
-make -j${NPROC}
+make -j${NPROC} VERBOSE=1
```
wanghan-iapcm, line 23: only for debug purpose?

wanghan-iapcm (approved):

source/install/package_c.sh
njzjz: Yes, it's used to debug GitHub Actions.

PR #2659, opened 2023-07-06
add precision arguments explicitly to examples

njzjz:
Set precision=float32 for NNs in examples, since we have found that float64 and float32 have the same accuracy in almost all the situation.

### DIFF
--- a/examples/dos/train/input.json
+++ b/examples/dos/train/input.json
@@ -17,6 +17,7 @@
       ],
       "resnet_dt": false,
       "axis_neuron": 8,
+      "precision": "float32",
       "seed": 1
     },
     "fitting_net": {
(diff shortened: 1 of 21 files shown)


wanghan-iapcm (commented):
I would recommend all examples preserves the old behavior and provide separate examples for float32. 
The same training/validating accuracy can not justify the same accuracy in real applications, for example, the calculations of elastic constant requires the high accuracy of second order derivatives of the energy. We have relatively little knowledge on the numerical influence on these properties.

njzjz:
> I would recommend all examples preserves the old behavior and provide separate examples for float32.

I change `float32` to `float64`. People can easily change `float64` to `float32` after we explicitly show the location of these arguments.

wanghan-iapcm (approved):

PR #2670, opened 2023-07-11
breaking: add energy bias to tab potential

njzjz:
See #2662.

Forces and virials will be changed accordingly.
One can turn off the new behavior by setting `srtab_add_bias` to False.


njzjz:
@wanghan-iapcm any comments on this PR? Maybe we can discuss it offline.

wanghan-iapcm:
> @wanghan-iapcm any comments on this PR? Maybe we can discuss it outside the GitHub.

Sure !

wanghan-iapcm (approved):

PR #2671, opened 2023-07-12
format ipynb files

njzjz:
Format ipynb files, supported in [ruff 0.0.276](https://github.com/astral-sh/ruff/releases/tag/v0.0.276).
Remove target-version, which is not required since [ruff v0.0.255](https://github.com/astral-sh/ruff/releases/tag/v0.0.255).


njzjz:
It seems not working, or nothing needs to format?

PR #2679, opened 2023-07-14
fix se_atten tabulate when `exclude_types` is given

njzjz:
This PR is similar to #2552. Usually, the input to tabulate OP is sorted, but if `exclude_types` is given, the input will not be sorted anymore.


nahso:
Does `se_a.py` has this issue too? They share same tabulation functions.

nahso (approved):

njzjz:
> Does `se_a.py` has this issue too? They share same tabulation functions.

The environment matrix in the standard se_a is split into different neighbor elements and each of them is sorted.

wanghan-iapcm (approved):

PR #2682, opened 2023-07-17
add pairwise DPRc

njzjz:
Add an experimental model called pairwise DPRc, which is fragment-based and integrated with QM/MM. Compression inference and training are supported.
Unit tests and documentation have been added.

Some features or bugfix to implement this PR have been merged in #2549, #2600, #2601, #2604, #2631, #2635, #2665, #2666, #2667, and #2679.
This PR makes some changes to `model.build_descrpt` additionally:
- fix errors when the suffix is not empty
- fix errors when `fparam` or `aparam` are given
- support model-customized `input_map`


deepmd/model/model.py
```diff
@@ -261,14 +267,31 @@ def build_descrpt(
                 suffix=suffix,
                 reuse=reuse,
             )
-            dout = tf.identity(dout, name="o_descriptor")
+            dout = tf.identity(dout, name="o_descriptor" + suffix)
```
wanghan-iapcm, line 270: the suffix is to distinguish the different descriptors of dm dmmm models, am I correct?

wanghan-iapcm (approved):

deepmd/model/model.py
njzjz: Yes. (qm and qmmm model)

PR #2690, opened 2023-07-20
support generalized force loss

njzjz:
Support the loss for generalized forces. Tests and examples have been added.

Generalized forces are given by
```math
Q_j = \sum_{i=1}^n \mathbf F_i \cdot \frac {\partial \mathbf r_i} {\partial q_j},\quad j=1,\ldots, m.
```
The loss for generalized forces is given by
```math
L_Q = \frac{1}{m} \sum_{j=1}^m (Q_j - Q_j^*)^2.
```

In the example, the generalized coordinates $q$ are the restraint coordinates in the enhanced sampling.

This PR also improves documentation for other arguments in the loss.


wanghan-iapcm (approved):
LGTM. 

I am not sure how would it be useful in practice, because a low normal force loss would naturally lead to a low generalized force loss if dr/dq is not degenerate.

njzjz:
> I am not sure how would it be useful in practice,

This is Tim's idea. Here is my understanding.

In umbrella sampling free energy calculation, the only quantity from simulations directly used to calculate free energy is the reaction coordinates (we make it the generalized coordinates $q$ in this context). Thus, the generalized forces $Q=\frac{\partial{E}}{\partial{q}}$ will be more important than the normal force in such an application.

> because a low normal force loss would naturally lead to a low generalized force loss if dr/dq is not degenerate.

I agree it has been low (the papers I published have proved it), but what we expect is that generalized force has even better accuracy than before, without significantly decreasing the accuracy of the normal force.

wanghan-iapcm:
Clear, Thanks!

PR #2717, opened 2023-08-04
bump lammps to stable_2Aug2023

njzjz:


Yi-FanLi (approved):
Looks good. I'm surprised that I did not explicitly use atom ids to compare expected quantities and lammps's output in the past.

wanghan-iapcm (approved):

PR #2731, opened 2023-08-09
fix bug in deepmd.infer.deep_pot.DeepPot

ChiahsinChu:
- fix bug in checking t_efield op in the graph
- delete arg atomic in `_prepare_feed_dict` (otherwise, efield cannot be imported correctly when using this method)


njzjz (changes requested):
os.path.join has different behaviors on Windows. Explicit `/` is preferred.

wanghan-iapcm (approved):

PR #2732, opened 2023-08-09
support mapping to ghost type

link89:
Currently  `pair_coeff` support customizing mapping rule by the following way
```
pair_coeff * * H O H
```
But there are cases that need to mapping specific atom type to ghost type, or type that doesn't existed in deepmd. 

This PR addresses the issue by treating `NULL` as ghost type, so that users can specify atoms type to be ghost type by the following way
```
pair coeff * * H O NULL
```


njzjz (changes requested):
Other pair styles in LAMMPS use NULL as the placeholder, for example, https://docs.lammps.org/pair_eam.html and https://docs.lammps.org/pair_reaxff.html. It's preferred to use the same key.
Documentation should be added like two above pair styles.

link89:
Update the patch will new placeholder `NULL` and its usage.

njzjz (approved):

PR #2747, opened 2023-08-18
Added atomic dipole to test.py

hanao2:
Added atomic dipole to test.py

The current code did not cover this case.

### DIFF
--- a/deepmd/entrypoints/test.py
+++ b/deepmd/entrypoints/test.py
@@ -933,9 +933,8 @@ def test_dipole(
     log.info("The unit of error is the same as the unit of provided label.")
 
     if detail_file is not None:
+        detail_path = Path(detail_file)
         if not atomic:
-            detail_path = Path(detail_file)
-
             pe = np.concatenate(
                 (
                     np.reshape(test_data["dipole"][:numb_test], [-1, 3]),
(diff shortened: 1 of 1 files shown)


deepmd/entrypoints/test.py
```diff
@@ -933,20 +933,37 @@ def test_dipole(
...
+                (
+                    test_data["atomic_dipole"][:numb_test],
+                    dipole,
+                ),
+                axis=1,
+            )
+
+            np.savetxt(
+                detail_path.with_suffix(".out"),
+                pe,
```
njzjz, line 965: `header` should be added here, like line 950.

deepmd/entrypoints/test.py
```diff
@@ -933,20 +933,37 @@ def test_dipole(
...
+                header="data_x data_y data_z pred_x pred_y pred_z",
+            )
+        else:
+            detail_path = Path(detail_file)
+
+            pe = np.concatenate(
+                (
+                    test_data["atomic_dipole"][:numb_test],
+                    dipole,
```
njzjz, line 958: `reshape` like line 941?

deepmd/entrypoints/test.py
njzjz: It seems that we do not need to write it twice. Just put these lines out of the `if...else`.

hanao2:
Thanks for the comments. Today I went through the code and some of my output files; I noticed that the atomic dipole array (from 'atomic_dipole.npy') undergoes a reindexing when reading the test data. Therefore, for consistency and proper comparison (with the model prediction), we need to address this reindexing in the output 'detailed_file'. I will update the code shortly.

njzjz (approved):

PR #2755, opened 2023-08-23
Change the smooth method in se_atten descriptor

iProzd:
This PR edits the smooth method of se_atten descriptor when using stripped type embedding:
1. Change the structure in stripped type embedding and compression ops.
2. Add smooth factor on the network output of type embedding.
3. Add a new descriptor se_atten_v2 to include these changes while keeping se_atten in old behaviors.


iProzd:
The default values of `stripped_type_embedding`, `smooth_type_embdding` and `set_davg_zero` should be discussed.

deepmd/descriptor/se_atten.py
```diff
@@ -133,6 +133,7 @@ def __init__(
         attn_mask: bool = False,
         multi_task: bool = False,
         stripped_type_embedding: bool = False,
+        smooth_type_embdding: bool = True,
```
njzjz, line 136: The default value here does not match the default value in `argcheck.py`.

source/tests/test_model_compression_se_atten.py
```diff
@@ -40,25 +40,27 @@ def _subprocess_run(command):
...
+# 4 tests:
+# - type embedding FP64, se_atten FP64
+# - type embedding FP64, se_atten FP32
+# - type embedding FP32, se_atten FP64
+# - type embedding FP32, se_atten FP32
+tests = [
+    {"se_atten precision": "float64", "type embedding precision": "float64"},
+    {"se_atten precision": "float64", "type embedding precision": "float32"},
+    # {"se_atten precision": "float32", "type embedding precision": "float64"},
+    # {"se_atten precision": "float32", "type embedding precision": "float32"},
```
wanghan-iapcm, line 52: Why these tests are disabled?

nahso: Because the `self.recovered_switch` in the expression `two_embd = two_embd * tf.reshape(self.recovered_switch, [-1, 1])
` in se_atten.py:1163 is always a float64, if the data type of `two_embd` set to float32, there will raise an TypeError: `TypeError: Input 'y' of 'Mul' Op has type float64 that does not match type float32 of argument 'x'.`

wanghan-iapcm: In this case `self.recovered_switch` should be cast to float32

wanghan-iapcm (approved):

nahso (approved):

deepmd/descriptor/se_atten_v2.py
```diff
@@ -0,0 +1,126 @@
...
+        activation_function: str = "tanh",
+        precision: str = "default",
+        uniform_seed: bool = False,
+        attn: int = 128,
+        attn_layer: int = 2,
+        attn_dotr: bool = True,
+        attn_mask: bool = False,
+        multi_task: bool = False,
+        stripped_type_embedding: bool = True,
+        smooth_type_embdding: bool = True,
```
wanghan-iapcm, line 96: we should not allow the user to change these two variables in v2

njzjz (approved):

PR #2756, opened 2023-08-23
breaking: dp test all data

njzjz:
Previously, we need to set `numb_test` to a quite large number to test all data. In this PR, passing 0 will lead to infinity.

Breaking change: the default value of `numb_test` is set to 0.

### DIFF
--- a/deepmd_cli/main.py
+++ b/deepmd_cli/main.py
@@ -276,7 +276,7 @@ def main_parser() -> argparse.ArgumentParser:
     parser_tst.add_argument(
         "-n",
         "--numb-test",
-        default=100,
+        default=0,
         type=int,
         help="The number of data for test. 0 means all data.",
     )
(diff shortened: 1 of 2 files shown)


deepmd_cli/main.py
```diff
@@ -274,7 +274,11 @@ def main_parser() -> argparse.ArgumentParser:
         "-S", "--set-prefix", default="set", type=str, help="The set prefix"
     )
     parser_tst.add_argument(
-        "-n", "--numb-test", default=100, type=int, help="The number of data for test"
+        "-n",
+        "--numb-test",
+        default=100,
```
wanghan-iapcm, line 279: is setting default to 0 a better option?

njzjz: I agree, although it might be a breaking change.

wanghan-iapcm (approved):

PR #2775, opened 2023-09-01
lmp: support `unit real`

njzjz:
Fix #1262.


source/lmp/pair_deepmd.cpp
```diff
@@ -774,7 +782,7 @@ void PairDeepMD::compute(int eflag, int vflag) {
   if (!atom->sp_flag) {
     for (int ii = 0; ii < nall; ++ii) {
       for (int dd = 0; dd < 3; ++dd) {
-        f[ii][dd] += scale[1][1] * dforce[3 * ii + dd];
+        f[ii][dd] += conversion_factor * dforce[3 * ii + dd];
```
wanghan-iapcm, line 785: I recommend to preserve scale, in case we need to revise it in the future.

source/lmp/pair_deepmd.cpp
```diff
@@ -344,11 +344,17 @@ PairDeepMD::PairDeepMD(LAMMPS *lmp)
...
+  if (strcmp(update->unit_style, "metal") == 0) {
+    unit_convert = utils::NOCONVERT;
+  } else if (strcmp(update->unit_style, "real") == 0) {
+    unit_convert = utils::METAL2REAL;
+  } else {
+    error->all(FLERR,
+               "Pair deepmd requires metal or real unit, please set it by "
+               "\"units metal\" or \"units real\"");
   }
+  conversion_factor = utils::get_conversion_factor(utils::ENERGY, unit_convert);
```
wanghan-iapcm, line 357: It looks OK. 

We need only one `conversion_factor`, because the length unit are the same and we only need to convert energy. It may be a better idea to express the fact in the name of the factor, e.g. change it to `ener_unit_cvt_factor`

wanghan-iapcm (changes requested):
The output of model deviations are also needed to be converted.

source/lmp/pair_deepmd.cpp
wanghan-iapcm: I see at the  virial we did not multiply the scale...

njzjz: I don't get what the scale is used for. But if the scale should be kept, a better way is something like `scale *= factor`.

wanghan-iapcm (approved):

PR #2776, opened 2023-09-01
lmp: `fix_dplr` use the same `type_map` from `pair_deepmd`

njzjz:
When `pair_coeff` has set LAMMPS `type_map` (e.g. `pair_coeff * * O H`), `fix_dplr` uses the same LAMMPS `type_map` (i.e. `O H` in this case) to map from LAMMPS types to the DP model types, considering LAMMPS types should always be the same among different commands.

### DIFF
--- a/source/lmp/fix_dplr.cpp
+++ b/source/lmp/fix_dplr.cpp
@@ -190,13 +190,12 @@ FixDPLR::FixDPLR(LAMMPS *lmp, int narg, char **arg)
         error->all(FLERR, "pair_coeff: element " + type_name +
                           " not found in the DPLR model");
       }
-      iarg += 1;
     }
     int numb_types = type_idx_map.size();
     if (numb_types < n) {
       type_idx_map.resize(n);
       for (int ii = numb_types; ii < n; ++ii) {
-        type_idx_map[ii] = -1;
+        type_idx_map[ii] = ii;
       }
     }
   }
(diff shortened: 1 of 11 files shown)


njzjz:
@Yi-FanLi, I am unsure whether `test_pair_deepmd_lr_type_map` has covered all the lines I modified.

source/lmp/fix_dplr.cpp
```diff
@@ -142,18 +138,70 @@ FixDPLR::FixDPLR(LAMMPS *lmp, int narg, char **arg)
...
+  std::vector<std::string> type_map;
+  std::string type_map_str;
+  dpt.get_type_map(type_map_str);
+  // convert the string to a vector of strings
+  std::istringstream iss(type_map_str);
+  std::string type_name;
+  while (iss >> type_name) {
+    type_map.push_back(type_name);
+  }
+  if (type_names.size() == 0 || type_map.size() == 0) {
```
Yi-FanLi, line 157: As I understand, type_map comes from the type map stored in the DeepTensor model. Is it possible that the DeepTensor model stores no information about type map so that type_map.size()==0?

Yi-FanLi (commented):
I reviewed this pull request. Most of them are fine. The test works as how it is designed.

source/lmp/fix_dplr.cpp
njzjz: If no type map is stored, it's expected to fall back to the old behavior, i.e. line 158-161.

Yi-FanLi: I see.

Yi-FanLi (approved):

PR #2800, opened 2023-09-09
lmp: support other units for both pair deepmd and fix dplr

Yi-FanLi:
This PR is intended to do 2 things:
1. support other unit styles like si and nano (except lj);
2. support the non-metal units for fix dplr and compute deeptensor/atom.

Unittests for both features mentioned above have also been added in this pull request.


source/lmp/pair_deepmd.cpp
```diff
@@ -354,8 +354,10 @@ PairDeepMD::PairDeepMD(LAMMPS *lmp)
                "Pair deepmd requires metal or real unit, please set it by "
                "\"units metal\" or \"units real\"");
   }
-  ener_unit_cvt_factor =
-      utils::get_conversion_factor(utils::ENERGY, unit_convert);
+  ener_unit_cvt_factor = force->boltz / 8.617343e-5;
```
njzjz, line 357:
```suggestion
  ener_unit_cvt_factor = force->qe2f;
```

njzjz:
Interesting！

source/lmp/pair_deepmd.cpp
Yi-FanLi: I understand why you are suggesting changing to force->qe2f: for "real", it is equal to 23.060549, which is force->boltz / 8.617343e-5. However, there are cases in which force->qe2f does not represent the correct energy conversion factor. For example, for units "si",1 eV = 1.60218e-19 Joule. It can be obtained with 
(force->boltz = 1.3806504e-23) / 8.617343e-5. However, force->qe2f for "si" is 1.0. Thus, I would keep this unchanged.

source/lmp/fix_dplr.cpp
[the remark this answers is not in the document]
njzjz: I can approve the change for `pair deepmd` ONLY if you remove the changes for `fix dplr` and `compute deeptensor/atom` and submit them in another PR.

source/lmp/compute_deeptensor_atom.cpp
```diff
@@ -155,7 +158,7 @@ void ComputeDeeptensorAtom::compute_peratom() {
     // record when selected and in group
     if (selected && ingroup) {
       for (int jj = 0; jj < size_peratom_cols; ++jj) {
-        tensor[ii][jj] = atensor[iter_tensor + jj];
+        tensor[ii][jj] = atensor[iter_tensor + jj] * dist_unit_cvt_factor;
```
njzjz, line 161: @Yi-FanLi and @wanghan-iapcm, I have a question here. Do we expect the output all tensorial models to have the unit of distance?

Yi-FanLi: Generally no. But at least in my knowledge and as introduced in the documentation, the usage of the compute deeptensor/atom feature is restricted to predicting the displacements of Wannier centroids, which do have the unit of distance. I suggest keeping the distance unit here at this moment, and see if there will be other ways of using compute deeptensor/atom. If there are, we should also add them in the document later on.

source/lmp/tests/test_dplr.py
```diff
@@ -313,6 +335,13 @@ def lammps_type_map():
     lmp.close()
 
 
+@pytest.fixture
+def lammps_si():
+    lmp = _lammps(data_file=data_file_si)
```
njzjz, line 340:
```suggestion
    lmp = _lammps(data_file=data_file_si, units="si")
```

njzjz (changes requested):

source/lmp/compute_deeptensor_atom.cpp
njzjz: In the current documentation, I don't see it restricted to predicting the displacements of Wannier centroids.

image

Yi-FanLi: In principle, the tensor can be any property. However, in practice, dipole (as shown in the example of the screenshot) is the only case I have ever seen. The practical use of this feature does not exceed the case of the dipole example. This, assuming that the tensor has the unit of displacement should be a reasonable assumption. I realize that the tensor can also be polarizability, as shown in the original paper of Deep Wannier. However, I did not see users actually using it in practice. In the future, if some users need to use non-metal units for different quantities, we can add an interface to type the name of the quantity so that we can determine the conversion factors accordingly. Currently, only supporting displacement should be enough.

njzjz: This behavior should be documented, since in the documentation, we say that the unit of the tensors can be any.

Yi-FanLi:
OK

source/lmp/compute_deeptensor_atom.cpp
Yi-FanLi: The features added in this PR have been discussed in detail in the document.

njzjz (approved):
LGTM

PR #2801, opened 2023-09-09
support atomic/relative model deviation in CLI

### LINKED ISSUE #2017: [Feature Request] support relative and atomic model deviation for CLI
### Summary

Currently, C++ API and the LAMMPS module have supported relative and atomic model deviation. The Python API and CLI can support this feature as well.

### Detailed Description

Something like `dp model-devi --relative 1.0 --atomic`.

- `dp model-devi`: https://docs.deepmodeling.com/projects/deepmd/en/latest/cli.html#model-devi

njzjz:
Fix #2017.

### DIFF
--- a/deepmd/infer/model_devi.py
+++ b/deepmd/infer/model_devi.py
@@ -38,7 +38,8 @@ def calc_model_devi_f(
     fs: np.ndarray,
     real_f: Optional[np.ndarray] = None,
     relative: Optional[float] = None,
-    *atomic: Literal[True],
+    *,
+    atomic: Literal[True],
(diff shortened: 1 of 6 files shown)


deepmd/infer/model_devi.py
```diff
@@ -42,17 +48,29 @@ def calc_model_devi_f(
         minimum deviation of force in all atoms
     avg_devi_f : numpy.ndarray
         average deviation of force in all atoms
+    fs_devi : numpy.ndarray
+        deviation of force in all atoms
```
wanghan-iapcm, line 52: This is a breaking change, we may want to add an option `atomic: bool=False`  to control the output of atomic force deviation.

wanghan-iapcm (approved):

PR #2827, opened 2023-09-15
support type embedding changing for frozen .pb

iProzd:


njzjz (changes requested):
To my understanding of this PR, the embedding network has the following requirements:

- ResNet does not apply;
- activation function should be an identity function;
- the number of layers should be 1.

These requirements should be documented.

A better way might be to directly update o_typeebd to a constant instead of updating the NN.

PR #2841, opened 2023-09-19
apply compression for se_e2_a_tebd

nahso:


deepmd/descriptor/se_a_v2.py
```diff
@@ -0,0 +1,66 @@
...
+)
+from .se_a import (
+    DescrptSeA,
+)
+
+log = logging.getLogger(__name__)
+
+
+@Descriptor.register("se_e2_a_v2")
+class DescrptSeAV2(DescrptSeA):
```
njzjz, line 23: Does it force one to enable type embedding? If so, I think a better name might be `se_e2_a_tebd_v2` or `se_a_tpe_v2` (there is `se_a_tpe` in `se_a_ebd.py`) since `se_e2_a` without type embedding should not use this type.

The example and documentation should be updated to the new type.

njzjz (changes requested):

examples/water/se_e2_a_tebd_compressible/input.json
njzjz, line 1: We don't need to keep the old input file, as there is no advantage compared to the new one. Just move this file to `examples/water/se_e2_a_tebd/input.json`, so it is always recommended.

njzjz (approved):
LGTM

wanghan-iapcm (approved):

PR #2844, opened 2023-09-19
merge cuda and rocm files

njzjz:
Merge `source/lib/src/cuda` and `source/lib/src/rocm` into `source/lib/src/gpu`.

- Define macros `gpuGetLastError`, `gpuDeviceSynchronize`, `gpuMemcpy`, `gpuMemcpyDeviceToHost`, `gpuMemcpyHostToDevice`, and `gpuMemset` to make them available for both CUDA and ROCm.
- Use `<<< >>> syntax` for both CUDA and ROCm. Per https://github.com/ROCm-Developer-Tools/HIP/commit/cf78d85638b22a586317dfee4514d4590e7b2eec, it has been supported in HIP since 2018.
- Fix several int const numbers that should be double or float.
- For tabulate:
  - Fix `WARP_SIZE` for ROCm. Per https://github.com/pytorch/pytorch/pull/64302, WARP_SIZE can be 32 or 64, so it should not be hardcoded to 64.
  - Add `GpuShuffleSync`.  Per https://github.com/ROCm-Developer-Tools/HIP/issues/1491, `__shfl_sync` is not supported by HIP.
  - After merging the code, #1274 should also work for ROCm.
  - Use the same `ii` for #830 and #2357. Although both of them work, `ii` has different meanings in these two PRs, but now it should be the same.
    - However, `ii` in `tabulate_fusion_se_a_fifth_order_polynomial` (rocm) added by #2532 is wrong. After merging the codes, it should be corrected.
  - Optimization in #830 was not applied to ROCm.
- There are several other differences between CUDA and ROCm.


source/lib/src/gpu/tabulate.cu
```diff
@@ -195,10 +249,18 @@ __global__ void tabulate_fusion_se_a_grad_fifth_order_polynomial(
     const int last_layer_size,
     const bool is_sorted) {
   bool enable_se_atten = two_embed != nullptr;
+#if GOOGLE_CUDA
   extern __shared__ int _data[];
+  int MTILE_OR_KTILE = MTILE;
+#elif TENSORFLOW_USE_ROCM
+  HIP_DYNAMIC_SHARED(int, _data)
+  int MTILE_OR_KTILE = KTILE;
```
njzjz, line 257: I must confirm if `KTILE` -> `MTILE` by #830 should also apply to ROCm.

source/lib/src/gpu/tabulate.cu
```diff
@@ -4,9 +4,30 @@
...
+#endif
 #define FULL_MASK 0xffffffff
 
+// Copyright 2017 The TensorFlow Authors.
+// Licensed under the Apache License, Version 2.0
+template <typename T>
+__device__ T
+GpuShuffleSync(unsigned mask, T value, int src_lane, int width = warpSize) {
+#if GOOGLE_CUDA
+  return __shfl_sync(mask, value, src_lane, width);
```
njzjz, line 23: It seems `__shfl_sync` and `__syncwarp` are new in CUDA 9. Should we update the documentation here from `7` to `9`? https://github.com/deepmodeling/deepmd-kit/blob/339ce4743194632ad6c4a906fb7c9fa4fe41dc32/doc/install/install-from-source.md?plain=1#L77

source/lib/src/gpu/tabulate.cu
```diff
@@ -110,8 +139,15 @@ __forceinline__ __device__ FPTYPE dot(FPTYPE ll[4], FPTYPE rr[4]) {
 
 template <typename FPTYPE>
 __forceinline__ __device__ void warp_reduce(FPTYPE& val) {
+#if GOOGLE_CUDA
   for (int offset = 16; offset > 0; offset >>= 1) {
     val += __shfl_down_sync(FULL_MASK, val, offset);
+#elif TENSORFLOW_USE_ROCM
+  for (int offset = 32; offset > 0; offset >>= 1) {
+    val += __shfl_down(val, offset);  // ########????
+#else
```
njzjz, line 148: Should the initial `offset` be `WARP_SIZE / 2`? (since it's not fixed in rocm)

denghuilu: Yes, It should be WARP_SIZE / 2 according to the metioned Torch PR

source/lib/src/gpu/tabulate.cu
denghuilu: Yes, they both have the same value 4, defined in the header of the file.
```
#define MM 4
#define KK 4
```

source/lib/src/gpu/tabulate.cu
denghuilu: Yes, I think we should update the doc

wanghan-iapcm (approved):

denghuilu (approved):

PR #2847, opened 2023-09-20
merge CUDA and ROCm codes in op

njzjz:


njzjz:
The OP codes contain `cudaMemcpy`, so we must wait for #2844.

wanghan-iapcm (approved):

PR #2848, opened 2023-09-20
add test cuda workflow

njzjz:


pyproject.toml
njzjz, line 1: #2850 should be merged before this PR.

wanghan-iapcm (changes requested):
I recommend writting in the doc how to activate the cuda UTs.

wanghan-iapcm (approved):

PR #2850, opened 2023-09-20
fix missing version file with setuptools-scm v8

njzjz:
See https://github.com/scikit-build/scikit-build-core/issues/507.

Now it's written to `{site_packages}/deepmd_cli/_version/__init__.py`. The previous implementation is too complex.

### DIFF
--- a/deepmd/__init__.py
+++ b/deepmd/__init__.py
@@ -32,7 +32,7 @@ from .infer.data_modifier import (
 set_mkl()
 
 try:
-    from ._version import version as __version__
+    from deepmd_cli._version import version as __version__
 except ImportError:
     from .__about__ import (
         __version__,
--- a/deepmd_cli/main.py
+++ b/deepmd_cli/main.py
@@ -1,31 +1,16 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 import argparse
-import importlib.util
 import logging
-import os
-import sys
 import textwrap
 from typing import (
     List,
     Optional,
 )
 
-
-def load_child_module(name):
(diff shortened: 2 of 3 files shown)


henryiii:
There shouldn't be a difference between `a/b/__init__.py` and `a/b.py`. `load_child_module` shouldn't be required in either case, lots of libraries do it this way. What's the issue when it's just `deepmd_cli/_version.py`?

njzjz:
> There shouldn't be a difference between `a/b/__init__.py` and `a/b.py` wrt `a`. `load_child_module` shouldn't be required in either case, lots of libraries do it this way. What's the issue when it's just `deepmd_cli/_version.py`?

@henryiii Thanks. I find that you are right. I thought it might not work for the editable mode, but it's working.

I have this idea because when using [`tf.load_op_library`](https://www.tensorflow.org/api_docs/python/tf/load_op_library) in the editable mode, it's not easy to find the libraries in `site-packages`. (these TensorFlow OP libraries cannot be directory imported) A `__init__.py` file is helpful to find them in this case.

`load_child_module` is for a historical reason. Previously, everything was in `deepmd` directory and needed to import TensorFlow. Someone complains the command line is too slow, so the command line module is split into a seperated namespace `deepmd_cli`, but `_version.py` is still in `deepmd`. I just realized the easiest way is to directly put it into `deepmd_cli` directory.

henryiii:
Ahh, I missed it was moving from `deepmd` to `deepmd_cli`! That's much nicer.

wanghan-iapcm (approved):

PR #2851, opened 2023-09-21
build linux-aarch64 wheel on self-hosted runner

njzjz:
..., which should be much faster than using QEMU on x86. The runner is a free Oracle ARM Ampere A1 Compute instance.


pyproject.toml
njzjz, line 1: Requires #2850 to be merged first.

wanghan-iapcm (approved):

PR #2857, opened 2023-09-22
cmake: use modern `HIP` language

njzjz:
Use modern `HIP` language instead of `hip_add_library`.
Use modern `find_package` to find `hip` and `hipcub` instead of `FindROCM.cmake` (which is removed in this PR).


wanghan-iapcm (approved):

backend/read_env.py
```diff
@@ -57,8 +57,8 @@ def get_argument_from_env() -> Tuple[str, list, list, dict, str]:
         if rocm_root:
             cmake_args.append(f"-DCMAKE_HIP_COMPILER_ROCM_ROOT:STRING={rocm_root}")
         hipcc_flags = os.environ.get("HIP_HIPCC_FLAGS")
```
wanghan-iapcm, line 59: I find this flag is not documented, could you please add it to the doc? Thanks!

njzjz: I don't prefer using this environment variable, though. The CMake natively supported [`HIPFLAGS`](https://cmake.org/cmake/help/latest/envvar/HIPFLAGS.html) should be used instead of our customized flag. The old flag is kept for compatibility. I'll add `HIPFLAGS` instead.

wanghan-iapcm (approved):

PR #2878, opened 2023-09-28
forward GPU error message

njzjz:
Forward detailed GPU error messages to DeePMD-kit (and then TensorFlow) instead of directly printing to the standard error.


njzjz:
CI needs to be fixed in #2879.

wanghan-iapcm (approved):

PR #2883, opened 2023-09-30
fix source distribution version in build-wheel.yml

njzjz:
#2324 sets `fetch-depth: 0` for wheels but not source distribution so that the wheel versions are correct, but the source distribution version is wrong.

image


njzjz:
Now it's good.

image

wanghan-iapcm (approved):

PR #2887, opened 2023-10-01
fix LAMMPS wheel with CUDA wheels

njzjz:
It's annoying that in the CUDA Python wheels, the library filename is `libcudart.so.11.0`, instead of `libcudart.so` or `libcudart.so.11` or `libcudart.11.0.xxx`. There are no symlinks in these CUDA Python wheels on PyPI. Thus, in this case, when DeePMD-kit tries to load `libcudart.so`, no file can be found.
Python interface doesn't have this issue, as it tries to use `glob` to find detailed library names. For the LAMMPS executable called from Python, `LD_LIBRARY_PATH` doesn't work with CUDA Python libraries (but works with an installation that contains symlinks). Thus, in this PR, `LD_PRELOAD` is set instead to preload CUDA symbols.
This fixes the docker images for the development branch.

### DIFF
--- a/deepmd/lmp.py
+++ b/deepmd/lmp.py
@@ -35,13 +35,15 @@ def get_env(paths: List[Optional[str]]) -> str:
     return ":".join(p for p in paths if p is not None)
 
 
-def get_library_path(module: str) -> List[str]:
+def get_library_path(module: str, filename: str) -> List[str]:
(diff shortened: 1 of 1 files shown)


njzjz:
cc @Yi-FanLi

wanghan-iapcm (approved):

PR #2890, opened 2023-10-02
fix TypeError when type_map is not given

### LINKED ISSUE #2889: [BUG]  `fparam` is invalid in DeePMD-kit 2.2.4 version.
### Bug summary

```bash
  File "/opt/deepmd-kit-2.2.4/bin/dp", line 10, in <module>
    sys.exit(main())
  File "/opt/deepmd-kit-2.2.4/lib/python3.10/site-packages/deepmd_cli/main.py", line 605, in main
    deepmd_main(args)
  File "/opt/deepmd-kit-2.2.4/lib/python3.10/site-packages/deepmd/entrypoints/main.py", line 73, in main
    train_dp(**dict_args)
  File "/opt/deepmd-kit-2.2.4/lib/python3.10/site-packages/deepmd/entrypoints/train.py", line 165, in train
    _do_work(jdata, run_opt, is_compress)
  File "/opt/deepmd-kit-2.2.4/lib/python3.10/site-packages/deepmd/entrypoints/train.py", line 189, in _do_work
    model = DPTrainer(jdata, run_opt=run_opt, is_compress=is_compress)
  File "/opt/deepmd-kit-2.2.4/lib/python3.10/site-packages/deepmd/train/trainer.py", line 86, in __init__
    self._init_param(jdata)
  File "/opt/deepmd-kit-2.2.4/lib/python3.10/site-packages/deepmd/train/trainer.py", line 104, in _init_param
    self.model = Model(**model_param)
  File "/opt/deepmd-kit-2.2.4/lib/python3.10/site-packages/deepmd/model/ener.py", line 89, in __init__
    super().__init__(
  File "/opt/deepmd-kit-2.2.4/lib/python3.10/site-packages/deepmd/model/model.py", line 524, in __init__
    **descriptor, ntypes=len(type_map), spin=self.spin
TypeError: object of type 'NoneType' has no len()
```

### DeePMD-kit Version

2.2.4

### TensorFlow Version

2.9.0

Offline packages

https://github.com/deepmodeling/deepmd-kit/tree/master/examples/fparam 

### Steps to Reproduce

```bash
cd examples/fparam/train;
dp train input.json;
```

_No response_

njzjz:
Fix #2889.

### DIFF
--- a/deepmd/descriptor/se_atten.py
+++ b/deepmd/descriptor/se_atten.py
@@ -120,6 +120,11 @@ class DescrptSeAtten(DescrptSeA):
             When using stripped type embedding, whether to dot smooth factor on the network output of type embedding
             to keep the network smooth, instead of setting `set_davg_zero` to be True.
             Default value will be True in `se_atten_v2` descriptor.
+
+    Raises
+    ------
+    ValueError
+        if ntypes is 0.
     """
 
     def __init__(
@@ -178,6 +183,8 @@ class DescrptSeAtten(DescrptSeA):
             assert Version(TF_VERSION) > Version(
                 "2"
             ), "se_atten only support tensorflow version 2.0 or higher."
+        if ntypes == 0:
+            raise ValueError("`model/type_map` is not set or empty!")
         self.stripped_type_embedding = stripped_type_embedding
         self.smooth = smooth_type_embdding
         self.ntypes = ntypes
--- a/deepmd/model/model.py
+++ b/deepmd/model/model.py
@@ -531,7 +531,7 @@ class StandardModel(Model):
             self.descrpt = descriptor
         else:
             self.descrpt = Descriptor(
-                **descriptor, ntypes=len(self.type_map), spin=self.spin
+                **descriptor, ntypes=len(self.get_type_map()), spin=self.spin
             )
 
         if isinstance(fitting_net, Fitting):
--- a/deepmd/model/model.py
+++ b/deepmd/model/model.py
@@ -531,7 +531,7 @@ class StandardModel(Model):
             self.descrpt = descriptor
         else:
             self.descrpt = Descriptor(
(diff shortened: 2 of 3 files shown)


wanghan-iapcm (commented):
shall we use self.get_type_map instead?

Yi-FanLi:
I have a question about this part: the argument to be passed is `ntypes`, so the error occurs since there is no `type_map.raw file provided` and  `type_map` is None. Even if we are using len(self.get_type_map), the `ntypes` is still set to be 0, but the actual number of types should be 1. Is it more reasonable to use something like `len(sel)` instead?

njzjz:
> I have a question about this part: the argument to be passed is `ntypes`, so the error occurs since there is no `type_map.raw file provided` and `type_map` is None. Even if we are using len(self.get_type_map), the `ntypes` is still set to be 0, but the actual number of types should be 1. Is it more reasonable to use something like `len(sel)` instead?

`sel` is not always list.

https://github.com/deepmodeling/deepmd-kit/blob/f7b87c384c1ca256973464e4549c27999c0f3b59/examples/water/se_atten_compressible/input.json#L10

Thus, `len(sel)` should be used only when sel is a list, e.g.

https://github.com/deepmodeling/deepmd-kit/blob/f7b87c384c1ca256973464e4549c27999c0f3b59/deepmd/descriptor/se_a.py#L220

wanghan-iapcm (approved):

wanghan-iapcm:
It will fail when we use `len(self.get_type_map())` to estimate the number of types if `type_map` is empty. 
It seems that when we use `se_attn`, the `type_map` must not be empty. 
It should be OK to set an empty `type_map` for `se_e2_a`, because it infers the number of types by the length of `sel`

njzjz:
> It seems that when we use `se_attn`, the `type_map` must not be empty.

I added an error message to `se_attn` when `type_map` is empty.

Issue #2618, opened 2023-06-16
[Feature Request] Compute `prod_env_mat` OP in parallel in the dimension of the frame


njzjz:
### Summary

Currently, the `prod_env_mat` OP and its kernel are only parallelized in the dimension of the atoms and are not parallelized in the dimension of the frame. This is not a problem if the training batch size is small or if running MD simulations, but it causes performance degradation when the training batch size is large or for inference (`dp test` and `dp model-devi`) on modern GPUs that have a large memory.

In #2600 and #2601, I refactored `prod_force` and `prod_force_grad`. A similar thing should be applied to `prod_env_mat`.

### Detailed Description

The current code is:

https://github.com/deepmodeling/deepmd-kit/blob/92ca097b66273c06c4f0cfe6858a742e4ccfe004/source/op/prod_env_mat_multi_device.cc#L1150-L1151

This loop should be avoided for at least GPUs.

PR #1624, opened 2022-04-08
implement parallelism for neighbor stat

njzjz:
benchmark on ANI-1x dataset: 19:25 -> 14:32
when setting `OMP_NUM_THREADS=1 TF_INTRA_OP_PARALLELISM_THREADS=1 TF_INTER_OP_PARALLELISM_THREADS=12`

aedc851c1fab2e04fb410988f952182
image

(I think `OMP_NUM_THREADS=2 TF_INTRA_OP_PARALLELISM_THREADS=2 TF_INTER_OP_PARALLELISM_THREADS=6` may have better performance, but I have no interest to test it)

PR #2897, opened 2023-10-04
support neighbor stat on GPUs

### LINKED ISSUE #2619: [Feature Request] support neighbor stat on GPUs
### Summary

Currently, the neighbor stat is conducted in CPUs. It is slower than that on GPUs, but we do not have GPU implementation.

### Detailed Description

The current neighbor stat code is

https://github.com/deepmodeling/deepmd-kit/blob/92ca097b66273c06c4f0cfe6858a742e4ccfe004/deepmd/utils/neighbor_stat.py#L75-L82

Ideally, parallelism in the dimension of the frames is also required to handle many structures. 

_No response_

njzjz:
Fix #2619.

The GPU implementation in this PR is usually faster than the CPU in one thread (i.e., not using the feature implemented in #1624). Still, it needs parallelism in the batch dimension, which is blocked by #2618, regarding building the neighbor list. The GPU utilization is less than 10% for the water system. It should be improved when #2618 makes progress.


njzjz:
Let's run Test CUDA after #2892 is merged.

wanghan-iapcm (commented):
I am not an expert on this part of the code. I will listen to Denghui's review.

denghuilu (approved):
Looks good to me! If we can implement batched parallelism for this function, we should see a significant improvement in performance.

Issue #2865, opened 2023-09-26
[BUG] Significant errors when compressed training se_atten_v2 model


njzjz:
### Bug summary

Significant errors when compressed training `se_atten_v2` model. The reason is that the `gradgrad` OP for `se_atten` is not implemented.

### DeePMD-kit Version

2.2.4

### TensorFlow Version

2.13.0

### How did you download the software?

Offline packages

### Input Files, Running Commands, Error Log, etc.

![GM7yQgVX5p](https://github.com/deepmodeling/deepmd-kit/assets/9496702/19e8899e-6565-4cfe-822c-56b0c783916d)

### Steps to Reproduce

Train a `se_atten_v2` model, and then freeze and compress it. Train a new model based on the compressed model (`--init_frz_model`).

PR #2898, opened 2023-10-04
Se atten grad grad

nahso:
fix https://github.com/deepmodeling/deepmd-kit/issues/2865

image


njzjz:
Let's rerun Test CUDA after #2892 is merged.

njzjz:
Let's rerun Test CUDA after #2892 is merged.

njzjz (approved):
It looks good to me. It's still unclear to me how the equation is derivated, though...

Issue #2811, opened 2023-09-12
Energy drift in NVE MD with potential generated with "se_atten" descriptor


tella94:
### Summary

Good Afternoon, 

when I train my model using the hybrid descriptors "se_e2_a" and "se_e3" the training goes perfectly fine and in the end i obtain a RMSE on energy of 7 meV/atom while on forces of 0.13 eV/ang. 

However if I substitute the descriptor "se_e2_a" with "se_atten", even if the final RMSE reduces to 4 meV/atom and 0.10 meV and the structural properties of the model are better represented, when I perform NVE simulation with the Lammps code (but I also tried ASE) the total energy drifts upwards. The drift is on the order of 10 meV/atom over 100 ps. Reducing the timestep does not reduce the drift. The curious thing is that the more I increase the number of attention layers, the more this drift becomes pronounced. 

Using "se_atten_v2" does not change anything.  Switching the keyword             "stripped_type_embedding" to true seems to reduce a little the drift but I have to test more. The database of the two different training are identical. Below I attach the two different part of input.json

### DeePMD-kit Version

2.2.3

### TensorFlow Version

2.13

### Python Version, CUDA Version, GCC Version, LAMMPS Version, etc

Python 3.10.8, Cuda 11.8, GCC 11.3.0 , Lammps 22Jun2022

### Details

Train 1:

        "descriptor" :{
            "type":             "hybrid",
            "list":  [
            {
                "type":             "se_e2_a",
                "sel":              [80,40],
                "rcut_smth":        2.00,
                "rcut":             7.00,
                "type_one_side":    false,
                "neuron":           [25,50,100],
                "resnet_dt":        true,
                "axis_neuron":      35,
                "seed":             1
            },
            {
                "type":             "se_e3",
                "sel":              [20,15],
                "rcut_smth":        2.00,
                "rcut":             4.00,
                "neuron":           [5,10,20],
                "resnet_dt":        true,
                "seed":             1
            }

            ]
        },

        "fitting_net" : {
            "neuron":           [240,240,240],
            "resnet_dt":        true,
            "atom_ener": [-102.29717333422299,-218.98450580235118],
            "seed":             1
        }

Train 2:

           "type":             "hybrid",
            "list":  [
            {
                "type":             "se_atten",
                "sel":              90,
                "rcut_smth":        2.00,
                "rcut":             7.00,
                "neuron":           [25,50,100],
                "resnet_dt":        true,
                "axis_neuron":      40,
[...]

Issue #2866, opened 2023-09-26
Critical bugs with no error messages


njzjz:
Incorrect results will be given with no error messages:

* incorrect results when compressed training `se_atten_v2` model; reported by #2865
* v2.2.0~v2.2.4: incorrect results for DPLR training; reported by #2862, fixed by #2864
* ~v2.2.3 with TF >=2.13: incorrect results on GPUs; reported by #2660, fixed by #2661
* ~v2.2.1: incorrect results from `se_e3` compressed model; reported by #2250, fixed by #2552
* v2.2.0: frozen models can not be performed on LAMMPS; reported by #2398, fixed by #2390
* v2.1.0~v2.1.4: incorrect results on CPUs when rcut * 2 > box size; fixed by #1862
* ~v2.1.0 with LAMMPS >=stable_30Jul2021: incorrect `pppm_dplr ` results; fixed by #1637
* ~v2.0.3 with LAMMPS >=stable_30Jul2021: incorrect `compute deeptensor/atom` results; reported by #1381, fixed by #1382
* ~v2.0.1 with LAMMPS >=stable_30Jul2021: incorrect `pair deepmd` results; reported by #1109, fixed by #1128

PR #2905, opened 2023-10-09
fix: make the se attn v2 descriptor energy conservative. 

wanghan-iapcm:
This PR fixes issue #2811
1. fix the auto-diff issue: input should be `descriptor_reshape`. 
2. fix the discontinuity introduced in the attention map.


deepmd/descriptor/se_atten.py
njzjz, line 1: To my understanding, the current compression kernel assumes the gradient of two_embed (`dy_dtwo`) is zero, so the forces will be incorrect when `dy_dtwo` is not zero.
Previously, due to this bug, `dy_dtwo` of the uncompressed model is also zero.
Luckily, it's caught by an indirect unit test...

njzjz (changes requested):
It seems to be a critical bug that the energy does not match the force. Do we need to add it to #2866?

deepmd/descriptor/se_atten.py
njzjz: I've done the `grad`. Need to check `gradgrad`.

source/lib/tests/test_tabulate_se_a.cc
```diff
@@ -802,18 +803,20 @@ TEST_F(TestTabulateSeA, tabulate_fusion_se_a_grad_gpu) {
...
   deepmd::malloc_device_memory_sync(dy_dem_x_dev, dy_dem_x);
   deepmd::malloc_device_memory_sync(dy_dem_dev, dy_dem);
   deepmd::malloc_device_memory_sync(table_dev, table);
   deepmd::malloc_device_memory_sync(em_x_dev, em_x);
   deepmd::malloc_device_memory_sync(em_dev, em);
   deepmd::malloc_device_memory_sync(dy_dev, dy);
   deepmd::tabulate_fusion_se_a_grad_gpu<double>(
-      dy_dem_x_dev, dy_dem_dev, table_dev, &info[0], em_x_dev, em_dev, nullptr,
-      dy_dev, nloc, nnei, last_layer_size);
+      dy_dem_x_dev, dy_dem_dev, dy_dtwo, table_dev, &info[0], em_x_dev, em_dev,
```
njzjz, line 818:
```suggestion
      dy_dem_x_dev, dy_dem_dev, dy_dtwo_dev, table_dev, &info[0], em_x_dev, em_dev,
```

deepmd/descriptor/se_atten.py
njzjz: Need help for gradgrad

njzjz: `gradgrad` done

iProzd (approved):

njzjz (approved):

nahso (approved):

source/lib/src/gpu/tabulate.cu
```diff
@@ -781,12 +798,15 @@ void tabulate_fusion_se_a_grad_gpu(FPTYPE* dy_dem_x,
   DPErrcheck(gpuDeviceSynchronize());
   DPErrcheck(gpuMemset(dy_dem_x, 0, sizeof(FPTYPE) * nloc * nnei));
   DPErrcheck(gpuMemset(dy_dem, 0, sizeof(FPTYPE) * nloc * nnei * 4));
+  if (two_embed != nullptr) {
+    DPErrcheck(gpuMemset(dy_dtwo, 0, sizeof(FPTYPE) * nloc * nnei));
```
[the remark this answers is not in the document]
nahso: This `memset` can be removed since all elements in `dy_dtwo` will be modified later, so the parameter of memory size is unimportant.

PR #2929, opened 2023-10-17
add `aparam_from_compute` to `pair deepmd`

ChiahsinChu:


source/lmp/pair_deepmd.h
```diff
@@ -121,8 +121,11 @@ class PairDeepMD : public Pair {
   double eps_v;
 
   void make_fparam_from_compute(std::vector<double> &fparam);
-  bool do_compute;
-  std::string compute_id;
+  bool do_compute_fparam;
+  std::string compute_fparam_id;
+  void make_aparam_from_compute(std::vector<double> &fparam);
```
njzjz, line 126:
```suggestion
  void make_aparam_from_compute(std::vector<double> &aparam);
```

source/lmp/pair_deepmd.cpp
```diff
@@ -1059,7 +1105,7 @@ void PairDeepMD::settings(int narg, char **arg) {
   if (do_ttm && aparam.size() > 0) {
     error->all(FLERR, "aparam and ttm should NOT be set simultaneously");
```
njzjz, line 1106: `do_compute_aparam` should be added
```suggestion
  if ((int)do_ttm + (int)do_compute_aparam + (int)(aparam.size() > 0) > 1) {
    error->all(FLERR, "aparam, aparam_from_compute, and ttm should NOT be set simultaneously");
```

njzjz (changes requested):

njzjz (approved):

PR #2951, opened 2023-10-26
move to ruff formatter

njzjz:
See https://astral.sh/blog/the-ruff-formatter


njzjz:
It looks like the current ruff version still has some issues...

njzjz:
> It looks like the current ruff version still has some issues...

I am going to mark this PR ready for review. Although there is still some deviations from black, it should be harmless.

wanghan-iapcm (approved):

PR #2953, opened 2023-10-27
add cosine restart learning rate

hellozhaoming:


deepmd/utils/learning_rate.py
```diff
@@ -105,3 +105,174 @@ def start_lr(self) -> float:
...
+            first_decay_steps=self.decay_steps_,
+            alpha=self.alpha_,
+            name="cosinerestart",
+        )
+
+    def start_lr(self) -> float:
+        """Get the start lr."""
+        return self.start_lr_
+
+    def value(self, step: int) -> float:
```
wanghan-iapcm, line 272: you may not need to implement the value method if you do not print the information regarding the learning rate at the beginning of the training:
https://github.com/hellozhaoming/deepmd-kit/blob/05052c195308f61b63ce2bab130ce0e8cba60604/deepmd/train/trainer.py#L566

deepmd/common.py
```diff
@@ -125,6 +125,7 @@ def gelu_wrapper(x):
     "softplus": tf.nn.softplus,
     "sigmoid": tf.sigmoid,
     "tanh": tf.nn.tanh,
+    "swish": tf.nn.swish,
```
njzjz, line 128: It seems that it has been renamed to `silu`: https://github.com/tensorflow/tensorflow/issues/41066

deepmd/train/trainer.py
```diff
@@ -553,29 +563,31 @@ def train(self, train_data=None, valid_data=None):
...
-                    run_sess(self.sess, self.learning_rate),
-                    self.lr.value(cur_batch),
-                    self.lr.decay_steps_,
-                    self.lr.decay_rate_,
-                    self.lr.value(stop_batch),
-                )
-            )
-        else:
-            for fitting_key in self.fitting:
+            if self.lr_type == "exp":
```
njzjz, line 566: It's not a good behavior to switch the learning rate in the `Trainer`. Instead, implement the method `LearningRate.log_start` (`LearningRate` should be an abstract base class and inherited by all learning rate classes) and call `self.lr.log_start(self.sess)` here.

deepmd/utils/argcheck.py
```diff
@@ -1010,13 +1010,42 @@ def learning_rate_exp():
...
 
 def learning_rate_variant_type_args():
     doc_lr = "The type of the learning rate."
 
     return Variant(
         "type",
-        [Argument("exp", dict, learning_rate_exp())],
+        [Argument("exp", dict, learning_rate_exp()),
+         Argument("cos", dict, learning_rate_cos()),
+         Argument("cosrestart", dict, learning_rate_cosrestarts())],
```
njzjz, line 1048: You may need to add some documentation to variants (`doc="xxx"`). Otherwise, no one knows what they are.

deepmd/utils/learning_rate.py
```diff
@@ -105,3 +105,174 @@ def start_lr(self) -> float:
...
+class LearningRateCos:
+    r"""The cosine decaying learning rate.
+
+  The function returns the decayed learning rate.  It is computed as:
+  ```python
+  global_step = min(global_step, decay_steps)
+  cosine_decay = 0.5 * (1 + cos(pi * global_step / decay_steps))
+  decayed = (1 - alpha) * cosine_decay + alpha
+  decayed_learning_rate = learning_rate * decayed
+  ```
```
njzjz, line 118: Please use this style: https://numpydoc.readthedocs.io/en/latest/format.html#other-points-to-keep-in-mind

deepmd/utils/learning_rate.py
```diff
@@ -105,3 +105,174 @@ def start_lr(self) -> float:
...
+        decayed_learning_rate = self.start_lr_ * decayed
+        return decayed_learning_rate
+
+
+class LearningRateCosRestarts:
+    r"""The cosine decaying restart learning rate.
+
+  The function returns the cosine decayed learning rate while taking into account
+  possible warm restarts.
+  ```
```
njzjz, line 201: This line should be removed.

njzjz (changes requested):
Please run `pre-commit` to format and lint the code: https://docs.deepmodeling.com/projects/deepmd/en/master/development/coding-conventions.html#run-scripts-to-check-the-code. Or you can submit from a non-protect branch and pre-commit.ci can do it for you.

Unit tests should be added for two new learning rate classes.

njzjz:
The feature has been implemented by #5142 and #5154.

PR #2971, opened 2023-11-01
lmp: refactor ixnode

njzjz:
The current implementation used a complex method to ensure `ixnode` (int) within `[0, nxnodes-1]`, but why not use mod ("%") to get the same behavior?


wanghan-iapcm (commented):
different from python, in c++, -1%10 == -1

njzjz:
> different from python, in c++, -1%10 == -1

Got it...

njzjz:
> different from python, in c++, -1%10 == -1

I follow the resolution here: https://stackoverflow.com/a/1907585/9567349

wanghan-iapcm (approved):

PR #3002, opened 2023-11-21
add cu11 prebuilt packages

njzjz:


wanghan-iapcm (commented):
we may need to write in the doc one needs to execute `pip install  deepmd-kit-cu11` to install deepmd-kit in an cuda11 environment.

njzjz:
> we may need to write in the doc one needs to execute `pip install deepmd-kit-cu11` to install deepmd-kit in an cuda11 environment.

I think I have written here:

image

CUDA 11 is unnecessary, as `[cu11]` can install CUDA 11 packages. The only required component is the driver.

wanghan-iapcm (approved):

PR #3035, opened 2023-12-05
skip pushing when an action is triggered by dependeabot

njzjz:
See https://github.com/deepmodeling/deepmd-kit/pull/3029
See also https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#responding-to-events


njzjz:
@wanghan-iapcm let's merge this PR. Then we can comment `@dependabot rebase` in #3029 to see if it works

wanghan-iapcm (approved):

wanghan-iapcm:
> @wanghan-iapcm let's merge this PR. Then we can comment `@dependabot rebase` in #3029 to see if it works

done. sry I missed it.

PR #3036, opened 2023-12-05
breaking(lmp): do not apply scale factor to model deviation

njzjz:
Resolves #3027.


wanghan-iapcm (commented):
The scale factor in the atomic force deviation should also be removed. 
https://github.com/deepmodeling/deepmd-kit/blob/66cc41ea26aabfed24f61373d37828bd27b7d8c5/source/lmp/pair_deepmd.cpp#L801-L802

wanghan-iapcm (approved):

PR #3037, opened 2023-12-05
fix se_a_ebd_v2 when nloc != nall

njzjz:
See also #2390 and #2505...

### DIFF
--- a/deepmd/descriptor/se_a.py
+++ b/deepmd/descriptor/se_a.py
@@ -782,16 +782,16 @@ class DescrptSeA(DescrptSe):
             type_i = -1
             if nvnmd_cfg.enable and nvnmd_cfg.quantize_descriptor:
                 inputs_i = descrpt2r4(inputs_i, natoms)
+            self.atype_nloc = tf.reshape(
+                tf.slice(atype, [0, 0], [-1, natoms[0]]), [-1]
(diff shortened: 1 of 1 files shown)


deepmd/descriptor/se_a.py
```diff
@@ -956,7 +956,7 @@ def _filter_lower(
                     extra_embedding_index = self.nei_type_vec
                 else:
                     padding_ntypes = type_embedding.shape[0]
-                    atype_expand = tf.reshape(self.atype, [-1, 1])
+                    atype_expand = tf.reshape(self.atype_nloc, [-1, 1])
                     idx_i = tf.tile(atype_expand * padding_ntypes, [1, self.nnei])
                     idx_j = tf.reshape(self.nei_type_vec, [-1, self.nnei])
                     idx = idx_i + idx_j
```
njzjz, line 962: @nahso It seems this block of codes is never used?

nahso: It is used in line 1061 below, which provide the index of `net_output`.

nahso (approved):

deepmd/descriptor/se_a.py
njzjz: I see!

PR #3047, opened 2023-12-09
print NaN loss when labeled data is not found

njzjz:
Currently, when loss is defined and the labeled data is not found, `lcurve.out` shows wrong RMSE (assuming data is all zero). In this case, printing NaN is better.


wanghan-iapcm (commented):
add a comment explaining the situation at the very first line of lcurve.out?

wanghan-iapcm (approved):

PR #3059, opened 2023-12-14
fix labeler.yml with actions/labeler v5

njzjz:
See https://github.com/actions/labeler/tree/main?tab=readme-ov-file#breaking-changes-in-v5; The action version was bumped by #3050


njzjz:
We need this PR itself to fix the CI.

wanghan-iapcm (approved):

PR #3075, opened 2023-12-20
add CodeQL checks

njzjz:
It can scan the codes and report as follows:

image


wanghan-iapcm (approved):

wanghan-iapcm:
The information is very helpful though they are more than 50 pages...

PR #3106, opened 2024-01-04
rm rcut from DeepmdDataSystem

wanghan-iapcm:


njzjz (approved):

njzjz:
rcut was used but removed by #2481 (released in v2.2.3)

PR #3112, opened 2024-01-05
fix native layer concat bug. 

wanghan-iapcm:
add UT testing all input and output shapes


njzjz:
What is the expected input shape?

njzjz (approved):

wanghan-iapcm:
> What is the expected input shape?

only the last dimension of the input is changed by the mlp. Do you add test of higher dimensional input like [xxx, yyy, ninp]?

wanghan-iapcm:
> > What is the expected input shape?
> 
> only the last dimension of the input is changed by the mlp. Do we add test of higher dimensional input like [xxx, yyy, ninp]?

done

njzjz (approved):

PR #3113, opened 2024-01-07
model format for the embedding net

wanghan-iapcm:
- refact the dir structure for the model_format
- change the input of `NativeNet` from `List[NativeLayer]` to the `List[dict]`


deepmd_utils/model_format/network.py
```diff
@@ -329,3 +329,62 @@ def call(self, x: np.ndarray) -> np.ndarray:
...
+        """Serialize the network to a dict.
+
+        Returns
+        -------
+        dict
+            The serialized network.
+        """
+        return {
+            "in_dim": self.in_dim,
+            "neuron": self.neuron,
```
njzjz, line 372:
```suggestion
            "neuron": self.neuron.copy(),
```

This is safer. If it is changed in other places, the default value of this class will not be affected.

deepmd_utils/model_format/network.py
```diff
@@ -329,3 +329,62 @@ def call(self, x: np.ndarray) -> np.ndarray:
         for layer in self.layers:
             x = layer.call(x)
         return x
+
+
+class EmbeddingNet(NativeNet):
+    def __init__(
+        self,
+        in_dim,
+        neuron: List[int] = [24, 48, 96],
```
njzjz, line 338: I think the first layer is 1? Should it be written explicitly?

wanghan-iapcm: Not really, we may pass the concatenated the s(r) and type through the embedding net.  see eq 5 of https://arxiv.org/pdf/2208.08236.pdf

njzjz (approved):

PR #3114, opened 2024-01-08
support numerical precision and env_mat

wanghan-iapcm:
- change of numerical precision is supported
- environment matrix is supported.

### DIFF
--- a/deepmd_utils/model_format/network.py
+++ b/deepmd_utils/model_format/network.py
@@ -160,6 +160,7 @@ class NativeLayer(NativeOP):
         self.idt = idt.astype(prec) if idt is not None else None
         self.activation_function = activation_function
         self.resnet = resnet
+        self.check_type_consistency()
 
     def serialize(self) -> dict:
         """Serialize the layer to a dict.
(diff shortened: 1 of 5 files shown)


deepmd_utils/model_format/network.py
```diff
@@ -198,6 +196,7 @@ def deserialize(cls, data: dict) -> "NativeLayer":
             idt=data["@variables"].get("idt", None),
             activation_function=data["activation_function"],
             resnet=data.get("resnet", False),
+            precision=data.get("precision", "default"),
```
njzjz, line 199: I propose
```suggestion
            precision=data["@variables"]["w"].dtype.name,
```

njzjz: Do we really need this key? I think we can get the type easily everywhere...

deepmd_utils/model_format/network.py
```diff
@@ -156,10 +150,13 @@ def __init__(
         idt: Optional[np.ndarray] = None,
         activation_function: Optional[str] = None,
         resnet: bool = False,
+        precision: str = "default",
```
njzjz, line 153: Why not just set `float64` as the default.. In deepmd tf, `default` may have different meanings, but I think we need the exact meaning when storing the model.

njzjz (approved):

njzjz:
I still have a question. Do we want to support `env_mat_r` in the future? The current `env_mat` is actually `env_mat_a`.

wanghan-iapcm:
> I still have a question. Do we want to support `env_mat_r` in the future? The current `env_mat` is actually `env_mat_a`.

To compute `env_mat_r`, we can cut the first column of env_mat_a out. The overhead of the extra three columns looks marginal.

PR #3117, opened 2024-01-09
add native Networks for mutiple Network classes

njzjz:


source/tests/test_model_format_utils.py
```diff
@@ -108,6 +109,61 @@ def test_embedding_net(self):
...
+                    "resnet": True,
+                    "@variables": {"w": w, "b": b},
+                },
+            ],
+        }
+
+    def test_two_dim(self):
+        networks = Networks(ndim=2)
+        networks[(0, 0)] = self.network
+        networks[(0, 1)] = self.network
```
wanghan-iapcm, line 134: should we assume the completeness of the networks? in case ndim==2 and ntypes==2, the networks (0, 0), (0, 1), (1, 0) and (1, 1) should be provided.

deepmd_utils/model_format/network.py
```diff
@@ -409,3 +411,98 @@ def deserialize(cls, data: dict) -> "EmbeddingNet":
         obj = cls(**data)
         super(EmbeddingNet, obj).__init__(layers)
         return obj
+
+
+class Networks:
```
wanghan-iapcm, line 416: perhaps the name `NetworkCollection` or `NetworkArray` is more self-explanatory.

wanghan-iapcm (approved):

PR #3123, opened 2024-01-09
Add dp model format sea

wanghan-iapcm:
- add precision test for embedding net

Limitations 
- only support `type_one_side`
- does not support type embedding and `stripped_type_embedding`
- does not support `exclude_types`
- does not support spin

### DIFF
--- a/deepmd_utils/model_format/se_e2_a.py
+++ b/deepmd_utils/model_format/se_e2_a.py
@@ -1,0 +1,195 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+import numpy as np
+
+try:
+    from deepmd_utils._version import version as __version__
+except ImportError:
+    __version__ = "unknown"
+
+from typing import (
+    Any,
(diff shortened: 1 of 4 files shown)


deepmd_utils/model_format/se_e2_a.py
```diff
@@ -0,0 +1,195 @@
...
+        for ii in range(self.ntypes):
+            self.embeddings.append(
+                EmbeddingNet(
+                    in_dim,
+                    self.neuron,
+                    self.activation_function,
+                    self.resnet_dt,
+                    self.precision,
+                )
+            )
```
njzjz, line 82: Does it only support `type_one_side = True`? Please see my PR #3117

wanghan-iapcm: forgot to mention the limitations of the current pr. please check the updated pr notes.

njzjz (approved):

PR #3128, opened 2024-01-10
add definition for the output of fitting and model

wanghan-iapcm:


deepmd_utils/model_format/output_def.py
```diff
@@ -0,0 +1,262 @@
...
+
+def model_check_output(cls):
+    class wrapper(cls):
+        def __init__(
+            self,
+            *args,
+            **kwargs,
+        ):
+            super().__init__(*args, **kwargs)
+            self.md = cls.output_def()
```
wanghan-iapcm, line 33: should not be a class method, will fix latter.

deepmd_utils/model_format/output_def.py
```diff
@@ -0,0 +1,278 @@
...
+    """Defines the shape and other properties of a variable.
+
+    Parameters
+    ----------
+    name
+          Name of the output variable. Notice that the xxxx_redu,
+          xxxx_derv_c, xxxx_derv_r are reserved names that should
+          not be used to define variables.
+    shape
+          The shape of the variable. e.g. energy should be [1],
```
njzjz, line 106: I think the shape of energy should be `[]`,

```py
>>> import numpy as np
>>> np.zeros(())
array(0.)
>>> np.zeros((1,))
array([0.])
```

wanghan-iapcm: Using [1] has the advantage that the shape handling is the same as other physical quantities that are higher than dimension 1, like the dipole ([3]).

njzjz:
@CodiumAI-Agent /review

njzjz:
@CodiumAI-Agent /improve --extend

njzjz (approved):
The suggestions by AI look not necessary...

PR #3127, opened 2024-01-10
[Paddle Backend][Fix] Fix wrong custom operator installation filename

HydrogenSulfate:
1. Fix wrong custom operator installation filename
2. Fix missing file `deepmd/entrypoints/gui.py`
3. Annotated `@cast_precision` and keep brackets `[]` util problems fixed @SigureMo

@njzjz Besides, can you please help to rename `paddle` branch to `paddle_1` or other backup name, and then change `paddle2` to `paddle` branch? Because user may get confused aboout the difference between `paddle2 and `paddle`. Thanks a lot.

PR #3130, opened 2024-01-11
[Paddle Backend] Deprecate old paddle branch

HydrogenSulfate:
Update caution notes and re-formatting README.md in old paddle branch as @njzjz advised at #3127 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,58 +1,71 @@
-<span style="font-size:larger;">DeePMD-kit Manual</span>
-========
+# DeePMD-kit Manual
+
 [![GitHub release](https://img.shields.io/github/release/deepmodeling/deepmd-kit.svg?maxAge=86400)](https://github.com/deepmodeling/deepmd-kit/releases)
 [![Documentation Status](https://readthedocs.org/projects/deepmd/badge/?version=latest)](https://deepmd.readthedocs.io/en/latest/?badge=latest)
 
+> [!Caution]
+> ⚠️⚠️⚠️
+>
+> **The current branch is no longer maintained due to the outdated Paddle version it adapts to and the significant differences from the DeePMD-kit main branch.**
+>
+> **The new paddle backend branch is [DeePMD-kit(paddle branch)](https://github.com/deepmodeling/deepmd-kit/tree/paddle)**
+
 # Table of contents
+
 - [About DeePMD-kit](#about-deepmd-kit)
- 	- [Highlighted features](#highlighted-features)
- 	- [Code structure](#code-structure)
- 	- [License and credits](#license-and-credits)
- 	- [Deep Potential in a nutshell](#deep-potential-in-a-nutshell)
+  - [Highlighted features](#highlighted-features)
+  - [Code structure](#code-structure)
+  - [License and credits](#license-and-credits)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,58 +1,71 @@
...
+
 [![GitHub release](https://img.shields.io/github/release/deepmodeling/deepmd-kit.svg?maxAge=86400)](https://github.com/deepmodeling/deepmd-kit/releases)
 [![Documentation Status](https://readthedocs.org/projects/deepmd/badge/?version=latest)](https://deepmd.readthedocs.io/en/latest/?badge=latest)
 
+> [!Caution]
+> ⚠️⚠️⚠️
+>
+> **The current branch is no longer maintained due to the outdated Paddle version it adapts to and the significant differences from the DeePMD-kit main branch.**
+>
+> **The new paddle backend branch is [DeePMD-kit(paddle branch)](https://github.com/deepmodeling/deepmd-kit/tree/paddle)**
```
njzjz, line 11: Shouldn't the branch be `paddle2`?

HydrogenSulfate: > Shouldn't the branch be `paddle2`?

I guess paddle2 is not suitable name in the future, and code in old paddle can be removed or moved to other branch name, so can we add extra deprecation note, such as: "As descripbed above, the current `paddle` branch will be replaced with the content in `paddle2` branch after 2024.x.y". and replace code in `paddle` branch with code in `padddle2 then delete `paddle2` branch on 2024.x.y that day?

njzjz: I think it is not necessary if we want to rename the branch

HydrogenSulfate: > I think it is not necessary if we want to rename the branch

URL has been updated

HydrogenSulfate: > I think it is not necessary if we want to rename the branch

paddle branch will not be maintain any more and the code in paddle branch is too old to run. so I guess it is necessary to remove paddle branch after certain day and move paddle2's code into paddle branch

njzjz: I will rename it tomorrow. I prefer keeping the old branch for the recording purposes. Even if no one uses it, people can know the history of development through the  old branch.

#By the way, we will focus on supporting multiple backends in the next five weeks. After we refactor several codes, it's more easier to support Paddle in the main development branch.

HydrogenSulfate: > I will rename it tomorrow. I prefer keeping the old branch for the recording purposes. Even if no one uses it, people can know the history of development through the old branch.
> 
> By the way, we will focus on supporting multiple backends in the next five weeks. After we refactor several codes, it's more easier to support Paddle in the main development branch.

1. I agree with you much that we need to record old code and history of it even if no one use(so there indeed unnecessary to delete old code), but I guess it is important to move the old code to some "bakcup" branch but not be kept in offcial branch(`paddle` branch), since an official branch is important or will cause recognition error, and we always prefer showing newest runnable code for those who use DeePMD-kit with paddle backend in official branch.

2. It is very glad to hear that a multi-backend project is released and you are working for it, willing to help to build togather after multi-bakend is finished.

Thanks for your reply.

njzjz: I've renamed `paddle` to `paddle-backup` and `paddle2` to `paddle`.

PR #3137, opened 2024-01-12
support fitting net

wanghan-iapcm:
- also add doc string for the embedding net

### DIFF
--- a/deepmd_utils/model_format/__init__.py
+++ b/deepmd_utils/model_format/__init__.py
@@ -8,6 +8,7 @@ from .env_mat import (
 )
 from .network import (
     EmbeddingNet,
+    FittingNet,
     NativeLayer,
     NativeNet,
     NetworkCollection,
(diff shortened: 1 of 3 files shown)


wanghan-iapcm:
I find we could have written a uniform implementation of embedding and fitting networks across platforms. We only have to make the interface of MLPLayer uniform across platforms. @njzjz what do you think?

njzjz:
> I find we could have written a uniform implementation of embedding and fitting networks across platforms. We only have to make the interface of MLPLayer uniform across platforms. @njzjz what do you think?

Agree

wanghan-iapcm:
> > I find we could have written a uniform implementation of embedding and fitting networks across platforms. We only have to make the interface of MLPLayer uniform across platforms. @njzjz what do you think?
> 
> Agree

I will contribute in a future PR.

wanghan-iapcm:
> > I find we could have written a uniform implementation of embedding and fitting networks across platforms. We only have to make the interface of MLPLayer uniform across platforms. @njzjz what do you think?
> 
> Agree

The first step: #3138

Issue #3118, opened 2024-01-09
[Feature Request] Move backend-independent utility codes to `deepmd_utils`


njzjz:
### Summary

Move backend-independent utility codes in the `deepmd` package to the `deepmd_utils` package so other backends can use them.

### Detailed Description

Compatibility should be kept.

#### Move without modification
- #3140
  - [x] `deepmd.common.add_data_requirement`
  - [x] `deepmd.common.select_idx_map`
  - [x] `deepmd.common.j_must_have`
  - [x] `deepmd.common.j_loader`
  - [x] `deepmd.common.expand_sys_str`
  - [x] `deepmd.common.get_np_precision`
  - [x] `deepmd.utils.argcheck`
  - [x] `deepmd.utils.compat`
  - [x] `deepmd.utils.data`
  - [x] `deepmd.utils.data_system`
  - [x] `deepmd.utils.pair_tab`
  - [x] `deepmd.utils.path`
  - [x] `deepmd.utils.plugin`
  - [x] `deepmd.utils.random`
  - [x] `deepmd.utils.weight_avg`
  - [x] `deepmd.model.model_stat`
  - [x] `deepmd.loggers`

#### Move with minor modification
- [x] `deepmd.utils.batch_size` #3143
  Two things are different: (1) is_gpu_available; (2) OOM exception.
- [x] `deepmd.model_devi` #3173
  A universal `DeepPot` class is required.
- [x] `deepmd.calculator` #3173
  A universal `DeepPot` class is required.
- [x] Several subcommands in `deepmd.entrypoints` #3144
- [x] <del>`deepmd.cluster`</del> #3239
  The way to get GPUs is different.
- [x] deepmd.entrypoints.test #3249

PR #3140, opened 2024-01-12
move utility to `deepmd_utils` (without modifaction)

njzjz:
Move framework-independent codes to the `deepmd_utils` module without modification, as a step of https://github.com/deepmodeling/deepmd-kit/issues/3118.


njzjz:
git cannot detect file renaming, per https://stackoverflow.com/questions/72618096/git-rename-file-and-create-file-with-old-name-in-same-commit :(

wanghan-iapcm (approved):

PR #3145, opened 2024-01-15
cc: merge `DeepPotBase` and `DeepTensor` member functions

njzjz:
Remove 18 member functions from `DeepPotBase` and 4 member functions from `DeepTensor`.


wanghan-iapcm:
Shouldn't we change the compute method of XXXTF class as private methods ?

njzjz:
> Shouldn't we change the compute method of XXXTF class as private methods ?

Meaningful

wanghan-iapcm (approved):

Issue #3119, opened 2024-01-09
[Feature Request] Multiple backend C++ API


njzjz:
### Summary

Propose a multiple backend C++ API. The backend can be switched according to the file format.

#### Backend independent library

When @wanghan-iapcm designed v1, `libdeepmd.so` has been backend-independent, so we can just reuse it.

#### Factory method pattern

Take the example of `DeepPot`. Move the current `DeepPot` to `DeepPotTf`. Create a base class `DeepPotBase`, with two subclasses inherited, `DeepPotTf` and `DeepPotPt`. Create a wrapper class, `DeepPot`, to keep the same API as the current one, so no downstream codes need to change.

The wrapper class may look like:
```cpp
enum Backend {TENSORFLOW, PYTORCH, UNKNOWN};

class DeepPot {
  public:
  DeepPot(const std::string& model,
                 const int& gpu_rank,
                 const std::string& file_content)
    Backend backend = detect_backend(model, file_content);
    if (Backend::TENSORFLOW == backend) {
        dp = std::make_unique<DeepPotTf>(model, gpu_rank, file_content);
    } else if (ProductId::PYTORCH == backend) {
        dp = std::make_unique<DeepPotPt>(model, gpu_rank, file_content);
    } else {
        throw deepmd::exception("Unknown file type");
    }
  }
  void compute(
      std::vector<double> &ener,
      std::vector<std::vector<VALUETYPE>> &force,
      std::vector<std::vector<VALUETYPE>> &virial,
      const std::vector<VALUETYPE> &coord,
      const std::vector<int> &atype,
      const std::vector<VALUETYPE> &box,
      const int nghost,
      const InputNlist &lmp_list,
      const int &ago,
      const std::vector<VALUETYPE> &fparam = std::vector<VALUETYPE>(),
      const std::vector<VALUETYPE> &aparam = std::vector<VALUETYPE>())
  {
    dp->compute(ener, force, virial, coord, atype, box, nghost, lmp_list, ago, fparam, aparam);
  }
  private:
    std::unique_ptr<DeepPotBase> dp;
};
```

The list of public APIs can be found in `deepmd.hpp` and `c_api.cc`. Not implemented methods can throw an exception.

#### Backend detection

The backend can be detected from the model file described below:
- PyTorch: in zip files. Zip files start with `PK`.
- TensorFlow: It's not easy to detect, but ProtoBuf should throw an error when parsing it.

If a backend is not enabled, we should skip it.
[...]

PR #3162, opened 2024-01-22
cc: add torch backend support for Multiple backend

CaRoLZhangxy:
need to test in union environment (tf and pt)
see https://github.com/deepmodeling/deepmd-kit/issues/3119


source/api_cc/src/DeepPotPT.cc
```diff
@@ -0,0 +1,255 @@
...
+  if (inited) {
+    std::cerr << "WARNING: deepmd-kit should not be initialized twice, do "
+                 "nothing at the second call of initializer"
+              << std::endl;
+    return;
+  }
+  std::cout << "load model from: " << model << " to gpu " << gpu_rank
+            << std::endl;
+  gpu_id = gpu_rank;
+  torch::Device device(torch::kCUDA, gpu_rank);
```
njzjz, line 27: What happens when the GPU is not available?

source/api_cc/src/DeepPotPT.cc
```diff
@@ -0,0 +1,255 @@
...
+                                    std::vector<double>& virial,
+                                    std::vector<double>& atom_energy,
+                                    std::vector<double>& atom_virial,
+                                    const int& nframes,
+                                    const std::vector<double>& coord,
+                                    const std::vector<int>& atype,
+                                    const std::vector<double>& box,
+                                    const std::vector<double>& fparam,
+                                    const std::vector<double>& aparam) {
+  throw;
```
njzjz, line 241: Always throw `deepmd::deepmd_exception`, which can be caught by the downstream software (e.g., LAMMPS).

source/api_cc/include/DeepPotPT.h
```diff
@@ -0,0 +1,328 @@
...
+    assert(inited);
+    return rcut;
+  };
+  /**
+   * @brief Get the number of types.
+   * @return The number of types.
+   **/
+  int numb_types() const {
+    assert(inited);
+    return ntypes;
```
njzjz, line 198: These variables (`ntypes`, `ntypes_spin`, `dim_fparam`, `dim_aparam`, `aparam_nall`) seem not to be initiated anywhere, so it's expected to see a segmentation fault.

njzjz (changes requested):

source/api_cc/src/commonPT.cc
```diff
@@ -0,0 +1,26 @@
...
+  int inum = inlist.inum;
+  ilist.resize(inum);
+  numneigh.resize(inum);
+  memcpy(&ilist[0], inlist.ilist, inum * sizeof(int));
+  int* max_element = std::max_element(inlist.numneigh, inlist.numneigh + inum);
+  max_num_neighbors = *max_element;
+  if (max_num_neighbors < nnei) {
+    max_num_neighbors = nnei;
+  }
+  jlist = (int*)malloc(inum * max_num_neighbors * sizeof(int));
```
njzjz, line 17: Doesn't it cause the memory leaking since it is never freed?

source/api_cc/src/DeepPotPT.cc
njzjz: We may use `torch::cuda::is_available` to check here.

deepmd/pt/model/model/dp_atomic_model.py
```diff
@@ -100,6 +101,10 @@ def get_sel(self) -> List[int]:
         """Get the neighbor selection."""
         return self.sel
 
+    def get_nnei(self) -> List[int]:
```
njzjz, line 104: Perhaps [`torch.jit.export`](https://pytorch.org/docs/stable/jit.html#torch.jit.export) should be used?

source/api_cc/src/commonPT.cc
```diff
@@ -0,0 +1,23 @@
...
+                                         int& max_num_neighbors) {
+  int inum = inlist.inum;
+  ilist.resize(inum);
+  numneigh.resize(inum);
+  memcpy(&ilist[0], inlist.ilist, inum * sizeof(int));
+  int* max_element = std::max_element(inlist.numneigh, inlist.numneigh + inum);
+  max_num_neighbors = *max_element;
+  unsigned long nlist_size = (unsigned long)inum * max_num_neighbors;
+  jlist.resize(nlist_size);
+  memset(&jlist[0], -1, nlist_size * sizeof(int));
```
njzjz, line 15: It seems wrong: https://stackoverflow.com/a/20951450/9567349

njzjz (changes requested):

njzjz (approved):
fparam and apraram are not supported yet. But I think it's okay for this PR.

source/api_cc/src/DeepPotPT.cc
```diff
@@ -1,8 +1,398 @@
...
+//     ENERGYTYPE& ener,
+//     std::vector<float>& force,
+//     std::vector<float>& virial,
+//     std::vector<float>& atom_energy,
+//     std::vector<float>& atom_virial,
+//     const std::vector<float>& coord,
+//     const std::vector<int>& atype,
+//     const std::vector<float>& box,
+//     const InputNlist& lmp_list,
+//     const int& ago);
```
wanghan-iapcm, line 170:
```suggestion

```

source/api_cc/src/DeepPotPT.cc
```diff
@@ -1,8 +1,398 @@
...
+//     const std::vector<double>& box);
+// template void DeepPotPT::compute<float, ENERGYTYPE>(
+//     ENERGYTYPE& ener,
+//     std::vector<float>& force,
+//     std::vector<float>& virial,
+//     std::vector<float>& atom_energy,
+//     std::vector<float>& atom_virial,
+//     const std::vector<float>& coord,
+//     const std::vector<int>& atype,
+//     const std::vector<float>& box);
```
wanghan-iapcm, line 287:
```suggestion

```

wanghan-iapcm (changes requested):

wanghan-iapcm (approved):

PR #3173, opened 2024-01-24
Move model deviation and ase calculator to `deepmd_utils`

njzjz:
..., so they can benifit from multiple-backend DeepPot. Update docs.


doc/inference/python.md
```diff
@@ -15,8 +15,8 @@ where `e`, `f` and `v` are predicted energy, force and virial of the system, res
 
 Furthermore, one can use the python interface to calculate model deviation.
 ```python
-from deepmd.infer import calc_model_devi
-from deepmd.infer import DeepPot as DP
+from deepmd_utils.infer import calc_model_devi
+from deepmd_utils.infer import DeepPot as DP
```
wanghan-iapcm, line 19: we may want to change deepmd_utils to deepmd, and deepmd to deepmd_tf. We can discuss.

PR #3174, opened 2024-01-24
add dpdata driver

njzjz:
Add a dpdata driver via the plugin mechanism (override that in the dpdata package) so it can benefit from the multiple-backend DeepPot. Currently, the driver in the dpdata package has to support both v1 and v2 for backward compatibility. When shipped within the deepmd-kit package, it only needs to support the current deepmd-kit version.


wanghan-iapcm (commented):
could you please provide a test case for cell==None?

njzjz:
> could you please provide a test case for cell==None?

Good idea. I see it is not covered.

wanghan-iapcm (approved):

PR #3186, opened 2024-01-27
Fix PT `DeepPot` and replace ASE calculator

njzjz:
- Set `deepmd.pt.utils.ase_calc.DPCalculator` as an alias of `deepmd.calculator.DP`;
- Replace `deepmd_pt` with  `deepmd.pt` in `deep_pot.py`; fix (atomic) virial output shape of `DeepPot`; add tests for them;
- Set `pbc` in `pt/test_calculator.py` as it requests stress.

### DIFF
--- a/deepmd/infer/deep_pot.py
+++ b/deepmd/infer/deep_pot.py
@@ -56,7 +56,7 @@ class DeepPot(ABC):
 
                 return super().__new__(DeepPotTF)
             elif backend == DPBackend.PyTorch:
-                from deepmd_pt.infer.deep_eval import DeepPot as DeepPotPT
+                from deepmd.pt.infer.deep_eval import DeepPot as DeepPotPT
 
                 return super().__new__(DeepPotPT)
             else:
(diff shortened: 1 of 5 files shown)


njzjz:
> Set pbc in pt/test_calculator.py as it requests stress

I think it's meaningless to calculate stress for a non-periodic system, right?

deepmd/infer/deep_pot.py
```diff
@@ -56,7 +56,7 @@ def __new__(cls, model_file: str, *args, **kwargs):
 
                 return super().__new__(DeepPotTF)
             elif backend == DPBackend.PyTorch:
-                from deepmd_pt.infer.deep_eval import DeepPot as DeepPotPT
+                from deepmd.pt.infer.deep_eval import DeepPot as DeepPotPT
```
wanghan-iapcm, line 59: I am wondering why it is not detected by a UT.

wanghan-iapcm:
> > Set pbc in pt/test_calculator.py as it requests stress
> 
> I think it's meaningless to calculate stress for a non-periodic system, right?

no it does not make sense

deepmd/infer/deep_pot.py
wanghan-iapcm: We may want to check with the pt backend in the UTs of python inference

wanghan-iapcm (commented):
please add the UT as commented.

wanghan-iapcm (approved):

PR #3187, opened 2024-01-27
Merge TF and PT CLI

njzjz:
Just merge in form. Several options or subcommands are only supported by TensorFlow or PyTorch.

Also, avoid import from `deepmd.tf` in `deepmd.utils.argcheck`.

```
Use --tf or --pt to choose the backend:
    dp --tf train input.json
    dp --pt train input.json
```

### DIFF
--- a/deepmd/main.py
+++ b/deepmd/main.py
@@ -84,7 +84,7 @@ def main_parser() -> argparse.ArgumentParser:
 
     # default backend is TF for compatibility
     default_backend = os.environ.get("DP_BACKEND", "tensorflow").lower()
-    if default_backend not in ["tensorflow", "pytorch"]:
+    if default_backend not in BACKEND_TABLE.keys():
         raise ValueError(
             f"Unknown backend {default_backend}. "
             "Please set DP_BACKEND to either tensorflow or pytorch."
--- a/deepmd/main.py
+++ b/deepmd/main.py
@@ -253,8 +253,8 @@ def main_parser() -> argparse.ArgumentParser:
     )
     parser_frz.add_argument(
         "-c",
-        "--checkpoint",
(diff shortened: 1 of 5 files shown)


deepmd/main.py
```diff
@@ -56,8 +57,48 @@ def main_parser() -> argparse.ArgumentParser:
...
+        raise ValueError(
+            f"Unknown backend {default_backend}. "
+            "Please set DP_BACKEND to either tensorflow or pytorch."
+        )
+
+    parser_backend = parser.add_mutually_exclusive_group()
+    parser_backend.add_argument(
+        "-b",
+        "--backend",
+        choices=["tensorflow", "pytorch"],
```
wanghan-iapcm, line 82: shall we provide short-hand keyword like `tf` and `pt` ?

deepmd/main.py
```diff
@@ -651,6 +733,16 @@ def main():
         if no command was input
     """
     args = parse_args()
-    from deepmd.tf.entrypoints.main import main as deepmd_main
+
+    if args.tf:
```
wanghan-iapcm, line 737: we may want to check `--tf`` and `--pt` should be be set at the same time.

njzjz: This is what `add_mutually_exclusive_group` has done

njzjz: It seems there is a better code here:
https://stackoverflow.com/a/62863301/9567349

deepmd/main.py
njzjz: I found an elegant way here: https://stackoverflow.com/a/72651934/9567349

wanghan-iapcm (approved):

PR #3191, opened 2024-01-27
PT: keep the same checkpoint behavior as TF

njzjz:
Set the default `save_ckpt` to `model.ckpt` as the prefix. When saving checkpoints, `model.ckpt-100.pt` will be saved, and `model.ckpt.pt` will be symlinked to `model.ckpt-100.pt`. A `checkpoint` file will be dedicated to record `model.ckpt-100.pt`.

This keeps the same behavior as the TF backend. One can do the below using the PT backend just like the TF backend:

```sh
dp --pt train input.json
# one can cancel the training before it finishes
dp --pt freeze
```

### DIFF
--- a/deepmd/common.py
+++ b/deepmd/common.py
@@ -1,5 +1,9 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
+import glob
 import json
+import os
+import platform
+import shutil
 import warnings
 from pathlib import (
     Path,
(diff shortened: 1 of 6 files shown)


wanghan-iapcm (approved):

thangckt:
hi @njzjz 

Can I know why you need different file extension `.pth` and `.pt` when using pytorch?

The files `*.pt` are generated when run
```
dp --pt train input.json
```
and the file `*.pth` when run
```
dp --pt freeze
```
can we just use one of these ext for convenient when collect files in dpegen?

njzjz:
No control flow is saved in the checkpoint file.

Issue #1895, opened 2022-09-06
[Feature Request] Validity check of the the short-range table and smooth it


wanghan-iapcm:
### Summary

Providing `use_srtab` will switch on the interpolation of a tabulated short-range interation with DP.  This issue asks for the validity check of the table and smooth it. 

### Detailed Description

1. The table provides a pairwise iteration saying u(r). The upper range of r in the table file should be larger than rcut, because the u(r) of the neighbors within rcut will be calculated. This is not checked by the current code. 
2. The short-range interaction should be smooth at the cut-off distance, and this may not be met by the user-provided tabulated potential. Therefore, DP should provide an option to smoothy the tabulated potential between `rcut_smth` and `rcut`.

PR #3192, opened 2024-01-28
Feat: add pair table model to pytorch

anyangml:
Migrated from this [PR](https://github.com/dptech-corp/deepmd-pytorch/pull/174). This is to reimplement the PairTab Model in Pytorch.

Notes:

1. different from the tensorflow version, the pytorch version abstracts away all the post energy conversion operations (force, virial).
2. the pytorch version implements additional checks on user provided table against the `rcut`, linear extrapolation will be performed when necessary. Check `_check_table_upper_boundary` method for details.

### DIFF
--- a/deepmd/pt/model/model/pair_tab.py
+++ b/deepmd/pt/model/model/pair_tab.py
@@ -1,0 +1,351 @@
+from .atomic_model import AtomicModel
+from deepmd.utils.pair_tab import (
+    PairTab,
+)
+import logging
+import torch
+from torch import nn
+import numpy as np
+from typing import Dict, List, Optional, Union
+
+from deepmd.model_format import FittingOutputDef, OutputVariableDef
+
+class PairTabModel(nn.Module, AtomicModel):
+    """Pairwise tabulation energy model.
+
+    This model can be used to tabulate the pairwise energy between atoms for either
+    short-range or long-range interactions, such as D3, LJ, ZBL, etc. It should not
+    be used alone, but rather as one submodel of a linear (sum) model, such as
+    DP+D3.
+
+    Do not put the model on the first model of a linear model, since the linear
+    model fetches the type map from the first model.
+
+    At this moment, the model does not smooth the energy at the cutoff radius, so
+    one needs to make sure the energy has been smoothed to zero.
+
+    Parameters
+    ----------
+    tab_file : str
+        The path to the tabulation file.
+    rcut : float
+        The cutoff radius.
+    sel : int or list[int]
+        The maxmum number of atoms in the cut-off radius.
+    """
+
+    def __init__(
+        self, tab_file: str, rcut: float, sel: Union[int, List[int]], **kwargs
+    ):
+        super().__init__() 
+        self.tab_file = tab_file
+        self.rcut = rcut
+
+        # check table data against rcut and update tab_file if needed.
+        self._check_table_upper_boundary()
+        
+        self.tab = PairTab(self.tab_file)
+        self.ntypes = self.tab.ntypes
+        
+
+        tab_info, tab_data = self.tab.get() # this returns -> Tuple[np.array, np.array]
+        self.tab_info = torch.from_numpy(tab_info) 
+        self.tab_data = torch.from_numpy(tab_data)
+
+        # self.model_type = "ener"
+        # self.model_version = MODEL_VERSION ## this shoud be in the parent class
+
+        if isinstance(sel, int):
+            self.sel = sel
+        elif isinstance(sel, list):
+            self.sel = sum(sel)
+        else:
+            raise TypeError("sel must be int or list[int]")
+    
+    def get_fitting_output_def(self)->FittingOutputDef:
+        return FittingOutputDef(
+            [OutputVariableDef(
+                name="energy",
+                shape=[1],
+                reduciable=True,
+                differentiable=True
+            )]
+        )
+
+    def get_rcut(self)->float:
+        return self.rcut
+    
+    def get_sel(self)->int:
+        return self.sel
+    
+    def distinguish_types(self)->bool:
+        # to match DPA1 and DPA2.
+        return False
+        
+    def forward_atomic(
+        self,
+        extended_coord, 
+        extended_atype, 
+        nlist,
+        mapping: Optional[torch.Tensor] = None,
+        do_atomic_virial: bool = False,
+        ) -> Dict[str, torch.Tensor]:
+        
+
+        nframes, nloc, nnei = nlist.shape
+
+        #this will mask all -1 in the nlist
+        masked_nlist = torch.clamp(nlist,0)
+
+        atype = extended_atype[:, :nloc] #(nframes, nloc)
+        pairwise_dr = self._get_pairwise_dist(extended_coord) # (nframes, nall, nall, 3)
+        pairwise_rr = pairwise_dr.pow(2).sum(-1).sqrt() # (nframes, nall, nall)
+
+        self.tab_data = self.tab_data.reshape(self.tab.ntypes,self.tab.ntypes,self.tab.nspline,4)
+
+        #to calculate the atomic_energy, we need 3 tensors, i_type, j_type, rr
+        #i_type : (nframes, nloc), this is atype.
+        #j_type : (nframes, nloc, nnei)
+        j_type = extended_atype[torch.arange(extended_atype.size(0))[:, None, None], masked_nlist]
+
+        #slice rr to get (nframes, nloc, nnei)
+        rr = torch.gather(pairwise_rr[:, :nloc, :],2, masked_nlist)
+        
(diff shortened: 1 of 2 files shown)


deepmd/pt/model/model/pair_tab.py
```diff
@@ -0,0 +1,374 @@
...
+                [0.01  0.8   1.6   2.4  ]
+                [0.015 0.    1.    1.5  ]]
+
+        rcut = 0.022
+
+        new_table = [[0.005 1.    2.    3.   ]
+                    [0.01  0.8   1.6   2.4  ]
+                    [0.015 0.    1.    1.5  ]
+                    [0.02  0.    0.5   0.75 ]
+                    [0.025 0.    0.    0.   ]]
```
wanghan-iapcm, line 164: We requires that the table strictly decays to zero at rcut, thus if the cut-off is 0.022, then the expected values are 
```
[0.02  0.    0.    0. ]
[0.025 0.    0.    0. ]]
```
otherwise we have a small but non-zero value at the rcut.

anyangml: i see.

deepmd/pt/model/model/pair_tab.py
```diff
@@ -0,0 +1,374 @@
...
+        atomic_energy = 0.5 * torch.sum(
+            torch.where(
+                nlist != -1, raw_atomic_energy, torch.zeros_like(raw_atomic_energy)
+            ),
+            dim=-1,
+        )
+
+        return {"energy": atomic_energy}
+
+    def _check_table_upper_boundary(self):
```
njzjz, line 143: Why not directly add to the `PairTab` class? This method just modifies the data but is not related to PT. TF can also use it.

deepmd/pt/model/model/pair_tab.py
```diff
@@ -0,0 +1,374 @@
...
+                    [
+                        np.linspace(start, 0, rcut_idx - upper_idx + 1)
+                        for start in upper_val
+                    ]
+                ).T
+                raw_data = np.concatenate((raw_data[:-1, :], pad_linear), axis=0)
+
+        # over writing file with padding if applicable.
+        with open(self.tab_file, "wb") as f:
+            np.savetxt(f, raw_data)
```
njzjz, line 234: Overwriting the user's provided file is not a good idea.

anyangml: three options here:

1. overwriting the file: I agree, this may not be the best practice, even though all the original information is still kept in the file.
2. write to a new file
3. change the `PairTab` and allow both `file path` and `np.array` as input. With this, we can easily add test cases, without writing new files.

deepmd/pt/model/model/pair_tab.py
anyangml: sure, if that's what we want.

deepmd/pt/model/model/pair_tab.py
```diff
@@ -0,0 +1,371 @@
...
+
+        rcut = 0.031
+
+        new_table = [[0.005 1.    2.    3.   ]
+                    [0.01  0.8   1.6   2.4  ]
+                    [0.015 0.5   1.    1.5  ]
+                    [0.02  0.25  0.4   0.75 ]
+                    [0.025 0.    0.1   0.   ]
+                    [0.03  0.    0.    0.   ]
+                    [0.035 0.    0.    0.   ]]
```
anyangml, line 183: @wanghan-iapcm I wonder for those padding zeros, when creating cubic spline, will that result in negative energy values?

anyangml: the easiest fix for this would be to get the number of zero padding rows from the updated table, and overwrite the cubic spline coefficients before calculating energy.

deepmd/pt/model/model/pair_tab.py
anyangml: @njzjz @wanghan-iapcm I lean towards option 3

njzjz: It should not be the input of `PairTab`. Instead, it should be internal within `PairTab`.

deepmd/pt/model/model/pair_tab.py
wanghan-iapcm: > @wanghan-iapcm It seems when padding with these zeros, it may cause the cubic spline go below zero and result in negative energy values.

I agree. it is a numerical issue.  The solution would be construct a smooth spline (continuous up to the 2nd order at table_up and rcut) between table_up and rcut.

deepmd/utils/pair_tab.py
```diff
@@ -73,11 +163,18 @@ def _make_data(self):
...
-                        -3 * vv[ii] + 3 * vv[ii + 1] - 2 * dd[ii] - dd[ii + 1]
+                        (-3 * vv[ii] + 3 * vv[ii + 1] - 2 * dd[ii] - dd[ii + 1])
+                        if vv[ii] != 0
+                        else 0
                     )
-                    dtmp[ii * 4 + 2] = dd[ii]
+                    dtmp[ii * 4 + 2] = dd[ii] if vv[ii] != 0 else 0
                     dtmp[ii * 4 + 3] = vv[ii]
                 data[
                     (t0 * self.ntypes + t1) * stride : (t0 * self.ntypes + t1) * stride
```
anyangml, line 180: if the table value at the beginning of the given spline is already `0` then, we set all 4 coefficients to `0`, to prevent negative energy.

deepmd/pt/model/model/pair_tab.py
```diff
@@ -0,0 +1,281 @@
...
+
+        uu -= idx
+
+        final_coef = self._extract_spline_coefficient(i_type, j_type, idx)
+
+        # here we need to do postprocess to overwrite coefficients when table values are not zero at rcut in linear extrapolation.
+        if self.tab.rmax < self.rcut:
+            post_mask = rr >= self.rcut
+            final_coef[post_mask] = torch.zeros(4)
+
```
anyangml, line 198: this differs from the tensorflow implementation, and may lead to inconsistent results. this is to handle a specific edge case, where rcut is slightly larger than the table upper boundary. for example, the table upper boundary is `0.2`, and the rcut is `0.22`, if the space between grids (hh) is `0.05`, in the current implementation, the table values do not decay to `0` at `0.22`, but at `0.25`. So for values between `0.2` and `0.22`, the values are calculated based on linear extrapolation in range 0.2~0.25, and the values at and beyond 0.22 are forced to `0`

deepmd/utils/pair_tab.py
anyangml: We decided to ignore this case, and allow negative energy.

deepmd/pt/model/model/pair_tab.py
anyangml: Note, this logic is not implemented in the Tensorflow version, need to add to the c++ operator

source/tests/tf/test_pairtab_preprocess.py
njzjz, line 1: This file should move to `common` directory

njzjz:
Does this PR fix #1895 or #1973?

anyangml:
> Does this PR fix #1895 or #1973?

It's related to these issues, and partially solves them. The extrapolation feature added in this PR only handles the case where `rcut` is larger than the table upper boundary, it will calculate atomic energy with cubic spline extrapolation for atoms in the range between table upper boundary and `rcut`.

njzjz (approved):

deepmd/pt/model/model/pair_tab.py
```diff
@@ -0,0 +1,375 @@
...
+        ener = self._calcualte_ener(table_coef, uu)
+
+        if self.tab.rmax <= self.rcut:
+            # here we need to overwrite energy to zero beyond rcut.
+            mask_beyond_rcut = rr > self.rcut
+            ener[mask_beyond_rcut] = 0
+
+            # here we use smooth extrapolation to replace linear extrapolation.
+            extrapolation = self._extrapolate_rmax_rcut()
+            if extrapolation is not None:
```
wanghan-iapcm, line 205: shouldn't the extrapolation implemented in the table preparation stage and made backend independent? 
https://github.com/deepmodeling/deepmd-kit/blob/2ba031889824120422b19e7b1b2290d170ade850/deepmd/utils/pair_tab.py#L141-L148

anyangml: The `linspace` in the current implementation is to maintain the grid space. The smooth extrapolation is overwriting the linear extrapolation. 
To use smooth extrapolation in `table preparation stage`, we need to use cubic spline to calculate those points, and then in the `_make_data` method use cubic spline again to calculate the coefficient.

wanghan-iapcm (approved):

PR #3193, opened 2024-01-28
docs: document PyTorch backend

### LINKED ISSUE #3121: [Feature Request] Documentation for multiple backends
### Summary

Propose a way to document multiple backends.

### Detailed Description

We may use font-awesome icons like [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/setup/#github-actions):

image
image

Readers can quickly know which backend is supported.

If commands are different on different backends, Sphinx provides a way:

image

The problem is that TF and PT icons are unavailable in font-awesome, per https://github.com/FortAwesome/Font-Awesome/issues/17377. We need to make them by ourselves...

_No response_

njzjz:
Fix #3121.

There are TODOs:
(1) PyTorch-backend specific features and arguments;
(2) Python interface installation. Currently, the TensorFlow backend is always installed, and I am considering rewriting the logic;
(3) Unsupported features - write docs when implemented.

### DIFF
--- a/doc/install/install-from-source.md
+++ b/doc/install/install-from-source.md
@@ -14,24 +14,63 @@ cd deepmd-kit
 deepmd_source_dir=`pwd`
 ```
 
-## Install the python interface
-### Install Tensorflow's python interface
-First, check the python version on your machine.
+## Install the Python interface
+### Install Backend's Python interface
+First, check the Python version on your machine.
 Python 3.8 or above is required.
 ```bash
 python --version
 ```
 
-We follow the virtual environment approach to install TensorFlow's Python interface. The full instruction can be found on the official [TensorFlow website](https://www.tensorflow.org/install/pip). TensorFlow 2.2 or later is supported. Now we assume that the Python interface will be installed to the virtual environment directory `$tensorflow_venv`
+We follow the virtual environment approach to install the backend's Python interface.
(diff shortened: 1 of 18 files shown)


doc/install/install-from-source.md
```diff
@@ -14,45 +14,74 @@ cd deepmd-kit
...
+python -c "import tensorflow as tf;print(tf.reduce_sum(tf.random.normal([1000, 1000])))"
 ```
-If one has multiple python interpreters named something like python3.x, it can be specified by, for example
+
+One can also [build the TensorFlow Python interface from source](https://www.tensorflow.org/install/source) for customized hardware optimization, such as CUDA, ROCM, or OneDNN support.
+
+:::
+
+:::{tab-item} PyTorch {{ pytorch_icon }}
+
```
anyangml, line 58: it seems some of the special characters are not rendered correctly.

anyangml (commented):
LGTM, except the rendering issue.

doc/install/install-from-source.md
njzjz: I can see it rendered correctly in the preview of the web page. If you mean the page rendered in GitHub, we have deprecated that one.

image

anyangml (approved):

doc/install/install-from-source.md
anyangml: Approved.

wanghan-iapcm (approved):

PR #3199, opened 2024-01-29
breaking: pt: add dp model format and refactor pt impl for the fitting net.

wanghan-iapcm:
- add dp model format (backend independent definition) for the fitting
- refactor torch support, compatible with dp model format
- fix mlp issue: the idt should only be used when a skip connection is available.
- add tools `to_numpy_array` and `to_torch_tensor`.


deepmd/pt/model/model/dp_atomic_model.py
```diff
@@ -94,7 +94,7 @@ def __init__(
 
         fitting_net["type"] = fitting_net.get("type", "ener")
         if self.descriptor_type not in ["se_e2_a"]:
-            fitting_net["ntypes"] = 1
+            fitting_net["ntypes"] = self.descriptor.get_ntype()
         else:
             fitting_net["ntypes"] = self.descriptor.get_ntype()
```
iProzd, line 99: Duplicate lines.

deepmd/model_format/fitting.py
```diff
@@ -0,0 +1,355 @@
...
+        return FittingOutputDef(
+            [
+                OutputVariableDef(
+                    self.var_name, [self.dim_out], reduciable=True, differentiable=True
+                ),
+            ]
+        )
+
+    def __setitem__(self, key, value):
+        if key in ("bias_atom_e"):
```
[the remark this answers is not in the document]
njzjz: Same below

deepmd/model_format/fitting.py
```diff
@@ -0,0 +1,355 @@
...
+    .. math::
+        \mathbf{y}=\mathcal{L}^{(n)}(\mathbf{x};\mathbf{w},\mathbf{b})=
+            \mathbf{x}^T\mathbf{w}+\mathbf{b}
+
+    where :math:`\mathbf{x} \in \mathbb{R}^{N_{n-1}}` is the input vector and :math:`\mathbf{y} \in \mathbb{R}`
+    is the output scalar. :math:`\mathbf{w} \in \mathbb{R}^{N_{n-1}}` and
+    :math:`\mathbf{b} \in \mathbb{R}` are weights and bias, respectively,
+    both of which are trainable if `trainable[n]` is `True`.
+
+    Parameters
```
njzjz, line 60: No spin here

deepmd/pt/model/task/ener.py
```diff
@@ -46,67 +69,325 @@ def __init__(
...
             [
-                OutputVariableDef("energy", [1], reduciable=True, differentiable=True),
+                OutputVariableDef(
+                    self.var_name, [self.dim_out], reduciable=True, differentiable=True
+                ),
             ]
         )
 
+    def __setitem__(self, key, value):
+        if key in ("bias_atom_e"):
```
njzjz, line 169: Wrong.

njzjz (changes requested):

deepmd/pt/utils/utils.py
```diff
@@ -41,3 +49,27 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
             return x
         else:
             raise RuntimeError(f"activation function {self.activation} not supported")
+
+
+def to_numpy_array(
+    xx: torch.Tensor,
+) -> np.ndarray:
+    if xx is not None:
+        prec = [key for key, value in PT_PRECISION_DICT.items() if value == xx.dtype]
```
njzjz, line 58: ChatGPT:

> The complexity of the provided code is O(n), where n is the number of items in `PT_PRECISION_DICT`. The list comprehension iterates through all key-value pairs in `PT_PRECISION_DICT` to find keys where the corresponding value is equal to `xx.dtype`.
> 
> To optimize this code, you can use a more efficient data structure for searching. Instead of iterating through all key-value pairs, you can use a dictionary reverse lookup to find the keys associated with a specific value in constant time. Here's an optimized version:
> 
> ```python
> # Create a reverse mapping of PT_PRECISION_DICT
> reverse_precision_dict = {v: k for k, v in PT_PRECISION_DICT.items()}
> 
> # Use the reverse mapping to find keys with the desired value
> prec = reverse_precision_dict.get(xx.dtype, None)
> ```
> 
> This optimization assumes that the values in `PT_PRECISION_DICT` are unique, as a dictionary requires unique keys. If the values are not unique, you may need to modify the optimization accordingly.

deepmd/model_format/fitting.py
wanghan-iapcm: the original (tf) fitting implementation does not have the fitting doc.. I would not add it in this PR.

deepmd/pt/model/task/ener.py
```diff
@@ -46,67 +69,325 @@ def __init__(
...
         - natoms: Tell atom count and element count. Its shape is [2+self.ntypes].
 
         Returns
         -------
         - `torch.Tensor`: Total energy with shape [nframes, natoms[0]].
         """
+        xx = descriptor
+        nf, nloc, nd = xx.shape
+        dtype = descriptor.dtype
+        device = env.DEVICE
```
njzjz, line 276: I guess this is the reason why the codeql is down. These two variables have been defined in line 40-41 and are redefined here

njzjz:
I will take time to figure out what is wrong with the codeql...

njzjz (approved):

PR #3208, opened 2024-01-31
feat: breaking: backend indepdent definition for dp model

wanghan-iapcm:
Features:
- abstract base classes for atomic model, fitting and descriptor.
- dp model format for atomic models
- dp model format for models.
- torch support for atomic model format. 
- torch support `fparam` and `aparam`.

This pr also introduces the following updates:
- support region and nlist in numpy code.
- class decorator like `fitting_check_output` gives human readable class names. 
- support int types in precision dict.
- fix descriptor interfaces.
- refactor torch atomic model impl. introduces dirty hacks to be fixed. 
- provide `format_nlist` that format the nlist in forward_lower method. 

Known limitations:
- torch atomic model has dirty hacks
- interfaces for descriptor, fitting and model statistics was not considered, should be fixed in future PRs. 

Will be fixed
-  [x] dp model module path is a mess to be refactorized.
-  [x] nlist consistency should be checked. if not format nlist.
-  [x] doc strings.
-  [x] `fparam` and `aparam` support.


deepmd/dpmodel/descriptor/se_e2_a.py
```diff
@@ -240,6 +259,7 @@ def call(
         sw
             The smooth switch function.
         """
+        del mapping
```
anyangml, line 262: why deleting mapping here?

njzjz (approved):

deepmd/dpmodel/model/make_model.py
```diff
@@ -0,0 +1,275 @@
...
+def make_model(T_AtomicModel):
+    """Make a model as a derived class of an atomic model.
+
+    The model provide two interfaces.
+
+    1. the `forward_common_lower`, that takes extended coordinates, atyps and neighbor list,
+    and outputs the atomic and property and derivatives (if required) on the extended region.
+
+    2. the `forward_common`, that takes coordinates, atypes and cell and predicts
+    the atomic and reduced property, and derivatives (if required) on the local region.
```
iProzd, line 34: May be 'call_lower' and 'call' here?

iProzd (approved):

anyangml (approved):

Issue #3099, opened 2023-12-31
[Feature Request] _Integrate Grimme DFTD-3 or DFTD-4 into DeepMD-kit


Epikurz:
### Summary

I would like to propose an enhancement for future versions of DeepMD-Kit. It would be highly beneficial if the software could integrate Grimme-D3 or D4 corrections directly into the potential energy functions trained with the PBE functional. This integration would eliminate the need to train potentials with van der Waals forces explicitly included.

From a theoretical standpoint, potentials without van der Waals forces could potentially utilize a smaller Rcut, offering computational advantages. I believe this enhancement would greatly improve the versatility and efficiency of DeepMD-Kit, particularly for users interested in training potentials using the PBE functional with van der Waals corrections.

### Detailed Description

Current versions allow for the simultaneous use of DeepMD-kit and DFTD3 as calculators through ASE to achieve certain functionalities. However, when running LAMMPS, there is currently no straightforward method to perform D3 corrections concurrently with MD simulations. I would like to propose the capability to easily incorporate van der Waals force corrections when using LAMMPS for DeepPotential simulations. I believe this approach would be more convenient and accurate than training potential functions using DFT data pre-corrected with D3 corrections.

njzjz:
I notice a recent work did the similar things: https://doing.org/10.1088/1361-648X/ad1278 The cutoff radius was added for simulations, but it's unclear whether there was a switch function.

I personally think it should be implemented into LAMMPS.

For the DeePMD-kit, the fastest way might be reimplementing the pairwise tabulate. So, the DeePMD-kit does not need to know the details about external potentials.

PR #3210, opened 2024-02-01
Feat: add ZBL weighted DP model

anyangml:
This PR is to implement general linear combination of several DPAtomicModels with user defined weights, as well as a special case ZBLModel with weights calculated based on this paper: `Appl. Phys. Lett. 114, 244101 (2019); doi: 10.1063/1.5098061`


deepmd/pt/model/model/linear_model.py
```diff
@@ -0,0 +1,194 @@
...
+
+        zbl_nframe, zbl_nloc, zbl_nnei = zbl_nlist.shape
+        dp_nframe, dp_nloc, dp_nnei = dp_nlist.shape
+        zbl_atype = extended_atype[
+            :, :zbl_nloc
+        ]  # nframe, nloc should all be the same, only difference is nnei based on rcut and nlist.
+        dp_atype = extended_atype[:, :dp_nloc]
+
+        # which rr should I use? this rr should be (nfrmaes, nloc, nnei)
+        zbl_weight = self._compute_weight(rr, ra, rb)
```
anyangml, line 147: @wanghan-iapcm If I understand this correctly, I think `nframe` and `nloc` should be the same for two sub atomic models, `nnei` will be different based on their respective `rcut`s and `nlist`s. When calculating the zbl weight, we need to use `rr`, which has the same shape as `nlist`. which nlist should we use?

wanghan-iapcm: Yes correct. we prefer the larger rcut to calculate the conf-determined weights.

deepmd/pt/model/model/linear_model.py
```diff
@@ -0,0 +1,230 @@
...
+        sigma = numerator / denominator
+        u = (sigma - ra) / (rb - ra)
+        coef = torch.zeros_like(u)
+        left_mask = sigma < ra
+        mid_mask = (ra <= sigma) & (sigma < rb)
+        right_mask = sigma >= rb
+        coef[left_mask] = 1
+        smooth = -6 * u**5 + 15 * u**4 - 10 * u**3 + 1
+        coef[mid_mask] = smooth[mid_mask]
+        coef[right_mask] = 0
```
anyangml, line 229: please double check this part.

source/tests/pt/test_linear_model.py
```diff
@@ -0,0 +1,157 @@
...
+                [0.9236, 0.0],
+                [0.6697, 0.0],
+                [0.3303, 0.0],
+                [0.0764, 0.0],
+                [0.0005, 0.0],
+                [0.0, 0.0],
+                [0.0, 0.0],
+            ],
+            dtype=torch.float64,
+        )
```
anyangml, line 106: I only tested the pairwise weight, the actual energy was not tested due to random initialization of DPmodel, we can set random seed if that's necessary.

deepmd/pt/model/model/linear_model.py
```diff
@@ -0,0 +1,230 @@
...
+from .model import (
+    BaseModel,
+)
+from .pair_tab_model import (
+    PairTabModel,
+)
+
+
+class LinearModel(BaseModel, BaseAtomicModel):
+    """Model linearly combine a list of AtomicModels.
```
anyangml, line 32: We probably don't need `BaseModel` here

deepmd/pt/model/model/linear_model.py
wanghan-iapcm: It is recommended to check against the corresponding tf operator 
https://github.com/deepmodeling/deepmd-kit/blob/cd77429dcbdaf42892ac3319b6fb72b9a2879a65/deepmd/tf/model/ener.py#L246-L256

source/tests/pt/test_linear_model.py
wanghan-iapcm: Ideally one check the cross consistency (one do not need to fix the rand seed) between pt and np, tf and np, then the pt and tf implementations are equivalent. 

You may check the smoothness and the permutational, translational and rotationally symmetries of the model. See pt/test_autodiff.py, pt/test_smooth.py, pt/test_permutation.py, pt/test_trans.py, pt/test_rot.py. 
They are all necessary but not sufficient conditions for a model.

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,234 @@
...
+        dp_energy = self.dp_model.forward_atomic(
+            extended_coord, extended_atype, dp_nlist
+        )["energy"]
+        # (nframes, nloc, 1)
+        zbl_energy = self.zbl_model.forward_atomic(
+            extended_coord, extended_atype, zbl_nlist
+        )["energy"]
+
+        fit_ret = {
+            "energy": (self.zbl_weight * zbl_energy + (1 - self.zbl_weight) * dp_energy)
```
anyangml, line 160: We might need to add the bias term here. It seems on `tf` side, the bias is calculated from data statistics `atom_bias_ener` in `Fitting`. I don't see a similar implementation in `pt`, should I leave it as an input parameter.

source/tests/pt/test_linear_model.py
anyangml: I will see if I can use the c++ ops to match `sw_lambda`.

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class LinearModel(BaseAtomicModel):
+    """Model linearly combine a list of AtomicModels.
+
+    Parameters
+    ----------
+    models
+            This linear model should take a DPAtomicModel and a PairTable model.
+    """
```
njzjz, line 35: Why is it called `LinearModel`? I don't see it as "linear".

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def __init__(
+        self,
+        dp_models: DPAtomicModel,
```
njzjz, line 39: It's confusing that the type of "dp_models" is `DPAtomicModel` instead of `list[DPAtomicModel]`.

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,234 @@
...
+        super().__init__()
+        self.dp_model = dp_models
+        self.zbl_model = zbl_model
+        self.rcut = self.get_rcut()
+        self.sel = self.get_sel()
+
+    def get_fitting_output_def(self) -> FittingOutputDef:
+        """Get the output def of the fitting net."""
+        return (
+            self.fitting_net.output_def()
```
njzjz, line 52: Where is `self.fitting_net` defined?

anyangml: this will be changed later

deepmd/dpmodel/model/linear_model.py
anyangml: In this implementation, we decided to only allow the linear combination of a `PairTabModel` and a `DPAtomicModel`.

deepmd/dpmodel/model/linear_model.py
anyangml: Just to match the tf side, a linear combination of two models.

deepmd/dpmodel/model/linear_model.py
wanghan-iapcm: Yes the data stat interface has not been added to the pt fittings. waiting for the design of @iProzd

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,226 @@
...
+        return self.dp_model.distinguish_types() and self.zbl_model.distinguish_types()
+
+    def forward_atomic(
+        self,
+        extended_coord,
+        extended_atype,
+        nlist,
+        sw_rmin: float,
+        sw_rmax: float,
+        smin_alpha: Optional[float] = 0.1,
```
wanghan-iapcm, line 76: this **breaks** the interface of `forward_atomic`. they are attributes of the class object, please set them at `__init__`

deepmd/dpmodel/model/linear_model.py
njzjz: The linear combination should accept any number of models, not only two. The weight can be any constant. Users have requested features including #2724, #2780, and #3099.
What you implement is more like interpolation but can be a particular case of linear combination.

wanghan-iapcm (changes requested):

njzjz:
A better design is first to define a common linear combination that allows any number of models (consistent with the TF side) and then to inherit from the common case to customize the weight for ZBL interpolation (limit the number of models to 2).

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,226 @@
...
+        self.rcut = self.get_rcut()
+        self.sel = self.get_sel()
+
+    def get_rcut(self) -> float:
+        """Get the cut-off radius."""
+        return self.get_rcuts()[-1]
+
+    def get_rcuts(self) -> float:
+        """Get the cut-off radius for each individual models in ascending order."""
+        return sorted([self.zbl_model.get_rcut(), self.dp_model.get_rcut()])
```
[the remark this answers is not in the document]
anyangml: removed the sorting, the order now matches the model input order.

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def get_sel(self) -> int:
+        """Get the neighbor selection."""
+        return self.get_sels()[-1]
```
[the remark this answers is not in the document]
anyangml: converted to List[int]

deepmd/dpmodel/model/linear_model.py
anyangml: fixed.

anyangml:
> A better design is first to define a common linear combination that allows any number of models (consistent with the TF side) and then to inherit from the common case to customize the weight for ZBL interpolation (limit the number of models to 2).

check if the current implementation matches what you have in mind.

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,415 @@
...
+        ]
+        ener_list = [
+            model.forward_atomic(
+                extended_coord,
+                extended_atype,
+                nlist,
+                mapping,
+            )["energy"]
+            for model, nlist in zip(self.models, nlists_)
+        ]
```
anyangml, line 155: We can potentially reuse some of the code here in the ZBLModel, will refactor in a Future PR, when we finalize the bias term.

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,322 @@
...
+    weights : list[float] or str
+        If the type is list[float], a list of weights for each model.
+        If "mean", the weights are set to be 1 / len(models).
+        If "sum", the weights are set to be 1.
+        If "zbl", the weights are calculated internally. This only allows the combination of a PairTabModel and a DPAtomicModel.
+    """
+
+    def __init__(
+        self,
+        models: List[Union[DPAtomicModel, PairTabModel]],
```
wanghan-iapcm, line 49: type hint as `List[BaseAtomicModel]`? the member in the list uses the interfaces defined by `BaseAtomicModel`

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def get_sel(self) -> List[int]:
+        return [max(self.get_sels())]
+
+    def get_sels(self) -> List[int]:
+        """Get the cut-off radius for each individual models in ascending order."""
+        return [
+            sum(model.get_sel())
+            if isinstance(model.get_sel(), list)
+            else model.get_sel()
```
wanghan-iapcm, line 80: for indistinguished type models, the get_sel method should also return List[int], the list is of length 1.

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,322 @@
...
+        """Get the cut-off radius for each individual models in ascending order."""
+        return [model.get_rcut() for model in self.models]
+
+    def get_sel(self) -> List[int]:
+        return [max(self.get_sels())]
+
+    def get_sels(self) -> List[int]:
+        """Get the cut-off radius for each individual models in ascending order."""
+        return [
+            sum(model.get_sel())
```
wanghan-iapcm, line 78: you may want to use `model.get_nsel()`

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,322 @@
...
+        If the type is list[float], a list of weights for each model.
+        If "mean", the weights are set to be 1 / len(models).
+        If "sum", the weights are set to be 1.
+        If "zbl", the weights are calculated internally. This only allows the combination of a PairTabModel and a DPAtomicModel.
+    """
+
+    def __init__(
+        self,
+        models: List[Union[DPAtomicModel, PairTabModel]],
+        weights: Union[str, List[float]],
```
wanghan-iapcm, line 50: not a good idea to maintain weight method by strings. 
I would suggest providing an abstract method for defining the weights. and different weight definitions are derived classed implementing the abstract method.

If you want to define the weight by str, the name of the weight for the zbl-interpolated model should not be "zbl", because the weights are nothing to do with the ZBL (which is a pair-wise energy model). You may want to use something like the switch_by_softmin_pair_distance.

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def forward_atomic(
+        self,
+        extended_coord,
+        extended_atype,
+        nlist,
+        mapping: Optional[np.ndarray] = None,
+        atomic_bias: Optional[np.ndarray] = None,
+        **kwargs,
+    ) -> Dict[str, np.ndarray]:
```
wanghan-iapcm, line 103: It is breaking the interface of atomic model, which is not preferred. the aparam and fparam should be passed as they are to the linearly interpolated models. 

If the `atomic_bias` should be set as an attribute of the class (because it would not change in the training or when doing MD simulations.

wanghan-iapcm (changes requested):

njzjz (changes requested):
I haven't carefully checked - but could you first add a test for JIT?
https://github.com/deepmodeling/deepmd-kit/blob/79f98cac8049b4ca22d2f42fe57bc715b5a65e01/source/tests/pt/model/test_jit.py#L34

deepmd/dpmodel/model/linear_model.py
anyangml: I thought about defining an interface `Weights`, but that means the user need to create a concrete `ZBLWeights` class before instantiating the `ZBLAtomicModel`. From a developer perspective, that does offer more flexibility, but from a user perspective, that may incur too much extra work.  I will use `str` for now to match the tf implementation.

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,326 @@
...
+    def get_rcut(self) -> float:
+        """Get the cut-off radius."""
+        return max(self.get_rcuts())
+
+    def get_rcuts(self) -> List[float]:
+        """Get the cut-off radius for each individual models in ascending order."""
+        return [model.get_rcut() for model in self.models]
+
+    def get_sel(self) -> List[int]:
+        return [max(self.get_sels())]
```
njzjz, line 73: Why does it contain only one element?

deepmd/pt/model/model/__init__.py
```diff
@@ -4,18 +4,57 @@
...
+    )
+    # pairtab
+    filepath = model_params["use_srtab"]
+    pt_model = PairTabModel(
+        filepath, model_params["descriptor"]["rcut"], model_params["descriptor"]["sel"]
+    )
+
+    rmin = model_params["sw_rmin"]
+    rmax = model_params["sw_rmax"]
+    return ZBLModel(dp_model, pt_model, rmin, rmax, "switch_by_softmin_pair_distance")
```
njzjz, line 55: Why not use the deserialize method?

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,326 @@
...
+            )["energy"]
+            for model, nl in zip(self.models, self.nlists_)
+        ]
+        weights = self._compute_weight()
+        self.atomic_bias = None
+        if self.atomic_bias is not None:
+            raise NotImplementedError("Need to add bias in a future PR.")
+        else:
+            fit_ret = {
+                "energy": sum([w * e for w, e in zip(weights, ener_list)]),
```
njzjz, line 155: `np.sum(np.array(weights, dtype=...) * np.array(ener_list, dtype=...))`

deepmd/pt/model/model/__init__.py
anyangml: not saving much efforts.

deepmd/dpmodel/model/linear_model.py
wanghan-iapcm: User interacts with deepmd-kit with json script.

The weight is not supposed to be implemented as a class, but a method for the derived method of a base linear model class.

wanghan-iapcm: I still find the current design strange: the `switch_by_softmin_pair_distance` style suppose that there are only two atomic models, and the order matters. but the constructor does not stop the user provide more than two models.

deepmd/dpmodel/model/pair_tab_model.py
wanghan-iapcm, line 1: how about changing pair_tab_model to pair_tab_atomic_model? to make the naming convention consistent with other atomic models

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,322 @@
...
+)
+from .dp_atomic_model import (
+    DPAtomicModel,
+)
+from .pair_tab_model import (
+    PairTabModel,
+)
+
+
+class LinearModel(BaseAtomicModel):
```
wanghan-iapcm, line 33: `LinearAtomicModel`?

deepmd/dpmodel/model/linear_model.py
```diff
@@ -0,0 +1,322 @@
...
+            This linear model should take a DPAtomicModel and a PairTable model.
+    """
+
+    def __init__(
+        self,
+        dp_model: DPAtomicModel,
+        zbl_model: PairTabModel,
+        sw_rmin: float,
+        sw_rmax: float,
+        weights="switch_by_softmin_pair_distance",
```
wanghan-iapcm, line 213: better to remove this arg. 
but we could not stop the user providing a weight to change the behavior of this class, i.e. not combined by the switch_with_softmin weight, but anything else. 

Again, I am against the current implementation of using a str to define the linear combination weights.

deepmd/dpmodel/model/linear_atomic_model.py
```diff
@@ -0,0 +1,288 @@
...
+            )["energy"]
+            for model, nl in zip(self.models, self.nlists_)
+        ]
+        self.weights = self._compute_weight()
+        self.atomic_bias = None
+        if self.atomic_bias is not None:
+            raise NotImplementedError("Need to add bias in a future PR.")
+        else:
+            fit_ret = {
+                "energy": sum(np.stack(ener_list) * np.stack(self.weights)),
```
njzjz, line 147: np.sum(..., axis=0) is much faster than sum for NumPy arrays

deepmd/pt/model/model/linear_atomic_model.py
```diff
@@ -0,0 +1,290 @@
...
+            )["energy"]
+            for model, nl in zip(self.models, self.nlists_)
+        ]
+        self.weights = self._compute_weight()
+        self.atomic_bias = None
+        if self.atomic_bias is not None:
+            raise NotImplementedError("Need to add bias in a future PR.")
+        else:
+            fit_ret = {
+                "energy": sum(torch.stack(ener_list) * torch.stack(self.weights)),
```
njzjz, line 153: torch.sum(..., 0)

deepmd/dpmodel/model/linear_atomic_model.py
```diff
@@ -0,0 +1,288 @@
...
+    def get_rcut(self) -> float:
+        """Get the cut-off radius."""
+        return max(self.get_rcuts())
+
+    def get_rcuts(self) -> List[float]:
+        """Get the cut-off radius for each individual models."""
+        return [model.get_rcut() for model in self.models]
+
+    def get_sel(self) -> List[int]:
+        return [model.get_nsel() for model in self.models]
```
wanghan-iapcm, line 69:
```suggestion
        return [max([model.get_nsel() for model in self.models])]
```
please check if it is correct.

deepmd/dpmodel/model/linear_atomic_model.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def distinguish_types(self) -> bool:
+        """If distinguish different types by sorting."""
+        return False
+
+    def get_rcut(self) -> float:
+        """Get the cut-off radius."""
+        return max(self.get_rcuts())
+
+    def get_rcuts(self) -> List[float]:
```
wanghan-iapcm, line 64:
```suggestion
    def get_models_rcuts(self) -> List[float]:
```

deepmd/dpmodel/model/linear_atomic_model.py
```diff
@@ -0,0 +1,288 @@
...
+        """Get the cut-off radius for each individual models."""
+        return [model.get_rcut() for model in self.models]
+
+    def get_sel(self) -> List[int]:
+        return [model.get_nsel() for model in self.models]
+
+    def get_original_sels(self) -> List[Union[int, List[int]]]:
```
wanghan-iapcm, line 71:
```suggestion
    def get_models_sels(self) -> List[Union[int, List[int]]]:
```

deepmd/dpmodel/model/linear_atomic_model.py
```diff
@@ -0,0 +1,288 @@
...
+                self.distinguish_type_list, raw_nlists, self.get_original_sels()
+            )
+        ]
+        ener_list = [
+            model.forward_atomic(
+                self.extended_coord,
+                extended_atype,
+                nl,
+                mapping,
+            )["energy"]
```
wanghan-iapcm, line 138: `fparam` and `aparam` should be passed to each `forward_atomic`

deepmd/dpmodel/model/linear_atomic_model.py
```diff
@@ -0,0 +1,288 @@
...
+            fit_ret = {
+                "energy": np.sum(np.stack(ener_list) * np.stack(self.weights), axis=0),
+            }  # (nframes, nloc, 1)
+        return fit_ret
+
+    def fitting_output_def(self) -> FittingOutputDef:
+        return FittingOutputDef(
+            [
+                OutputVariableDef(
+                    name="energy", shape=[1], reduciable=True, differentiable=True
```
wanghan-iapcm, line 155:
```suggestion
                    name="energy", shape=[1], reduciable=True, r_differentiable=True, c_differentiable=True
```
according to #3240

deepmd/dpmodel/model/linear_atomic_model.py
```diff
@@ -0,0 +1,288 @@
...
+        self.zbl_model = zbl_model
+
+        self.sw_rmin = sw_rmin
+        self.sw_rmax = sw_rmax
+        self.smin_alpha = smin_alpha
+
+    def serialize(self) -> dict:
+        return {
+            "dp_model": self.dp_model.serialize(),
+            "zbl_model": self.zbl_model.serialize(),
```
wanghan-iapcm, line 206: better to use the `serialize` of the `LinearAtomicModel` to serialize models.

deepmd/dpmodel/model/linear_atomic_model.py
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+    @classmethod
+    def deserialize(cls, data) -> "DPZBLLinearAtomicModel":
+        sw_rmin = data["sw_rmin"]
+        sw_rmax = data["sw_rmax"]
+        smin_alpha = data["smin_alpha"]
+
+        dp_model = DPAtomicModel.deserialize(data["dp_model"])
+        zbl_model = PairTabModel.deserialize(data["zbl_model"])
```
wanghan-iapcm, line 219: better to use the deserialize of the LinearAtomicModel to deserialize models.

deepmd/dpmodel/model/linear_atomic_model.py
```diff
@@ -0,0 +1,288 @@
...
+        )
+
+    def serialize(self) -> dict:
+        return {
+            "models": [model.serialize() for model in self.models],
+        }
+
+    @classmethod
+    def deserialize(cls, data) -> "LinearAtomicModel":
+        models = [DPAtomicModel.deserialize(model) for model in data["models"]]
```
wanghan-iapcm, line 167: it seems that the `deserialization` would not work if one model is `DPAtomicModel` and another is `PairTabModel` 

please refer to the serialize and deserialize methods used in below 
https://github.com/deepmodeling/deepmd-kit/blob/f1baf198b2c703637215942428d2ef24575b0e92/deepmd/dpmodel/model/dp_atomic_model.py#L122-L141

deepmd/pt/model/model/linear_atomic_model.py
```diff
@@ -0,0 +1,292 @@
...
+)
+from .model import (
+    BaseModel,
+)
+from .pairtab_atomic_model import (
+    PairTabModel,
+)
+
+
+class LinearAtomicModel(BaseModel, BaseAtomicModel):
```
wanghan-iapcm, line 39: if on the pt side the implementation shares most method as the np side, we may write class functional like `make_linear_atomic_model` to crease the both pt and np classes.

I would contribute to this refactorization if you need.

deepmd/pt/model/model/linear_atomic_model.py
wanghan-iapcm, line 1: the pt shares the same comments as the np linear atomic model

wanghan-iapcm (changes requested):

njzjz:
> I haven't carefully checked - but could you first add a test for JIT?
> 
> https://github.com/deepmodeling/deepmd-kit/blob/79f98cac8049b4ca22d2f42fe57bc715b5a65e01/source/tests/pt/model/test_jit.py#L34

Have jit passed?

deepmd/pt/model/model/pairtab_atomic_model.py
```diff
@@ -230,42 +234,31 @@ def _pair_tabulated_inter(
...
+        batch_indices = torch.arange(nlist.shape[0])[:, None, None]
+        expanded_batch_indices = batch_indices.expand_as(nlist)
+        neighbor_atoms = coords[expanded_batch_indices, nlist]
+        loc_atoms = coords[:, : nlist.shape[1], :]
+        loc_atoms = loc_atoms.unsqueeze(2).expand_as(neighbor_atoms)
+        pairwise_dr = loc_atoms - neighbor_atoms
+        # still have to add an epsilon to handle autograd, due to -1 in the nlist,
+        # they don't contribute to the final energy since they are masked.
+        pairwise_rr = torch.clamp(pairwise_dr.square().sum(-1), 1e-20).sqrt()
+        return pairwise_rr
```
anyangml, line 261: Still need to add a epsilon.

anyangml:
@njzjz Jit passed, check the UTs in this file `source/tests/pt/model/test_linear_atomic_model.py`, I believe all the data are kept when you jit a deserialized model.  Please double check this code block https://github.com/deepmodeling/deepmd-kit/blob/ccf63ceea587190f256e00ebeb5e1c7f65ea65ed/deepmd/pt/model/model/pairtab_atomic_model.py#L50-L84

source/tests/pt/model/test_permutation.py
wanghan-iapcm, line 1: missing permutation test for the ZBL model?

source/tests/pt/model/test_pairtab_atomic_model.py
```diff
@@ -114,9 +114,8 @@ def test_cross_deserialize(self):
         expected_result = self.model.forward_atomic(
             self.extended_coord, self.extended_atype, torch.from_numpy(self.nlist)
         )
-
         np.testing.assert_allclose(
-            result["energy"], expected_result["energy"], 0.0001, 0.0001
+            result["energy"], expected_result["energy"].detach().numpy(), 0.0001, 0.0001
```
wanghan-iapcm, line 118: you may use to_numpy_array in pt/utils/utils.py, which takes care of device transfer.

wanghan-iapcm (approved):

njzjz (approved):

PR #3213, opened 2024-02-01
refactor DeepEval

njzjz:
Implement high-level `DeepEval` and low-level `DeepEvalBackend`.

Ugly things:
(1) `DipoleChargeModifier` is not updated in this PR. Thus, it still inherits from the old `DeepEval`. (It should not inherit from `DeepEval`!!!)
(2) There are no unit tests or testing models for DeepGlobarPolar or DeepWFC.
(3) The shape of the atomic tensor looks different from forces and atomic virials.

TODO:
- [x] Add docs


deepmd/infer/deep_eval.py
```diff
@@ -0,0 +1,467 @@
...
+        >>> dp = DeepPotential('graph.pb')
+        >>> dp.eval_typeebd()
+        """
+        return self.deep_eval.eval_typeebd()
+
+    def _standard_input(self, coords, cells, atom_types, fparam, aparam, mixed_type):
+        coords = np.array(coords)
+        if cells is not None:
+            cells = np.array(cells)
+        atom_types = np.array(atom_types)
```
wanghan-iapcm, line 444:
```suggestion
        atom_types = np.array(atom_types, dtype=np.int32)
```

deepmd/pt/infer/deep_eval.py
```diff
@@ -64,29 +101,55 @@ def __init__(
...
+
+    def get_has_efield(self):
+        """Check if the model has efield."""
+        return False
+
+    def get_ntypes_spin(self):
+        """Get the number of spin atom types of this model."""
+        return 0
+
+    _OUTDEF_DP2PT: ClassVar[dict] = {
```
wanghan-iapcm, line 145: should be placed in backend indepdent infer?

deepmd/pt/infer/deep_eval.py
```diff
@@ -109,9 +207,19 @@ def eval(
...
-        return self._eval_func(self._eval_model, numb_test, natoms)(
+        out = self._eval_func(self._eval_model, numb_test, natoms)(
             coords, cells, atom_types, atomic
         )
+        return dict(
+            zip(
+                [
+                    x.name
+                    for x in self.output_def.var_defs.values()
+                    if atomic or "_redu" in x.name or "_derv_r" in x.name
```
wanghan-iapcm, line 218: _derv_r is atomic?

wanghan-iapcm: what is the condition actually is ? provide a method to get the condition out of a model output def?

deepmd/pt/infer/deep_eval.py
```diff
@@ -195,54 +303,33 @@ def _eval_model(
...
+        return tuple(results)
 
-    def get_dim_fparam(self) -> int:
-        """Get the number (dimension) of frame parameters of this DP."""
-        return 0
-
-    def get_dim_aparam(self) -> int:
-        """Get the number (dimension) of atomic parameters of this DP."""
-        return 0
+    def _get_output_shape(self, name, nframes, natoms, shape):
```
wanghan-iapcm, line 318: suggest to clarify all the conditions needed by this method.

deepmd/pt/infer/deep_eval.py
```diff
@@ -195,54 +303,33 @@ def _eval_model(
...
-    def get_ntypes(self) -> int:
-        """Get the number of atom types of this model."""
-        return len(self.type_map)
-
-    def get_type_map(self) -> List[str]:
-        """Get the type map (element name of the atom types) of this model."""
-        return self.type_map
+        results = []
+        for ii, odef in enumerate(self.output_def.var_defs.values()):
+            if not atomic and "_redu" not in odef.name and "_derv_r" not in odef.name:
```
wanghan-iapcm, line 309: same condition (~) as that used in `eval`

wanghan-iapcm (changes requested):

deepmd/pt/infer/deep_eval.py
njzjz: Currently they have different names, atom_virial vs atomic_virial

deepmd/infer/deep_eval.py
```diff
@@ -0,0 +1,467 @@
...
+        Raises
+        ------
+        KeyError
+            If the model does not enable type embedding.
+        """
+        raise NotImplementedError
+
+    def _check_distinguished_types(self, atom_types: np.ndarray) -> bool:
+        """Check if atom types of each frame."""
+        return np.all(np.equal(atom_types, atom_types[0]))
```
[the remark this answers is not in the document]
njzjz: Here the shape of atom_type is nframes * natoms, and it checks if all frames have the same atomic types. Only se_atten supports different atomic types.

deepmd/pt/infer/deep_eval.py
njzjz: When atomic is True, all outputs are required; otherwise, only energy, force, and virial are required.

deepmd/infer/deep_eval.py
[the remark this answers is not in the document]
njzjz: I may still use mixed_type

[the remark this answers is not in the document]
wanghan-iapcm: so what is the desired input type for dpa1 ? faked types (all 0) or real types?

deepmd/pt/infer/deep_eval.py
wanghan-iapcm: I would suggest unifying the names to **atomic_xxx**. what do you think?

deepmd/pt/infer/deep_eval.py
wanghan-iapcm: map all outputs and pop the atomic energy and virial when `atomic==False` ? 
or provide a method like `get_reduced_and_force`

deepmd/infer/deep_eval.py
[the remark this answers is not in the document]
njzjz: Real types.

deepmd/pt/infer/deep_eval.py
njzjz: `o_atom_virial` has been hardcoded in all existing TF models, though.

wanghan-iapcm: then we have to made them **atom_xxx***...

deepmd/infer/deep_eval.py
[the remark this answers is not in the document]
wanghan-iapcm: I still find it a little strange. In the case of se_atten and you have ntypes > 1, distinguished-type nlists will be made, although it does not affect the final result, one is granted to use indistinguished-type nlists

deepmd/pt/infer/deep_eval.py
njzjz: I added a method called `_get_request_defs` and directly passed `request_defs` (instead of `atomic`) to the inner method.

deepmd/pt/infer/deep_eval.py
njzjz: https://github.com/deepmodeling/deepmd-kit/blob/b875ea8f6661b6e1567537ead7e2b4a8b14ea113/source/tests/infer/se_atten_no_atten_1.pbtxt#L12996-L12997

njzjz: > then we have to made them **atom_xxx***...

It will also affect PT models that have been saved. @CaRoLZhangxy also needs to change the name in #3162. Perhaps it should be conducted in a separated PR?

wanghan-iapcm: we change the convention in the PT models. let make it in a separate PR.

deepmd/pt/infer/deep_eval.py
```diff
@@ -195,54 +303,33 @@ def _eval_model(
...
+                return [nframes, *shape, 1]
+        else:
+            if "_derv_c" in name:
+                return [nframes, *shape[:-2], natoms, 9]
+            elif "_derv_r" in name:
+                return [nframes, *shape[:-1], natoms, 3]
+            else:
+                # Something wrong here?
+                # return [nframes, *shape, natoms, 1]
+                return [nframes, natoms, *shape, 1]
```
wanghan-iapcm, line 357: could you fix the wrong shape in a new PR?

deepmd/pt/infer/deep_eval.py
```diff
@@ -195,54 +303,33 @@ def _eval_model(
...
-    def get_dim_fparam(self) -> int:
-        """Get the number (dimension) of frame parameters of this DP."""
-        return 0
-
-    def get_dim_aparam(self) -> int:
-        """Get the number (dimension) of atomic parameters of this DP."""
-        return 0
+    def _get_output_shape(self, name, nframes, natoms, shape):
+        if "_redu" in name:
+            if "_derv_c" in name:
```
wanghan-iapcm, line 340: do you need methods like `is_derviative_wrt_cell` ?

deepmd/infer/deep_eval.py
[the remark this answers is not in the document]
njzjz: This method doesn't check whether ntypes > 1. Instead, it checks types in frame 0 == types in frame 1 == ... == types in frame Nframes.

If False, the TF backend will generate a special mesh implemented in #2481, and only `se_atten` can handle it.
If True, all descriptors from the TF backend can accept it.

I think the PT backend doesn't have this limitation, but need to confirm.

deepmd/pt/infer/deep_eval.py
njzjz: The old TF UT failed if the shape is `[nframes, *shape, natoms, 1]`. Currently, I am not sure what the correct result is, i.e. whether the result matches its shape.

deepmd/pt/infer/deep_eval.py
njzjz: Yes, it is really ugly to check the name...

njzjz: Indeed, I need some methods that can be used to distinguish each of them.

njzjz: Perhaps using [`enum`](https://docs.python.org/3/library/enum.html)?

deepmd/infer/deep_eval.py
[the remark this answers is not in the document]
wanghan-iapcm: get it. thanks a lot for the explanation. 

PT does not have the limitation. but when building with `distinguished_types`, the efficiency is lower.  you may check
https://github.com/deepmodeling/deepmd-kit/blob/cd77429dcbdaf42892ac3319b6fb72b9a2879a65/source/tests/pt/test_nlist.py#L52-L57

deepmd/pt/infer/deep_eval.py
wanghan-iapcm: I agree.

wanghan-iapcm (approved):
LGTM

deepmd/tf/infer/deep_eval.py
```diff
@@ -29,7 +58,1040 @@
...
+        else:
+            coords = np.reshape(np.array(coords), [-1, natoms * 3])
+        nframes = coords.shape[0]
+        return natoms, nframes
+
+    def eval(
+        self,
+        coords: np.ndarray,
+        cells: np.ndarray,
+        atom_types: List[int],
```
iProzd, line 684: Here `atom_types` should be np.ndarray? Because it's after DeepEvalWrapper's standardization.

deepmd/infer/deep_eval.py
```diff
@@ -0,0 +1,497 @@
...
+            else:
+                raise NotImplementedError("Unsupported backend: " + str(backend))
+        return super().__new__(cls)
+
+    @abstractmethod
+    def eval(
+        self,
+        coords: np.ndarray,
+        cells: np.ndarray,
+        atom_types: List[int],
```
iProzd, line 110: Here atom_types should be np.ndarray? Because it's after DeepEvalWrapper's standardization.

iProzd (approved):

PR #3220, opened 2024-02-02
pt: apply global user set precision to pt

njzjz:


njzjz:
We may also rename other environmental variables to the name with a prefix of `DP_`. @iProzd could you point out which environmental variables should be documented?

deepmd/pt/utils/env.py
```diff
@@ -5,14 +5,19 @@
...
+    GLOBAL_NP_FLOAT_PRECISION,
     get_default_nthreads,
     set_default_nthreads,
 )
 
-PRECISION = os.environ.get("PRECISION", "float64")
-GLOBAL_NP_FLOAT_PRECISION = getattr(np, PRECISION)
-GLOBAL_PT_FLOAT_PRECISION = getattr(torch, PRECISION)
-GLOBAL_ENER_FLOAT_PRECISION = getattr(np, PRECISION)
+numpy_to_torch_dtype_dict = {
```
wanghan-iapcm, line 14: we may implement it by reversing dicts from str->np prec in dpmodel and str->torch prec.

njzjz: np->str can be simply written as `np.dtype(GLOBAL_NP_FLOAT_PRECISION).name`. I'll reuse `str->torch`.

wanghan-iapcm (approved):

PR #3223, opened 2024-02-03
c: fix all memory leaks; add sanitizer checks

njzjz:
Fix #3045. All memory leaks have been fixed!


wanghan-iapcm (approved):

wanghan-iapcm:
cherry pick to v2.2 ?

njzjz:
> cherry pick to v2.2 ?

Agree.

PR #3226, opened 2024-02-04
pt: rename atomic_virial to atom_virial in the model output

njzjz:
To be consistent with TF, as discussed in https://github.com/deepmodeling/deepmd-kit/pull/3213#discussion_r1477183841. Old PT models are expected to be incompatible.


njzjz:
cc: @CaRoLZhangxy @iProzd

wanghan-iapcm (approved):

PR #3228, opened 2024-02-05
add category property to OutputVariableDef

njzjz:


deepmd/dpmodel/output_def.py
```diff
@@ -107,6 +111,34 @@ def __call__(
...
+    """Derivative w.r.t. coordinates."""
+    DERV_C = 4
+    """Derivative w.r.t. cell."""
+
+
+class OutputVariableCategory(IntEnum):
+    """Defines the category of the output variable."""
+
+    OUT = OutputVariableOperation.NONE
+    """Output variable."""
```
wanghan-iapcm, line 131:
```suggestion
    """Output variable. (e.g. atom energy)"""
```

deepmd/dpmodel/output_def.py
```diff
@@ -107,6 +111,34 @@ def __call__(
...
+    REDU = OutputVariableOperation.REDU
+    """Reduced output variable."""
```
wanghan-iapcm, line 133:
```suggestion
    """Reduced output variable. (e.g. system energy)"""
```

deepmd/dpmodel/output_def.py
```diff
@@ -107,6 +111,34 @@ def __call__(
...
+    DERV_R = OutputVariableOperation.DERV_R
+    """Derivative w.r.t. coordinates."""
```
wanghan-iapcm, line 135:
```suggestion
    """Negative derivative w.r.t. coordinates. (e.g. force)"""
```

deepmd/dpmodel/output_def.py
```diff
@@ -107,6 +111,34 @@ def __call__(
...
+    DERV_C = OutputVariableOperation.DERV_C
+    """Derivative w.r.t. cell."""
+    DERV_C_REDU = OutputVariableOperation.DERV_C | OutputVariableOperation.REDU
+    """Reduced derivative w.r.t. cell."""
```
wanghan-iapcm, line 139:
```suggestion
    """Virial, the transposed negative gradient with cell tensor times cell tensor, see eq 40 JCP 159, 054801 (2023). """
```

deepmd/dpmodel/output_def.py
```diff
@@ -261,9 +295,15 @@ def do_reduce(
...
+            assert vv.category & OutputVariableOperation.REDU.value == 0
             rk = get_reduce_name(kk)
             def_redu[rk] = OutputVariableDef(
-                rk, vv.shape, reduciable=False, differentiable=False, atomic=False
+                rk,
+                vv.shape,
+                reduciable=False,
+                differentiable=False,
+                atomic=False,
+                category=vv.category | OutputVariableOperation.REDU.value,
```
njzjz, line 306: Torchscript is a shit! I have to use `OutputVariableOperation.REDU.value` instead of `OutputVariableOperation.REDU` to let the jit compile.

deepmd/dpmodel/output_def.py
```diff
@@ -275,19 +315,23 @@ def do_derivative(
     def_derv_c: Dict[str, OutputVariableDef] = {}
     for kk, vv in def_outp_data.items():
         if vv.differentiable:
+            assert vv.category & OutputVariableOperation.DERV_R.value == 0
+            assert vv.category & OutputVariableOperation.DERV_C.value == 0
```
wanghan-iapcm, line 319: we may take twice derivative over r in the future to compute the hessian. we may want to consider such extension.

deepmd/dpmodel/output_def.py
wanghan-iapcm: introduce something like 
`apply_operation(vv, OutputVariableOperation.XXX)` and `check_operation_applied(vv, OutputVariableOperation.XXX)`  to solve the issue?

deepmd/dpmodel/output_def.py
```diff
@@ -289,21 +293,69 @@ def get_deriv_name(name: str) -> Tuple[str, str]:
...
+
+    Returns
+    -------
+    int
+        The new category of the variable definition.
+    """
+    return var_def.category | op.value
+
+
+def check_operation_applied(
```
wanghan-iapcm, line 314: I am confused by the behavior of the function. could you please provide uts so I can better understand?

deepmd/dpmodel/output_def.py
```diff
@@ -289,21 +293,69 @@ def get_deriv_name(name: str) -> Tuple[str, str]:
...
+        The variable definition.
+    op : OutputVariableOperation
+        The operation to be checked.
+
+    Returns
+    -------
+    bool
+        True if the operation has been applied, False otherwise.
+    """
+    if op in (OutputVariableOperation.DERV_REDU, OutputVariableOperation.DERV_C):
```
wanghan-iapcm, line 331: `(OutputVariableOperation.DERV_REDU` is not defined.

deepmd/dpmodel/output_def.py
```diff
@@ -307,7 +307,23 @@ def apply_operation(var_def: OutputVariableDef, op: OutputVariableOperation) ->
...
+        and exceed the maximum limitation.
     """
+    if op == OutputVariableOperation.REDU or op == OutputVariableOperation.DERV_C:
+        if check_operation_applied(var_def, op):
+            raise ValueError(f"operation {op} has been applied")
+    elif op == OutputVariableOperation.DERV_R:
+        if check_operation_applied(var_def, OutputVariableOperation.DERV_R):
+            op = OutputVariableOperation.SEC_DERV_R
+            if check_operation_applied(var_def, OutputVariableOperation.SEC_DERV_R):
+                raise ValueError(f"operation {op} has been applied twice")
```
wanghan-iapcm, line 324:
```suggestion
    elif op == OutputVariableOperation.DERV_R:
        if check_operation_applied(var_def, OutputVariableOperation.DERV_R):
            raise ValueError(f"operation {op} has been applied")
    elif op = OutputVariableOperation.SEC_DERV_R:
        if not check_operation_applied(var_def, OutputVariableOperation.DERV_R):
            raise ValueError(f"operation OutputVariableOperation.DERV_R should be applied before {op}")
        if check_operation_applied(var_def, OutputVariableOperation.SEC_DERV_R):
            raise ValueError(f"operation {op} has been applied")
```

source/tests/common/dpmodel/test_output_def.py
```diff
@@ -103,6 +106,101 @@ def test_model_output_def(self):
...
+        # raise ValueError
+        with self.assertRaises(ValueError):
+            apply_operation(md["energy_redu"], OutputVariableOperation.REDU)
+        with self.assertRaises(ValueError):
+            apply_operation(md["energy_derv_c"], OutputVariableOperation.DERV_C)
+        with self.assertRaises(ValueError):
+            apply_operation(md["energy_derv_c_redu"], OutputVariableOperation.REDU)
+        # hession
+        hession_cat = apply_operation(
+            md["energy_derv_r"], OutputVariableOperation.DERV_R
```
wanghan-iapcm, line 189: what is the convention of getting hessian?
1. apply_operation(apply_operation(energy, derv_r), derv_r)
2. apply_operation(apply_operation(energy, derv_r), sec_derv_r)

deepmd/dpmodel/output_def.py
njzjz: Do you think we should explicitly pass SEC_DERV to this method? Then we need to check whether it has been applied derivatives in `do_derivative`.

source/tests/common/dpmodel/test_output_def.py
njzjz: (1) is implicit, and (2) is explicit. But even for the explicit case, we still have to check in `do_derivative`.

njzjz: I don't oppose the explicit case...

njzjz: Or, does `do_derivative` only work for the original atomic energy/tensor?

wanghan-iapcm: The force derivative works for any tensor. while the cell derivative works only on the force does not make sense. 
It seems that we should introduce differentiable_r and differentiable_c instead of differentiable. 
I would prefer the implicit case, the derv_r can be use for both the first and second order derivatives.

deepmd/dpmodel/output_def.py
wanghan-iapcm: It seems that the `SEC_DERV` operation should not be exposed to users..

wanghan-iapcm (approved):

PR #3240, opened 2024-02-06
support separate r_differentiable and c_differentiable

wanghan-iapcm:
solve some virial shape issue: should be [9] rather than [3,3]

### DIFF
--- a/deepmd/dpmodel/output_def.py
+++ b/deepmd/dpmodel/output_def.py
@@ -68,9 +68,11 @@ def model_check_output(cls):
                 if dd.reduciable:
                     rk = get_reduce_name(kk)
                     check_var(ret[rk], self.md[rk])
-                if dd.differentiable:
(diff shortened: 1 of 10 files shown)


deepmd/dpmodel/output_def.py
```diff
@@ -371,21 +383,26 @@ def do_derivative(
...
-        if vv.differentiable:
-            rkr, rkc = get_deriv_name(kk)
+        rkr, rkc = get_deriv_name(kk)
+        if vv.r_differentiable:
             def_derv_r[rkr] = OutputVariableDef(
                 rkr,
                 vv.shape + [3],  # noqa: RUF005
                 reduciable=False,
-                differentiable=False,
+                r_differentiable=False,
```
njzjz, line 392: Should it be True for the hession? Or change it in the future?

PR #3233, opened 2024-02-05
breaking: pt: remove data stat from model init

iProzd:
This PR:
- clean up the data stat process from model init.
- fix import error and skipped uts.

Please note that this code PR is just an initial cleanup and refinement of the data stat, and a more detailed design of the data stat will be completed in the next PR.

PR #3245, opened 2024-02-07
breaking: pt: remove data stat from model init

iProzd:
Restore #3233 with resolved conflicts and conversations.

This PR clean up the data stat process from model init.

Please note that this code PR is just an initial cleanup and refinement of the data stat, and a more detailed design of the data stat will be completed in the next PR:

- independent data stat from dataloader
- data stat support for hybrid descriptors


deepmd/dpmodel/model/make_base_atomic_model.py
```diff
@@ -84,6 +84,32 @@ def serialize(self) -> dict:
         def deserialize(cls):
             pass
 
+        def compute_or_load_stat(
```
wanghan-iapcm, line 87: is it supposed to be an interface? if so make it an abstract method

deepmd/dpmodel/model/make_base_atomic_model.py
```diff
@@ -84,6 +84,32 @@ def serialize(self) -> dict:
...
+            ----------
+            type_map
+                Mapping atom type to the name (str) of the type.
+                For example `type_map[1]` gives the name of the type 1.
+            sampled
+                The sampled data frames from different data systems.
+            stat_file_path
+                The path to the statistics files.
+            """
+            pass
```
wanghan-iapcm, line 111: if not an abstract method, raise NotImplementedError

deepmd/pt/model/descriptor/descriptor.py
```diff
@@ -56,15 +60,127 @@ class SomeDescript(Descriptor):
...
+        Usually use the combination of descriptor name, rcut, rcut_smth and sel as the statistic file name.
+        """
+        if cls is not Descriptor:
+            raise NotImplementedError("get_stat_name is not implemented!")
         descrpt_type = config["type"]
-        return Descriptor.__plugins.plugins[descrpt_type].get_stat_name(config)
+        return Descriptor.__plugins.plugins[descrpt_type].get_stat_name(config, ntypes)
 
     @classmethod
     def get_data_process_key(cls, config):
```
wanghan-iapcm, line 74: very bad design of passing-in a dict. explicitly passing the parameters is preferred.

deepmd/pt/model/descriptor/descriptor.py
```diff
@@ -56,15 +60,127 @@ class SomeDescript(Descriptor):
...
+        Get the keys for the data preprocess.
+        Usually need the information of rcut and sel.
+        TODO Need to be deprecated when the dataloader has been cleaned up.
+        """
+        if cls is not Descriptor:
+            raise NotImplementedError("get_data_process_key is not implemented!")
         descrpt_type = config["type"]
         return Descriptor.__plugins.plugins[descrpt_type].get_data_process_key(config)
 
+    def get_data_stat_key(self):
```
wanghan-iapcm, line 85: should be abstract method?

deepmd/pt/model/descriptor/descriptor.py
```diff
@@ -56,15 +60,127 @@ class SomeDescript(Descriptor):
...
+        """
+        Get the keys for the data statistic of the descriptor.
+        Return a list of statistic names needed, such as "sumr", "suma" or "sumn".
+        """
+        raise NotImplementedError("get_data_stat_key is not implemented!")
+
+    def set_stats(
```
wanghan-iapcm, line 92: `compute_or_load_data_stats`?

deepmd/pt/model/descriptor/descriptor.py
```diff
@@ -162,7 +278,7 @@ def compute_input_stats(self, merged):
         pass
 
     @abstractmethod
-    def init_desc_stat(self, sumr, suma, sumn, sumr2, suma2):
+    def init_desc_stat(self, stat_dict):
         """Initialize the model bias by the statistics."""
```
wanghan-iapcm, line 282: model bias? wrong doc

deepmd/pt/model/task/ener.py
```diff
@@ -200,6 +203,31 @@ def __getitem__(self, key):
...
+        mixed_type = "real_natoms_vec" in merged[0]
+        if mixed_type:
+            input_natoms = [item["real_natoms_vec"] for item in merged]
+        else:
+            input_natoms = [item["natoms"] for item in merged]
+        tmp = compute_output_bias(energy, input_natoms)
+        bias_atom_e = tmp[:, 0]
+        return {"bias_atom_e": bias_atom_e}
+
+    def init_fitting_stat(self, result_dict):
```
wanghan-iapcm, line 224: better not passing dicts. explicitly tell what does this method need.

deepmd/pt/model/task/ener.py
```diff
@@ -394,6 +422,16 @@ def __init__(
             **kwargs,
         )
 
+    @classmethod
+    def get_stat_name(cls, config, ntypes):
```
wanghan-iapcm, line 426: not a good idea of passing dict.

wanghan-iapcm (changes requested):
Are the data stats of descriptor and fitting tested by UTs?

deepmd/pt/model/model/model.py
```diff
@@ -1,148 +1,8 @@
...
-
-log = logging.getLogger(__name__)
-
 
 class BaseModel(torch.nn.Module):
     def __init__(self):
         """Construct a basic model for different tasks."""
         super().__init__()
-
-    def compute_or_load_stat(
```
njzjz, line 23: Should keep this method as the interface, if not all `BaseModel` is `DPAtomicModel`.

deepmd/pt/entrypoints/main.py
```diff
@@ -122,49 +125,39 @@ def prepare_trainer_input_single(
...
+                    default_stat_file_name_descrpt = Descriptor.get_stat_name(
+                        model_params_single["descriptor"],
+                        len(model_params_single["type_map"]),
+                    )
+                    stat_file["descriptor"] = default_stat_file_name_descrpt
+                if "fitting_net" in model_params_single:
+                    default_stat_file_name_fitting = Fitting.get_stat_name(
+                        model_params_single["fitting_net"],
+                        len(model_params_single["type_map"]),
+                    )
```
njzjz, line 144: Should not written here, but implement `Model.get_stat_name`.

deepmd/pt/model/descriptor/descriptor.py
```diff
@@ -56,15 +60,127 @@ class SomeDescript(Descriptor):
...
+        Parameters
+        ----------
+        result_dict
+            The dictionary of statistics results.
+        stat_file_path
+            The path to the statistics file(s).
+        """
+        if not isinstance(stat_file_path, list):
+            log.info(f"Saving stat file to {stat_file_path}")
+            np.savez_compressed(stat_file_path, **result_dict)
```
njzjz, line 139: I prefer saving into an HDF5 file, so more complex models can be handled and more metadata can be saved. It doesn't need to be refactored in this PR, but can be in the future.

iProzd:
> Are the data stats of descriptor and fitting tested by UTs?

source/tests/pt/test_stat.py has the UTs to compare the consistence of data stats between tf and pt, on both descriptor and fitting but only on se_a model.

deepmd/pt/model/model/model.py
iProzd: Yes, I agree, but the `compute_or_load_stat` here was previously designed for models with descriptors, while `BaseModel` may not necessarily have descriptors. However, `DPAtomicModel` definitely has descriptors, so this function has been moved to DPAtomicModel. But I also agree that it's good to keep a parent class function interface here, which makes it convenient for models like `PairTabModel` to inherit.

deepmd/dpmodel/model/make_base_atomic_model.py
iProzd: Maybe `BaseAtomicModel` is a better choice to have `compute_or_load_stat` as the interface, since both`DPAtomicModel` and `PairTabModel` inherit from it. But it's maybe not an abstract method, because numpy doesn't need this method. So maybe here raise NotImplementedError is better.

deepmd/pt/model/descriptor/hybrid.py
```diff
@@ -153,23 +153,37 @@
...
+        return {
+            "sumr": sumr,
+            "suma": suma,
+            "sumn": sumn,
+            "sumr2": sumr2,
+            "suma2": suma2,
+        }
 
-    def init_desc_stat(self, sumr, suma, sumn, sumr2, suma2):
+    def init_desc_stat(self, stat_dict):
```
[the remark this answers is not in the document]
njzjz: Since the parent method allows `**kwargs`, you may need to allow `**kwargs` for its child method as well.

njzjz (approved):

deepmd/dpmodel/model/make_base_atomic_model.py
wanghan-iapcm: The problem is that all backends are supposed to use the stat in exactly the same way, thus all the backends are assumed to provide the same interface. I support design it as abstract method of  some base class, but may not be `BaseAtomicModel`, because it does not necessarily have a descriptor, thus the statistics is different from those needs descriptor statistics. 

It should be abstract method, because the stat interfaces will be used in the same way cross descriptors and fittings. numpy may pass null stats (mean 0 std 1).

deepmd/dpmodel/descriptor/make_base_descriptor.py
```diff
@@ -69,15 +69,13 @@ def distinguish_types(self) -> bool:
             """
             pass
 
-        @abstractmethod
         def compute_input_stats(self, merged):
```
wanghan-iapcm, line 72: should be abstract method. the developer must provide the implementation of this interface.

deepmd/dpmodel/descriptor/make_base_descriptor.py
```diff
@@ -69,15 +69,13 @@ def distinguish_types(self) -> bool:
...
             """Update mean and stddev for descriptor elements."""
-            pass
+            raise NotImplementedError
 
-        @abstractmethod
-        def init_desc_stat(self, sumr, suma, sumn, sumr2, suma2):
+        def init_desc_stat(self, **kwargs):
```
wanghan-iapcm, line 76: should be abstract method. the developer must provide the implementation of this interface.

deepmd/dpmodel/fitting/make_base_fitting.py
```diff
@@ -52,6 +52,14 @@ def fwd(
             """Calculate fitting."""
             pass
 
+        def compute_output_stats(self, merged):
```
wanghan-iapcm, line 55: should be abstract method

deepmd/dpmodel/fitting/make_base_fitting.py
```diff
@@ -52,6 +52,14 @@ def fwd(
...
+            """Update the output bias for fitting net."""
+            raise NotImplementedError
+
+        def init_fitting_stat(self, **kwargs):
```
wanghan-iapcm, line 59: should be abstract method

deepmd/dpmodel/descriptor/make_base_descriptor.py
wanghan-iapcm: please not that the pt backend should preserve the same interface as the dpmodel.

wanghan-iapcm (changes requested):

deepmd/pt/model/descriptor/descriptor.py
iProzd: This has already been changed to `compute_or_load_stat`.

deepmd/dpmodel/model/make_base_atomic_model.py
iProzd: `compute_or_load_stat` method has been changed into `BaseModel` now, which is a base class of pytorch model, but not as an abstract method.
The inheritance relationship between those class is as following:
DPAtomicModel(BaseModel, BaseAtomicModel)

iProzd: `compute_or_load_stat` method has been changed in to `BaseModel`, which is a base class of pytorch model, but as a method with `NotImplement` error and not as an abstract method. 

The inheritance relationship of these classed is as following:
`DPAtomicModel`(`BaseModel`, `BaseAtomicModel`(torch.Tensor))
`PairTabModel`(`BaseModel`, `BaseAtomicModel`(torch.Tensor))

Now only `DPAtomicModel` has implemented the `compute_or_load_stat` method with descriptors and fittings.

I think this way may allow users implement different stat methods for model such as `PairTabModel` and not implement the numpy stat interface.

But now in this PR may be not suitable directly as an abstract method in `BaseModel`? (Or I can make it an abstract method and raise `NotImplement` in `PairTabModel` waiting for @anyangml fix?).

wanghan-iapcm (approved):
Per discussion, further refactorization will be provided by a future PR.

PR #3247, opened 2024-02-08
add `get_type_map` method to model; export model methods 

njzjz:


deepmd/dpmodel/model/linear_atomic_model.py
```diff
@@ -62,6 +62,10 @@ def get_rcut(self) -> float:
         """Get the cut-off radius."""
         return max(self.get_model_rcuts())
 
+    def get_type_map(self) -> Optional[List[str]]:
+        """Get the type map."""
+        return self.models[0].get_type_map()
```
wanghan-iapcm, line 67: should be considered if the models does not have the same typemap, then how to handle. 
1. does not accept and raise an error
2. building a new typemap on top of all the models and correctly map the types to submodels?

deepmd/dpmodel/model/pairtab_atomic_model.py
```diff
@@ -82,6 +82,9 @@ def fitting_output_def(self) -> FittingOutputDef:
     def get_rcut(self) -> float:
         return self.rcut
 
+    def get_type_map(self) -> Optional[List[str]]:
+        return None
```
wanghan-iapcm, line 86: should have a type map i guess.

wanghan-iapcm (changes requested):

deepmd/dpmodel/model/linear_atomic_model.py
njzjz: There is no interface for users to assign a type map...

deepmd/dpmodel/model/pairtab_atomic_model.py
njzjz: There is no interface for users to assign a type map for pairtab atomic model...

deepmd/dpmodel/model/linear_atomic_model.py
wanghan-iapcm: I would suggest raising `NotImplementedError` to indicate the model does not provide the interface at the moment.

deepmd/dpmodel/model/pairtab_atomic_model.py
wanghan-iapcm: The same suggestion

deepmd/dpmodel/model/linear_atomic_model.py
njzjz: Done in e8ceb781b7bb77e5e7892bc87760d32b426a4180. It still needs to be implemented, otherwise `DeepEval` won't work.

wanghan-iapcm (approved):

PR #3261, opened 2024-02-11
breaking: pt: remove data preprocess from data stat

iProzd:
This PR:


deepmd/pt/model/descriptor/hybrid.py
```diff
@@ -103,6 +103,14 @@ def get_dim_in(self) -> int:
     def get_dim_emb(self):
         return self.dim_emb
 
+    def distinguish_types(self) -> bool:
+        """Returns if the descriptor requires a neighbor list that distinguish different
+        atomic types or not.
+        """
+        return True in [
```
njzjz, line 110: Use `any(...)`

deepmd/pt/model/descriptor/se_a.py
```diff
@@ -36,6 +37,11 @@
 from deepmd.pt.model.network.network import (
     TypeFilter,
 )
+from deepmd.pt.utils.nlist import (
+    process_input,
+)
+
+log = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
njzjz: Duplicated line

CaRoLZhangxy (approved):

wanghan-iapcm (changes requested):

njzjz (approved):

deepmd/pt/utils/dataset.py
```diff
@@ -694,13 +532,8 @@ def single_preprocess(self, batch, sid):
                             f"Add noise times beyond max tries {self.max_fail_num}!"
                         )
                     continue
-                batch["atype"] = masked_type
+                batch["type"] = masked_type
                 batch["coord"] = noised_coord
```
Chengqian-Zhang, line 536: My understanding is that this part needs to be removed completely, and the placement of the addition of noise needs to be discussed later.

Chengqian-Zhang (approved):

deepmd/pt/utils/dataset.py
iProzd: This interface and process code may leave for you to clean up in an individual PR.

wanghan-iapcm (approved):

PR #3262, opened 2024-02-11
Implement hessian autodiff calculation

wanghan-iapcm:
restrictions:
- cannot jit
- only the `forward_common` interface has its hessian calculation. not for `forward_common_lower`.
- may give nan when nall == nloc. specifically when nloc==1

also fix bug in pt: transform_output. The output shape will be wrong when the dimension of output variable is larger than 1.

### DIFF
--- a/deepmd/pt/model/model/make_hessian_model.py
+++ b/deepmd/pt/model/model/make_hessian_model.py
@@ -1,23 +1,25 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+import copy
 from typing import (
     Dict,
+    List,
     Optional,
+    Union,
 )
 
 import torch
-import copy
 
 from deepmd.dpmodel import (
-    ModelOutputDef,
-    FittingOutputDef,
+    get_hessian_name,
 )
 
 
 def make_hessian_model(T_Model):
-    """Make a model that can compute Hessian. 
-    
+    """Make a model that can compute Hessian.
+
     LIMITATION: this model is not jitable due to the restrictions of torch jit script.
 
-    LIMITATION: only the hessian of `forward_common` is available. 
+    LIMITATION: only the hessian of `forward_common` is available.
 
     Parameters
     ----------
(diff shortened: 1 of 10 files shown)


deepmd/pt/model/model/make_hessian_model.py
```diff
@@ -0,0 +1,217 @@
...
+                    hess_keys.append(kk)
+            # result dict init by empty lists
+            res = {get_hessian_name(kk): [] for kk in hess_keys}
+            # loop over variable
+            for kk in hess_keys:
+                vdef = fdef[kk]
+                vshape = vdef.shape
+                vsize = 1
+                for ii in vshape:
+                    vsize *= ii
```
njzjz, line 144: `vsize = math.prod(vshape)` or `vsize = np.prod(vshape)`

deepmd/pt/model/model/make_hessian_model.py
```diff
@@ -0,0 +1,217 @@
...
+                # loop over frames
+                for ii in range(nf):
+                    icoord = coord[ii]
+                    iatype = atype[ii]
+                    ibox = box[ii] if box is not None else None
+                    ifparam = fparam[ii] if fparam is not None else None
+                    iaparam = aparam[ii] if aparam is not None else None
+                    # loop over all components
+                    for idx in range(vsize):
```
njzjz, line 153: Will it be slow when using `for ... in` twice?

wanghan-iapcm: I haven't benchmarked, but i guess there should not be a significant difference.

deepmd/pt/model/model/make_hessian_model.py
wanghan-iapcm: for the sake of jit... now we use a more elegant way

njzjz (approved):

iProzd (approved):

PR #3263, opened 2024-02-12
support TF se_e2_a serialization; add a common test fixture to compare TF, PT, and DP models

njzjz:
Some codes are split from #2987.

### DIFF
--- a/source/tests/consistent/common.py
+++ b/source/tests/consistent/common.py
@@ -1,5 +1,8 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
+import itertools
 import os
+import sys
+import unittest
 from abc import (
     ABC,
     abstractmethod,
@@ -12,6 +15,7 @@ from importlib.util import (
 )
 from typing import (
     Any,
+    Callable,
     ClassVar,
     List,
     Optional,
@@ -318,3 +322,55 @@ class CommonTest(ABC):
         """Clear the TF session."""
         if not self.skip_tf:
             clear_session()
+
+
+def parameterized(*attrs: tuple) -> Callable:
+    """Parameterized test.
+
+    Orginal class will not be actually generated. Avoid inherbiting from it.
+    New classes are generated with the name of the original class and the
+    parameters.
+
+    Parameters
+    ----------
+    *attrs : tuple
+        The attributes to be parameterized.
+
+    Returns
+    -------
+    object
+        The decorator.
+
+    Examples
+    --------
+    >>> @parameterized(
+    ...     (True, False),
+    ...     (True, False),
+    ... )
+    ... class TestSeA(CommonTest, unittest.TestCase):
+    ...     @property
+    ...     def data(self) -> dict:
+    ...         (
+    ...             param1,
+    ...             param2,
+    ...         ) = self.param
+    ...         return {
+    ...             "param1": param1,
+    ...             "param2": param2,
+    ...         }
+    """
+
+    def decorator(base_class: type):
+        class_module = sys.modules[base_class.__module__].__dict__
+        for pp in itertools.product(*attrs):
+
+            class TestClass(base_class, unittest.TestCase):
+                param: ClassVar = pp
+
+            name = f"{base_class.__name__}_{'_'.join(str(x) for x in pp)}"
+
+            class_module[name] = TestClass
+        # make unittest module happy by ignoring the original one
+        return object
+
+    return decorator
--- a/source/tests/consistent/common.py
+++ b/source/tests/consistent/common.py
@@ -3,6 +3,9 @@ from abc import (
     ABC,
     abstractmethod,
 )
+from enum import (
+    Enum,
+)
 from typing import (
     Any,
     ClassVar,
@@ -44,6 +47,8 @@ class CommonTest(ABC):
     """Arguments that maps to the `data`."""
     skip_dp: ClassVar[bool] = False
     """Whether to skip the native DP model."""
+    skip_tf: ClassVar[bool] = False
+    """Whether to skip the TensorFlow model."""
     skip_pt: ClassVar[bool] = False
     """Whether to skip the PyTorch model."""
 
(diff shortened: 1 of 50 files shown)


source/tests/consistent/common.py
```diff
@@ -0,0 +1,269 @@
...
+        run_sess(sess, tf.global_variables_initializer())
+        return run_sess(
+            sess,
+            t_out_indentity,
+            feed_dict=feed_dict,
+        )
+
+    def test_tf_dp_consistent(self):
+        """Test whether TF and DP are consistent."""
+        if self.skip_dp:
```
iProzd, line 146: if self.skip_dp or self.skip_tf?

iProzd (commented):
Should the status of pt, tf, and dp in testing be considered equivalent? Should we set self.skip_tf, self.skip_pt, and self.skip_dp all? In future model implementations, there might be parts that are not available in tf as well.

source/tests/consistent/test_se_e2_a.py
```diff
@@ -0,0 +1,207 @@
...
+            dtype=GLOBAL_NP_FLOAT_PRECISION,
+        )
+        self.atype = np.array([0, 1, 1, 0, 1, 1], dtype=np.int32)
+        self.box = np.array(
+            [13.0, 0.0, 0.0, 0.0, 13.0, 0.0, 0.0, 0.0, 13.0],
+            dtype=GLOBAL_NP_FLOAT_PRECISION,
+        )
+        self.natoms = np.array([6, 6, 2, 4], dtype=np.int32)
+
+    def build_tf(self, obj: Any, suffix: str) -> Tuple[list, dict]:
```
iProzd, line 83: This method should be in common.py? Most of descriptors may be the same.

source/tests/consistent/test_se_e2_a.py
```diff
@@ -0,0 +1,207 @@
...
+        )
+        return [t_des], {
+            t_coord: self.coords,
+            t_type: self.atype,
+            t_natoms: self.natoms,
+            t_box: self.box,
+            t_mesh: make_default_mesh(True, False),
+        }
+
+    def eval_dp(self, dp_obj: Any) -> Any:
```
iProzd, line 106: Maybe `eval_dp`, `eval_pt` and `compare_tf_dp_ret` should be in common.py as well?

wanghan-iapcm, line 106: would it better to explictly expose all inputs to developers? prevent using `self.coord`, `self.atype` and so on

source/tests/consistent/common.py
```diff
@@ -0,0 +1,269 @@
...
+        data2 = dp_obj.serialize()
+        dp_obj2 = self.dp_class.deserialize(data2)
+        ret3 = self.eval_dp(dp_obj2)
+        for rr2, rr3 in zip(ret2, ret3):
+            if isinstance(rr2, np.ndarray) and isinstance(rr3, np.ndarray):
+                np.testing.assert_allclose(rr2, rr3)
+            else:
+                self.assertEqual(rr2, rr3)
+
+    def test_tf_pt_consistent(self):
```
wanghan-iapcm, line 216: should'nt the cross backend test be designed as test_dp_xxx_consistent?

wanghan-iapcm (commented):
I am not sure if the current design is a good idea, because to test the implementation of pt, one have to install the tf. suppose we have 5 backends, the env would be very heavy for the developers. 

Ideally, the dp backend can be tested without any deep learning platforms, while any implementation, e.g. pt, can be tested without any other backend.

source/tests/consistent/test_se_e2_a.py
njzjz: I don't think a descriptor and a fitting net can share the input. One may also want to test additional inputs.

source/tests/consistent/common.py
njzjz: After refactoring, the reference backend is dynamic, according to whether they are available.

https://github.com/deepmodeling/deepmd-kit/blob/a3b7e0f2d778892f101fe3f8888cfc89ca255eb5/source/tests/consistent/common.py#L191-L198

source/tests/consistent/test_se_e2_a.py
njzjz: This class is not in common.py

source/tests/consistent/common.py
```diff
@@ -0,0 +1,404 @@
...
+        data = obj.serialize()
+        return ret, data
+
+    def get_dp_ret_serialization_from_cls(self, obj):
+        ret = self.eval_dp(obj)
+        data = obj.serialize()
+        return ret, data
+
+    def get_reference_backend(self):
+        if not self.skip_dp:
```
wanghan-iapcm, line 221:
```suggestion
    def get_reference_backend(self):
        """order of checking for ref: dp, tf, dp""" 
        if not self.skip_dp:
```

source/tests/consistent/common.py
```diff
@@ -0,0 +1,404 @@
...
+            else:
+                self.assertEqual(rr1, rr2)
+
+    def tearDown(self) -> None:
+        """Clear the TF session."""
+        if not self.skip_tf:
+            clear_session()
+
+
+class DescriptorTest:
```
wanghan-iapcm, line 343: provide in a separate file for descriptors? or shall we have a folder structure for the tests of different components?

wanghan-iapcm (commented):
would it be possible to loop over a list of possible model parameters like 
https://github.com/deepmodeling/deepmd-kit/blob/392b9e0c5e44c1f1f93b2db2e63fa36ebcfb7914/source/tests/pt/model/test_se_e2_a.py#L43-L45

source/tests/consistent/common.py
njzjz: A folder structure is excellent.

source/tests/consistent/common.py
wanghan-iapcm: sry a typo

wanghan-iapcm (approved):
I like the design of decorator `parameterized`  very much.

iProzd (approved):

PR #3266, opened 2024-02-13
Chore: refactor InvarFitting

anyangml:


anyangml:
@wanghan-iapcm, this is only a draft, not tested, but it's what i have in mind. feels like we can reuse a large portion of the code from InvarFitting, we only need to change the output dim from `1` to `3`or `9`, and transform the output with `gr`. If that's the case, we might want to create a GeneralFitting to encapsulate the common methods, and refactor InvarFitting as a child class.

wanghan-iapcm:
please note that the predictions of dipole and polar are rotationally equivariant.

anyangml:
> please note that the predictions of dipole and polar are rotationally equivariant.

I thought this handles the rotational equivariance, https://github.com/deepmodeling/deepmd-kit/blob/9bc9900d27d4a9217dadf1bcdf9506b14aa759c3/deepmd/pt/model/task/dipole.py#L203

deepmd/pt/model/task/dipole.py
```diff
@@ -55,13 +90,34 @@ def forward(self, inputs, atype, atype_tebd, rot_mat):
...
+        gr = gr.view(nframes, nloc, -1, 3)
+        out_dim = gr.shape[2]  # m1
+        self.filter_layers = NetworkCollection(
+            1 if self.distinguish_types else 0,
+            self.ntypes,
+            network_type="fitting_network",
+            networks=[
+                FittingNet(
+                    in_dim,
+                    out_dim,
```
anyangml, line 105: @wanghan-iapcm ,the `out_dim` here should be `m1` from `gr`, right? I am having shape mismatch with the mask in GeneralFitting `mask = (atype == type_i).unsqueeze(-1)`.

deepmd/pt/model/task/fitting.py
```diff
@@ -316,3 +340,299 @@ def change_energy_bias(
...
+        else:
+            self.filter_layers = NetworkCollection(
+                1 if self.distinguish_types else 0,
+                self.ntypes,
+                network_type="fitting_network",
+                networks=[
+                    FittingNet(
+                        in_dim,
+                        self.dim_out,
+                        self.neuron,
```
anyangml, line 435: we plan to only support the new implementation for dipole and polar, the `dim_out` is different for those fitting net, we might want to do something like this
```
if gr is not None:
    m1 = gr.shape[2]
    expected_dim_out = m1
else:
    expected_dim_out = self.dim_out
....
FittingNet(in_dim, expected_dim_out.....)
```

wanghan-iapcm: We may provide an abstract method like `_net_out_dim`, so all inherited classes should provide this information

deepmd/pt/model/task/fitting.py
```diff
@@ -316,3 +340,299 @@ def change_energy_bias(
...
+            },
+            # "rcond": self.rcond ,
+            # "tot_ener_zero": self.tot_ener_zero ,
+            # "trainable": self.trainable ,
+            # "atom_ener": self.atom_ener ,
+            # "layer_name": self.layer_name ,
+            # "use_aparam_as_mask": self.use_aparam_as_mask ,
+            # "spin": self.spin ,
+            ## NOTICE:  not supported by far
+            "rcond": None,
```
wanghan-iapcm, line 480: will be updated when https://github.com/deepmodeling/deepmd-kit/pull/3279 is merged.

deepmd/pt/model/task/fitting.py
```diff
@@ -316,3 +340,299 @@ def change_energy_bias(
...
+        neuron: List[int] = [128, 128, 128],
+        resnet_dt: bool = True,
+        numb_fparam: int = 0,
+        numb_aparam: int = 0,
+        activation_function: str = "tanh",
+        precision: str = DEFAULT_PRECISION,
+        distinguish_types: bool = False,
+        **kwargs,
+    ):
+        """Construct a fitting net for energy.
```
iProzd, line 361: docstr mismatched

deepmd/pt/model/task/fitting.py
```diff
@@ -316,3 +340,299 @@ def change_energy_bias(
...
+
+        Args:
+        - ntypes: Element count.
+        - embedding_width: Embedding width per atom.
+        - neuron: Number of neurons in each hidden layers of the fitting net.
+        - bias_atom_e: Average enery per atom for each element.
+        - resnet_dt: Using time-step in the ResNet construction.
```
iProzd, line 368: docstr mismatched

deepmd/pt/model/task/fitting.py
```diff
@@ -316,3 +340,299 @@ def change_energy_bias(
...
+                    )
+                    for ii in range(self.ntypes if self.distinguish_types else 1)
+                ],
+            )
+            self.filter_layers_old = None
+
+        # very bad design...
+        if "seed" in kwargs:
+            log.info("Set seed to %d in fitting net.", kwargs["seed"])
+            torch.manual_seed(kwargs["seed"])
```
iProzd, line 449: we may explicitly add `seed` into `__init__` method?

deepmd/pt/model/task/fitting.py
```diff
@@ -316,3 +340,299 @@ def change_energy_bias(
...
+        gr: Optional[torch.Tensor] = None,
+        g2: Optional[torch.Tensor] = None,
+        h2: Optional[torch.Tensor] = None,
+        fparam: Optional[torch.Tensor] = None,
+        aparam: Optional[torch.Tensor] = None,
+    ):
+        xx = descriptor
+        nf, nloc, nd = xx.shape
+
+        if hasattr(self, "bias_atom_e"):
```
iProzd, line 549: shall we change name `bias_atom_e` into something like `out_bias`?

anyangml: this is probably fine for now.

deepmd/pt/model/task/fitting.py
```diff
@@ -356,16 +359,21 @@ def __init__(
         activation_function: str = "tanh",
         precision: str = DEFAULT_PRECISION,
         distinguish_types: bool = False,
```
iProzd, line 361: Parameters missed in docstr

iProzd (approved):

deepmd/pt/model/task/fitting.py
```diff
@@ -316,3 +343,311 @@ def change_energy_bias(
...
+        Only atoms with selected atom types have atomic contribution
+        to the result of the model.
+        If returning an empty list, all atom types are selected.
+        """
+        return []
+
+    @abstractmethod
+    def _net_out_dim(self):
+        """Set the FittingNet output dim."""
+        raise NotImplementedError
```
wanghan-iapcm, line 530:
```suggestion
        pass
```
otherwise 530 will be recorded as not covered by UT.

wanghan-iapcm (approved):

deepmd/pt/model/task/fitting.py
```diff
@@ -316,3 +343,311 @@ def change_energy_bias(
...
+            nb, nloc, _ = aparam.shape
+            t_aparam_avg = self._extend_a_avg_std(self.aparam_avg, nb, nloc)
+            t_aparam_inv_std = self._extend_a_avg_std(self.aparam_inv_std, nb, nloc)
+            aparam = (aparam - t_aparam_avg) * t_aparam_inv_std
+            xx = torch.cat(
+                [xx, aparam],
+                dim=-1,
+            )
+
+        outs = torch.zeros(nf, nloc, self.dim_out)  # jit assertion
```
njzjz, line 612: dtype is not set

njzjz, line 612: device is not set

njzjz (changes requested):

deepmd/pt/model/task/fitting.py
anyangml: this will be overwritten.

deepmd/pt/model/task/ener.py
```diff
@@ -63,118 +54,64 @@ def __init__(
...
     ):
         """Construct a fitting net for energy.
 
-        Args:
-        - ntypes: Element count.
-        - embedding_width: Embedding width per atom.
-        - neuron: Number of neurons in each hidden layers of the fitting net.
-        - bias_atom_e: Average enery per atom for each element.
-        - resnet_dt: Using time-step in the ResNet construction.
+        Parameters
```
njzjz, line 62: The constructor's parameters should be documented in the class docstring.
See https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

anyangml:
UT 'test_make_hessian_model` fails on CUDA, not sure if it was tested before.

deepmd/pt/model/task/fitting.py
```diff
@@ -316,3 +343,332 @@ def change_energy_bias(
...
+            t_aparam_inv_std = self._extend_a_avg_std(self.aparam_inv_std, nb, nloc)
+            aparam = (aparam - t_aparam_avg) * t_aparam_inv_std
+            xx = torch.cat(
+                [xx, aparam],
+                dim=-1,
+            )
+
+        outs = torch.zeros(
+            (nf, nloc, self.dim_out), dtype=env.GLOBAL_PT_FLOAT_PRECISION
+        ).to(env.DEVICE)  # jit assertion
```
njzjz, line 642: This creates the tensor on the CPU and copies it to the GPU, which is obviously slower than directly creating on the GPU.
See https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html#create-tensors-directly-on-the-target-device

source/tests/pt/model/test_make_hessian_model.py
njzjz, line 1: This can be fixed in a separated PR, as it's not related to this PR. cc @wanghan-iapcm

njzjz (approved):

source/tests/pt/model/test_make_hessian_model.py
anyangml: Resolved.

iProzd (approved):

PR #3268, opened 2024-02-13
dp: pt: rearrange folder structure of atomic_model and model.

wanghan-iapcm:
- create models `dpmodel.atomic_model` and `pt.model.atomic_model`. 
- place all the atomic model related codes into the new modules. 
- `PairTabModel` -> `PairTabAtomicModel`
- `ZBLModel` -> `DPZBLModel`


deepmd/dpmodel/atomic_model/__init__.py
njzjz, line 1: Module-level docstring is required, so the developer can distinguish `atomic_model` from `model`.

deepmd/dpmodel/model/__init__.py
njzjz, line 1: Module-level docstring is required.

deepmd/pt/model/atomic_model/pairtab_atomic_model.py
```diff
@@ -19,12 +19,9 @@
 from .base_atomic_model import (
     BaseAtomicModel,
 )
-from .model import (
-    BaseModel,
-)
 
 
-class PairTabModel(BaseModel, BaseAtomicModel):
+class PairTabAtomicModel(torch.nn.Module, BaseAtomicModel):
```
anyangml, line 24: I changed this to BaseModel in the CUDA PR, I think this is fine, but we should probably test it to be sure.

wanghan-iapcm: I find the current `BaseModel` (after #3245) provides nothing and can be safely removed.

wanghan-iapcm: By name `BaseModel` seems to be the common base class for all models but not for all atomic models.

anyangml:
LGTM

anyangml (approved):

PR #3271, opened 2024-02-14
add neighbor stat support with NumPy and PyTorch implementation

njzjz:
I also tested `examples/water`,`examples/nopbc`, and the ANI-1x dataset (compared to the screenshot in #1624) to confirm consistent results.

Besides, as the OP supports multiple frames, the PT implementation only takes 9 s on ANI-1x, which is much faster than the TF implementation, which took over 10 min as shown in #1624.

image


deepmd/pt/utils/neighbor_stat.py
```diff
@@ -0,0 +1,190 @@
...
+from deepmd.pt.utils.nlist import (
+    extend_coord_with_ghosts,
+)
+from deepmd.utils.data_system import (
+    DeepmdDataSystem,
+)
+from deepmd.utils.neighbor_stat import NeighborStat as BaseNeighborStat
+
+
+class NeighborStatOP(torch.nn.Module):
```
njzjz, line 26: It seems TorchScript does not work with code coverage. xref: https://github.com/pytorch/pytorch/issues/43146

njzjz:
> which is much faster than the TF implementation

We may consider applying the implementation in this PR to TF.

wanghan-iapcm (approved):

PR #3274, opened 2024-02-14
fix bug of output dimension in pt InvarFitting 

wanghan-iapcm:


anyangml (commented):
can we simply use dim_out, instead of creating another variable?

wanghan-iapcm:
> can we simply use dim_out, instead of creating another variable?

sure

wanghan-iapcm:
> can we simply use dim_out, instead of creating another variable?

introduced to align with `in_dim` :)

anyangml (approved):

PR #3280, opened 2024-02-15
feat: dp and pt: implement exclude types in descriptor se_a

wanghan-iapcm:


source/tests/pt/model/test_se_e2_a.py
```diff
@@ -40,9 +40,10 @@ def test_consistency(
         dstd = rng.normal(size=(self.nt, nnei, 4))
         dstd = 0.1 + np.abs(dstd)
 
-        for idt, prec in itertools.product(
+        for idt, prec, em in itertools.product(
             [False, True],
             ["float64", "float32"],
+            [[[0, 1]], [[1, 1]]],
```
njzjz, line 46: Should we keep the test for `[]` (the original case)?

wanghan-iapcm: sure

njzjz:
GPU tests failed.

deepmd/pt/model/descriptor/descriptor.py
```diff
@@ -206,6 +208,32 @@ class DescriptorBlock(torch.nn.Module, ABC):
...
+                [
+                    1 if (tt_i, tt_j) not in _exclude_types else 0
+                    for tt_i in range(ntypes + 1)
+                ]
+                for tt_j in range(ntypes + 1)
+            ],
+            dtype=np.int32,
+        )
+        # (ntypes+1 x ntypes+1)
+        self.type_mask = torch.from_numpy(self.type_mask).view([-1])
```
iProzd, line 234: Should `self.type_mask` be saved through `register_buffer`, so it can be saved into `model.pt`? Or we allow user to change `exclude_types` each time when restart from saved checkpoint, even if it doesn't match each other?

iProzd: Additionally, `device` missed here for `torch.from_numpy`, GPU may fail.

deepmd/pt/model/descriptor/descriptor.py
```diff
@@ -332,7 +360,50 @@ def forward(
...
         """Calculate DescriptorBlock."""
-        raise NotImplementedError
+        pass
+
+    # may have a better place for this method...
+    def build_type_exclude_mask(
+        self,
+        nlist: torch.Tensor,
+        atype_ext: torch.Tensor,
+    ) -> torch.Tensor:
```
iProzd, line 370: Maybe we can make `ExcludeMask` class in utils and as an attribute (if needed) in atomic model level? Then we can get `exlude_mask` at the begining of model `forward` and pass the mask to `forward` of both descriptor and fitting.

iProzd: Anyway, this can also be done in spin PR.

wanghan-iapcm: A `ExcludeMask` class like that in the `dpmodel` cannot be jitted...
the exclude_mask is only needed by descriptors, because it is pairwise. The fitting needs only single atom masks.

njzjz (approved):

iProzd (approved):

PR #3285, opened 2024-02-17
pt: refactor data stat

njzjz:


deepmd/pt/utils/env_mat_stat.py
```diff
@@ -0,0 +1,205 @@
...
+                self.descriptor.get_rcut(),
+                # TODO: export rcut_smth from DescriptorBlock
+                self.descriptor.rcut_smth,
+            )
+            env_mat = env_mat.view(
+                coord.shape[0], coord.shape[1], self.descriptor.get_nsel(), 4
+            )
+            env_mats = {}
+
+            if "real_natoms_vec" not in system:
```
wanghan-iapcm, line 136: shouldn't be determined by the descriptor's attribute `distinguish_type`?

deepmd/pt/utils/env_mat_stat.py
```diff
@@ -0,0 +1,205 @@
...
+                for frame_item in range(env_mat.shape[0]):
+                    dd_ff = env_mat[frame_item]
+                    atype_frame = atype[frame_item]
+                    for type_i in range(self.descriptor.get_ntypes()):
+                        type_idx = atype_frame == type_i
+                        dd = dd_ff[type_idx]
+                        dd = dd.reshape([-1, 4])  # typen_atoms * nnei, 4
+                        env_mats[f"r_{type_i}"] = dd[:, :1]
+                        env_mats[f"a_{type_i}"] = dd[:, 1:]
+                        yield self.compute_stat(env_mats)
```
wanghan-iapcm, line 161: I am worry about the efficiency of this part of code. 
actually we can use type mask to process across frames in parallel, which is supposed to be more efficient than the for loop. 

The efficiency issue may be left for a future PR.

source/tests/pt/test_stat.py
```diff
@@ -170,20 +170,15 @@ def test_descriptor(self):
...
-        self.assertTrue(
-            np.allclose(
-                self.dp_d.dstd.reshape([-1]),
-                my_en.stddev.cpu().reshape([-1]),
-                rtol=0.01,
-            )
+        np.testing.assert_allclose(
+            self.dp_d.dstd.reshape([-1]),
+            my_en.stddev.cpu().reshape([-1]),
+            rtol=0.01,
```
wanghan-iapcm, line 181: not sure if the rtol is too loose.

wanghan-iapcm (commented):
It seems that the case of `mixed_type`, or not `distinguish_type` is not covered by UTs.

deepmd/pt/utils/env_mat_stat.py
njzjz: No. To my understanding, `mixed_type` (produce `real_natoms_vec` here) describes the central atom i from the data, while `distinguish_type` describes the neighboring atoms. Here, the central atoms matter.

source/tests/pt/test_stat.py
njzjz: Absulotely! When I increased rtol, I found a bug in the code. The original tol is too low to find the bug.

njzjz: I found there are another issue...

deepmd/pt/utils/env_mat_stat.py
njzjz: No. To my understanding, mixed_type (produce real_natoms_vec here) describes the central atom i from the data, while distinguish_type describes the neighboring atoms from the descriptor. Here, the central atoms matter.

source/tests/pt/test_stat.py
njzjz: I found coordinates fed to TF (line 150) is different from those fed to PT (line 171).

Need help here...

njzjz: I make it work in a hack way: picking only 2 frames in to a new data directory.

njzjz:
> It seems that the case of `mixed_type`, or not `distinguish_type` is not covered by UTs.

Resolved.

source/tests/pt/test_stat.py
wanghan-iapcm: > I make it work in a hack way: picking only 2 frames in to a new data directory.

confused by the hacking. If i understand correctly, the statistics should be the same if datasets and numpy random seeds are the same.

njzjz: > confused by the hacking. If i understand correctly, the statistics should be the same if datasets and numpy random seeds are the same.

It looks like TF and PT don't have consistent random behavior. It may need @CaRoLZhangxy to resolve.

deepmd/pt/utils/env_mat_stat.py
```diff
@@ -0,0 +1,206 @@
...
+
+        Returns
+        -------
+        str
+            The hash of the environment matrix.
+        """
+        return get_hash(
+            {
+                "type": "se_a",
+                "ntypes": self.descriptor.get_ntypes(),
```
iProzd, line 175: Descriptors that have different `type_map` but the same ntypes will get the same hash, which will mismatch the stat.

deepmd/pt/model/descriptor/descriptor.py
```diff
@@ -312,7 +204,6 @@ def share_params(self, base_class, shared_level, resume=False):
                     "sumr2": sumr2_base + sumr2,
                     "suma2": suma2_base + suma2,
                 }
-                base_class.init_desc_stat(**stat_dict)
```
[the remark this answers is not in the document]
iProzd:
```Python
base_env = EnvMatStatSeA(base_class)
for kk in base_class.stats:
    tmp_env = EnvMatStatSeA(self)
    base_env.stats[kk] += tmp_env.stats[kk]
mean, stddev = base_env()
if not base_class.set_davg_zero:
    base_class.mean.copy_(torch.tensor(mean, device=env.DEVICE))
base_class.stddev.copy_(torch.tensor(stddev, device=env.DEVICE))
self.mean = base_class.mean
self.stddev = base_class.stddev
```

deepmd/pt/utils/env_mat_stat.py
njzjz: I see. Since a descriptor doesn't have `type_map`, I will not add it here but add it in the model level.

deepmd/pt/model/descriptor/descriptor.py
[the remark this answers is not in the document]
njzjz: Fixed in ba688a9892f4c9bb9786e8ec4bc5a90be6cd649a

deepmd/pt/utils/env_mat_stat.py
njzjz: Fixed in 6a83465c7447a5aa159bbb892cc4b2b08e4dc843

deepmd/pt/model/descriptor/descriptor.py
```diff
@@ -183,27 +196,13 @@ def share_params(self, base_class, shared_level, resume=False):
...
-                stat_dict = {
-                    "sumr": sumr_base + sumr,
-                    "suma": suma_base + suma,
-                    "sumn": sumn_base + sumn,
-                    "sumr2": sumr2_base + sumr2,
-                    "suma2": suma2_base + suma2,
-                }
+                base_env = EnvMatStatSeA(base_class)
+                for kk in base_class.get_stats():
+                    base_env.stats[kk] += self.get_stats()[kk]
```
iProzd, line 201: `base_env.stats` is still empty at the begining, rather than start from `base_class.get_stats()`?

njzjz: Fixed in https://github.com/deepmodeling/deepmd-kit/pull/3285/commits/7c9a66e544a371cc33dae7a03e4e7a7ebf66d950

iProzd (approved):

source/tests/pt/test_stat.py
wanghan-iapcm: could you please raise an issue?

PR #3286, opened 2024-02-17
consistent energy fitting

njzjz:


source/tests/consistent/fitting/test_ener.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @property
+    def skip_pt(self) -> bool:
+        (
+            resnet_dt,
+            precision,
+            distinguish_types,
+            numb_fparam,
+        ) = self.param
+        # TODO: float32 has bug
```
njzjz, line 83: @anyangml It should be fixed

deepmd/pt/model/model/__init__.py
```diff
@@ -54,6 +54,7 @@ def get_zbl_model(model_params):
     fitting_net["ntypes"] = descriptor.get_ntypes()
     fitting_net["distinguish_types"] = descriptor.distinguish_types()
     fitting_net["embedding_width"] = descriptor.get_dim_out()
```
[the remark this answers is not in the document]
njzjz: It's still used by the dipole fitting.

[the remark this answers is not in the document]
wanghan-iapcm: @anyangml please not here we should unify the name of the parameter. `embedding_width`  is prefered to `dim_rot_mat`.

wanghan-iapcm (approved):

source/tests/consistent/fitting/test_ener.py
anyangml: what's the issue here?

anyangml (approved):

source/tests/consistent/fitting/test_ener.py
wanghan-iapcm: it seems that float 32 does not pass the consistency test. what is the difference? do we reduce the tolerance accordingly when using float32?

njzjz: > do we reduce the tolerance accordingly when using float32?

It just complains that a float64 tensor cannot be operated with a float32 tensor.

wanghan-iapcm: please raise an issue for it. Thanks!

njzjz: I found only pt has the bug. The dp backend had the problem of tolerance, and I fixed that.

source/tests/consistent/fitting/test_ener.py
```diff
@@ -0,0 +1,185 @@
...
+        (
+            resnet_dt,
+            precision,
+            distinguish_types,
+            numb_fparam,
+        ) = self.param
+        if precision == "float64":
+            return 1e-10
+        elif precision == "float32":
+            return 1e-5
```
njzjz, line 183: It looks `1e-5` may still fail the test...
```suggestion
            return 1e-4
```

PR #3287, opened 2024-02-17
pt:Removed the data system implemented in PyTorch

CaRoLZhangxy:
The old code has been retained for reference by @Chengqian-Zhang for de-noise implementation.


deepmd/utils/data.py
```diff
@@ -129,6 +129,15 @@ def __init__(
...
         self.modifier = modifier
+        # calculate prefix sum for get_item method
+        self.nframes = 0
+        i = 1
+        self.prefix_sum = [0] * (len(self.dirs) + 1)
+        for item in self.dirs:
+            frames = self._get_nframes(item)
+            self.prefix_sum[i] = self.prefix_sum[i - 1] + frames
+            i += 1
+            self.nframes += frames
```
wanghan-iapcm, line 140: what is `prefix_sum` ? explanation in doc is needed.

deepmd/utils/data.py
```diff
@@ -440,6 +468,30 @@ def _shuffle_data(self, data):
...
+        if self.data_dict["coord"]["high_prec"]:
+            coord = path.load_numpy().astype(GLOBAL_ENER_FLOAT_PRECISION)
+        else:
+            coord = path.load_numpy().astype(GLOBAL_NP_FLOAT_PRECISION)
+        if coord.ndim == 1:
+            coord = coord.reshape([1, -1])
+        nframes = coord.shape[0]
+        return nframes
+
+    def preprocess(self, data):
```
wanghan-iapcm, line 485: doc string needed. 
is the preprocess used only for the torch code? 
suggest to rename the function in a more self-explanatory way.

deepmd/utils/data.py
```diff
@@ -129,6 +129,16 @@ def __init__(
...
+        # calculate prefix sum for get_item method
+        self.nframes = 0
+        i = 1
+        # The prefix sum stores the range of indices contained in each directory, which is needed by get_item method
+        self.prefix_sum = [0] * (len(self.dirs) + 1)
+        for item in self.dirs:
+            frames = self._get_nframes(item)
+            self.prefix_sum[i] = self.prefix_sum[i - 1] + frames
+            i += 1
+            self.nframes += frames
```
njzjz, line 141: It looks you just rewrote np.cumsum and np.sum

caic99:
```suggestion
        # calculate prefix sum for get_item method
        frames_list = [self._get_nframes(item) for item in self.dirs]
        self.nframes = np.sum(frames_list)
        # The prefix sum stores the range of indices contained in each directory, which is needed by get_item method
        self.nframes = np.cumsum([0] + frames_list).tolist()
```

@CaRoLZhangxy Plz try if this is OK

njzjz: Or use `itertools.accumulate` and `sum` without numpy

deepmd/pt/utils/dataset.py
```diff
@@ -575,44 +24,22 @@ def __init__(
...
+        self._data_system = DeepmdData(
+            sys_path=system, shuffle_test=shuffle, type_map=self._type_map
         )
+        self._data_system.add("energy", 1, atomic=False, must=False, high_prec=True)
+        self._data_system.add("force", 3, atomic=True, must=False, high_prec=False)
+        self._data_system.add("virial", 9, atomic=False, must=False, high_prec=False)
         self.mixed_type = self._data_system.mixed_type
         self._ntypes = self._data_system.get_ntypes()
-        self._natoms = self._data_system._natoms
+        self._natoms = self._data_system.natoms
```
njzjz, line 35: The public method is preferred
```suggestion
        self._natoms = self._data_system.get_natoms()
```

deepmd/pt/utils/dataset.py
```diff
@@ -575,44 +24,22 @@ def __init__(
...
-            type_map=self._type_map,
-            type_split=type_split,
-            noise_settings=noise_settings,
-            shuffle=shuffle,
+        self._data_system = DeepmdData(
+            sys_path=system, shuffle_test=shuffle, type_map=self._type_map
         )
+        self._data_system.add("energy", 1, atomic=False, must=False, high_prec=True)
+        self._data_system.add("force", 3, atomic=True, must=False, high_prec=False)
+        self._data_system.add("virial", 9, atomic=False, must=False, high_prec=False)
```
njzjz, line 32: I don't quite understand why these are hard-coded here - what will happen if we fit the dipole other than energy?

deepmd/utils/data.py
```diff
@@ -249,6 +254,25 @@ def check_test_size(self, test_size):
         else:
             return None
 
+    def get_item(self, index: int) -> dict:
+        """Get a single frame data . The frame is picked from the data system by index.
```
wanghan-iapcm, line 258:
```suggestion
        """Get a single frame data . The frame is picked from the data system by index. The index is coded across all the sets.
```

deepmd/utils/data.py
```diff
@@ -249,6 +254,25 @@ def check_test_size(self, test_size):
...
+        Parameters
+        ----------
+        index
+            index of the frame
+        """
+        for i in range(
+            0, len(self.dirs)
+        ):  # note: if different sets can be merged, prefix sum is unused to calculate
+            if index < self.prefix_sum[i]:
+                break
```
njzjz, line 269: I guess you need [bisect](https://docs.python.org/3/library/bisect.html). This can be simply written as
```suggestion
        i = bisect.bisect_right(self.prefix_sum, index)
```

deepmd/utils/data.py
```diff
@@ -249,6 +254,25 @@ def check_test_size(self, test_size):
         else:
             return None
 
+    def get_item(self, index: int) -> dict:
```
njzjz, line 257: It seems this method can only be used by pt (since it calls `reformat_data_torch`).

deepmd/pt/utils/dataset.py
njzjz: @iProzd do you know what the design is here?

wanghan-iapcm (approved):

njzjz (approved):

Chengqian-Zhang (approved):

caic99 (approved):

deepmd/pt/utils/dataset.py
iProzd: This should be placed into loss or model requirements indeed. The hard-coded here was then not well-designed.
Btw in tf it's a global variable called `data_requirement`, which is not flexible way for multi-task training when different models have different requirements. 

I think I will fix this part in training code cleaning PR.

iProzd (approved):

PR #3289, opened 2024-02-18
fix: replace all distinguished types by mixed types.

wanghan-iapcm:


deepmd/dpmodel/atomic_model/linear_atomic_model.py
```diff
@@ -50,13 +50,11 @@ def __init__(
...
         super().__init__()
         self.models = models
-        self.distinguish_type_list = [
-            model.distinguish_types() for model in self.models
-        ]
+        self.mixed_types_list = [model.mixed_types() for model in self.models]
 
-    def distinguish_types(self) -> bool:
+    def mixed_types(self) -> bool:
         """If distinguish different types by sorting."""
```
iProzd, line 56: docstr may be 'If ignore differences in types when sorting.'

deepmd/dpmodel/fitting/invar_fitting.py
```diff
@@ -126,7 +126,7 @@ def __init__(
         layer_name: Optional[List[Optional[str]]] = None,
         use_aparam_as_mask: bool = False,
         spin: Any = None,
-        distinguish_types: bool = False,
+        mixed_types: bool = False,
```
iProzd, line 129: Update the default value.

deepmd/pt/model/descriptor/dpa1.py
```diff
@@ -108,11 +108,17 @@ def get_dim_out(self) -> int:
     def get_dim_emb(self) -> int:
         return self.se_atten.dim_emb
 
-    def distinguish_types(self) -> bool:
-        """Returns if the descriptor requires a neighbor list that distinguish different
-        atomic types or not.
+    def mixed_types(self) -> bool:
+        """If true, the discriptor
+        1. assumes total numbe of atoms aligned across frames;
```
iProzd, line 113: 'numbe' typo

deepmd/dpmodel/model/make_model.py
```diff
@@ -111,7 +111,7 @@ def call(
                 nloc,
                 self.get_rcut(),
                 self.get_sel(),
-                distinguish_types=self.distinguish_types(),
+                distinguish_types=not self.mixed_types(),
```
iProzd, line 114: key not changed for `build_neighbor_list`

iProzd: maybe it's ok here

deepmd/pt/model/descriptor/dpa1.py
njzjz: Same issue below

deepmd/pt/model/descriptor/dpa1.py
```diff
@@ -108,11 +108,17 @@ def get_dim_out(self) -> int:
...
+        2. requires a neighbor list that does not distinguish different atomic types.
+
+        If false, the discriptor
+        1. assumes total numbe of atoms of each atom type aligned across frames;
```
njzjz, line 117: I think it also requires the atom type to be aligned, not only the number.

deepmd/pt/model/descriptor/dpa1.py
```diff
@@ -108,11 +108,17 @@ def get_dim_out(self) -> int:
...
+        2. requires a neighbor list that distinguishes different atomic types.
```
njzjz, line 118: TF requires both central and neighbor atoms to be sorted by atom types. Does PT have this requirement?

deepmd/pt/model/descriptor/dpa1.py
wanghan-iapcm: what is the different ? i did not get.

deepmd/pt/model/descriptor/dpa1.py
wanghan-iapcm: pt does not support "type two side", i.e. it only use different nets for different neighbor types, but not for different center atom types, thus the center atoms are not assumed to be sorted.

deepmd/pt/model/descriptor/dpa1.py
njzjz: Is this allowed?

Frame 0: 0 0 0 1 1 1
Frame 1: 1 1 1 0 0 0

njzjz (approved):

anyangml (approved):

deepmd/pt/model/descriptor/dpa1.py
wanghan-iapcm: yes see https://github.com/deepmodeling/deepmd-kit/pull/3295

PR #3303, opened 2024-02-19
pt: avoid `set_default_dtype` in tests

njzjz:
Why it is bad:
The default dtype in the producution code is still float32. When setting it to float64 during tests, the actual producation behavior may not be properly tested. (for example, if the production code misses dtype for pt.zeros, we are not able to find it in these tests)

### DIFF
--- a/source/tests/pt/model/test_descriptor_dpa1.py
+++ b/source/tests/pt/model/test_descriptor_dpa1.py
@@ -22,9 +22,6 @@ from deepmd.pt.utils.nlist import (
     extend_input_and_build_neighbor_list,
 )
 
-dtype = torch.float64
-torch.set_default_dtype(dtype)
-
 CUR_DIR = os.path.dirname(__file__)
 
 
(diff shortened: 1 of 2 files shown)


wanghan-iapcm:
> Why it is bad: The default dtype in the producution code is still float32. When setting it to float64 during tests, the actual producation behavior may not be properly tested. (for example, if the production code misses dtype for pt.zeros, we are not able to find it in these tests)

pt uses float32 only when `DP_INTERFACE_PREC` is set to low ?

wanghan-iapcm (approved):

njzjz:
> pt uses float32 only when `DP_INTERFACE_PREC` is set to low ?

When `DP_INTERFACE_PREC` is low, `GLOBAL_PT_FLOAT_PRECISION` is float32. Otherwise, `GLOBAL_PT_FLOAT_PRECISION` is float64. That's it.

If `dtype` is not set, a tensor (`torch.tensor`, `torch.ones`, `torch.zeros`, etc) still uses float32, no matter what `GLOBAL_PT_FLOAT_PRECISION` is.

PR #3304, opened 2024-02-19
fix neighbor stat mixed_types input

njzjz:
It was broken by #3289, which replaced distinguished types by mixed types but didn't change the input.

Add tests for two types.

### DIFF
--- a/deepmd/dpmodel/utils/neighbor_stat.py
+++ b/deepmd/dpmodel/utils/neighbor_stat.py
@@ -126,7 +126,7 @@ class NeighborStat(BaseNeighborStat):
         one_type: bool = False,
     ) -> None:
         super().__init__(ntypes, rcut, one_type)
-        self.op = NeighborStatOP(ntypes, rcut, not one_type)
+        self.op = NeighborStatOP(ntypes, rcut, one_type)
 
     def iterator(
         self, data: DeepmdDataSystem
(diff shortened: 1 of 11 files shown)


deepmd/dpmodel/utils/neighbor_stat.py
```diff
@@ -126,7 +126,7 @@ def __init__(
         one_type: bool = False,
     ) -> None:
         super().__init__(ntypes, rcut, one_type)
-        self.op = NeighborStatOP(ntypes, rcut, not one_type)
+        self.op = NeighborStatOP(ntypes, rcut, one_type)
```
wanghan-iapcm, line 129: does `one_type` have the same meaning as `mixed_types`? if yes changing `one_type` to `mixed_types`?

njzjz: I agree, but I am not sure if it should be done in this PR.

njzjz: It seems not too hard; it's just replacing all...

njzjz: https://github.com/deepmodeling/deepmd-kit/pull/3304/commits/530a3ce11bfd45776448c928d3b9ac8de974afce

wanghan-iapcm (approved):
LGTM

PR #3306, opened 2024-02-20
consistent energy model

njzjz:
Many hacking things...


deepmd/tf/model/model.py
```diff
@@ -730,3 +787,65 @@ def update_sel(cls, global_jdata: dict, local_jdata: dict):
...
+            The deserialized descriptor
+        """
+        data = copy.deepcopy(data)
+
+        data["descriptor"]["type"] = {
+            "DescrptSeA": "se_e2_a",
+        }[data.pop("descriptor_name")]
+        data["fitting"]["type"] = {
+            "EnergyFittingNet": "ener",
+        }[data.pop("fitting_name")]
```
njzjz, line 817: The hack thing is the manual conversion here... It needs to be discussed.

source/tests/consistent/model/common.py
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+    def eval_dp_model(self, dp_obj: Any, natoms, coords, atype, box) -> Any:
+        return dp_obj(coords, atype, box=box)
+
+    def eval_pt_model(self, pt_obj: Any, natoms, coords, atype, box) -> Any:
+        return {
+            kk: vv.detach().cpu().numpy() if torch.is_tensor(vv) else vv
+            for kk, vv in pt_obj(
+                torch.from_numpy(coords).to(PT_DEVICE),
```
wanghan-iapcm, line 61: use `pt.utils.utils.to_numpy_array` ?

source/tests/consistent/model/test_ener.py
```diff
@@ -0,0 +1,183 @@
...
+    tf_class = EnergyModelTF
+    dp_class = EnergyModelDP
+    pt_class = EnergyModelPT
+    args = model_args()
+
+    def pass_data_to_cls(self, cls, data) -> Any:
+        """Pass data to the class."""
+        data = data.copy()
+        if cls is EnergyModelDP:
+            # should not do it here...
```
wanghan-iapcm, line 88: I agree... we need a `get_model` in dpmodel.

source/tests/consistent/model/common.py
njzjz: Need `pt.utils.utils.from_numpy_array` for this line, not `to_numpy_array`.

wanghan-iapcm (approved):

PR #3307, opened 2024-02-20
pt: explicitly set device

njzjz:
Forcely requires setting the device to `env.DEVICE` or `cpu` explicitly for functions like `torch.tensor`, `torch.zeros`, `torch.ones`, `torch.rand`, `torch.eye`, `LayerNorm`, and data loader.
This ensures that no OP runs on the wrong device. The trick here is `torch.set_default_device("cuda:9999999")` in the tests, so errors will be thrown if the default device is used.

Tips:
(1) Avoid `torch.zeros(...).to(device=...)`. This first initializes memory on CPUs and copies it to GPUs.
(2) Use `with torch.device(...)` for a module that cannot set a device (i.e., the data loader).


wanghan-iapcm (approved):

deepmd/pt/train/training.py
```diff
@@ -519,7 +521,8 @@ def step(_step_id, task_key="Default"):
                     if not torch.isfinite(grad_norm).all():
                         # check local gradnorm single GPU case, trigger NanDetector
                         raise FloatingPointError("gradients are Nan/Inf")
-                self.optimizer.step()
+                with torch.device("cpu"):
+                    self.optimizer.step()
```
iProzd, line 525: why use device settings here?

deepmd/pt/utils/dataloader.py
```diff
@@ -139,8 +139,9 @@ def construct_dataset(system):
             self.total_batch += len(system_dataloader)
         # Initialize iterator instances for DataLoader
         self.iters = []
-        for item in self.dataloaders:
-            self.iters.append(iter(item))
+        with torch.device("cpu"):
+            for item in self.dataloaders:
+                self.iters.append(iter(item))
```
iProzd, line 144: Does this device setting matter here?

anyangml (approved):

deepmd/pt/train/training.py
njzjz: https://github.com/pytorch/pytorch/blob/c28a43988e7f3c630e5c50717c146c2cd936eabf/torch/optim/adam.py#L104-L111

deepmd/pt/utils/dataloader.py
njzjz: https://github.com/pytorch/pytorch/blob/c28a43988e7f3c630e5c50717c146c2cd936eabf/torch/utils/data/dataloader.py#L602

iProzd: got it

iProzd (approved):

PR #3309, opened 2024-02-20
Feat: add DipoleModel and PolarModel

anyangml:
This PR is to provide model wrappers for `DipoleFittingNet` and `PolarFittingNet`, such that the saved model can be used directly in inference with `DeepDiole` and `DeepPolar`.


anyangml:
@njzjz @wanghan-iapcm , not sure if my OutDef shape is wrong 
image
I am expecting `1, 5, 3, 3` and `1, 5, 3, 9`.
we may need to change this function to accommodate `dipole` shape.

Also, we have to add `dipole_derv_c` otherwise, we have length mismatch here https://github.com/deepmodeling/deepmd-kit/blob/4b994df1d4a99b78efd9f00a32e4d4b66d84d6d3/deepmd/pt/infer/deep_eval.py#L245-L251

anyangml:
this will return `0`, since get_sel_type() returns an empty list in GeneralFitting. https://github.com/deepmodeling/deepmd-kit/blob/4b994df1d4a99b78efd9f00a32e4d4b66d84d6d3/deepmd/infer/deep_tensor.py#L107

njzjz:
> @njzjz @wanghan-iapcm , not sure if my OutDef shape is wrong 
> I am expecting `1, 5, 3, 3` and `1, 5, 3, 9`. we may need to change this function to accommodate `dipole` shape.

We have to use [`torch.permute`](https://pytorch.org/docs/stable/generated/torch.permute.html) to change the output shape, so it can be compatible with the current DPLR code. (obviously, we don't want to rewrite these codes or break the existing TF models)

For how OutDef defines, I will follow @wanghan-iapcm's suggestion.

njzjz:
> this will return `0`, since get_sel_type() returns an empty list in GeneralFitting.

`_get_sel_natoms` needs to handle this particular situation.

deepmd/pt/model/model/dipole_model.py
```diff
@@ -0,0 +1,89 @@
...
+            mapping,
+            fparam=fparam,
+            aparam=aparam,
+            do_atomic_virial=do_atomic_virial,
+        )
+        if self.fitting_net is not None:
+            model_predict = {}
+            model_predict["dipole"] = model_ret["dipole"]
+            model_predict["global_dipole"] = model_ret["dipole_redu"]
+            if self.do_grad("dipole"):
```
wanghan-iapcm, line 80: here we should distinguish the do_grad_r and do_grad_c from do_grad. I will contribute.

anyangml: I think we might have to do grad for both c and r. In the comment I made above, `request_defs` and `out` may contain different number of elements, which results in a name-value mismatch.

```
return dict( 
     zip( 
         [x.name for x in request_defs], 
         out, 
     ) 
 ) 
 ```

njzjz (approved):

wanghan-iapcm (approved):

PR #3314, opened 2024-02-21
Fix: float32 bug in UTs

anyangml:


deepmd/pt/model/network/mlp.py
njzjz, line 1: This might be not correct, since the tensor is not converted back to the original precision after this module.

njzjz: You may check whether the output value is float64 or float32

anyangml: so the output should always be `float64` or whatever the precision  of `xx` originally is?

njzjz: The precision here should only be applied to the neural networks, but not the output (such as energy, which usually has lots of digit)

njzjz (approved):

wanghan-iapcm (approved):

PR #3315, opened 2024-02-21
pt: fix se_e2_a precision cast

njzjz:
The output should cast back to the global precision.

### DIFF
--- a/source/tests/consistent/descriptor/test_se_e2_a.py
+++ b/source/tests/consistent/descriptor/test_se_e2_a.py
@@ -39,6 +39,7 @@ from deepmd.utils.argcheck import (
     (True, False),  # resnet_dt
     (True, False),  # type_one_side
     ([], [[0, 1]]),  # excluded_types
+    ("float32", "float64"),  # precision
 )
 class TestSeA(CommonTest, DescriptorTest, unittest.TestCase):
     @property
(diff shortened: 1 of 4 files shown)


wanghan-iapcm (approved):

iProzd (approved):

source/tests/consistent/descriptor/test_se_e2_a.py
```diff
@@ -147,3 +152,35 @@ def eval_pt(self, pt_obj: Any) -> Any:
...
+            type_one_side,
+            excluded_types,
+            precision,
+        ) = self.param
+        if precision == "float64":
+            return 1e-10
+        elif precision == "float32":
+            return 1e-4
+        else:
+            raise ValueError(f"Unknown precision: {precision}")
```
iProzd, line 186: These two properties not used yet?

iProzd (approved):

source/tests/consistent/descriptor/test_se_e2_a.py
njzjz: It's used here: https://github.com/deepmodeling/deepmd-kit/blob/604679c8a38eaa2403a9233c8e4d3a6c54792617/source/tests/consistent/common.py#L257

PR #3327, opened 2024-02-23
Feat: add polar consistency test

anyangml:
This PR is to add cross framework consistency test on PolarFittingNet.

Note: `shift_diag` not yet implemented in PT.

### DIFF
--- a/deepmd/utils/argcheck.py
+++ b/deepmd/utils/argcheck.py
@@ -718,9 +718,7 @@ def fitting_polar():
         Argument("resnet_dt", bool, optional=True, default=True, doc=doc_resnet_dt),
         Argument("precision", str, optional=True, default="default", doc=doc_precision),
         Argument("fit_diag", bool, optional=True, default=True, doc=doc_fit_diag),
-        Argument(
-            "scale", [List[float], float], optional=True, default=1.0, doc=doc_scale
-        ),
+        Argument("scale", List[float], optional=True, default=None, doc=doc_scale),
         # Argument("diag_shift", [list,float], optional = True, default = 0.0, doc = doc_diag_shift),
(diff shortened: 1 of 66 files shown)


source/tests/consistent/fitting/test_polar.py
```diff
@@ -0,0 +1,192 @@
...
+        (
+            resnet_dt,
+            precision,
+            mixed_types,
+        ) = self.param
+        if precision == "float64":
+            return 1e-10
+        elif precision == "float32":
+            return 1e-3
+        else:
```
anyangml, line 176: won't pass `1e-4`

deepmd/utils/argcheck.py
```diff
@@ -718,9 +718,7 @@ def fitting_polar():
         Argument("resnet_dt", bool, optional=True, default=True, doc=doc_resnet_dt),
         Argument("precision", str, optional=True, default="default", doc=doc_precision),
         Argument("fit_diag", bool, optional=True, default=True, doc=doc_fit_diag),
-        Argument(
-            "scale", [List[float], float], optional=True, default=1.0, doc=doc_scale
-        ),
+        Argument("scale", List[float], optional=True, default=None, doc=doc_scale),
```
njzjz, line 721: Is there any reason to change it? The argument is exposed to users. Changing it will cause a breaking change.

njzjz: It's okay if `None` is equivalent to `1.0`, but it's a breaking change to disallow `float`.

anyangml: I see, I will update torch to match this behavior.

njzjz (approved):

wanghan-iapcm (approved):

PR #3330, opened 2024-02-24
pt: remove env.DEVICE in all `forward` functions

njzjz:
Ensure the saved JIT model can run on both CPUs and GPUs.

### DIFF
--- a/source/tests/pt/model/test_deeppot.py
+++ b/source/tests/pt/model/test_deeppot.py
@@ -115,3 +115,11 @@ class TestDeepPotFrozen(TestDeepPot):
         )
         freeze(ns)
         self.model = frozen_model
+
+    # Note: this can not actually disable cuda device to be used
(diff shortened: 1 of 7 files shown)


source/tests/pt/model/test_deeppot.py
```diff
@@ -115,3 +115,11 @@ def setUp(self):
         )
         freeze(ns)
         self.model = frozen_model
+
+    # Note: this can not actually disable cuda device to be used
+    # only can be used to test whether devices are dismatched
```
caic99, line 120:
```suggestion
    # only can be used to test whether devices are mismatched
```

caic99 (approved):

wanghan-iapcm (approved):

iProzd (approved):

CaRoLZhangxy (approved):

anyangml (approved):

PR #3301, opened 2024-02-19
Support DPSpin for AtomicModel

iProzd:
This PR support DPSpin for AtomicModel:

- [x] `Spin` base class to handle spin-related information.

- [x] `SpinModel` implementation in pt, which can generally take any backbone model, with input and output process for output like `energy`, `dipole` and etc. 

- [x] `SpinEnergyModel` implementation inherited from `SpinModel` for `energy` calclulation.

- [x] Input improvement, now DPSpin model can take real `spin` as input with `vitual_scale`, which can support much more flexible input. (Also with dataset support for `spin` input.)

- [x] `EnergySpinLoss` class to process `energy`, `force_real`, `force_mag`, `virial`.

- [x] Add `protection` for environment calculations.

TODO:
- [x] Support `forward_lower` for `SpinModel`.
- [x] Add `data_stat` with general `exclude_types` for descriptors and fittings.
- [x] Add examples and UTs.
- [x] numpy model
- [x] Support multi-task training for `SpinModel` models.

Will fix in next PRs:
- tf consistency test (?)

PR #3335, opened 2024-02-25
add BaseModel; store type in serialization

njzjz:


deepmd/dpmodel/model/base_model.py
```diff
@@ -0,0 +1,160 @@
...
+from deepmd.utils.plugin import (
+    Plugin,
+)
+
+
+class BaseBaseModel(ABC):
+    """Base class for final exported model that will be directly used for inference.
+
+    The class defines some abstractmethods that will be directly called by the
+    inference interface. If the final model class inherbits some of those methods
```
wanghan-iapcm, line 25:
```suggestion
    inference interface. If the final model class inherits some of those methods
```

deepmd/pt/model/model/model.py
```diff
@@ -36,3 +149,22 @@ def compute_or_load_stat(
...
+    @abstractmethod
+    def get_nnei(self) -> int:
+        """Returns the total number of selected neighboring atoms in the cut-off radius."""
+        # for C++ interface
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_nsel(self) -> int:
+        """Returns the total number of selected neighboring atoms in the cut-off radius."""
+        raise NotImplementedError
```
wanghan-iapcm, line 170:
```suggestion
        pass
```

deepmd/pt/model/model/model.py
```diff
@@ -36,3 +149,22 @@ def compute_or_load_stat(
...
+    @abstractmethod
+    def get_model_def_script(self) -> str:
+        """Get the model definition script."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_nnei(self) -> int:
+        """Returns the total number of selected neighboring atoms in the cut-off radius."""
+        # for C++ interface
+        raise NotImplementedError
```
wanghan-iapcm, line 165:
```suggestion
              pass
```

deepmd/pt/model/model/model.py
```diff
@@ -36,3 +149,22 @@ def compute_or_load_stat(
...
         """
         raise NotImplementedError
+
+    model_def_script: str
+
+    # currently, only pt needs the following methods
+    @abstractmethod
+    def get_model_def_script(self) -> str:
+        """Get the model definition script."""
+        raise NotImplementedError
```
wanghan-iapcm, line 159:
```suggestion
        pass
```

wanghan-iapcm (commented):
- I do not understand why the `serialization` and `deserialization` cannot be put int `BaseBaseModel`. 
- would the following inheritance work?
```
#       torch.nn.Module        BaseAtomicModel        BaseBaseModel
#             |                       |                    |
#             -------------------------                    |
#                         |                                |
#                    DPAtomicModel                         |
#                         |                                |
#                make_model(DPAtomicModel)                 |
#                         |                                |
#                         ----------------------------------
#                                           |
#                                         DPModel
```
- in our concept a model is an atomic model, therefore, the base model seems to be a facilitate that provides the plugin, other methods like the `get_rcut`, `get_dim_fparam`... seems to be inherited from atomic model rather than model .

njzjz:
> * I do not understand why the `serialization` and `deserialization` cannot be put int `BaseBaseModel`.

Done.

> * would the following inheritance work?
> 
> ```
> #       torch.nn.Module        BaseAtomicModel        BaseBaseModel
> #             |                       |                    |
> #             -------------------------                    |
> #                         |                                |
> #                    DPAtomicModel                         |
> #                         |                                |
> #                make_model(DPAtomicModel)                 |
> #                         |                                |
> #                         ----------------------------------
> #                                           |
> #                                         DPModel
> ```
> 

We cannot store the plugins of dpmodel and pt in the same class. There must be a class to initialize the plugin separately.

> * in our concept a model is an atomic model, therefore, the base model seems to be a facilitate that provides the plugin, other methods like the `get_rcut`, `get_dim_fparam`... seems to be inherited from atomic model rather than model .

It looks to me `SpinModel` in #3301 is not inherited from an atomic model.

deepmd/pt/model/model/model.py
```diff
@@ -1,16 +1,111 @@
...
+    Plugin,
+)
 
 
-class BaseModel(torch.nn.Module):
+# trick: torch.nn.Module should not be inherbited here, otherwise,
+# the abstract method will override the method from the atomic model
+# as Python resolves method lookups using the C3 linearisation.
+# See https://stackoverflow.com/a/47117600/9567349
+# Take an example, this is sitatuion for only inherbiting BaseBaseModel:
```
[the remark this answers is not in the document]
njzjz:
```suggestion
# Take an example, this is the situation for only inheriting BaseBaseModel:
```

deepmd/pt/model/model/model.py
```diff
@@ -1,16 +1,111 @@
...
+#                    DPAtomicModel                      BaseModel
+#                         |                                |
+#                make_model(DPAtomicModel)                 |
+#                         |                                |
+#                         ----------------------------------
+#                                           |
+#                                         DPModel
+#
+# The order is: DPModel -> make_model(DPAtomicModel) -> DPAtomicModel ->
+# torch.nn.Module -> BaseAtomicModel -> BaseBaseModel
```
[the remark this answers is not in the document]
njzjz:
```suggestion
# torch.nn.Module -> BaseAtomicModel -> BaseModel -> BaseBaseModel
```

wanghan-iapcm:
> We cannot store the plugins of dpmodel and pt in the same class. There must be seperately classes to initialize the plugin separately.

it should be possible to `make_base_model` for different backends, see #3340

deepmd/dpmodel/model/base_model.py
```diff
@@ -6,136 +6,142 @@
...
-        pass
-
-    @abstractmethod
-    def get_type_map(self) -> List[str]:
-        """Get the type map."""
-
-    @abstractmethod
-    def get_rcut(self):
-        """Get the cut-off radius."""
+def make_base_model() -> Type[object]:
```
njzjz, line 18: Added `make_base_model`.

deepmd/utils/plugin.py
```diff
@@ -93,3 +96,60 @@ class PluginVariant(metaclass=VariantABCMeta):
     """A class to remove `type` from input arguments."""
 
     pass
+
+
+def make_plugin_registry(name: Optional[str] = None) -> Type[object]:
```
njzjz, line 101: I also added a `make_plugin_registry` method and will apply it to other classes after this PR.

deepmd/dpmodel/model/base_model.py
```diff
@@ -151,38 +157,6 @@ class BaseModel(BaseBaseModel):
...
-            cls = cls.get_class_by_type(j_get_type(kwargs, cls.__name__))
-        return super().__new__(cls)
-
-    @classmethod
-    def get_class_by_type(cls, model_type: str) -> Type["BaseModel"]:
-        if model_type in BaseModel.__plugins.plugins:
-            return BaseModel.__plugins.plugins[model_type]
-        else:
-            raise RuntimeError("Unknown model type: " + model_type)
+    def get_model_def_script(self) -> str:
```
wanghan-iapcm, line 160: The method may be moved to make base atomic model, the same designed as the pt backend.

wanghan-iapcm (approved):

PR #3339, opened 2024-02-26
feat(pt/dpmodel): support type_one_side in se_e2_a

njzjz:


deepmd/dpmodel/descriptor/se_e2_a.py
```diff
@@ -313,6 +327,17 @@ def call(
...
+            for embedding_idx in itertools.product(range(self.ntypes), repeat=2):
+                # not actually used; to match serilization data from TF to pass the test
+                if embedding_idx in self.emask:
+                    for ilayer in range(len(self.neuron)):
+                        layer = self.embeddings[embedding_idx][ilayer]
+                        layer.w.fill(0.0)
+                        layer.b.fill(0.0)
+                        if layer.idt is not None:
+                            layer.idt.fill(0.0)
+
```
wanghan-iapcm, line 340: ugly code. can tf avoids such serialization?

wanghan-iapcm (approved):

deepmd/dpmodel/descriptor/se_e2_a.py
wanghan-iapcm: or the hacking code should be moved to network collection?

wanghan-iapcm: why we did not see the hack in pt...

njzjz: PT doesn't compare with TF directly.

njzjz: DP and PT will store this network, although it is never used. TF will lose the information after deserializing + serializing.

njzjz: I move it to `Network.clear()`. See b3160c878aaed8187cc60589bdcd70f7be872dda.

PR #3340, opened 2024-02-26
refact: pt: mv all plugin support to base descriptor. 

wanghan-iapcm:
thus pt reusing the dp code.

### DIFF
--- a/deepmd/dpmodel/descriptor/make_base_descriptor.py
+++ b/deepmd/dpmodel/descriptor/make_base_descriptor.py
@@ -156,26 +156,6 @@ def make_base_descriptor(
                 return BD.get_class_by_type(data["type"]).deserialize(data)
             raise NotImplementedError("Not implemented in class %s" % cls.__name__)
 
-        @classmethod
-        def get_data_process_key(cls, config):
-            """
-            Get the keys for the data preprocess.
-            Usually need the information of rcut and sel.
-            TODO Need to be deprecated when the dataloader has been cleaned up.
-            """
-            if cls is not BD:
-                raise NotImplementedError("get_data_process_key is not implemented!")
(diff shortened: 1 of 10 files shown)


deepmd/dpmodel/descriptor/make_base_descriptor.py
```diff
@@ -156,6 +156,26 @@ def deserialize(cls, data: dict) -> "BD":
                 return BD.get_class_by_type(data["type"]).deserialize(data)
             raise NotImplementedError("Not implemented in class %s" % cls.__name__)
 
+        @classmethod
+        def get_data_process_key(cls, config):
+            """
+            Get the keys for the data preprocess.
+            Usually need the information of rcut and sel.
+            TODO Need to be deprecated when the dataloader has been cleaned up.
```
njzjz, line 164: Is it still be used?

njzjz (approved):

deepmd/dpmodel/descriptor/make_base_descriptor.py
njzjz: I think no

deepmd/dpmodel/descriptor/make_base_descriptor.py
```diff
@@ -156,6 +156,26 @@ def deserialize(cls, data: dict) -> "BD":
...
+            """
+            if cls is not BD:
+                raise NotImplementedError("get_data_process_key is not implemented!")
+            descrpt_type = config["type"]
+            return BD.__plugins.plugins[descrpt_type].get_data_process_key(config)
+
+        @property
+        def data_stat_key(self):
```
njzjz, line 172: This method is not used as well.

PR #3348, opened 2024-02-27
update sel

njzjz:


deepmd/pt/model/descriptor/dpa2.py
```diff
@@ -396,3 +399,32 @@ def forward(
...
+            The local data refer to the current class
+        """
+        local_jdata_cpy = local_jdata.copy()
+        update_sel = UpdateSel()
+        local_jdata_cpy = update_sel.update_one_sel(
+            global_jdata,
+            local_jdata_cpy,
+            True,
+            rcut_key="repinit_rcut",
+            sel_key="repinit_nsel",
```
wanghan-iapcm, line 421: the interface of update_sel may be refactorized. left for a future PR.

wanghan-iapcm (approved):

deepmd/pt/model/descriptor/dpa2.py
njzjz: Perhaps the interface of dpa2 should be refactorized. For example,

```json
{
  "type": "dpa2",
  "repinit": {
    "rcut": 6,
    "rcut_smth": 0.5,
    "sel": "auto",
  },
  "repformer": {
    "rcut": 6,
    "rcut_smth": 0.5,
    "sel": "auto",
  }
}
```

njzjz: cc @iProzd

iProzd (approved):

deepmd/pt/model/descriptor/dpa2.py
wanghan-iapcm: the input output of update_sel are dicts, which is not a good design.

PR #3349, opened 2024-02-27
add @version to serialization data

njzjz:
At this time, all current versions and the minimal and maximum versions are set to 1. If any changes, whether breaking or not, are made to the serialization data, the version should be bumped.


wanghan-iapcm (approved):

wanghan-iapcm:
What kind of convention shall we adapt to numbering the model version? 
shall we support separate version control for fine-grained modules?  imagine the case that the descriptor of a model may be used to initialized the descriptor of another model.

njzjz:
> What kind of convention shall we adapt to numbering the model version?

The version should be bumped when compatibility is broken with either **forward** or **backward**.

> shall we support separate version control for fine-grained modules? imagine the case that the descriptor of a model may be used to initialized the descriptor of another model.

I think we should support separate versions for each `serialize` and `deserialize` method, which is the most convenient for developers.

PR #3351, opened 2024-02-28
Fix: se_r prod_env_mat

anyangml:
This should fix the bug.
image

### DIFF
--- a/deepmd/pt/utils/env_mat_stat.py
+++ b/deepmd/pt/utils/env_mat_stat.py
@@ -101,6 +101,12 @@ class EnvMatStatSe(EnvMatStat):
             dtype=env.GLOBAL_PT_FLOAT_PRECISION,
             device=env.DEVICE,
         )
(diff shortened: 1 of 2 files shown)


deepmd/pt/utils/env_mat_stat.py
```diff
@@ -101,6 +101,14 @@ def iter(
...
             device=env.DEVICE,
         )
+        if self.last_dim == 4:
+            radial_only = False
+        elif self.last_dim == 1:
+            radial_only = True
+        else:
+            raise ValueError(
+                "last_dim should be 1 for raial-only or 4 for full descriptor."
+            )
```
anyangml, line 111: assuming all the data share the same type of descriptor, otherwise, we need to move this into the for loop.

njzjz (commented):
Could you add a test to cover the line you added?

njzjz (approved):

PR #3357, opened 2024-02-28
feat: support exclude atypes in atomic model

wanghan-iapcm:
With this pr, we do not need to implement the exclude types for each descriptor and fitting.


njzjz (approved):

deepmd/utils/argcheck.py
```diff
@@ -866,6 +866,9 @@ def model_args(exclude_hybrid=False):
     doc_srtab_add_bias = "Whether add energy bias from the statistics of the data to short-range tabulated atomic energy. It only takes effect when `use_srtab` is provided."
     doc_compress_config = "Model compression configurations"
     doc_spin = "The settings for systems with spin."
+    doc_atom_exclude_types = "Exclude the atomic contribution of the listed atom types"
+    doc_pair_exclude_types = "The atom pairs of the listed types are not treated to be neighbors, i.e. they do not see each other."
```
iProzd, line 870: Add (Supported Backend: PyTorch) after #3342 merged.

iProzd (approved):

deepmd/utils/argcheck.py
njzjz: I clicked the merge button..

PR #3361, opened 2024-02-28
pt: fix se_a type_one_side performance degradation

njzjz:
The code in this PR is ugly, but applying a mask is causing performance degradation for ~3 ms/step.

When applying a mask, `aten::nonzero` has a high host time, as it causes host-device synchronization:
image

After fixing:
image

See https://github.com/pytorch/pytorch/issues/12461 for more information.


wanghan-iapcm (approved):

wanghan-iapcm:
Is the exclude_mask has side effect on the performance?

njzjz:
> Is the exclude_mask has side effect on the performance? 

It has a different problem. It seems the integer index is slow: https://github.com/pytorch/pytorch/issues/15245 I haven't tested it.

PR #3362, opened 2024-02-29
Chore: Update doc for PyTorch

anyangml:
This PR is to perform regression test on `PT backend` against `TF backend`, and update doc to provide instructions for using `PT backend`.

### Regression Test on EnergyFitting
image
image
image

### DIFF
--- a/doc/freeze/freeze.md
+++ b/doc/freeze/freeze.md
@@ -2,9 +2,13 @@
 
 The trained neural network is extracted from a checkpoint and dumped into a protobuf(.pb) file. This process is called "freezing" a model. The idea and part of our code are from [Morgan](https://blog.metaflow.fr/tensorflow-how-to-freeze-a-model-and-serve-it-with-a-python-api-d4f3596b3adc). To freeze a model, typically one does
 ```bash
-$ dp freeze -o graph.pb
+$ dp freeze -o model.pb
 ```
-in the folder where the model is trained. The output model is called `graph.pb`.
+or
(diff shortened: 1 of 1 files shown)


njzjz (approved):

doc/model/train-se-atten.md
```diff
@@ -98,6 +104,43 @@ An example of the DPA-1 descriptor is provided as follows
...
+          "rcut":		6.00,
+          "sel":		120,
+          "neuron":		[25, 50, 100],
+          "tebd_dim": 8,
+          "axis_neuron":	16,
+          "attn":	128,
+          "attn_layer":	2,
+          "attn_mask": false,
+          "attn_dotr": true,
+          "post_ln": true,
```
anyangml, line 124: @iProzd take a look at the paras

doc/model/train-se-atten.md
```diff
@@ -98,6 +104,43 @@ An example of the DPA-1 descriptor is provided as follows
...
+	}
+```
+
+* The {ref}`type <model/descriptor/type>` of the descriptor is set to `"se_atten"`, which will use DPA-1 structures.
```
iProzd, line 128: set to "dpa1"

doc/model/train-se-atten.md
njzjz, line 124: This is not a valid JSON - `,` after the last parameter is disallowed.

njzjz (approved):

PR #3365, opened 2024-02-29
Hybrid descriptor

njzjz:


deepmd/dpmodel/descriptor/hybrid.py
```diff
@@ -0,0 +1,186 @@
...
+            elif isinstance(ii, dict):
+                formatted_descript_list.append(BaseDescriptor(**ii))
+            else:
+                raise NotImplementedError
+        self.descrpt_list = formatted_descript_list
+        self.numb_descrpt = len(self.descrpt_list)
+        for ii in range(1, self.numb_descrpt):
+            assert (
+                self.descrpt_list[ii].get_ntypes() == self.descrpt_list[0].get_ntypes()
+            ), f"number of atom types in {ii}th descrptor does not match others"
```
anyangml, line 57: maybe it's more informative to show the descriptor type in the error msg, something like `descriptor.__class__.__name__`

deepmd/dpmodel/descriptor/hybrid.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+@BaseDescriptor.register("hybrid")
+class DescrptHybrid(BaseDescriptor, NativeOP):
+    """Concate a list of descriptors to form a new descriptor.
+
+    Parameters
+    ----------
+    list : list
+            Build a descriptor from the concatenation of the list of descriptors.
```
wanghan-iapcm, line 30: please explicitly say what kind of object is allow as the list item.

deepmd/dpmodel/descriptor/hybrid.py
```diff
@@ -0,0 +1,186 @@
...
+            self.nlist_cut_idx.append(cut_idx)
+
+    def get_rcut(self) -> float:
+        """Returns the cut-off radius."""
+        return np.max([descrpt.get_rcut() for descrpt in self.descrpt_list]).item()
+
+    def get_sel(self) -> List[int]:
+        """Returns the number of selected atoms for each type."""
+        return np.max(
+            [descrpt.get_sel() for descrpt in self.descrpt_list], axis=0
```
wanghan-iapcm, line 77: does it handle the case that some of the descriptors are `mixed_type` while others are not?

deepmd/dpmodel/descriptor/hybrid.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def get_dim_emb(self) -> int:
+        """Returns the output dimension."""
+        return np.sum([descrpt.get_dim_emb() for descrpt in self.descrpt_list]).item()
+
+    def mixed_types(self):
+        """Returns if the descriptor requires a neighbor list that distinguish different
+        atomic types or not.
+        """
+        return all(descrpt.mixed_types() for descrpt in self.descrpt_list)
```
wanghan-iapcm, line 96: should not be `any`? the mixed type nlist would be formatted anyway for a type distinguished nlist.

deepmd/dpmodel/descriptor/hybrid.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        out_descriptor = []
+        for descrpt, nci in zip(self.descrpt_list, self.nlist_cut_idx):
+            # cut the nlist to the correct length
+            odescriptor, _, _, _, _ = descrpt(
+                coord_ext, atype_ext, nlist[:, :, nci], mapping
+            )
+            out_descriptor.append(odescriptor)
+        out_descriptor = np.concatenate(out_descriptor, axis=-1)
+        return out_descriptor, None, None, None, None
```
wanghan-iapcm, line 147: plz check if they are correct 
gr: concate along the -2 dim if not None
g2: concate along the -1 dim if not None
h2: use the one returned by largest rcut descriptor if not None
sw: use the one returned by largest rcut descriptor if not None

wanghan-iapcm (changes requested):

deepmd/dpmodel/descriptor/hybrid.py
njzjz: It might behave differently from the theory when the `sel` is not enough. Considering a simple case, an atom has 4 type-0 neighbors and 2 type-1 neighbors, in the order of [1,0,0,0,0,1] (assume the indexes of these neighbors are 0~5). Descript 0 requests sel=[2, 3], and Descriptor 1 requests nsel=5. The expected neighbor indexes for them are `[1, 2, 0, 5, -1]` and `[0, 1, 2, 3, 4]`, respectively. When we request 5 neighbors (`nsel` of both them is 5), we cannot cover them.

If `sel` is enough, it is not a problem.

However, I noticed that the current `nlist_distinguish_types` has the same behavior... (meaning different behavior from the theory), so it might be okay.

wanghan-iapcm: I agree, the distinguished type is not well supported by the pt backend.

wanghan-iapcm (approved):
The `mixed_types` and `get_sel` methods of the linear combination atomic model may be want to aligned with the behavior of the hybrid descriptor. @anyangml

PR #3370, opened 2024-03-01
feat: atom_ener in energy fitting

njzjz:
Also, fix the TF serialization issue (it tried to store a tensor instead of a NumPy array).

### DIFF
--- a/deepmd/dpmodel/fitting/general_fitting.py
+++ b/deepmd/dpmodel/fitting/general_fitting.py
@@ -75,8 +75,8 @@ class GeneralFitting(NativeOP, BaseFitting):
             Atomic contributions of the excluded atom types are set zero.
     remove_vaccum_contribution: List[bool], optional
         Remove vaccum contribution before the bias is added. The list assigned each
-        type. If not mixed_types, only remove the vaccum contribution for the atom types
-        in the list.
+        type. For `mixed_types` provide `[True]`, otherwise it should be a list of the same
+        length as `ntypes` signaling if or not removing the vaccum contribution for the atom types in the list.
     """
 
     def __init__(
--- a/deepmd/dpmodel/fitting/general_fitting.py
+++ b/deepmd/dpmodel/fitting/general_fitting.py
@@ -8,7 +8,6 @@ from typing import (
     Dict,
     List,
     Optional,
-    Union,
 )
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


wanghan-iapcm (approved):

deepmd/dpmodel/fitting/general_fitting.py
```diff
@@ -73,7 +73,10 @@ class GeneralFitting(NativeOP, BaseFitting):
             different fitting nets for different atom types.
     exclude_types: List[int]
             Atomic contributions of the excluded atom types are set zero.
-
+    remove_vaccum_contribution: List[bool], optional
```
wanghan-iapcm, line 76: how should one write the `remove_vaccum_contribution` if `mixed_types`?  like [True]?  It seems that True is better.

njzjz: This option is for internal use, called by the energy fitting (which users define `atom_ener`), so only the developers will touch it.

`torch.jit` doesn't like `Union[bool, List[bool]]`.

deepmd/dpmodel/fitting/general_fitting.py
```diff
@@ -73,7 +73,10 @@ class GeneralFitting(NativeOP, BaseFitting):
...
+        Remove vaccum contribution before the bias is added. The list assigned each
+        type. If not mixed_types, only remove the vaccum contribution for the atom types
+        in the list.
```
wanghan-iapcm, line 79:
```suggestion
        Remove vaccum contribution before the bias is added. The list assigned each
        type. For `mixed_types` provide `[True]`, otherwise it should be a list of the same length as `ntypes` signaling if or not removing the vaccum contribution for the atom types
        in the list.
```

deepmd/pt/model/task/fitting.py
```diff
@@ -246,7 +246,10 @@ class GeneralFitting(Fitting):
         Random seed.
     exclude_types: List[int]
         Atomic contributions of the excluded atom types are set zero.
-
+    remove_vaccum_contribution: List[bool], optional
+        Remove vaccum contribution before the bias is added. The list assigned each
+        type. If not mixed_types, only remove the vaccum contribution for the atom types
+        in the list.
```
njzjz, line 252:
```suggestion
        Remove vaccum contribution before the bias is added. The list assigned each
        type. For `mixed_types` provide `[True]`, otherwise it should be a list of the same
        length as `ntypes` signaling if or not removing the vaccum contribution for the atom types in the list.
```

PR #3371, opened 2024-03-01
Add trainable settings for pt

iProzd:


njzjz (approved):

wanghan-iapcm (commented):
Could you please add UT for this parameter?

deepmd/pt/model/task/fitting.py
```diff
@@ -282,6 +285,11 @@ def __init__(
         self.rcond = rcond
         # order matters, should be place after the assignment of ntypes
         self.reinit_exclude(exclude_types)
+        self.trainable = trainable
+        # need support for each layer settings
+        self.trainable = (
+            all(self.trainable) if isinstance(self.trainable, list) else self.trainable
```
wanghan-iapcm, line 291: doc and type hint should be updated.  The limitation of the current support should be mentioned in doc.

wanghan-iapcm (approved):

PR #3375, opened 2024-03-01
pt: add real atoms select in c++ interface

CaRoLZhangxy:
select real atoms may be merged in DeepPot class for both backends in the future.


njzjz (approved):
LGTM

wanghan-iapcm (approved):
LGTM

PR #3378, opened 2024-03-01
Merge `change_energy_bias` and fix finetune

iProzd:


njzjz (approved):

wanghan-iapcm (commented):
how do you validate that the change_energy_bias  works as you want? How is it covered by UT?

wanghan-iapcm (approved):

PR #3393, opened 2024-03-02
throw errros if rmin is no less than rmax

njzjz:
when rmin==rmax, the previous implementation of compute_smooth_weight will give all nan. In theory, it should not happen.


wanghan-iapcm (commented):
an error should raise when rcut_smth >= rcut.

wanghan-iapcm (approved):

wanghan-iapcm:
should be fixed by #3413

PR #3398, opened 2024-03-03
Feat: add zbl training

anyangml:


deepmd/pt/model/atomic_model/linear_atomic_model.py
```diff
@@ -307,6 +310,34 @@ def __init__(
...
+        stat_file_path
+            The dictionary of paths to the statistics files.
+        """
+        if stat_file_path is not None and self.type_map is not None:
+            # descriptors and fitting net with different type_map
+            # should not share the same parameters
+            stat_file_path /= " ".join(self.type_map)
+        self.dp_model.descriptor.compute_input_stats(sampled_func, stat_file_path)
+        if self.dp_model.fitting_net is not None:
+            self.dp_model.fitting_net.compute_output_stats(sampled_func, stat_file_path)
```
njzjz, line 339: Will this work?
self.dp_model.compute_or_load(sampled_func, stat_file_path)

deepmd/pt/model/atomic_model/pairtab_atomic_model.py
```diff
@@ -154,6 +169,65 @@ def deserialize(cls, data) -> "PairTabAtomicModel":
         tab_model.register_buffer("tab_data", torch.from_numpy(tab_model.tab.tab_data))
         return tab_model
 
+    def compute_output_stats(
+        self,
```
anyangml, line 173: I am not sure about this. @wanghan-iapcm

deepmd/pt/model/atomic_model/pairtab_atomic_model.py
```diff
@@ -154,6 +169,65 @@ def deserialize(cls, data) -> "PairTabAtomicModel":
...
+            # shape: (nframes, ndim)
+            merged_energy = to_numpy_array(torch.cat(energy))
+            # shape: (nframes, ntypes)
+            merged_natoms = to_numpy_array(torch.cat(input_natoms)[:, 2:])
+
+            bias_atom_e, _ = compute_stats_from_redu(
+                merged_energy,
+                merged_natoms,
+                assigned_bias=None,
+                rcond=None,
```
anyangml, line 218: are we going to add these inputs to PairtabAtomicModel?

deepmd/pt/model/atomic_model/pairtab_atomic_model.py
```diff
@@ -70,6 +84,7 @@ def __init__(
         else:
             self.register_buffer("tab_info", None)
             self.register_buffer("tab_data", None)
+        self.bias_atom_e = None
```
anyangml, line 87: It seems I need to create a placeholder for `bias_atom_e` to make jit happy, that requires knowing `ntypes`. In the case where we don't have input file, I am not sure what we be the best way to get that information, may be add as user input?

source/tests/pt/test_finetune.py
```diff
@@ -109,6 +110,21 @@ def tearDown(self) -> None:
         FinetuneTest.tearDown(self)
 
 
+class TestEnergyZBLModelSeA(unittest.TestCase, FinetuneTest):
+    def setUp(self):
```
anyangml, line 114: need to fix this.

deepmd/pt/model/atomic_model/pairtab_atomic_model.py
```diff
@@ -65,11 +79,16 @@ def __init__(
...
             self.register_buffer("tab_info", torch.from_numpy(tab_info))
-            self.register_buffer("tab_data", torch.from_numpy(tab_data))
+            self.register_buffer(
+                "tab_data",
+                torch.from_numpy(tab_data).reshape(ntypes, ntypes, nspline, 4),
+            )
         else:
             self.register_buffer("tab_info", None)
             self.register_buffer("tab_data", None)
+        self.bias_atom_e = None
```
anyangml, line 91: it's probably better to create a placeholder with a proper shape `ntypes, 1`. However, in the case we don't have an valid input file, the ntype is passed in via deserialization after initialization. We might need a way to fetch that information during initialization.

source/tests/pt/test_finetune.py
```diff
@@ -30,20 +32,47 @@
...
+            energy_bias_before = deepcopy(
+                model.fitting_net.bias_atom_e.detach().cpu().numpy().reshape(-1)
+            )
+            bias_atom_e_input = deepcopy(
+                model.fitting_net.bias_atom_e.detach().cpu().numpy().reshape(-1)
+            )
+        elif isinstance(model, DPZBLModel):
+            model.dp_model.fitting_net.bias_atom_e = torch.rand_like(
+                model.dp_model.fitting_net.bias_atom_e
+            )
```
anyangml, line 59: currently, pairtab. bias_atom_e is in initialized to None, we better initialize it to a proper shape.

deepmd/pt/model/atomic_model/pairtab_atomic_model.py
wanghan-iapcm: 1. At the atomic_model level, one needs to implement the `compute_or_load_stat`. 
2. We may need to add the abstract method to the `atomic_model`. 
3. The torch implementation of energy stat seems to be abstracted to some level rather than copying the code.

deepmd/pt/model/atomic_model/pairtab_atomic_model.py
wanghan-iapcm: Yes, they seem to be necessary

source/tests/pt/test_finetune.py
wanghan-iapcm: Yes, we may have to consider the case that the type maps of different model are distinct..

deepmd/pt/model/atomic_model/pairtab_atomic_model.py
wanghan-iapcm: We may want to ask the user to pass the type_map to `__init__`

deepmd/pt/train/training.py
```diff
@@ -247,7 +249,10 @@ def get_sample():
         def get_single_model(
             _model_params,
         ):
-            model = get_model(deepcopy(_model_params)).to(DEVICE)
+            if "use_srtab" in _model_params:
+                model = get_zbl_model(deepcopy(_model_params)).to(DEVICE)
+            else:
+                model = get_model(deepcopy(_model_params)).to(DEVICE)
```
iProzd, line 255: Maybe we use only one func called get_model, and wrap all these judgement into get_model.

source/tests/pt/test_finetune.py
```diff
@@ -109,6 +138,22 @@
...
 
 
+## disabled for now.
+# class TestEnergyZBLModelSeA(unittest.TestCase, FinetuneTest):
+#     def setUp(self):
+#         self.data_file = [str(Path(__file__).parent / "water/data/data_0")]
+#         self.data = DeepmdDataSystem(
+#             self.data_file,
+#             batch_size=1,
+#             test_size=1,
```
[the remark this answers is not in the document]
njzjz: Use skip to disable a test.

deepmd/pt/model/atomic_model/linear_atomic_model.py
```diff
@@ -198,16 +202,56 @@ def forward_atomic(
...
-                "energy": torch.sum(
-                    torch.stack(ener_list) * torch.stack(weights), dim=0
-                ),
-            }  # (nframes, nloc, 1)
+        atype = extended_atype[:, :nloc]
+        for idx, model in enumerate(self.models):
+            if isinstance(model, DPAtomicModel):
+                bias_atom_e = model.fitting_net.bias_atom_e
+            elif isinstance(model, PairTabAtomicModel):
+                bias_atom_e = model.bias_atom_e
```
wanghan-iapcm, line 210: It seems that we need to provide interfaces for atomic models to read/manipulate bias_atom_e

njzjz (approved):

deepmd/pt/model/atomic_model/linear_atomic_model.py
njzjz: It may be marked with the `TODO` tag

deepmd/pt/model/atomic_model/linear_atomic_model.py
```diff
@@ -185,10 +188,12 @@ def forward_atomic(
         ener_list = []
 
         for i, model in enumerate(self.models):
+            ori_map = model.get_type_map()
+            updated_atype = self.remap_atype(extended_atype, ori_map, self.type_map)
```
wanghan-iapcm, line 192: the map should be prepared and upload to GPU at `__init__`, so we can avoid host-GPU data transfer in forward.

wanghan-iapcm (changes requested):

wanghan-iapcm (approved):

iProzd (approved):

PR #3418, opened 2024-03-05
convert exclude_types to sel_type

njzjz:
This can give the correct result for `dp test`,
```sh
cd examples/water_tensor/dipole
dp --pt train input_torch.json
dp --pt freeze
dp test -m frozen_model.pth -s validation_data/global_system/
```

### DIFF
--- a/deepmd/pt/model/task/fitting.py
+++ b/deepmd/pt/model/task/fitting.py
@@ -388,6 +388,9 @@ class GeneralFitting(Fitting):
         """Get the number (dimension) of atomic parameters of this atomic model."""
         return self.numb_aparam
 
+    # make jit happy
+    exclude_types: List[int]
+
     def get_sel_type(self) -> List[int]:
         """Get the selected atom types of this model.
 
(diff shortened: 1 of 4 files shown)


deepmd/pt/model/task/fitting.py
```diff
@@ -388,14 +388,22 @@ def get_dim_aparam(self) -> int:
         """Get the number (dimension) of atomic parameters of this atomic model."""
         return self.numb_aparam
 
+    # make jit happy
+    exclude_types: List[int]
```
wanghan-iapcm, line 392: could you plz check if jit passes if we change the line 
https://github.com/deepmodeling/deepmd-kit/blob/d81443ba37adab034d7ffa9e0523b141cd96a1c4/deepmd/pt/model/task/fitting.py#L329
to 
```python
self.exclude_types : List[int] = exclude_types
```

wanghan-iapcm (approved):

deepmd/pt/model/task/fitting.py
njzjz: xref: https://github.com/pytorch/pytorch/pull/67051

PR #3419, opened 2024-03-06
revert test Python to 3.11

njzjz:
Python 3.12 or TF 2.16 seems to be slowing down...


njzjz:
Yes, it is faster than Python 3.12...

wanghan-iapcm (approved):

PR #3424, opened 2024-03-06
pt: avoid D2H in se_e2_a

njzjz:
sec is used as a slice index, so it should not stored on the GPU, otherwise, D2H will happen to create the tensor with the shape.

Before:
image

After:
image


wanghan-iapcm (approved):
That's surprising...

PR #3425, opened 2024-03-06
pt: improve nlist performance

njzjz:
1. use inv_ex instead of inv. `inv_ex` does not check errors. We can assume the input is correct.
2. pass CPU box for `torch.arange`;
3. avoid torch.tensor.

### DIFF
--- a/deepmd/pt/utils/nlist.py
+++ b/deepmd/pt/utils/nlist.py
@@ -27,14 +27,16 @@ def extend_input_and_build_neighbor_list(
 ):
     nframes, nloc = atype.shape[:2]
     if box is not None:
+        box_gpu = box.to(coord.device, non_blocking=True)
         coord_normalized = normalize_coord(
             coord.view(nframes, nloc, 3),
-            box.reshape(nframes, 3, 3),
+            box_gpu.reshape(nframes, 3, 3),
         )
     else:
+        box_gpu = None
         coord_normalized = coord.clone()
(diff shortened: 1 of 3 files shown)


deepmd/pt/utils/nlist.py
```diff
@@ -299,26 +304,23 @@ def extend_coord_with_ghosts(
...
-            [0, 0, 1], dtype=env.GLOBAL_PT_FLOAT_PRECISION, device=device
-        )
+        nbuff_cpu = nbuff.cpu()
+        xi = torch.arange(-nbuff_cpu[0], nbuff_cpu[0] + 1, 1, device=device)
+        yi = torch.arange(-nbuff_cpu[1], nbuff_cpu[1] + 1, 1, device=device)
+        zi = torch.arange(-nbuff_cpu[2], nbuff_cpu[2] + 1, 1, device=device)
+        eye_3 = torch.eye(3, dtype=env.GLOBAL_PT_FLOAT_PRECISION, device=device)
+        xyz = xi.view(-1, 1, 1, 1) * eye_3[0]
+        xyz = xyz + yi.view(1, -1, 1, 1) * eye_3[1]
+        xyz = xyz + zi.view(1, 1, -1, 1) * eye_3[2]
```
wanghan-iapcm, line 323: how about leaving all the operations on cpu and push xyz back to device?

wanghan-iapcm (approved):

PR #3426, opened 2024-03-06
Feat: Add polar stat constant matrix calculation to PT

anyangml:


deepmd/tf/fit/polar.py
```diff
@@ -193,18 +193,16 @@ def compute_output_stats(self, all_stat):
...
                         sys_matrix[-1][0, itype] = len(index_lis)
 
                         polar_bias.append(
                             np.sum(
-                                all_stat["atomic_polarizability"][ss].reshape((-1, 9))[
-                                    index_lis
-                                ],
-                                axis=0,
+                                all_stat["atomic_polarizability"][ss][:, index_lis, :],
+                                axis=(0, 1),
```
anyangml, line 205: currently, the `TF` only considers the first frame when doing aggregation. This need to consider all frames.

Also, there is a factor of `nframe` differences between the `TF` and `NP` implementation.

deepmd/pt/model/task/polarizability.py
```diff
@@ -184,7 +190,50 @@ def compute_output_stats(
...
+                else:
+                    sampled = merged
+
+                sys_constant_matrix = []
+                for sys in range(len(sampled)):
+                    nframs = sampled[sys]["type"].shape[0]
+
+                    if sampled[sys]["find_atomic_polarizability"] > 0.0:
+                        sys_atom_polar = compute_stats_from_atomic(
+                            sampled[sys]["atomic_polarizability"], sampled[sys]["type"]
```
anyangml, line 211: this might cause issue if the `sampled[sys]["type"].max() + 1` != `self.ntypes`

deepmd/pt/model/task/polarizability.py
```diff
@@ -184,7 +190,62 @@ def compute_output_stats(
...
+                                force=True
+                            )
+
+                        sys_bias_redu = sampled[sys]["polarizability"].numpy(force=True)
+
+                        sys_atom_polar = compute_stats_from_redu(
+                            sys_bias_redu, sys_type_count
+                        )[0]
+                    cur_constant_matrix = np.zeros(self.ntypes)
+                    if sys_atom_polar.shape[0] < self.ntypes:
```
wanghan-iapcm, line 230: it seems that the case should not happen. The dataloader is initialized by the model's type_map, so it is expected to return the data containing the same number of types as the model. @iProzd @CaRoLZhangxy please check. 
see also 
https://github.com/deepmodeling/deepmd-kit/blob/fa8e6453b5cc88ae575284c4c0ded06758c6de82/deepmd/pt/utils/dataloader.py#L60-L67

deepmd/pt/model/task/polarizability.py
```diff
@@ -184,7 +190,62 @@ def compute_output_stats(
...
+                        # pad zeros in case atype.max() + 1 != ntypes
+                        sys_atom_polar = np.concatenate(
+                            [
+                                sys_atom_polar,
+                                np.zeros((self.ntypes - sys_atom_polar.shape[0], 9)),
```
njzjz, line 235: dtype should give

deepmd/pt/model/task/polarizability.py
```diff
@@ -184,7 +190,62 @@ def compute_output_stats(
...
+                        cur_constant_matrix[itype] = np.mean(
+                            np.diagonal(sys_atom_polar[itype].reshape(3, 3))
+                        )
+                    sys_constant_matrix.append(cur_constant_matrix)
+                constant_matrix = np.stack(sys_constant_matrix).mean(axis=0)
+                # handle nan values.
+                constant_matrix = np.nan_to_num(constant_matrix)
+            self.constant_matrix = torch.tensor(constant_matrix, device=env.DEVICE)
+            if stat_file_path is not None:
+                stat_file_path.save_numpy(self.constant_matrix.detach().cpu().numpy())
```
njzjz, line 248: This causes an extra device-to-host copy... We have had the array on the host.

njzjz (changes requested):

deepmd/pt/model/task/polarizability.py
```diff
@@ -184,7 +190,62 @@ def compute_output_stats(
...
+                            type_mask = sampled[sys]["type"] == itype
+                            sys_type_count[:, itype] = type_mask.sum(dim=1).numpy(
+                                force=True
+                            )
+
+                        sys_bias_redu = sampled[sys]["polarizability"].numpy(force=True)
+
+                        sys_atom_polar = compute_stats_from_redu(
+                            sys_bias_redu, sys_type_count
+                        )[0]
```
njzjz, line 228: Why not pass `cond`?

deepmd/tf/fit/polar.py
njzjz, line 1: It seems to be a bug that TF doesn't save constant_matrix...

deepmd/dpmodel/fitting/polarizability_fitting.py
```diff
@@ -174,7 +175,9 @@ def serialize(self) -> dict:
         data["embedding_width"] = self.embedding_width
         data["old_impl"] = self.old_impl
         data["fit_diag"] = self.fit_diag
+        data["shift_diag"] = self.shift_diag
         data["@variables"]["scale"] = self.scale
+        data["@variables"]["constant_matrix"] = self.constant_matrix
```
njzjz, line 180: We may need to bump the version, considering new serialized data can be not be deserialized by the old code.

anyangml: you mean model plugin?

njzjz: I mean the version system introduced in #3349.

deepmd/tf/fit/polar.py
anyangml: not going to fix this.

njzjz (approved):

deepmd/pt/model/task/fitting.py
```diff
@@ -419,6 +415,8 @@ def __setitem__(self, key, value):
             self.aparam_inv_std = value
         elif key in ["scale"]:
             self.scale = value
+        elif key in ["constant_matrix"]:
+            self.constant_matrix = value
```
wanghan-iapcm, line 419: the variable should be maintained by the polarizability_fitting, not by the general fitting.

wanghan-iapcm (changes requested):

wanghan-iapcm (approved):

deepmd/pt/model/task/polarizability.py
```diff
@@ -184,7 +216,60 @@ def compute_output_stats(
...
+                            np.diagonal(sys_atom_polar[itype].reshape(3, 3))
+                        )
+                    sys_constant_matrix.append(cur_constant_matrix)
+                constant_matrix = np.stack(sys_constant_matrix).mean(axis=0)
+
+                # handle nan values.
+                constant_matrix = np.nan_to_num(constant_matrix)
+            if stat_file_path is not None:
+                stat_file_path.save_numpy(self.constant_matrix)
+            self.constant_matrix = torch.tensor(constant_matrix, device=env.DEVICE)
```
iProzd, line 272: typo?  
```suggestion
            if stat_file_path is not None:
                stat_file_path.save_numpy(constant_matrix)
            self.constant_matrix = torch.tensor(constant_matrix, device=env.DEVICE)
```

iProzd (changes requested):

iProzd (approved):

Issue #3427, opened 2024-03-07
clean up the init interface of pt.dataloader


wanghan-iapcm:
### Summary

The initialization of a dataloader needs a dict named `model_param` . 
https://github.com/deepmodeling/deepmd-kit/blob/fa8e6453b5cc88ae575284c4c0ded06758c6de82/deepmd/pt/utils/dataloader.py#L60-L67

this is not a good design, we should explicitly write what information is necessary to init the dataloader. 

### DeePMD-kit Version

v3.0.0.a0

### TensorFlow Version

2.14.0

### Details

As described in summary.

PR #3434, opened 2024-03-08
clean up the init interface of pt.dataloader

CaRoLZhangxy:
https://github.com/deepmodeling/deepmd-kit/issues/3427


deepmd/pt/utils/dataset.py
```diff
@@ -16,24 +16,16 @@
 
 
 class DeepmdDataSetForLoader(Dataset):
-    def __init__(
-        self,
-        system: str,
-        type_map: str,
-        shuffle=True,
-    ):
+    def __init__(self, system: str, type_map: str):
```
njzjz, line 19: Why `type_map` is str?

wanghan-iapcm (commented):
The same comment as Jinzhe.

wanghan-iapcm (approved):

njzjz (approved):

PR #3435, opened 2024-03-08
Doc: ZBL pytorch

anyangml:
pytorch and tf should be able to use the same zbl input.json https://github.com/deepmodeling/deepmd-kit/issues/3498

## Summary by CodeRabbit

- **Documentation**
	- Expanded the list of supported backends for interpolation or combination with a pairwise potential in DeePMD-kit documentation to include PyTorch and DP models alongside TensorFlow. Icons for each backend have been added for clearer representation.


anyangml:
@wanghan-iapcm feels like we can leave this unchanged based on our discussion. 

1. during training, user knows what `type_map` to use, thus, three models should share the same `type_map`.
2. during fine tuning, the new `type_map` should also be the same for three models, and the old type_map stored in sub-models will be remapped to the new `type_map`.

njzjz (approved):

PR #3438, opened 2024-03-09
split Model and AtomicModel

njzjz:
Model is not inherited from AtomicModel anymore.

### DIFF
--- a/deepmd/dpmodel/model/make_model.py
+++ b/deepmd/dpmodel/model/make_model.py
@@ -1,19 +1,29 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
+from abc import (
+    abstractmethod,
+)
 from typing import (
     Dict,
     List,
     Optional,
     Tuple,
+    Type,
 )
 
 import numpy as np
 
+from deepmd.dpmodel.atomic_model.base_atomic_model import (
+    BaseAtomicModel,
+)
 from deepmd.dpmodel.common import (
     GLOBAL_ENER_FLOAT_PRECISION,
     GLOBAL_NP_FLOAT_PRECISION,
     PRECISION_DICT,
     RESERVED_PRECISON_DICT,
     NativeOP,
+)
+from deepmd.dpmodel.model.base_model import (
+    BaseModel,
 )
 from deepmd.dpmodel.output_def import (
     ModelOutputDef,
@@ -34,7 +44,7 @@ from .transform_output import (
 )
 
 
-def make_model(T_AtomicModel):
+def make_model(T_AtomicModel: Type[BaseAtomicModel]):
     """Make a model as a derived class of an atomic model.
 
     The model provide two interfaces.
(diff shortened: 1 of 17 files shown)


deepmd/dpmodel/model/make_model.py
```diff
@@ -377,4 +385,94 @@ def _format_nlist(
...
+            """
+            return self.atomic_model.do_grad_c(var_name)
+
+        def serialize(self) -> dict:
+            return self.atomic_model.serialize()
+
+        @classmethod
+        @abstractmethod
+        def deserialize(cls, data) -> "CM":
+            pass
```
wanghan-iapcm, line 412: Any idea of implementing the class method.  in principle, we only need to deserialize the `self.atomic_model`

deepmd/pt/model/model/make_model.py
```diff
@@ -382,4 +394,111 @@ def _format_nlist(
...
+            """
+            return self.atomic_model.do_grad_c(var_name)
+
+        def serialize(self) -> dict:
+            return self.atomic_model.serialize()
+
+        @classmethod
+        @abstractmethod
+        def deserialize(cls, data) -> "CM":
+            pass
```
wanghan-iapcm, line 421: same question as the dpmodel.make_model

deepmd/pt/model/model/dp_zbl_model.py
```diff
@@ -120,3 +127,27 @@ def update_sel(cls, global_jdata: dict, local_jdata: dict):
             global_jdata, local_jdata["dpmodel"]
         )
         return local_jdata_cpy
+
+    @classmethod
+    def deserialize(cls, data) -> "DPZBLModel":
```
wanghan-iapcm, line 132: the method is simply a copy of the atomic dp zbl model.

deepmd/dpmodel/model/make_model.py
njzjz: self.atomic_model.deserialize returns a BaseAtomicModel, but we want a BaseModel.

deepmd/pt/model/model/make_model.py
njzjz: same above

deepmd/pt/model/model/dp_zbl_model.py
njzjz: I have no idea how to convert a DP AtomicZBLModel to a DPZBLModel

deepmd/dpmodel/model/make_model.py
njzjz: Oh, I come up with a trick way to do so.

wanghan-iapcm (approved):

deepmd/pt/model/model/model.py
```diff
@@ -11,61 +13,14 @@
...
-#           DPAtomicModel                   BaseModel
-#                |                              |
-#                |                              |
-#       make_model(DPAtomicModel)               |
-#                |                              |
-#                |                              |
-#                --------------------------------
-#                         |
-#                         |
-#                      DPModel
```
anyangml, line 52: I think it might be helpful to document the new design pattern somewhere appropriate for reference.

anyangml (approved):

deepmd/pt/model/model/model.py
njzjz: Fixed in https://github.com/deepmodeling/deepmd-kit/pull/3438/commits/965f471876d1ba69013eea0c4387ba7502f4056e.

PR #3442, opened 2024-03-09
pt: cleanup tester

iProzd:


njzjz (approved):

njzjz:
Perhaps need a better name in the future. This method is now only used to recover a model object from a checkpoint file.

wanghan-iapcm (approved):

PR #3445, opened 2024-03-11
pt: add explicit decay_rate for lr

iProzd:
This is for multitask training, when explicitly setting decay_rate is much more convenient for long training.


deepmd/pt/utils/learning_rate.py
```diff
@@ -3,14 +3,36 @@
 
 
 class LearningRateExp:
-    def __init__(self, start_lr, stop_lr, decay_steps, stop_steps, **kwargs):
-        """Construct an exponential-decayed learning rate.
+    def __init__(
+        self,
+        start_lr,
+        stop_lr,
```
wanghan-iapcm, line 9: `stop_lr` may be made optional, user may not want to set it if decay_rat is set.

deepmd/pt/utils/learning_rate.py
```diff
@@ -20,16 +42,15 @@ def __init__(self, start_lr, stop_lr, decay_steps, stop_steps, **kwargs):
         self.decay_rate = np.exp(
             np.log(stop_lr / self.start_lr) / (stop_steps / self.decay_steps)
         )
-        if "decay_rate" in kwargs:
-            self.decay_rate = kwargs["decay_rate"]
-        if "min_lr" in kwargs:
-            self.min_lr = kwargs["min_lr"]
+        if decay_rate is not None:
+            self.decay_rate = decay_rate
+            self.min_lr = min_lr
```
wanghan-iapcm, line 47: raise error or warn the user if stop_lr is set.

deepmd/pt/utils/learning_rate.py
```diff
@@ -20,16 +42,15 @@ def __init__(self, start_lr, stop_lr, decay_steps, stop_steps, **kwargs):
...
         else:
-            self.min_lr = 3e-10
+            self.min_lr = None
```
wanghan-iapcm, line 49: what would happen if stop_lr, min_lr are set but decay_rate is none? it is not handled.

deepmd/tf/utils/learning_rate.py
```diff
@@ -41,6 +41,7 @@ def __init__(
         stop_lr: float = 5e-8,
         decay_steps: int = 5000,
         decay_rate: float = 0.95,
+        **kwargs,
```
wanghan-iapcm, line 44: accept and del extra keys. stop using **kwargs.

wanghan-iapcm (changes requested):

njzjz (changes requested):

deepmd/pt/utils/learning_rate.py
njzjz: Theoretically, one only needs to set there among `start_lr`, `stop_lr`, `decay_steps`, and `decay_rate`.

iProzd: I see, we can use `stop_lr` param instead of the `min_lr` param when users set `decay_rate` explicitly.
1. when users set `start_lr`, `decay_steps`, `stop_lr`, it will follow the traditional way to compute the `decay_rate`
2. when users set `start_lr`, `decay_steps`, `stop_lr` and `decay_rate` in pytorch, it will take the user-defined `decay_rate` and with the minimum learning rate to be `stop_lr`.

njzjz: I am not sure whether we should merge these two modes into one doc, or document them separately using `Variant`.

njzjz (approved):

source/tests/pt/test_lr.py
wanghan-iapcm, line 1: The case when setting decay step and the learning rate reaches stop_lr is not tested.

wanghan-iapcm (changes requested):

wanghan-iapcm (approved):

PR #3451, opened 2024-03-12
Chore: refactor LinearAtomicModel serialize/deserialize

anyangml:
This PR refactors the serialization/ deserialization in LinearEnergyAtomicModel using the plugin mechanism introduced in #3438 .


deepmd/dpmodel/atomic_model/linear_atomic_model.py
```diff
@@ -225,32 +225,30 @@ def fitting_output_def(self) -> FittingOutputDef:
...
-            for name, model in zip(model_names, data["models"])
+            BaseAtomicModel.get_class_by_type(model["type"]).deserialize(model)
+            for model in data["models"]
         ]
-        return models, type_map
+        data.pop("models")
+        return cls(models, type_map, **data)
 
-    @abstractmethod
+    # @abstractmethod
```
anyangml, line 247: If we want to use the serialize/deserialize method from LinearAtomicModel in ZBLAtomicModel, we have to instantiate a concrete class, thus need to remove the `@abstractmethod`

deepmd/dpmodel/atomic_model/linear_atomic_model.py
```diff
@@ -336,9 +330,9 @@ def serialize(self) -> dict:
                 "@class": "Model",
                 "type": "zbl",
                 "@version": 1,
```
anyangml, line 332: will bump version

deepmd/dpmodel/atomic_model/linear_atomic_model.py
njzjz: It might not be a good idea to instantiate it. We may need another method just to handle data, but doesn't instantiate the object.

anyangml: @njzjz I updated the PT side, see if that's what you have in mind.

wanghan-iapcm (approved):

wanghan-iapcm (approved):
This PR provides a default weight computing method for the linear model (average over all sub-models) to make it instantiable. @njzjz please check if it a good design

deepmd/dpmodel/atomic_model/linear_atomic_model.py
```diff
@@ -225,40 +221,38 @@ def fitting_output_def(self) -> FittingOutputDef:
...
     def _compute_weight(
         self,
         extended_coord: np.ndarray,
         extended_atype: np.ndarray,
         nlists_: List[np.ndarray],
     ) -> np.ndarray:
         """This should be a list of user defined weights that matches the number of models to be combined."""
-        raise NotImplementedError
+        nmodels = len(self.models)
+        return [np.ones(1) / nmodels for _ in range(nmodels)]
```
njzjz, line 255: I am not sure what should be returned, but at least this is List[np.ndarray] instead of np.ndarray.

njzjz (changes requested):

deepmd/pt/model/atomic_model/linear_atomic_model.py
```diff
@@ -260,35 +256,35 @@ def fitting_output_def(self) -> FittingOutputDef:
...
+        return cls(models, type_map, **data)
 
-    @abstractmethod
     def _compute_weight(
         self, extended_coord, extended_atype, nlists_
     ) -> List[torch.Tensor]:
         """This should be a list of user defined weights that matches the number of models to be combined."""
-        raise NotImplementedError
+        nmodels = len(self.models)
+        return [torch.ones(1) / nmodels for _ in range(nmodels)]
```
njzjz, line 287: device and dtype are not given.

njzjz (changes requested):

deepmd/dpmodel/atomic_model/linear_atomic_model.py
anyangml: List[np.ndarray] is expected.

deepmd/dpmodel/atomic_model/linear_atomic_model.py
```diff
@@ -225,40 +221,38 @@ def fitting_output_def(self) -> FittingOutputDef:
...
+        data.pop("models")
+        return cls(models, type_map, **data)
 
-    @abstractmethod
     def _compute_weight(
         self,
         extended_coord: np.ndarray,
         extended_atype: np.ndarray,
         nlists_: List[np.ndarray],
     ) -> np.ndarray:
```
njzjz, line 252:
```suggestion
    ) -> List[np.ndarray]:
```

njzjz (approved):

PR #3452, opened 2024-03-12
Feat: add DOS net

anyangml:
This PR provides DOS fitting net in Pytorch.

Future TODO:
- [ ] Loss implementation
  - [ ] Training/Fine-tuning test
  - [ ] Jit test
- [ ] Doc


source/tests/consistent/common.py
```diff
@@ -264,6 +264,11 @@ def test_tf_consistent_with_ref(self):
 
         np.testing.assert_equal(data1, data2)
         for rr1, rr2 in zip(ret1, ret2):
+            if tf_obj.__class__.__name__.startswith("DOS"):
+                rr1 = rr1.ravel()
+                rr1.sort()
+                rr2 = rr2.ravel()
+                rr2.sort()
```
anyangml, line 271: Ideally, the TF backend should match the PT backend, and we do not need this bypass. Currently, using the UT example where the input descriptor is `np.ones(1,6,20)`, atype is `[0,0,1,1,1,1]`, we should expected the output to be (1, 6, ndos=3) and for the same type, the `nods` should be identical. Below is the results of TF and PT.

```
# TF
array([[[ 1.38051963],
        [-4.33505583],
        [-1.00053191],
        [-0.58370066],
        [-1.13134444],
        [-1.00053191]],

       [[ 2.9075191 ],
        [ 1.38051963],
        [-0.58370066],
        [-1.13134444],
        [-1.00053191],
        [-0.58370066]],

       [[-4.33505583],
        [ 2.9075191 ],
        [-1.13134444],
        [-1.00053191],
        [-0.58370066],
        [-1.13134444]]])

#PT
array([[[ 1.38051954, -4.33505519,  2.90751917],
        [ 1.38051954, -4.33505519,  2.90751917],
        [-1.00053181, -0.58370096, -1.13134455],
        [-1.00053181, -0.58370096, -1.13134455],
        [-1.00053181, -0.58370096, -1.13134455],
        [-1.00053181, -0.58370096, -1.13134455]]])
```

anyangml: @njzjz not sure if it has something to do with the atype order for TF.

deepmd/tf/fit/dos.py
njzjz, line 1: Need to confirm:
(1) whether the shape and data in `examples/dos/data/heat-221/set.000/dos.npy` is still correct.

njzjz, line 1: (2) whether `doc/model/train-fitting-dos.md` needs to be updated;

njzjz, line 1: (4) whether the output shape of `dp test` is correct.

anyangml: the data still works for TF.

anyangml: image

anyangml: yes, we probably need to update the output description.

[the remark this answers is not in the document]
anyangml: let me double check on thi. I believe the out put tensor should still retain the same shape, just the order was updated.

njzjz: I mean the detail file specified by -d, --detail-file

anyangml: the `dos.out` has shape (2, 250); `ados.out` has shape (2,8000)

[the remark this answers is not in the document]
anyangml: I don't think it's a breaking change. the inputs from the user and outputs to the user are still the same.

njzjz (approved):

deepmd/dpmodel/fitting/general_fitting.py
```diff
@@ -85,6 +87,7 @@ def __init__(
         resnet_dt: bool = True,
         numb_fparam: int = 0,
         numb_aparam: int = 0,
+        bias_atom_e: Optional[np.ndarray] = None,
```
wanghan-iapcm, line 90: we may rename all `bias_atom_e` to `bias_atom`..

deepmd/pt/model/model/__init__.py
```diff
@@ -140,7 +140,8 @@ def get_standard_model(model_params):
     fitting_net["type"] = fitting_net.get("type", "ener")
     fitting_net["ntypes"] = descriptor.get_ntypes()
     fitting_net["mixed_types"] = descriptor.mixed_types()
-    fitting_net["embedding_width"] = descriptor.get_dim_emb()
+    if fitting_net["type"] != "dos":
+        fitting_net["embedding_width"] = descriptor.get_dim_emb()
```
wanghan-iapcm, line 144: it seems that `embedding_width` is only needed by polar and dipole fitting. 
```python 
if fitting_net["type"]  in ["dipole", "polar"]:
    fitting_net["embedding_width"] = descriptor.get_dim_emb()
```
may be bettter

deepmd/tf/fit/dos.py
```diff
@@ -521,7 +528,11 @@ def build(
 
                 final_layer = tf.reshape(
                     final_layer,
-                    [tf.shape(inputs)[0] * self.numb_dos, natoms[2 + type_i]],
+                    [
+                        tf.shape(inputs)[0],
+                        natoms[2 + type_i],
+                        self.numb_dos,
+                    ],
```
wanghan-iapcm, line 535: LOL

deepmd/dpmodel/fitting/general_fitting.py
anyangml: I will update this in the following PR.

wanghan-iapcm (approved):

PR #3456, opened 2024-03-13
pt: add index input for `use_spin`

iProzd:
It's convenient for spin multitask input file to handle different spin inputs.


deepmd/pt/model/model/__init__.py
```diff
@@ -57,6 +59,12 @@
 
 def get_spin_model(model_params):
     model_params = copy.deepcopy(model_params)
+    if model_params["spin"]["use_spin"] and isinstance(
+        model_params["spin"]["use_spin"][0], int
+    ):
```
njzjz, line 64: When `use_spin` is `[]` (I don't know whether it's allowed), it cannot be converted to the mask (all false).

njzjz (approved):

Issue #3455, opened 2024-03-13
[BUG] Finetuned model has wrong type_map


iProzd:
### Bug summary

When doing finetuing, the user-defined `type_map` (e.g. ['H', 'O']) will be covered by the `type_map` in the pretrained model (e.g. the whole periodic table) , which is confusing for users.

### DeePMD-kit Version

3.0.0a

### TensorFlow Version

2.6.0

### How did you download the software?

Built from source

### Input Files, Running Commands, Error Log, etc.

See above.

### Steps to Reproduce

See above.

njzjz:
Idea: the easiest way is to add a virtual `Model`, "adapt type map model", which just adapts the input `atom_type` from the outer model `type_map` to the inner model `type_map` and forwards everything else like #3450.

iProzd:
> Idea: the easiest way is to add a virtual `Model`, "adapt type map model", which just adapts the input `atom_type` from the outer model `type_map` to the inner model `type_map` and forwards everything else like #3450.
I see, but this will cause the model to be wrapped repeatedly each time it is fine-tuned, as discussed with @wanghan-iapcm and @anyangml.

njzjz:
>  repeatedly each time it is fine-tuned

Indeed, I don't understand why it needs to change the type map each time it is fine-tuned...

anyangml:
> > repeatedly each time it is fine-tuned
> 
> Indeed, I don't understand why it needs to change the type map each time it is fine-tuned...

For the LinearModel, let's say we have two pre-trained models: model A with ["H", "O", "Na"], model B with ["H", "O", "K"]. Now if we want to finetune a LinearModel with this, the new type map becomes ["H", "O"].

njzjz:
> > > repeatedly each time it is fine-tuned
> 
> > 
> 
> > Indeed, I don't understand why it needs to change the type map each time it is fine-tuned...
> 
> 
> 
> For the LinearModel, let's say we have two pre-trained models: model A with ["H", "O", "Na"], model B with ["H", "O", "K"]. Now if we want to finetune a LinearModel with this, the new type map becomes ["H", "O"].

This is not correct. The type map should be the union of two models.

anyangml:
> > > > repeatedly each time it is fine-tuned
> > 
> > 
> > > 
> > 
> > 
> > > Indeed, I don't understand why it needs to change the type map each time it is fine-tuned...
> > 
> > 
[...]

PR #3480, opened 2024-03-17
pt: support multitask finetune

iProzd:
This PR：
1. merge `change_energy_bias` into `compute_output_stats` and reformat it into `change_out_bias` of `model` level.
2. support single-task/multi-task finetuning from single-task/multi-task pretrained model.

Need fix in future PR:
1. Finetuned model has covered `type_map`. (If fixed, `change_out_bias` func will not need the input params `origin_type_map` and `full_type_map`.) See also #3455.
2. `change_out_bias` support for other models.(e.g. Spin, ZBL, Polar, Dipole and Dos.)


deepmd/pt/utils/stat.py
```diff
@@ -129,13 +136,67 @@ def compute_output_stats(
...
+                    if system["box"] is not None
+                    else None
+                )
+                if data_mixed_type:
+                    atype = to_numpy_array(system["atype"]).reshape(nframes, -1)
+                else:
+                    atype = to_numpy_array(system["atype"]).reshape(nframes, -1)[0]
+                fparam = (
+                    to_numpy_array(system["fparam"])
+                    if "fparam" in system is not None
```
njzjz, line 164: This looks wrong? `a in b` is a bool, so it is always not None.

deepmd/pt/utils/stat.py
```diff
@@ -129,13 +136,67 @@ def compute_output_stats(
...
+                    else None
+                )
+                aparam = (
+                    to_numpy_array(system["aparam"])
+                    if "aparam" in system is not None
```
njzjz, line 169: +1 here

deepmd/pt/utils/stat.py
```diff
@@ -129,13 +136,67 @@ def compute_output_stats(
...
+            )
+            unbias_e = energy_predict + merged_natoms @ bias_atom_e
+            atom_numbs = merged_natoms.sum(-1)
+            rmse_ae = np.sqrt(
+                np.mean(
+                    np.square((unbias_e.ravel() - merged_energy.ravel()) / atom_numbs)
+                )
+            )
+            log.info(
+                f"RMSE of atomic energy after linear regression is: {rmse_ae} eV/atom."
```
njzjz, line 197: It's energy per atom but not atomic energy.

deepmd/pt/utils/finetune.py
```diff
@@ -10,88 +13,171 @@
...
+
+    Parameters
+    ----------
+    finetune_model:
+        The pretrained model.
+    model_config
+        The fine-tuning input parameters.
+    model_branch
+        The model branch chosen in command-line mode, only for single-task fine-tuning.
+    This function modifies the fine-tuning input in different modes as follows:
```
njzjz, line 101: This should not be within the parameters section

deepmd/pt/model/model/ener_model.py
```diff
@@ -97,3 +113,58 @@ def forward_lower(
...
+            The full type_map in pre-trained model
+        bias_shift : str
+            The mode for changing energy bias : ['delta', 'statistic']
+            'delta' : perform predictions on energies of target dataset,
+                    and do least sqaure on the errors to obtain the target shift as bias.
+            'statistic' : directly use the statistic energy bias in the target dataset.
+        """
+        sorter = np.argsort(full_type_map)
+        idx_type_map = sorter[
+            np.searchsorted(full_type_map, origin_type_map, sorter=sorter)
```
njzjz, line 143: Before using this method, we need to assert that the original type map is a subset of the full type map.

njzjz (changes requested):

deepmd/pt/model/model/ener_model.py
```diff
@@ -97,3 +113,58 @@ def forward_lower(
...
+        ]
+        original_bias = self.get_fitting_net()["bias_atom_e"]
+        if bias_shift == "delta":
+            tmp_model = tempfile.NamedTemporaryFile(delete=False, suffix=".pth")
+            model = torch.jit.script(self)
+            torch.jit.save(model, tmp_model.name)
+            dp = DeepEval(tmp_model.name)
+            os.unlink(tmp_model.name)
```
wanghan-iapcm, line 151: why not using the self forward interface ?

wanghan-iapcm (changes requested):
- It seems that the change output bias should be place in atomic model (current bias can also be computed by the forward_atomic method)
- the implementation should be able to reused by other atomic models, including the zbl and linear models, thus it is either placed in base_atomic_model, or be designed as a common method called by models, just like the compute_output_stat.

deepmd/pt/utils/stat.py
```diff
@@ -129,13 +136,67 @@ def compute_output_stats(
...
-            merged_energy,
-            merged_natoms,
-            assigned_bias=assigned_atom_ener,
-            rcond=rcond,
-        )
+        if model is None:
+            # only use statistics result
+            bias_atom_e, _ = compute_stats_from_redu(
+                merged_energy,
+                merged_natoms,
```
anyangml, line 143: I noticed that this function is hardcoded with the `energy` key. Since we are refactoring, do we want to make it a bit more flexible, for example for `bias_dos`. Although, technically, we can hack the `bias_dos` to use `energy` as the keyword.

iProzd: We can use a new PR to add support for finetuning in other models.

iProzd: We can use a new PR to add support for finetuning in other models.

njzjz (approved):

anyangml (approved):

deepmd/pt/model/atomic_model/dp_atomic_model.py
```diff
@@ -223,6 +233,86 @@ def wrapped_sampler():
         if self.fitting_net is not None:
             self.fitting_net.compute_output_stats(wrapped_sampler, stat_file_path)
 
+    def change_out_bias(
```
wanghan-iapcm, line 236: how can other atomic model reuse this method? should it be based in base_atomic_model?

wanghan-iapcm (changes requested):

deepmd/pt/model/atomic_model/dp_atomic_model.py
iProzd: Updated in https://github.com/deepmodeling/deepmd-kit/pull/3480/commits/bdf74c916a1490c9be459234339b0411bec43ab2

deepmd/pt/train/training.py
```diff
@@ -486,60 +487,114 @@ def get_loss(loss_params, start_lr, _ntypes, _model):
...
-                    ):
-                        self.model.atomic_model.fitting_net.change_energy_bias(
-                            config,
-                            self.model,
-                            old_type_map,
-                            new_type_map,
-                            ntest=ntest,
-                            bias_shift=model_params.get("bias_shift", "delta"),
+                    if isinstance(_model, EnergyModel):
+                        _model.atomic_model.change_out_bias(
```
wanghan-iapcm, line 569: bad design. you are accessing the datamember of the model!.  we should provide the `change_out_bias` method for model that is forwarding the atomic model's `change_out_bias`

wanghan-iapcm (changes requested):

deepmd/pt/model/atomic_model/base_atomic_model.py
```diff
@@ -176,6 +190,59 @@ def serialize(self) -> dict:
...
+                    nlist,
+                    mapping=mapping,
+                    fparam=fparam,
+                    aparam=aparam,
+                )
+                return atomic_ret[out_name].detach()
+
+        return model_forward
+
     def compute_or_load_stat(
```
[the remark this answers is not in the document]
iProzd: I think we may need this func. The common function for data statistics and fine-tuning is `compute_output_stats`, which is capable of managing both scenarios. Both `compute_or_load_stat` and `change_out_bias` utilize `compute_output_stats`. However, separate functions are necessary to handle the input for `compute_output_stats`, such as preparing model forward wrapper functions, managing type maps in `change_out_bias`, and organizing stat files in `compute_or_load_stat`.

wanghan-iapcm (approved):

PR #3566, opened 2024-03-19
Feat pt: Support multi structures input for single frame

Chengqian-Zhang:
Support multi structures input for one frame:

background:
When we predict the properties of small molecules, sometimes multiple 3D conformations are generated from the SMILES format structure, but their targets are the same. In each epoch we would like to pick only one of the 3D configurations as input.

Impletation:
When the file "multistru" is present in the data folder (like nopbc), "coord.npy" can be written in the format (nframe * struc_num * natom * 3), where *struc_num* is the number of 3D conformations for each structure.

### DIFF
--- a/deepmd/utils/data.py
+++ b/deepmd/utils/data.py
@@ -272,9 +272,8 @@ class DeepmdData:
         i = bisect.bisect_right(self.prefix_sum, index)
         frames = self._load_set(self.dirs[i])
         if self.multistru:
-            size = frames["coord"].shape[1]
-            rng = np.random.default_rng()
-            sample_idx = rng.integers(low=0, high=size, size=1)[0]
+            num_stru = frames["coord"].shape[1]
+            sample_idx = dp_random.choice(range(num_stru), size=1)[0]
             frames["coord"] = frames["coord"][:, sample_idx, :]
         frame = self._get_subdata(frames, index - self.prefix_sum[i])
         frame = self.reformat_data_torch(frame)
--- a/deepmd/utils/data.py
+++ b/deepmd/utils/data.py
@@ -518,7 +518,9 @@ class DeepmdData:
             coord = coord.reshape([1, -1])
         nframes = coord.shape[0]
         if self.multistru:
-            assert (coord.shape[1] / self.data_dict["coord"]["ndof"] * self.natoms).is_integer()
+            assert (
+                coord.shape[1] / self.data_dict["coord"]["ndof"] * self.natoms
+            ).is_integer()
         else:
             assert coord.shape[1] == self.data_dict["coord"]["ndof"] * self.natoms
         # load keys
--- a/deepmd/utils/data.py
+++ b/deepmd/utils/data.py
@@ -9,6 +9,7 @@ from typing import (
 )
 
 import numpy as np
+import contextlib
 
 from deepmd.env import (
     GLOBAL_ENER_FLOAT_PRECISION,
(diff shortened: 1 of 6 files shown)


deepmd/utils/data.py
```diff
@@ -269,6 +271,11 @@ def get_item_torch(self, index: int) -> dict:
         """
         i = bisect.bisect_right(self.prefix_sum, index)
         frames = self._load_set(self.dirs[i])
+        if self.multistru:
+            size = frames["coord"].shape[1]
+            rng = np.random.default_rng()
+            sample_idx = rng.integers(low=0, high=size, size=1)[0]
```
Chengqian-Zhang, line 277: Is it really random?

deepmd/utils/data.py
njzjz, line 1: Shouldn't it be documented? Otherwise no one knows how to use it.

deepmd/utils/data.py
njzjz: We use the random generator in `deepmd.utils.rndom`. The generator here cannot be controlled.

deepmd/utils/data.py
```diff
@@ -621,43 +633,49 @@ def _load_data(
...
-                                )
-                                data = tmp
-                            else:
-                                natoms = natoms_sel
-                                idx_map = idx_map_sel
-                                ndof = ndof_ * natoms
-                        elif data.size == nframes * natoms * ndof_:
-                            if output_natoms_for_type_sel:
-                                pass
+                if key == "coord" and self.multistru:
```
njzjz, line 636: Do we assume aparam and other properties are the same for all conformations?

njzjz (changes requested):

deepmd/utils/data.py
Chengqian-Zhang: I have change it to 
`from deepmd.utils import random as dp_random`
`sample_idx = dp_random.choice(range(num_stru), size=1)[0]`
But I find that when num_stru = 11, the `sample_idx` is 1,4,8,10 all the time. I hope that the value is randomly selected from `range(0--10)`. How can I solve it?

deepmd/utils/data.py
Chengqian-Zhang: I add doc to `deepmd/doc`. Should I add doc to other places?

deepmd/utils/data.py
Chengqian-Zhang: I haven't thought of a situation where there are different other properties(aparam fparam) of all conformations. The reason I add this feature is that when one only have molecules in SMILES format, one may generate several possible conformations using tools such as rdkit. Do you have more better ideas?

wanghan-iapcm (commented):
I do not find the the feature is necessary. if different structures have different labels, they should not be placed in one dataset, because the dp model cannot fit  to diverging labels.

Chengqian-Zhang:
@wanghan-iapcm  Different 3D conformations of a molecule have exactly the same label, and each epoch takes one of all 3D conformations of this molecule.

wanghan-iapcm:
> @wanghan-iapcm Different 3D conformations of a molecule have exactly the same label, and each epoch takes one of all 3D conformations of this molecule. Because molecule may only has SMILES, the exact 3D sturcture is unknown.

I do not find any reason why different 3D configurations cannot be place in one batch.

Chengqian-Zhang:
There is currently no need for this PR, I will close it.

PR #3569, opened 2024-03-19
pt: refactor loss

iProzd:
This PR updates the loss interface to allow for a more flexible design. 
It enables processing input tensors before feeding them into the model, such as denoising operations (fyi @Chengqian-Zhang ). Previously, this was done in the data loader, which was less intuitive and more confusing. Now, users can easily handle these tasks within the loss function itself, as demonstrated in similar implementations in uni-mol: https://github.com/dptech-corp/Uni-Mol/blob/main/unimol/unimol/losses/unimol.py#L20.

### DIFF
--- a/deepmd/pt/loss/ener.py
+++ b/deepmd/pt/loss/ener.py
@@ -90,20 +90,26 @@ class EnergyStdLoss(TaskLoss):
         self.use_l1_all = use_l1_all
         self.inference = inference
 
-    def forward(self, model_pred, label, natoms, learning_rate, mae=False):
-        """Return loss on loss and force.
+    def forward(self, input_dict, model, label, natoms, learning_rate, mae=False):
+        """Return loss on energy and force.
 
-        Args:
-        - natoms: Tell atom count.
-        - p_energy: Predicted energy of all atoms.
-        - p_force: Predicted force per atom.
-        - l_energy: Actual energy of all atoms.
-        - l_force: Actual force per atom.
+        input_dict: dict[str, torch.Tensor]
+            Model inputs.
+        model: torch.nn.Module
+            Model to be used to output the predictions.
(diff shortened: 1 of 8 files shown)


njzjz:
Could you make an explanation why it needs to change the interface?

iProzd:
> Could you make an explanation why it needs to change the interface?

I've updated the comment.

deepmd/pt/loss/ener.py
```diff
@@ -90,20 +90,26 @@ def __init__(
...
+        natoms : int
+            The local atom number.
 
         Returns
         -------
-        - loss: Loss to minimize.
+        loss: torch.Tensor
+            Loss for model to minimize.
+        more_loss: dict[str, torch.Tensor]
+            Other losses for display.
```
njzjz, line 110: The return values need to be updated.

deepmd/pt/loss/ener.py
```diff
@@ -90,20 +90,26 @@ def __init__(
...
+    def forward(self, input_dict, model, label, natoms, learning_rate, mae=False):
+        """Return loss on energy and force.
 
-        Args:
-        - natoms: Tell atom count.
-        - p_energy: Predicted energy of all atoms.
-        - p_force: Predicted force per atom.
-        - l_energy: Actual energy of all atoms.
-        - l_force: Actual force per atom.
+        input_dict: dict[str, torch.Tensor]
```
njzjz, line 96:
```suggestion
        Parameters
        ----------
        input_dict: dict[str, torch.Tensor]
```

njzjz (approved):

Chengqian-Zhang (approved):

wanghan-iapcm (approved):

PR #3576, opened 2024-03-21
lmp: support input mapping for single process dpa2 model inference

CaRoLZhangxy:


deepmd/pt/model/model/make_model.py
```diff
@@ -469,6 +469,11 @@ def get_nnei(self) -> int:
             """Returns the total number of selected neighboring atoms in the cut-off radius."""
             return self.atomic_model.get_nnei()
 
+        @torch.jit.export
+        def get_ntypes(self) -> int:
```
njzjz, line 473: Not all models are made from the atomic model. This method should be added into the BaseModel.

source/api_c/include/c_api.h
```diff
@@ -24,7 +24,8 @@ extern DP_Nlist* DP_NewNlist(int inum_,
                              int* ilist_,
                              int* numneigh_,
                              int** firstneigh_);
-
+extern DP_Nlist* DP_NewNlist_mapping(
```
njzjz, line 27: Documentation is required, especially for the new argument mapping.

source/lmp/pair_deepmd.cpp
```diff
@@ -551,7 +557,7 @@ void PairDeepMD::compute(int eflag, int vflag) {
       (numb_models > 1 && (out_freq > 0 && update->ntimestep % out_freq == 0));
   if (do_ghost) {
     deepmd_compat::InputNlist lmp_list(list->inum, list->ilist, list->numneigh,
-                                       list->firstneigh);
+                                       list->firstneigh, mapping);
```
njzjz, line 560: If mapping is only correct for the serial run, we may only pass mapping when the number of ranks is 1 to prevent incorrect results.

source/lmp/pair_deepmd.cpp
njzjz, line 1: A test is required.

njzjz (changes requested):

PR #3589, opened 2024-03-23
Chore: refactor PT compute_stat

anyangml:
After #3593, need to remove bias in fitting net, and redesign UT's calling methods from AtomicModels.

- [ ] update doc --> bias file name
- [ ] refactor UTs


deepmd/pt/utils/stat.py
```diff
@@ -200,3 +238,153 @@ def model_forward_auto_batch_size(*args, **kwargs):
             stat_file_path.save_numpy(bias_atom_e)
     assert all(x is not None for x in [bias_atom_e])
     return to_torch_tensor(bias_atom_e)
+
+
+def compute_output_stats_with_atomic(
```
wanghan-iapcm, line 243: I find the current implementation has too much hard coding, which makes it not easily extensible to new fitting targets.

PR #3602, opened 2024-03-25
enable TCH rule to move imports to TYPE_CHECKING

njzjz:


njzjz:
This PR is not compatible with PyTorch. Closing.

PR #3613, opened 2024-03-27
fix c++ interface bug

CaRoLZhangxy:
https://github.com/deepmodeling/deepmd-kit/issues/3578
https://github.com/deepmodeling/deepmd-kit/issues/3579


njzjz (commented):
Could you remove the `skip` and `TODO` in #3578 and #3579? If the skipped tests can pass, this PR does work.

source/api_cc/src/DeepPotPT.cc
```diff
@@ -47,6 +47,12 @@ void DeepPotPT::init(const std::string& model,
     std::cout << "load model from: " << model << " to gpu " << gpu_rank
               << std::endl;
   }
+  int gpu_num = -1;
+  DPGetDeviceCount(gpu_num);
+  if (gpu_id > gpu_num) {
+    throw deepmd::deepmd_exception(
+        "current rank" + gpu_id + "is larger than the number of gpu" + gpu_num);
+  }
```
njzjz, line 55: We should use `gpu_id %= gpu_num` like TF

https://github.com/deepmodeling/deepmd-kit/blob/b875ea8f6661b6e1567537ead7e2b4a8b14ea113/source/api_cc/src/DeepPotTF.cc#L447-L449

njzjz (changes requested):

njzjz (approved):

Issue #3475, opened 2024-03-17
[BUG] pt: setting `batch_size` to `list` throws errors


njzjz:
### Bug summary

In the PyTorch backend, setting `batch_size` to `list` throws errors as shown below.

### DeePMD-kit Version

v3.0.0a0-28-ged831c88

### TensorFlow Version

PT v2.2.0+cu121-g8ac9b20d4b0

### How did you download the software?

Built from source

### Input Files, Running Commands, Error Log, etc.

```
Traceback (most recent call last):
  File "/home/jz748/anaconda3/bin/dp", line 8, in <module>
    sys.exit(main())
  File "/home/jz748/codes/deepmd-kit/deepmd/main.py", line 807, in main
    deepmd_main(args)
  File "/home/jz748/anaconda3/lib/python3.10/site-packages/torch/distributed/elastic/multiprocessing/errors/__init__.py", line 347, in wrapper
    return f(*args, **kwargs)
  File "/home/jz748/codes/deepmd-kit/deepmd/pt/entrypoints/main.py", line 306, in main
    train(FLAGS)
  File "/home/jz748/codes/deepmd-kit/deepmd/pt/entrypoints/main.py", line 270, in train
    trainer = get_trainer(
  File "/home/jz748/codes/deepmd-kit/deepmd/pt/entrypoints/main.py", line 166, in get_trainer
    ) = prepare_trainer_input_single(
  File "/home/jz748/codes/deepmd-kit/deepmd/pt/entrypoints/main.py", line 149, in prepare_trainer_input_single
    train_data_single = DpLoaderSet(
  File "/home/jz748/codes/deepmd-kit/deepmd/pt/utils/dataloader.py", line 129, in __init__
    system_dataloader = DataLoader(
  File "/home/jz748/anaconda3/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 356, in __init__
    batch_sampler = BatchSampler(sampler, batch_size, drop_last)
  File "/home/jz748/anaconda3/lib/python3.10/site-packages/torch/utils/data/sampler.py", line 267, in __init__
    raise ValueError(f"batch_size should be a positive integer value, but got batch_size={batch_size}")
ValueError: batch_size should be a positive integer value, but got batch_size=[1, 1, 1]
```

### Steps to Reproduce

```sh
cd examples/water/se_atten
```
Do the following modifications:

```diff
[...]

PR #3614, opened 2024-03-27
pt: support list format  batch size

CaRoLZhangxy:
https://github.com/deepmodeling/deepmd-kit/issues/3475

### DIFF
--- a/deepmd/pt/utils/dataloader.py
+++ b/deepmd/pt/utils/dataloader.py
@@ -125,6 +125,7 @@ class DpLoaderSet(Dataset):
             self.batch_sizes = batch_size
         else:
             self.batch_sizes = batch_size * np.ones(len(systems), dtype=int)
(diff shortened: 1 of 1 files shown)


deepmd/pt/utils/dataloader.py
```diff
@@ -106,29 +107,33 @@ def construct_dataset(system):
...
+                ni = ii._natoms
+                bsi = rule // ni
+                if bsi * ni < rule:
+                    bsi += 1
+                self.batch_sizes.append(bsi)
+        elif isinstance(batch_size, list):
+            self.batch_sizes = batch_size
+        else:
+            self.batch_sizes = batch_size * np.ones(len(systems), dtype=int)
+        for system, batch_size in zip(self.systems, self.batch_sizes):
```
njzjz, line 128: We may need to check and ensure systems and batch sizes have the same length.

njzjz (approved):

PR #3617, opened 2024-03-28
feat: consistent type embedding

njzjz:


source/tests/tf/test_model_se_a.py
```diff
@@ -299,7 +299,10 @@ def test_model_atom_ener_type_embedding(self):
         test_data = data.get_test()
         numb_test = 1
 
-        typeebd = TypeEmbedNet(**jdata["model"]["type_embeding"])
+        typeebd = TypeEmbedNet(
+            ntypes=len(jdata["model"]["descriptor"]["type_map"]),
```
njzjz, line 303:
```suggestion
            ntypes=len(jdata["model"]["descriptor"]["sel"]),
```

source/tests/tf/test_model_se_a_ebd_v2.py
```diff
@@ -60,6 +60,7 @@ def test_model(self):
         jdata["model"]["type_embedding"]["seed"] = 1
         typeebd_param = jdata["model"]["type_embedding"]
         typeebd = TypeEmbedNet(
+            ntypes=len(jdata["model"]["descriptor"]["type_map"]),
```
njzjz, line 63:
```suggestion
            ntypes=len(jdata["model"]["descriptor"]["sel"]),
```

wanghan-iapcm (commented):
It seems that by this PR, when `neuron == []`, the embedding is equivalent to identity, rather than a linear mapping. It seems that the implementation should be via the `FittingNet` not the `EmbeddingNet`. 
Please check if I am wrong.

deepmd/pt/model/network/network.py
```diff
@@ -591,6 +598,126 @@ def share_params(self, base_class, shared_level, resume=False):
...
+            ntypes,
+            self.neuron,
+            self.activation_function,
+            self.resnet_dt,
+            self.precision,
+        )
+        for param in self.parameters():
+            param.requires_grad = trainable
+
+    def forward(self, device: torch.device):
```
wanghan-iapcm, line 658: The input of the type embedding should be the atom types, an integer tensor.

njzjz: Currently, it's consistent with TF. (I don't change TF in this PR.) Do we need to make changes to TF?

wanghan-iapcm: > Currently, it's consistent with TF. (I didn't change TF in this PR.) Do we need to make changes to TF?

I get the point. the implementation is correct.

deepmd/pt/model/network/network.py
```diff
@@ -591,6 +598,126 @@ def share_params(self, base_class, shared_level, resume=False):
...
+        Random seed for initializing the network parameters.
+    padding
+        Concat the zero padding to the output, as the default embedding of empty type.
+    """
+
+    def __init__(
+        self,
+        *,
+        ntypes: int,
+        neuron: List[int] = [],
```
wanghan-iapcm, line 628: empty neuron seems not to be valid. otherwise we should provide the UT for the empty case.

wanghan-iapcm (changes requested):

deepmd/pt/model/network/network.py
njzjz: I remove the default value.

njzjz: Considering it is not valid, I think it has never been used.

wanghan-iapcm (approved):

iProzd (approved):
I agree with this PR but with one comment, will we keep `bias` for `TypeEmbedNet` in each layer? Because it may be confusing for some one-hot analysis on embedding weights, such as interpolation on different elements and etc.

njzjz:
> I agree with this PR but with one comment, will we keep `bias` for `TypeEmbedNet` in each layer? Because it may be confusing for some one-hot analysis on embedding weights, such as interpolation on different elements and etc.

Keeping and removing `bias` are equivalent only when the activation function is linear.

We should not remove the bias if we do not fix the activation function to linear.

The configuration of the type embedding may need further discussion, i.e., whether we allow flexible configurations for the type embedding.

PR #3619, opened 2024-03-28
pt: use unified activation

iProzd:


deepmd/pt/utils/utils.py
njzjz, line 1: I have several questions about this PR.
(1) Are you sure this is enough? I remember that TorchScript does not support returning a method.
(2) I think `ActivationFn` below is used instead of `get_activation_fn`?
(2) Why don't other activation functions (tanh, linear, etc) need it?

iProzd: (1)(3) Because only gelu here uses lambda expression, which is not supported by jit. I've test to freeze and test models using gelu and it works well, maybe I'll add a ut for this.
(2) These two methods are both used, `ActivationFn` is used in SimpleLinear and  `get_activation_fn ` is used in some descriptor level. Maybe we need to unite them, e.g. only keep `ActivationFn`.

njzjz: https://github.com/deepmodeling/deepmd-kit/blob/78189568e9285362f4cfc75ca3ebbb48983695bc/deepmd/pt/utils/utils.py#L40-L41

njzjz (approved):

PR #3625, opened 2024-03-30
feat: apply descriptor exclude_types to env mat stat

njzjz:


deepmd/pt/utils/env_mat_stat.py
```diff
@@ -146,6 +146,9 @@ def iter(
                 radial_only,
                 protection=self.descriptor.env_protection,
             )
+            # apply excluded_types
+            exclude_mask = self.descriptor.emask(nlist, extended_atype)
+            env_mat *= exclude_mask.unsqueeze(-1)
```
wanghan-iapcm, line 151: may be a better option to use the same mechanism as the atomic model 
https://github.com/deepmodeling/deepmd-kit/blob/cc30f275f34fd999658025446703289c4718aa31/deepmd/pt/model/atomic_model/dp_atomic_model.py#L209-L222

wanghan-iapcm (commented):
I agree with this PR, but please not that would deprecate the exclude type in the future.

deepmd/tf/descriptor/se_r.py
```diff
@@ -196,6 +196,18 @@ def __init__(
                 rcut_smth=self.rcut_smth,
                 sel=self.sel_r,
             )
+            if len(self.exclude_types):
+                # exclude types applied to data stat
+                mask = self.build_type_exclude_mask(
+                    self.exclude_types,
+                    self.ntypes,
+                    self.sel_a,
```
njzjz, line 204:
```suggestion
                    self.sel_r,
```

njzjz:
> would deprecate the exclude type in the future.

What's the reason for that?

deepmd/pt/utils/env_mat_stat.py
njzjz: After I check the code, I think the zeros generated by the descriptor exclude types will still be passed to the neural network, so I think my implementation is correct.
The zeros generated by `pair_exclude_types` seems not used for stat. Need to confirm whether it is correct.

wanghan-iapcm:
> > would deprecate the exclude type in the future.
> 
> What's the reason for that?

because not all atomic model has descriptor.  the dp atomic model's exclude pair is equivalent to the descriptors exclude pair.

njzjz:
> > > would deprecate the exclude type in the future.
> 
> > 
> 
> > What's the reason for that?
> 
> 
> 
> because not all atomic model has descriptor.  the dp atomic model's exclude pair is equivalent to the descriptors exclude pair. 

An atomic model may not only contain one descriptor. Note that two descriptors in a DPRc model have different exclude pairs.

wanghan-iapcm:
> > > > would deprecate the exclude type in the future.
> > 
> > 
> > > 
> > 
> > 
> > > What's the reason for that?
> > 
> > 
> > because not all atomic model has descriptor.  the dp atomic model's exclude pair is equivalent to the descriptors exclude pair.
> 
> An atomic model may not contain only one descriptor. Note that two descriptors in a DPRc model have different exclude pairs. So it may not be equivalent.

I see in multiple descriptor scenario they are not equivalent.

wanghan-iapcm (approved):

deepmd/pt/utils/env_mat_stat.py
wanghan-iapcm: As for descriptors, the pair exclusion is not used in stat.

iProzd (commented):
I notice that the `excluded_types` only masks the input `env_mat`, while still keeps the total number of atoms in `excluded_types` (`sumn`). Another way is to mask `excluded_types` in both numerator (`sumr` etc.) and denominator (`sumn`) . Which one should we prefer? @wanghan-iapcm @njzjz

wanghan-iapcm:
> I notice that the `excluded_types` only masks the input `env_mat`, while still keeps the total number of atoms in `excluded_types` (`sumn`). Another way is to mask `excluded_types` in both numerator (`sumr` etc.) and denominator (`sumn`) . Which one should we prefer? @wanghan-iapcm @njzjz

I would prefer mask both the numerator and the denominator, which gives consistent result as if the masked pairs do not exists.

njzjz:
> > I notice that the `excluded_types` only masks the input `env_mat`, while still keeps the total number of atoms in `excluded_types` (`sumn`). Another way is to mask `excluded_types` in both numerator (`sumr` etc.) and denominator (`sumn`) . Which one should we prefer? @wanghan-iapcm @njzjz
> 
> I would prefer mask both the numerator and the denominator, which gives consistent result as if the masked pairs do not exists.

They are not equivalent. When masked pairs do not exist, zeros (meaning no neighbor) will be used for the stat.

The current masking behavior has some problems. The env mat is always masked to zero, but if `set_davg` is false, it should be masked to `davg`.

PR #3626, opened 2024-03-30
feat: auto batch size supports methods that return a dict

wanghan-iapcm:
- also add missing UT for execute_all

### DIFF
--- a/deepmd/pt/utils/auto_batch_size.py
+++ b/deepmd/pt/utils/auto_batch_size.py
@@ -12,6 +12,32 @@ from deepmd.utils.batch_size import AutoBatchSize as AutoBatchSizeBase
 
 
 class AutoBatchSize(AutoBatchSizeBase):
+    """Auto batch size.
+
+    Parameters
+    ----------
+    initial_batch_size : int, default: 1024
+        initial batch size (number of total atoms) when DP_INFER_BATCH_SIZE
+        is not set
+    factor : float, default: 2.
+        increased factor
+    returned_dict:
+        if the batched method returns a dict of arrays.
+
+    """
+
+    def __init__(
+        self,
+        initial_batch_size: int = 1024,
+        factor: float = 2.0,
+        returned_dict: bool = False,
+    ):
+        super().__init__(
+            initial_batch_size=initial_batch_size,
+            factor=factor,
+        )
+        self.returned_dict = returned_dict
+
     def is_gpu_available(self) -> bool:
         """Check if GPU is available.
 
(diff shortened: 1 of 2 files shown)


deepmd/pt/utils/auto_batch_size.py
```diff
@@ -78,26 +104,44 @@ def execute_with_batch_size(
...
+                ret = torch.cat(r, dim=0)
             else:
                 raise RuntimeError(f"Unexpected result type {type(r[0])}")
-        r = tuple(r_list)
-        if len(r) == 1:
-            # avoid returning tuple if callable doesn't return tuple
-            r = r[0]
+            return ret
+
+        if not self.returned_dict:
```
njzjz, line 139: I think we don't need this argument? Just check the type of results

deepmd/pt/utils/auto_batch_size.py
```diff
@@ -12,6 +12,32 @@
...
+
+    Parameters
+    ----------
+    initial_batch_size : int, default: 1024
+        initial batch size (number of total atoms) when DP_INFER_BATCH_SIZE
+        is not set
+    factor : float, default: 2.
+        increased factor
+    returned_dict:
+        if the batched method returns a dict of arrays.
```
anyangml, line 25: bool, default: false

deepmd/pt/utils/auto_batch_size.py
```diff
@@ -78,26 +104,44 @@ def execute_with_batch_size(
...
-            if n_batch:
-                for rr in result:
-                    rr.reshape((n_batch, -1))
-                results.append(result)
-        r_list = []
-        for r in zip(*results):
+
+            def append_to_list(res_list, res):
+                if n_batch:
+                    res_list.append(res)
```
anyangml, line 116: no need to reshape?

deepmd/pt/utils/auto_batch_size.py
wanghan-iapcm: good idea plz check [3c4ff2c](https://github.com/deepmodeling/deepmd-kit/pull/3626/commits/3c4ff2c1c34675c93b853902fd6b0d699034479a)

deepmd/pt/utils/auto_batch_size.py
wanghan-iapcm: the argument is removed according to @njzjz 's suggestion.

deepmd/pt/utils/auto_batch_size.py
wanghan-iapcm: no the shape of the returned tensors are preserved. A concatenation is performed on the first (batch size) dimension of batch calculated tensors. plz see 
https://github.com/deepmodeling/deepmd-kit/blob/3c4ff2c1c34675c93b853902fd6b0d699034479a/deepmd/pt/utils/auto_batch_size.py#L132-L139

njzjz (approved):

anyangml (approved):

PR #3628, opened 2024-03-31
feat: compute output stat for a dict of labels.

wanghan-iapcm:
- add UT for it.
- at the moment, only energy is supported in the `base_atomic_model`. handling of multiple output stat will be implemented in a future PR.

### DIFF
--- a/deepmd/pt/utils/stat.py
+++ b/deepmd/pt/utils/stat.py
@@ -1,5 +1,8 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 import logging
+from pathlib import (
+    Path,
+)
 from typing import (
     Callable,
     List,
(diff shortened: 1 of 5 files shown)


deepmd/pt/utils/stat.py
```diff
@@ -78,9 +81,39 @@ def make_stat_input(datasets, dataloaders, nbatches):
...
 
+def restore_from_file(
+    stat_file_path: Path,
+    keys: List[str] = ["energy"],
+) -> Optional[dict]:
+    if stat_file_path is None:
+        return None
+    stat_files = [stat_file_path / f"bias_atom_{kk}.npy" for kk in keys]
+    if any(not (ii.is_file()) for ii in stat_files):
+        return None
```
anyangml, line 92: should we raise an error here?

wanghan-iapcm: when the stat file is empty or not complete, it is expected that the stat is computed. 

please check `source.tests.pt.test_stat.TestOutputStat` the first computation 
https://github.com/wanghan-iapcm/deepmd-kit/blob/88c1add3dacb6fe682881673d9bbe9e7c965f40e/source/tests/pt/test_stat.py#L361-L368
the stat file is empty, it is expected to compute the stat and store the result in the file.

anyangml (approved):

njzjz (approved):

PR #3630, opened 2024-04-01
remove multiprocess and buffer in multitask mode

CaRoLZhangxy:
the trainer choose random model for each step. Prefetching for each model may be inefficient.


iProzd (commented):
Is there efficiency comparison before and after this modification in multitask training with numerous datasets?

PR #3633, opened 2024-04-01
feat(pt): allow using directories to store stat

njzjz:
If the stat file ends with .h5 or .hdf5, an HDF5 file is used; otherwise, a directory containing NumPy binary files is used. This gives flexibility to users.


wanghan-iapcm (approved):

iProzd (approved):

iProzd:
Is 'saving to numpy directory' tested in UT?

njzjz:
> Is 'saving to numpy directory' tested in UT?

This was added in #3631

PR #3636, opened 2024-04-02
fix: set rpath for libtorch and protobuf

njzjz:
Fix #3568.

Set run path for libtorch and protobuf. Refactor the original `TensorFlow_LIBRARY_PATH` to `BACKEND_LIBRARY_PATH` (list).


njzjz:
cc: @Yi-FanLi

wanghan-iapcm (approved):

PR #3642, opened 2024-04-04
compute output stat for atomic model

wanghan-iapcm:
This PR:
- breaking change: the base atomic model is now a module. 
  - reason: the out stat is a data attribute of the base atomic model. 
- implement the `compute_or_load_output_stat` for the base atomic model. the method computes both bias and std. 
- the derived atomic models call the `compute_or_load_output_stat`  method for computing output stat. 
- atomic model provides the `apply_out_stat`, the derived class may override the method to define how the statistics is applied to an atomic model's output. @anyangml may need. 
- `out_stat` support statistics of output tensor of any shape. 

@iProzd  please check if i took it correctly in [ce7ec1f](https://github.com/deepmodeling/deepmd-kit/pull/3642/commits/ce7ec1f39be669fcc63a05e9474bc248d9a54d8a)

To be done:
- atomic statistics of the bias and std. @anyangml 
- erialization and deserialization.

### DIFF
--- a/deepmd/pt/model/atomic_model/base_atomic_model.py
+++ b/deepmd/pt/model/atomic_model/base_atomic_model.py
@@ -23,6 +23,7 @@ from deepmd.dpmodel.output_def import (
 from deepmd.pt.utils import (
     AtomExcludeMask,
     PairExcludeMask,
+    env,
 )
 from deepmd.pt.utils.nlist import (
     extend_input_and_build_neighbor_list,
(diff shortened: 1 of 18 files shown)


njzjz (approved):

anyangml (approved):

deepmd/pt/model/atomic_model/base_atomic_model.py
```diff
@@ -210,9 +281,60 @@ def compute_or_load_stat(
...
+        out_bias, out_std = self._fetch_out_stat(self.bias_keys)
+        for kk in self.bias_keys:
+            # nf x nloc x odims, out_bias: ntypes x odims
+            ret[kk] = ret[kk] + out_bias[kk][atype]
+        return ret
+
     def change_out_bias(
         self,
         sample_merged,
+        stat_file_path: Optional[DPPath] = None,
```
iProzd, line 337: docstr missed

source/tests/pt/test_multitask.py
```diff
@@ -128,17 +128,17 @@ def test_multitask_train(self):
...
                     multi_state_dict_finetuned[state_key],
                 )
-            elif "model_3" in state_key and "bias_atom_e" not in state_key:
+            elif "model_3" in state_key and "out_bias" not in state_key:
                 torch.testing.assert_close(
                     multi_state_dict[state_key.replace("model_3", "model_2")],
                     multi_state_dict_finetuned[state_key],
                 )
-            elif "model_4" in state_key and "fitting_net" not in state_key:
+            elif "model_4" in state_key and "atomic_model" not in state_key:
```
iProzd, line 141: "model_4" has the same descriptor but random fitting net and bias.

```suggestion
            elif "model_4" in state_key and "fitting_net" not in state_key and "out_bias" not in state_key:
```

iProzd (changes requested):

iProzd (approved):

deepmd/pt/model/atomic_model/base_atomic_model.py
```diff
@@ -35,19 +36,88 @@
...
+
+    """
+
     def __init__(
         self,
+        type_map: List[str],
         atom_exclude_types: List[int] = [],
         pair_exclude_types: List[Tuple[int, int]] = [],
+        rcond: Optional[float] = None,
+        preset_out_bias: Optional[Dict[str, torch.Tensor]] = None,
```
njzjz, line 74: line 74 is different from line 60

wanghan-iapcm: originated from the difference between https://github.com/wanghan-iapcm/deepmd-kit/blob/2171e19cec5b5f9c07f8d530e210d1539e5f33ef/deepmd/pt/utils/stat.py#L233 and https://github.com/wanghan-iapcm/deepmd-kit/blob/2171e19cec5b5f9c07f8d530e210d1539e5f33ef/deepmd/pt/utils/stat.py#L212

Issue #3641, opened 2024-04-04
any c_api for model deviation compute without nlist?


robinzyb:
### Summary

nope

### DeePMD-kit Version

xxx

### TensorFlow Version

xx

### Details

I would like to add model devi in cp2k while the nlist is required in 
https://github.com/deepmodeling/deepmd-kit/blob/b875ea8f6661b6e1567537ead7e2b4a8b14ea113/source/api_c/src/c_api.cc#L1065-L1081
How do I bypass the Nlist input?

njzjz:
I remember no one has written it, including the C++ API.

PR #3647, opened 2024-04-05
feat: model devi C/C++ API without nlist

robinzyb:
To solve this issue:
https://github.com/deepmodeling/deepmd-kit/issues/3641

### DIFF
--- a/source/api_cc/include/DeepPot.h
+++ b/source/api_cc/include/DeepPot.h
@@ -660,6 +660,41 @@ class DeepPotModelDevi {
                const std::vector<VALUETYPE>& aparam = std::vector<VALUETYPE>());
 
   /**
+   * @brief Evaluate the energy, force, virial, atomic energy, and atomic virial
+   *by using these DP models.
+   * @param[out] all_ener The system energies of all models.
+   * @param[out] all_force The forces on each atom of all models.
+   * @param[out] all_virial The virials of all models.
+   * @param[out] all_atom_energy The atomic energies of all models.
+   * @param[out] all_atom_virial The atomic virials of all models.
+   * @param[in] coord The coordinates of atoms. The array should be of size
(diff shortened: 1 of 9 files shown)


source/api_cc/include/DeepPot.h
```diff
@@ -629,6 +629,36 @@ class DeepPotModelDevi {
...
+   * nframes x natoms x dim_aparam.
+   * natoms x dim_aparam. Then all frames are assumed to be provided with the
+   *same aparam. dim_aparam. Then all frames and atoms are provided with the
+   *same aparam.
+   **/
+  template <typename VALUETYPE>
+  void compute(std::vector<ENERGYTYPE>& all_ener,
+               std::vector<std::vector<VALUETYPE> >& all_force,
+               std::vector<std::vector<VALUETYPE> >& all_virial,
+               const std::vector<VALUETYPE>& coord,
```
robinzyb, line 656: I am consued that why the variables `coord` and `atype` are not the same as they are in the implementation `dcoord_`, `datype` @njzjz . I just follow the convention. Is it a typo or something else?

njzjz: The convention exists in the initial version: https://github.com/deepmodeling/deepmd-kit/blame/e86cf20b4a9ad6ae629ec16c566ed3b086cb4b75/source/lib/src/NNPInter.cc
where dcoord is a vector and coord is a tensor. @wanghan-iapcm may answer.

njzjz (approved):
Looks good to me. I think you are expecting a new version, right?

robinzyb:
> Looks good to me. I think you are expecting a new version, right?

Many thanks. Recently I will refactor the dpmd api in cp2k. Move the dpmd section from the forcefield section to a new section under force_eval.

PR #3654, opened 2024-04-08
Chore: refactor atomic bias

anyangml:
`var_name` is no longer required as user input for `polar` and `dipole`.


deepmd/entrypoints/test.py
```diff
@@ -870,7 +870,7 @@ def test_polar(
         arrays with results and their shapes
     """
     data.add(
-        "polarizability" if not atomic else "atomic_polarizability",
```
njzjz, line 873: It seems a breaking change. What's the reason to introduce it? Besides, lots of codes still use `atomic_`, such as those in `data/`, `examples/`, `doc/`, `tf/loss/`.

anyangml: > It seems a breaking change. What's the reason to introduce it? Besides, lots of codes still use `atomic_`, such as those in `data/`, `examples/`, `doc/`, `tf/loss/`.

currently some fitting nets use `atomic_***` as the label while others use `atom_***`, when calculating stat with atomic labels, we have to introduce hardcoded keys depending on the fitting nets, which will be hard to maintain.

anyangml: maybe we can keep using `atomic_***` in tf, if we separate the data.

njzjz: Could we support both? Just need to implement an alias feature in the data requirement.

anyangml: I have reverted these changes. Will address this issue when migrating polar bias.

njzjz (approved):

source/tests/pt/model/test_atomic_model_stat.py
wanghan-iapcm, line 1: may rename as test_atomic_model_global_stat.py

source/tests/pt/model/test_atomic_bias.py
wanghan-iapcm, line 1: may rename as test_atomic_model_atomic_stat.py

deepmd/pt/utils/stat.py
```diff
@@ -246,87 +274,308 @@ def compute_output_stats(
...
+        )
 
-        # unbias_e is only used for print rmse
-        if model_forward is None:
-            unbias_e = {
-                kk: merged_natoms @ bias_atom_e[kk].reshape(ntypes, -1) for kk in keys
+        # concat all frames within those systmes
+        model_pred_g = (
+            {
+                kk: np.concatenate(model_pred_g[kk])
```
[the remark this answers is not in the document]
anyangml: the aggregation is in the global function 

```
model_pred = {kk: np.sum(model_pred[kk], axis=1) for kk in keys}
        stats_input = {
            kk: merged_output[kk] - model_pred[kk] for kk in keys if kk in merged_output
        }
```

deepmd/pt/utils/stat.py
```diff
@@ -246,87 +274,308 @@ def compute_output_stats(
...
-                (unbias_e[kk].reshape(nf, -1) - merged_output[kk].reshape(nf, -1))
-                / atom_numbs[:, None]
-            )
-            log.info(
-                f"RMSE of {kk} per atom after linear regression is: {rmse_ae} in the unit of {kk}."
-            )
+            # use atomic bias whenever available
+            if kk in bias_atom_a:
+                bias_atom_e[kk] = bias_atom_a[kk]
+                std_atom_e[kk] = std_atom_a[kk]
```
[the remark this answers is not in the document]
anyangml: it is processed within the global function.
https://github.com/deepmodeling/deepmd-kit/blob/99f18dabac4dcb86f7a7481aa2cc0c9bc11a47af/deepmd/pt/utils/stat.py#L489

PR #3662, opened 2024-04-10
Chore: migrate and refactor polar and dos bias

anyangml:
- [x] remove polar and dos bias from fitting net
- [x] update apply bias
- [x] update UT for polar stat


anyangml:
@njzjz take a look, I think this will convert all `atomic_***` labels to `atom_***` without a breaking change.

njzjz (approved):

njzjz:
> @njzjz take a look, I think this will convert all `atomic_***` labels to `atom_***` without a breaking change.

I approve, but it seems the PR title is unrelated, right?

deepmd/pt/model/task/invar_fitting.py
```diff
@@ -146,40 +138,6 @@ def deserialize(cls, data: dict) -> "GeneralFitting":
         check_version_compatibility(data.pop("@version", 1), 1, 1)
         return super().deserialize(data)
 
-    def compute_output_stats(
-        self,
-        merged: Union[Callable[[], List[dict]], List[dict]],
```
anyangml, line 151: @wanghan-iapcm should I remove all fitting net level stat UTs, `tests/pt/test_stat.py`, and `tests/pt/model/test_polar_stat.py`?

wanghan-iapcm: I agree

source/tests/pt/model/test_atomic_model_atomic_stat.py
```diff
@@ -212,6 +212,8 @@ def cvt_ret(x):
             self.merged_output_stat, stat_file_path=self.stat_file_path
         )
         ret1 = md0.forward_common_atomic(*args)
+        expected_std = np.ones((2, 2, 2))  # 2 keys, 2 atypes, 2 max dims.
```
wanghan-iapcm, line 215: shouldn't the std of type 0  be 0? All atomic labels of type 0 are 5.0.

deepmd/utils/data.py
wanghan-iapcm, line 1: is the current implementation support both the atom_ and atomic_ prefix of data files?

source/tests/pt/model/test_atomic_model_atomic_stat.py
anyangml: > shouldn't the std of type 0 be 0? All atomic labels of type 0 are 5.0.

the std are updated to 1 in `_post_process_stat`, if we want the actual input, we would need to turn that off.

deepmd/utils/data.py
anyangml: > is the current implementation support both the atom_ and atomic_ prefix of data files?

I think so. It will convert the name of the ingested data to `atom_**` no matter what the file name is.

deepmd/pt/model/atomic_model/polar_atomic_model.py
```diff
@@ -24,5 +24,38 @@ def apply_out_stat(
...
+
+        """
+        out_bias, out_std = self._fetch_out_stat(self.bias_keys)
+
+        if self.fitting_net.shift_diag:
+            nframes, nloc = atype.shape
+            device = out_bias[self.bias_keys[0]].device
+            for kk in self.bias_keys:
+                ntypes = out_bias[kk].shape[0]
+                temp = torch.zeros(ntypes, device=device)
```
njzjz, line 44: should set dtype explicitly

deepmd/pt/model/atomic_model/polar_atomic_model.py
```diff
@@ -24,5 +24,38 @@ def apply_out_stat(
...
+                for i in range(ntypes):
+                    temp[i] = torch.mean(torch.diagonal(out_bias[kk][i].reshape(3, 3)))
+                modified_bias = temp[atype]
+
+                # (nframes, nloc, 1)
+                modified_bias = (
+                    modified_bias.unsqueeze(-1) * self.fitting_net.scale[atype]
+                )
+
+                eye = torch.eye(3, device=device)
```
njzjz, line 54: should set dtype explicitly

source/tests/pt/model/test_atomic_model_atomic_stat.py
njzjz, line 215: should set dtype explicitly

wanghan-iapcm: the `_post_process_stat` should only be applied to global stat, because in that case the std of type is not available. In the case of atomic stat, we use the std from data for each atom type.

deepmd/pt/utils/stat.py
```diff
@@ -567,9 +567,8 @@ def compute_output_stats_atomic(
                 nan_padding = np.empty((missing_types, bias_atom_e[kk].shape[1]))
                 nan_padding.fill(np.nan)
                 bias_atom_e[kk] = np.concatenate([bias_atom_e[kk], nan_padding], axis=0)
-                std_atom_e[kk] = np.concatenate([bias_atom_e[kk], nan_padding], axis=0)
+                std_atom_e[kk] = np.concatenate([std_atom_e[kk], nan_padding], axis=0)
```
wanghan-iapcm, line 570: Yes you find a bug :P

wanghan-iapcm (approved):

njzjz (approved):

PR #3663, opened 2024-04-11
Chore: refactor dpmodel

anyangml:
To decouple Model from DPAtomicModel, allowing make_model on different atomic model.


deepmd/pt/model/model/ener_model.py
```diff
@@ -6,12 +6,23 @@
...
 from .dp_model import (
     DPModel,
 )
+from .make_model import (
+    make_model,
+)
 
 
-class EnergyModel(DPModel):
+@BaseModel.register("standard")
```
njzjz, line 24: Inconsistent with dipole

https://github.com/deepmodeling/deepmd-kit/blob/56accdfaa1779377d57913b1b05f77a011e69aee/deepmd/pt/model/model/dipole_model.py#L24-L25

njzjz (changes requested):

deepmd/pt/model/model/ener_model.py
njzjz: What is the proposed design here?

deepmd/pt/model/model/dp_model.py
```diff
@@ -95,30 +23,3 @@ def update_sel(cls, global_jdata: dict, local_jdata: dict):
...
         )
         return local_jdata_cpy
-
-    def get_fitting_net(self):
-        """Get the fitting network."""
-        return self.atomic_model.fitting_net
-
-    def get_descriptor(self):
-        """Get the descriptor."""
-        return self.atomic_model.descriptor
```
wanghan-iapcm, line 105: fitting and descriptor access methods may be preserved here. The `dp` model assumes a descriptor and a fitting net as members.

deepmd/pt/model/model/dp_model.py
```diff
@@ -1,83 +1,11 @@
...
-                cls = EnergyModel
-            elif isinstance(fitting, DipoleFittingNet):
-                cls = DipoleModel
-            elif isinstance(fitting, PolarFittingNet):
-                cls = PolarModel
-            elif isinstance(fitting, DOSFittingNet):
-                cls = DOSModel
-            # else: unknown fitting type, fall back to DPModel
-        return super().__new__(cls)
+class DPModel:
```
wanghan-iapcm, line 7: the name `DPModel` is a bit misleading, because the class it self is not a model

deepmd/pt/model/model/ener_model.py
wanghan-iapcm: the principle is that each model is derived from its own atomic model. we need to implement energy, dos, models at the atomic level. Then we can remove the ugly code in the dp_model
https://github.com/deepmodeling/deepmd-kit/blob/a26b680308f02b4c0a3648aac7b76bebef0ae0ef/deepmd/pt/model/model/dp_model.py#L48-L80

njzjz: I see what you mean. But is it good to move all stuff to the top module?

https://github.com/deepmodeling/deepmd-kit/blob/56accdfaa1779377d57913b1b05f77a011e69aee/deepmd/pt/model/model/__init__.py#L172-L181

anyangml: > Inconsistent with dipole
> 
> https://github.com/deepmodeling/deepmd-kit/blob/56accdfaa1779377d57913b1b05f77a011e69aee/deepmd/pt/model/model/dipole_model.py#L24-L25

registered as "energy" and "standard", in the numpy backend, we only have implementation for `EnergyModel` which is `standard`.

njzjz: >  In the numpy backend, we only have implementation for `EnergyModel` which is `standard`.

I think we don't even have `EnergyModel` in the numpy backend... Please check below.

https://github.com/deepmodeling/deepmd-kit/blob/a26b680308f02b4c0a3648aac7b76bebef0ae0ef/deepmd/dpmodel/model/dp_model.py#L20-L21

I believe it doesn't represent any specific property.

njzjz: I mean, is `standard` required by something?

anyangml: > I mean, is `standard` required by something?

I guess I will have to change all `standard` to `ener`, which also includes the default tag in argcheck?

njzjz: How does standard + dipole fitting work in the current implementation?

anyangml: > How does standard + dipole fitting work in the current implementation?

in pytorch backend, `standard` is basically an alias for `ener`. I don't think we can register a model as `standard` + `dipole`, the old standard model (which was the DPModel) is removed.

njzjz: So again, what's the reason for removing DPModel, especially `__new__`? I don't think `__new__` conflicts with `make_model` or other methods.

anyangml: > So again, what's the reason for removing DPModel, especially `__new__`? I don't think `__new__` conflicts with `make_model` or other methods.

In the old design, DPModel inherit from make_model(DPAtomicModel), we want more flexibility, such that each Model can inherit from it's corresponding make_model(AtomicModel).

njzjz: We can keep __new__ and plugin behaviors but just remove inheritance.

anyangml: > We can keep `__new__` and plugin behaviors but just remove inheritance.

In that case, how do we handle the `deserialize` method in DPModel? If we remove inheritance, there is no deserialize for base model to call. 

```
@classmethod
    def deserialize(cls, data: dict) -> "BaseBaseModel":
        """Deserialize the model.
    
        Parameters
        ----------
        data : dict
            The serialized data
    
        Returns
        -------
        BaseModel
            The deserialized model
        """
        if inspect.isabstract(cls):
>           return cls.get_class_by_type(data["type"]).deserialize(data)
```

njzjz: We should choose one of below:
(1) Deprecate and remove `standard` in serialization.
(2) Keep `standard`, but have an abstract class to redirect to the actual class.

We shouldn't mix them.

deepmd/pt/model/atomic_model/dipole_atomic_model.py
```diff
@@ -0,0 +1,20 @@
...
+)
+
+import torch
+
+from .dp_atomic_model import (
+    DPAtomicModel,
+)
+
+
+class DPDipoleAtomicModel(DPAtomicModel):
```
wanghan-iapcm, line 13: we may assert the type of the fitting net at the `__inti__` of the derived atomic models.

deepmd/pt/model/model/dipole_model.py
```diff
@@ -6,12 +6,23 @@
...
+    DPModelCommon,
+)
+from .make_model import (
+    make_model,
 )
 
 
-class DipoleModel(DPModel):
+@BaseModel.register("dipole")
+class DipoleModel(DPModelCommon, make_model(DPDipoleAtomicModel)):
```
wanghan-iapcm, line 25: the `super` in the `__init__` only initialize the `DPModelCommon`, not the `make_model(DPDipoleAtomicModel)`.

please check the example
```python
class A:                                                                                                                                                                                       
  def __init__(self, a):                                                                                                                                                                       
    self.a = a                                                                                                                                                                                 
                                                                                                                                                                                               
class B:                                                                                                                                                                                       
  def __init__(self, b):                                                                                                                                                                       
    self.b = b                                                                                                                                                                                 
                                                                                                                                                                                               
class C(A, B):                                                                                                                                                                                 
  def __init__(self, *args, **kwargs):                                                                                                                                                         
    super().__init__(*args, **kwargs)                                                                                                                                                          
                                                                                                                                                                                               
                                                                                                                                                                                               
c = C(1.5)                                                                                                                                                                                     
print(c.a)                                                                                                                                                                                     
print(c.b)    
```
you will have the error
```
  File "/home/wanghan/tmp/a.py", line 16, in <module>
    print(c.b)
          ^^^
AttributeError: 'C' object has no attribute 'b'
```

deepmd/pt/model/model/dp_zbl_model.py
```diff
@@ -24,7 +24,7 @@
 
 
 @BaseModel.register("zbl")
-class DPZBLModel(DPZBLModel_):
+class DPZBLModel(DPModelCommon, DPZBLModel_):
```
wanghan-iapcm, line 27: DPZBLModel should not be derived from `DPModelComm`, because it is not a DP model.

anyangml:
@njzjz , it seems removing `standard` caused an inconsistency in deserialization between pt and tf. I am not entirely sure what is missing, can you take a look

deepmd/dpmodel/model/base_model.py
```diff
@@ -35,7 +35,7 @@ class BaseBaseModel(ABC, PluginVariant, make_plugin_registry("model")):
 
         def __new__(cls, *args, **kwargs):
             if inspect.isabstract(cls):
-                cls = cls.get_class_by_type(kwargs.get("type", "standard"))
+                cls = cls.get_class_by_type(kwargs.get("type", "ener"))
```
njzjz, line 38: should get the model type from the fitting type instead of `ener`

njzjz: Or modify `get_class_by_type`

deepmd/dpmodel/model/base_model.py
```diff
@@ -151,7 +151,7 @@ def update_sel(cls, global_jdata: dict, local_jdata: dict):
             local_jdata : dict
                 The local data refer to the current class
             """
-            cls = cls.get_class_by_type(local_jdata.get("type", "standard"))
+            cls = cls.get_class_by_type(local_jdata.get("type", "ener"))
```
njzjz, line 154: +1 here

deepmd/utils/argcheck.py
```diff
@@ -1372,7 +1372,7 @@ def model_args(exclude_hybrid=False):
                     *hybrid_models,
                 ],
                 optional=True,
-                default_tag="standard",
+                default_tag="ener",
```
njzjz, line 1375: Should be reverted.

deepmd/tf/model/model.py
```diff
@@ -589,7 +589,6 @@ def serialize(self, suffix: str = "") -> dict:
         raise NotImplementedError("Not implemented in class %s" % self.__name__)
 
 
-@Model.register("standard")
```
njzjz, line 592: Should be reverted.

deepmd/utils/argcheck.py
```diff
@@ -1383,7 +1383,7 @@ def standard_model_args() -> Argument:
     doc_fitting = "The fitting of physical properties."
 
     ca = Argument(
-        "standard",
+        "ener",
```
njzjz, line 1386: Should be reverted.

njzjz (changes requested):

deepmd/dpmodel/model/base_model.py
anyangml: in case there is no fitting type, which Model do we use?

wanghan-iapcm (approved):
GLTM

deepmd/tf/model/tensor.py
```diff
@@ -22,6 +22,8 @@
 )
 
 
+@StandardModel.register("dipole")
+@StandardModel.register("polar")
```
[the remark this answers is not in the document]
anyangml: @njzjz resolved.

njzjz (approved):

PR #3684, opened 2024-04-18
Deterministic version of CUDA forces and stresses kernels

mtaillefumier:
Calculations of the forces and stress are deterministic on GPU. It does not imply that the DeepMD code is deterministic by default as TensorFlow also requires to be set up properly either at runtime or during the initialization phase. 

To obtain the same model parameters, add the following variables to the job scripts

```[sh]
export TF_DETERMINISTIC_OPS=1
export TF_INTER_OP_PARALLELISM_THREADS=0
export TF_INTER_OP_PARALLELISM_THREADS=0
```
Details of the changes:

- Remove the use of atomic operations in the forces and stress kernels.
- Use template programming to minimize code duplication and minor refactoring

Authors :
- (@mtaillefumier) M. Taillefumier (ETH Zurich / CSCS) 
- (@asedova) A. Sedova (ORNL)


njzjz:
Please rebase the PR against the `devel` branch. We don't accept PRs on other branches.

source/lib/src/gpu/prod_force.cu
```diff
@@ -43,92 +43,146 @@ __global__ void force_deriv_wrt_center_atom(FPTYPE* force,
...
+
+    const int* nei_nei_list_ = &frame_neighbor_list_[atom_j * nnei];
+    int atom_id_position = 0;
+
+    // search the index of the atom i in the local neighbor list of atom j
+    for (atom_id_position = 0; atom_id_position < nnei; atom_id_position++) {
+      if (nei_nei_list_[atom_id_position] == atom_id) {
+        break;
+      }
+    }
```
wanghan-iapcm, line 106: The complexity of the index searching is of order N_nei, which does not present in the atomic operation implementation. Does it have an observable side effect on the performance of the `prod_force` operator?

mtaillefumier: I could not observe any side effect except fluctuations of 5 % in performance. The gain is that these operators are deterministic by default which is worth the 5% potential penalty (or less) introduced by this code change.

mtaillefumier:
I opened a new PR #3693 on the develop branch starting the latest devel branch as well. Same title, the same content

PR #3696, opened 2024-04-21
refact: the DPA1 descriptor

iProzd:
This PR reformat DPA1 descriptor consistent for TF/PT/DP

- Clean up the implementation of DPA1 descriptor in TF/PT
- Add numpy implementation for DPA1 descriptor
- Add `serialize` and `deserialize` for DPA1 descriptor in TF/PT/DP
- Add consistence test for TF/PT/DP

Now DPA1 descriptor in TF and PT&DP are consistent when:
`smooth_type_embdding` == False
`env_protection` == 0.0
`stripped_type_embedding` == False

 DPA1 descriptor in PT and DP are always consistent.

Need to fix 
- consistence in TF: when `excluded_types` != [] and `attn_layer` > 0.


node_modules/.cache/prettier/.prettier-caches/9be9545d53bb64e65febe2ff48926b4145285f3a.json
njzjz, line 1: should be deleted and added to gitignore

deepmd/tf/env.py
```diff
@@ -178,18 +178,18 @@ def dlopen_library(module: str, filename: str):
...
+    r"attention_layer_(\d+)/(c_key)/(bias)|"
+    r"attention_layer_(\d+)/(c_value)/(matrix)|"
+    r"attention_layer_(\d+)/(c_value)/(bias)|"
+    r"attention_layer_(\d+)/(c_out)/(matrix)|"
+    r"attention_layer_(\d+)/(c_out)/(bias)|"
+    r"attention_layer_(\d+)/(layer_normalization)/(beta)|"
+    r"attention_layer_(\d+)/(layer_normalization)/(gamma)|"
+    # r"attention_layer_(\d+)/(layer_normalization)_\d+/(beta)|"
+    # r"attention_layer_(\d+)/(layer_normalization)_\d+/(gamma)|"
 )
```
njzjz, line 193:
```suggestion
)[:-1]
```

I am surprised that an empty pattern doesn't break anything.

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -0,0 +1,911 @@
...
+        # nfnl x tebd_dim
+        atype_embd = atype_embd_ext[:, :nloc, :].reshape(nf * nloc, -1)
+        # nfnl x nnei x tebd_dim
+        atype_embd_nnei = np.tile(atype_embd[:, np.newaxis, :], (1, nnei, 1))
+        # nfnl x nnei
+        nlist_mask = nlist != -1
+        nlist_masked = np.copy(nlist)
+        # nfnl x nnei x 1
+        sw = np.where(nlist_mask[:, :, None], sw, 0.0)
+        nlist_masked[nlist_masked == -1] = 0
```
wanghan-iapcm, line 463:
```suggestion
        nlist_masked = np.where(nlist_mask, nlist, 0)
```

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -0,0 +1,911 @@
...
+        # add type embedding into input
+        # nf x nall x tebd_dim
+        atype_embd_ext = self.type_embedding.call()[atype_ext]
+        # nfnl x tebd_dim
+        atype_embd = atype_embd_ext[:, :nloc, :].reshape(nf * nloc, -1)
+        # nfnl x nnei x tebd_dim
+        atype_embd_nnei = np.tile(atype_embd[:, np.newaxis, :], (1, nnei, 1))
+        # nfnl x nnei
+        nlist_mask = nlist != -1
+        nlist_masked = np.copy(nlist)
```
wanghan-iapcm, line 460:
```suggestion
```

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -0,0 +1,911 @@
...
+        rr = rr * exclude_mask[:, :, None]
+        # nfnl x nnei x 1
+        ss = rr[..., 0:1]
+        if self.tebd_input_mode in ["concat"]:
+            if not self.type_one_side:
+                # nfnl x nnei x (1 + 2 * tebd_dim)
+                ss = np.concatenate([ss, atype_embd_nlist, atype_embd_nnei], axis=-1)
+            else:
+                # nfnl x nnei x (1 + tebd_dim)
+                ss = np.concatenate([ss, atype_embd_nlist], axis=-1)
```
wanghan-iapcm, line 484: it seems that you are supporting type_one_size

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -0,0 +1,911 @@
...
+        # Linear projection
+        q, k, v = np.split(self.in_proj(query), 3, axis=-1)
+        # Reshape and normalize
+        q = q.reshape(-1, self.nnei, self.hidden_dim)
+        k = k.reshape(-1, self.nnei, self.hidden_dim)
+        v = v.reshape(-1, self.nnei, self.hidden_dim)
+        if self.normalize:
+            q = np_normalize(q, axis=-1)
+            k = np_normalize(k, axis=-1)
+            v = np_normalize(v, axis=-1)
```
wanghan-iapcm, line 862: really need to normalize q, k and v?

deepmd/dpmodel/utils/network.py
```diff
@@ -381,6 +381,154 @@ def fn(x):
...
+
+    @staticmethod
+    def layer_norm_numpy(x, shape, weight, bias, eps):
+        # mean and variance
+        mean = np.mean(x, axis=tuple(range(-len(shape), 0)), keepdims=True)
+        var = np.var(x, axis=tuple(range(-len(shape), 0)), keepdims=True)
+        # normalize
+        x_normalized = (x - mean) / np.sqrt(var + eps)
+        # shift and scale
+        x_ln = x_normalized * weight + bias
```
wanghan-iapcm, line 528: weight and bias may be `None`

deepmd/pt/model/network/layernorm.py
```diff
@@ -0,0 +1,126 @@
...
+        yy: torch.Tensor
+            The output.
+        """
+        mean = xx.mean(dim=-1, keepdim=True)
+        variance = xx.var(dim=-1, unbiased=False, keepdim=True)
+        yy = (xx - mean) / torch.sqrt(variance + self.eps)
+        if self.matrix is not None:
+            yy = yy * self.matrix
+        if self.bias is not None:
+            yy = yy + self.bias
```
wanghan-iapcm, line 79: it seems that the `self.matrix` and `self.bias` are not updated at all. If it is the situation, then they should be removed.

wanghan-iapcm (commented):
stripped type embedding should be added in a future PR.

source/tests/pt/model/test_dpa1.py
```diff
@@ -0,0 +1,214 @@
...
+    def test_consistency(
+        self,
+    ):
+        rng = np.random.default_rng(100)
+        nf, nloc, nnei = self.nlist.shape
+        davg = rng.normal(size=(self.nt, nnei, 4))
+        dstd = rng.normal(size=(self.nt, nnei, 4))
+        dstd = 0.1 + np.abs(dstd)
+
+        for idt, prec, sm, to in itertools.product(
```
wanghan-iapcm, line 51: add comment explaining the meaning of the variables.

wanghan-iapcm: I find the test may not fully cover all the possibilities of the input parameters of dpa1. 
1. type_one_side
2. attn_layer 0 or > 0
3. attn_dotr
4. attn_mask (do you really use it??) 
5. exclude types
6. set_davg_zero
7. scaling_factor
8. normalize
9. temperature (really used?)
10. smooth_type_embedding
11. concat_output_tebd

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -0,0 +1,911 @@
...
+    4. tebd_input_mode != 'concat'
+
+    References
+    ----------
+    .. [1] Duo Zhang, Hangrui Bi, Fu-Zhi Dai, Wanrun Jiang, Linfeng Zhang, and Han Wang. 2022.
+       DPA-1: Pretraining of Attention-based Deep Potential Model for Molecular Simulation.
+       arXiv preprint arXiv:2208.08236.
+    """
+
+    def __init__(
```
wanghan-iapcm, line 197: you should carefully check and explain which parameters are really in use  and which parameters are kept for compatibility.

source/tests/pt/model/test_dpa1.py
```diff
@@ -0,0 +1,214 @@
...
+            )
+            np.testing.assert_allclose(
+                rd0.detach().cpu().numpy(),
+                rd2,
+                rtol=rtol,
+                atol=atol,
+                err_msg=err_msg,
+            )
+            # dp impl serialization
+            dd3 = DPDescrptDPA1.deserialize(dd2.serialize())
```
wanghan-iapcm, line 110: the self consistency check of DPDescrptDPA1 should be place in the dpmodel folder.

source/tests/pt/model/test_dpa1.py
iProzd: 1，2，5，6，10，11 are used to be tested in source/tests/consistent/descriptor/test_dpa1.py

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -0,0 +1,911 @@
...
+            ## to be updated when the options are supported.
+            "trainable": True,
+            "spin": None,
+        }
+
+    @classmethod
+    def deserialize(cls, data: dict) -> "DescrptDPA1":
+        """Deserialize from dict."""
+        data = data.copy()
+        check_version_compatibility(data.pop("@version", 1), 1, 1)
```
njzjz, line 562: We don't need a default value, as it's a new class, and can assume that the version is always stored.

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -0,0 +1,911 @@
...
+        data.pop("@class", None)
+        data.pop("type", None)
```
njzjz, line 564: We don't need a default value.

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -0,0 +1,911 @@
...
+    concat_output_tebd: bool
+            Whether to concat type embedding at the output of the descriptor.
+    spin
+            The old implementation of deepspin (deprecated in the descriptor).
+
+    Limitations
+    -----------
+    The currently implementation does not support the following features
+
+    1. type_one_side == True
```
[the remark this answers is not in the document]
iProzd: Already supported and docstr updated.

deepmd/dpmodel/utils/network.py
iProzd: I've added `trainable` options in https://github.com/deepmodeling/deepmd-kit/pull/3696/commits/85c7d6e1bc11665038a73f06a72bd64bb9988a15 for all layernorm instead of making them None. It's more convenient to be consistent with TF using `tf.get_variable`.

source/tests/pt/model/test_dpa1.py
iProzd: All params above, except `attn_mask`, are added to consistent tests.

deepmd/dpmodel/descriptor/dpa1.py
iProzd: Docs updated, there are now two kinds of **special** parameters:
1. Implemented in this class but only specific options can be consistent with other backends.(e.g. `attn_mask`(only False), `smooth_type_embedding` (only False) of `DescrptDPA1Compat`  in TF to be consistent.) They are marked in docstr like: 
```
    attn_mask: bool
            (Only support False to keep consistent with other backend references.)
            If mask the diagonal of attention weights
```
2. Not implemented in the class but are only kept for compatibility. (e.g. `scaling_factor`, `normalize`, `temperature` of `DescrptDPA1Compat`  in TF and `attn_mask` of `DescrptDPA1` in PT/DP.) They are marked in docstr like:
```
    normalize: bool
            (Only support True to keep consistent with other backend references.)
            (Not used in this version.)
            Whether to normalize the hidden vectors in attention weights calculation.
```

deepmd/dpmodel/descriptor/dpa1.py
iProzd: When developing dpa1, it was found that without normalization here, the attention weights are often dominated by `q` and `k` values from the largest `g_ij` (typically the nearest neighbor). This can result in the attention mechanism failing to capture relevant dependencies among other neighbors. 

I also tried adding normalization in various places, and found that applying it here helps stabilize the attention weights and training process.

This phenomenon is likely due to the fact that our `g_ij` uses distance scale information as input, which differs from the typical attention mechanisms in NLP. Of course, there are now more changes in other parts of the system, so if have time, we may explore other potential solutions as well.

deepmd/pt/model/descriptor/se_atten.py
```diff
@@ -347,13 +452,396 @@ def forward(
...
+        nei_mask,
+        input_r: Optional[torch.Tensor] = None,
+        sw: Optional[torch.Tensor] = None,
+        attnw_shift: float = 20.0,
+    ):
+        """
+        Args:
+            query: input G, [nframes * nloc, nnei, embed_dim].
+            nei_mask: neighbor mask, [nframes * nloc, nnei].
+            input_r: normalized radial, [nframes, nloc, nei, 3].
```
njzjz, line 756: Could you write doc in numpydoc format?

source/tests/tf/test_data_large_batch.py
```diff
@@ -201,37 +201,37 @@ def test_data_mixed_type(self):
         np.savetxt("f.out", f.reshape([1, -1]), delimiter=",")
         np.savetxt("v.out", v.reshape([1, -1]), delimiter=",")
 
-        refe = [6.121172052273665543e01]
+        refe = [6.12116933882038480874e01]
```
njzjz, line 204: Is it a breaking change or just a random behavior change? When restarting from an old model, will anything be broken?

deepmd/dpmodel/descriptor/dpa1.py
wanghan-iapcm: Interesting, we may reconsider the attentions in dpa2, which did not use this normalization.

source/tests/tf/test_data_large_batch.py
iProzd: This is a breaking change because the **layernorm** in TF (keras) is not consistent with those in PT/DP ( has numerical difference). It may be little confusing to keep two inplementations of Layernorm, should we keep old layernorm for TF? @njzjz @wanghan-iapcm

deepmd/pt/model/descriptor/se_atten.py
```diff
@@ -141,19 +223,41 @@ def __init__(
...
-        filter_layers.append(one)
-        self.filter_layers = torch.nn.ModuleList(filter_layers)
+        if self.tebd_input_mode in ["concat"]:
+            if not self.type_one_side:
+                self.embd_input_dim = 1 + self.tebd_dim * 2
+            else:
+                self.embd_input_dim = 1 + self.tebd_dim
+        else:
+            self.embd_input_dim = 1
+
```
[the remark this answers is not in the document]
iProzd: Yes, please open a GitHub issue to track this.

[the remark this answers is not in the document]
iProzd: > The handling of `tebd_input_mode` with conditional logic suggests that other modes might be planned for future implementation. Consider adding a TODO or FIXME to track this enhancement.
> 
> Would you like me to open a GitHub issue to track the implementation of additional `tebd_input_mode` options?

Yes, please open a GitHub issue to track this. @coderabbitai

source/tests/tf/test_data_large_batch.py
iProzd: It seems like a way to keep compatibility of TF while being consistent with DP/PT:
1. Add option for eps value in PT/DP, and they are consistent with TF when eps = 1e-3 (need test).
2. Remap the variable names to be compat with old models using keras.

source/tests/pt/model/test_dpa1.py
```diff
@@ -92,37 +82,9 @@ def test_consistency(
...
-                self.nlist,
-            )
-            np.testing.assert_allclose(
-                rd0.detach().cpu().numpy(),
-                rd2,
-                rtol=rtol,
-                atol=atol,
-                err_msg=err_msg,
-            )
-            # dp impl serialization
```
wanghan-iapcm: why do you remove the consistency check?

source/tests/pt/model/test_dpa1.py
```diff
@@ -92,37 +82,9 @@ def test_consistency(
...
-                self.atype_ext,
-                self.nlist,
-            )
-            np.testing.assert_allclose(
-                rd0.detach().cpu().numpy(),
-                rd3,
-                rtol=rtol,
-                atol=atol,
-                err_msg=err_msg,
-            )
```
wanghan-iapcm: this test should be place in the common/dpmodel subfolder of the test, not removed!

source/tests/pt/model/test_dpa1.py
wanghan-iapcm: your change does NOT fix this comment. see the new comments.

deepmd/dpmodel/descriptor/dpa1.py
wanghan-iapcm: I find the attn_mask == True is not implemented at all, this should be explicitly pointed out in the doc.

iProzd: It was pointed out in the `Limitations` at the end of the doc and will also raise `NotImplementedError` at the front lines of `__init__`. I also added more explicit doc in [add docs for not-used param](https://github.com/deepmodeling/deepmd-kit/pull/3696/commits/9662a2c97528448f6065deaddcc50ac87ffa91cc).

source/tests/consistent/descriptor/test_dpa1.py
```diff
@@ -0,0 +1,333 @@
...
+            excluded_types,
+            env_protection,
+            set_davg_zero,
+            scaling_factor,
+            normalize,
+            temperature,
+            ln_eps,
+            smooth_type_embedding,
+            concat_output_tebd,
+            precision,
```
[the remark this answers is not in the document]
njzjz: This method is not needed at all.

njzjz (approved):
LGTM

wanghan-iapcm (approved):

PR #3698, opened 2024-04-21
Fix typo in `smooth_type_embdding`

iProzd:


deepmd/utils/argcheck.py
```diff
@@ -503,11 +503,11 @@ def descrpt_se_atten_args():
             doc=doc_only_tf_supported + doc_stripped_type_embedding,
         ),
         Argument(
-            "smooth_type_embdding",
+            "smooth_type_embedding",
```
njzjz, line 506: We may keep the original one as an alias, so there's no breaking change.

wanghan-iapcm: +1

njzjz (approved):

wanghan-iapcm (approved):

PR #3712, opened 2024-04-25
feat: Support stripped type embedding in DPA1 of PT/DP

iProzd:
This PR supports stripped type embedding in DPA1 of PT/DP:

- Remove `stripped_type_embedding` params in all classes and use `tebd_input_mode` == "strip" instead.
- Add stripped type embedding inplementation for DPA1 of PT/DP.
- Add serialize and deserialize for stripped type embedding.

Note: 
- Old TF inplementation has not consistent behaivior when `type_one_side`==True and `tebd_input_mode` == "strip", it always uses two_side type stripped embeddings input, which is also inconsistent with `DescrptSeAEbdV2` in TF (but the training still works and only raise `NotImplementedError` when doing serialization now) may need support from @nahso . 
- Old TF inplementation `init_variables` will not init `idt` weights from graph for `two_side_embeeding_net_variables` (fixed), I'm surprised that no ut failed before (maybe all tests use `resnet_dt` == False).
- The TF implementation of `DescrptSeAtten` does not support serialization when `tebd_input_mode` == "strip". This limitation arises because the shape of `type_embedding` cannot be determined after init, as it is decided at runtime. While the consistent version `DescrptDPA1Compat` is compatible with this configuration.

## Summary by CodeRabbit

- **New Features**
	- Enhanced model flexibility with new type embedding input modes: `concat` and `strip`.
- **Bug Fixes**
	- Improved model compression logic alignment with new type embedding modes for more efficient operations.
- **Documentation**
	- Updated documentation to explain the impact of new type embedding input modes on model descriptors.
- **Tests**
	- Adjusted test cases to reflect changes in type embedding input modes for robust testing.


deepmd/utils/argcheck.py
```diff
@@ -2311,6 +2311,25 @@ def gen_args(**kwargs) -> List[Argument]:
...
 
 
+def backend_compat(data):
+    data = data.copy()
+    # stripped_type_embedding in old DescrptSeAtten
+    if (
+        "descriptor" in data["model"]
+        and data["model"]["descriptor"].get("type", "se_e2_a") == "se_atten"
+        and data["model"]["descriptor"].pop("stripped_type_embedding", False)
+    ):
```
njzjz, line 2321: It does not support the hybrid descriptor.

njzjz: Perhaps support and add a customized normalize hook to dargs

examples/water/se_atten_dpa1_compat/input.json
njzjz, line 1: Have you tested the DPA-1 model?

njzjz (changes requested):

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -290,25 +285,41 @@ def __init__(
             activation_function="Linear",
             precision=precision,
         )
+        if not self.type_one_side:
+            self.tebd_dim_input = self.tebd_dim * 2
+        else:
+            self.tebd_dim_input = self.tebd_dim
```
wanghan-iapcm, line 291:
```suggestion
        self.tebd_dim_input = self.tebd_dim if self.type_one_side else self.tebd_dim * 2
```

deepmd/utils/argcheck.py
iProzd: I tried to add `extra_check` in Argument (outside in the descriptor args level), it works but there are still problems:
1. `hybrid` descriptor still need to be individually considered.
2. This keeps `stripped_type_embedding` in `descrpt_se_atten_args`. In other places (such as tests in `source/tests/consistent/descriptor`) where using list of `descrpt_args_plugin`, the `extra_check` can not be reused.

njzjz: > `hybrid` descriptor still need to be individually considered.

It looks to me `hybrid` is defined here, where `extra_check` may be applied.
https://github.com/deepmodeling/deepmd-kit/blob/f3056ee5003b2d33a1394857ee380c0364f405f4/deepmd/utils/argcheck.py#L382-L395

I tried to move `extra_check` to the variant choice argument, i.e., Argument for `se_atten`, but it seems to me the dargs package never executes it. @y1xiaoc any idea about it?

> This keeps `stripped_type_embedding` in `descrpt_se_atten_args`. In other places (such as tests in `source/tests/consistent/descriptor`) where using list of `descrpt_args_plugin`, the `extra_check` can not be reused.

We may deprecate the usage of the `list[Argument]`.

wanghan-iapcm (approved):

deepmd/utils/argcheck.py
y1xiaoc: It's been so long I almost forgot everything lol. But you're right, the choice argument is just dummy argument as a collection of its sub arguments, no real check is performed at that level. In the checking process it will be flattened and checked at the parent level (at the argument that defines all the sub variants). So maybe you want to add an extra check there.

deepmd/utils/argcheck.py
```diff
@@ -495,16 +494,28 @@ def descrpt_se_atten_args():
...
+        "- 'concat': Concatenate the type embedding with the smoothed radial information as the union input for the embedding network. "
+        "When `type_one_side` is False, the input is `input_ij = concat([r_ij, tebd_j, tebd_i])`. When `type_one_side` is True, the input is `input_ij = concat([r_ij, tebd_j])`. "
+        "The output is `out_ij = embeding(input_ij)` for the pair-wise representation of atom i with neighbor j."
+        "- 'strip': Use a separated embedding network for the type embedding and combine the output with the radial embedding network output. "
+        f"When `type_one_side` is False, the input is `input_t = concat([tebd_j, tebd_i])`. {doc_only_pt_supported} When `type_one_side` is True, the input is `input_t = tebd_j`. "
+        "The output is `out_ij = embeding_t(input_t) * embeding_s(r_ij) + embeding_s(r_ij)` for the pair-wise representation of atom i with neighbor j."
+    )
+    doc_stripped_type_embedding = (
+        "(Deprecated, kept only for compatibility.) Whether to strip the type embedding into a separated embedding network. "
+        "Setting this to `True` is equivalent to setting `tebd_input_mode` to 'strip'."
```
njzjz, line 508: It seems that setting it to `False` also has an effect; only setting it to None has not effect.

njzjz (approved):

PR #3750, opened 2024-05-06
tf: improve finetune params to keep from user input

iProzd:
## Summary by CodeRabbit

- **Refactor**
	- Improved clarity in model parameter replacement functionality by removing outdated comments and unnecessary logic.

### DIFF
--- a/deepmd/tf/utils/finetune.py
+++ b/deepmd/tf/utils/finetune.py
@@ -99,7 +99,6 @@ def replace_model_params_with_pretrained_model(
         ):
             target_para = pretrained_jdata["model"][config_key]
             cur_para = jdata["model"][config_key]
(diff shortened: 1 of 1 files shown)


iProzd:
I've checked for #3551 , there are several params that need discussed whether to keep from new input script (now we only keep `trainable` params and replace others with those in the pre-trained model):
- `exclude_types` from `descriptor`
- `atom_ener` from `EnerFitting`

Do we need to keep these params from user-defined input script? @njzjz @wanghan-iapcm If not, there's nothing to change for this issue.

njzjz:
> Do we need to keep these params from user-defined fine-tuning input script?

Why not provide an option to users?

iProzd:
Use this issue to track this need: https://github.com/deepmodeling/deepmd-kit/issues/3869

PR #3767, opened 2024-05-10
test(pt): add common test case for model/atomic model

### LINKED ISSUE #3501: [Feature Request] test: test serialization & deserialization for models and atomic models

### LINKED ISSUE #3517: [Feature Request] test: test get_sel, get_type, etc for models and atomic models

### LINKED ISSUE #3518: [Feature Request] test: test JIT interface for all models

njzjz:
Fix #3501. Fix #3517. Fix #3518.

## Summary by CodeRabbit

- **Tests**
    - Expanded testing capabilities for atomic and energy models to improve accuracy and reliability in energy calculations.
    - Implemented new test cases for atomic and energy models, along with common model test cases, to validate diverse functionalities and calculations.
    - Introduced test case classes for atomic and energy models with methods to assess parameters, types, outputs, and forward computations.
    - Added utility functions for testing PyTorch-based deep learning models with a custom backend.


njzjz:
#3763 has fixed the tests, so I'll not resolve it in this PR.

iProzd (commented):
1. Do we need a similar test for DP models? or we use a consistent test.
2. The forward and forward_lower test is different for spin_models, see source/tests/pt/model/test_forward_lower.py

wanghan-iapcm (approved):

wanghan-iapcm:
I agree with @iProzd , a multi-backend test support is necessary.

wanghan-iapcm (commented):
I agree with @iProzd , a multi-backend test support is necessary.

source/tests/universal/common/cases/atomic_model/ener_model.py
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+
+from .utils import (
+    AtomicModelTestCase,
+)
+
+
+class EnerAtomicModelTest(AtomicModelTestCase):
```
wanghan-iapcm, line 9: How does one want to add a new test like one for permutation in-variance of the energy? It may not be a good idea to derive from an `AtomicModelTestCase` although the name somehow indicating the derivation.

njzjz: Why cannot one add to this class?

wanghan-iapcm: different test are better provided in different test cases. I do not want to test e.g. autodiff when I am testing translational invariance.

njzjz: It's unexpected that they will run together. Each test in a test case should run individually.

njzjz: Do you mean you want to select a certain test? `-k` can be used to do so.

wanghan-iapcm: by your definition of the base class , all the test will run

wanghan-iapcm (approved):

iProzd (approved):
Approve but two comments:
1. As @wanghan-iapcm said, if we want to add a test, such as permutation invariance, the standard way is to add a common test method in `AtomicModelTestCase` right?
2. If one model has different behavior to test a common test method, such as `forward_lower` test for SpinModel, it's expected to overwrite the method in SpinModel's test case. Is it right?

iProzd (commented):
1. As @wanghan-iapcm said, if we want to add a test, such as permutation invariance, the standard way is to add a common test method in `AtomicModelTestCase` right?
2. If one model has different behavior to test a common test method, such as `forward_lower` test for SpinModel, it's expected to overwrite the method in SpinModel's test case. Is it right?

njzjz:
> 1. As @wanghan-iapcm said, if we want to add a test, such as permutation invariance, the standard way is to add a common test method in `AtomicModelTestCase` right?
> 
>     2. If one model has different behavior to test a common test method, such as `forward_lower` test for SpinModel, it's expected to overwrite the method in SpinModel's test case. Is it right?

You are correct.

PR #3768, opened 2024-05-10
breaking: seperate params in dpa2

iProzd:
## Summary by CodeRabbit

- **New Features**
  - Enhanced descriptor configuration with the introduction of `RepinitArgs` and `RepformerArgs` classes for more flexible and structured argument handling.
  - Improved JSON configuration structure for descriptors, aligning properties under `repinit` and `repformer` into nested objects for clarity.

- **Refactor**
  - Updated the initialization process in the model to utilize the new argument classes, ensuring more robust setup.
  - Refactored argument checking functions to support the new class-based configuration.

- **Documentation**
  - Streamlined and clarified documentation related to descriptor arguments to better align with the new configuration approach.

- **Tests**
  - Extended testing suite to include new argument classes, enhancing test coverage and consistency checks.


deepmd/utils/argcheck.py
```diff
@@ -736,270 +636,367 @@ def descrpt_dpa2_args():
...
+        # descriptor args
+        Argument(
+            "concat_output_tebd",
+            bool,
+            optional=True,
+            default=True,
+            doc=doc_concat_output_tebd,
+        ),
+        Argument("precision", str, optional=True, default="default", doc=doc_precision),
+        Argument("smooth", bool, optional=True, default=False, doc=doc_smooth),
```
njzjz, line 651: Why is default False?

njzjz (approved):
Good. Just a question not related to this PR.

deepmd/utils/argcheck.py
iProzd: I've tracked about this parameter. I find it was transfered from the default value (False) of `smooth_type_embedding` param in DPA1 descriptor. To have a better performance, I suggest to change the default value here to `True`. What do you think? @wanghan-iapcm @njzjz

deepmd/dpmodel/descriptor/dpa2.py
```diff
@@ -516,42 +624,8 @@ def serialize(self) -> dict:
...
-            "repformer_attn2_hidden": repformers.attn2_hidden,
-            "repformer_attn2_nhead": repformers.attn2_nhead,
-            "repformer_attn2_has_gate": repformers.attn2_has_gate,
-            "repformer_activation_function": repformers.activation_function,
-            "repformer_update_style": repformers.update_style,
-            "repformer_set_davg_zero": repformers.set_davg_zero,
-            "repformer_trainable_ln": repformers.trainable_ln,
-            "repformer_ln_eps": repformers.ln_eps,
+            "repinit": self.repinit_args.serialize(),
+            "repformer": self.repformer_args.serialize(),
```
wanghan-iapcm, line 628: we may want to name them by repinit_args and repformer_args?

deepmd/utils/argcheck.py
wanghan-iapcm: default `True` is definitely a better option. 
what was the setting in dpa-2 manuscript? have you set a smooth_type_embedding?

iProzd: Yes, It was `True` in all settings we used. Because this parameter only has default `True` in class `__init__` method and was never added into argcheck.py before #3758.

deepmd/dpmodel/descriptor/dpa2.py
iProzd: I agree, https://github.com/deepmodeling/deepmd-kit/pull/3768/commits/7936f9d6357db2bc6aa4512ce9312952c386f83c changes their names in serialized dict. But in input.json, I think users prefer to use "repinit": {} and "repformer": {} for convenience, so I kept them.

wanghan-iapcm (approved):

PR #3773, opened 2024-05-12
feat: support seed for pt/dp models

iProzd:
## Summary by CodeRabbit

- **New Features**
  - Added various neural network weight initialization methods: zeros, ones, constants, normal distributions, truncated normal distributions, Kaiming distributions, and Xavier distributions.

- **Improvements**
  - Included optional `seed` parameter in initialization methods and classes.
  - Implemented a `get_generator` function for random seed management.

- **Bug Fixes**
  - Addressed potential unintended behavior by ensuring proper random seed setting during training processes.


njzjz (changes requested):

deepmd/pt/model/network/init.py
```diff
@@ -0,0 +1,451 @@
...
+import math
+import warnings
+from typing import Optional as _Optional
+
+import torch
+from torch import (
+    Tensor,
+)
+
+# Initialization methods copied from torch.nn.init from version v2.3.0
```
njzjz, line 11: PyTorch uses the BSD-3-Clause license, which contains

```
1. Redistributions of source code must retain the above copyright
   notice, this list of conditions and the following disclaimer.
```

we should obey the license requirement.

iProzd: Is https://github.com/deepmodeling/deepmd-kit/pull/3773/commits/bfdc38330f73281a8211ba0eba9919b48876fc15 ok?

iProzd: Noticed that 2.3.0 is released in 2024, changed in https://github.com/deepmodeling/deepmd-kit/pull/3773/commits/5f09c92b470f675a9bbbb3ada5e60c031540a0e2

deepmd/pt/model/descriptor/dpa1.py
```diff
@@ -225,12 +227,12 @@ def __init__(
         smooth_type_embedding: bool = True,
         type_one_side: bool = False,
         stripped_type_embedding: Optional[bool] = None,
+        seed: Optional[int] = None,
```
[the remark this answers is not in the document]
njzjz: @iProzd, does this PR only fix PT? It seems that DP still needs to be fixed.

njzjz (approved):
I approve this PR, but it looks to me that `seed` is not passed from descriptors/fittings to the network in dpmodel, which needs to be fixed in the future.

wanghan-iapcm:
> I approve this PR, but it looks to me that `seed` is not passed from descriptors/fittings to the network in dpmodel, which needs to be fixed in the future.

an issue is necessary before this PR can be merged.

PR #3803, opened 2024-05-22
feat(pt): consistent fine-tuning with init-model

### LINKED ISSUE #3455: [BUG] Finetuned model has wrong type_map

### LINKED ISSUE #3747: [Feature Request] Consistent user experience for finetuning and init-model
### Summary

Consistent user experience for finetuning and init-model

### Detailed Description

Currently, finetuning uses model structure defined in the pretrained model, ignoring those in `input.json`. In comparison, training with init-model uses model structure defined in `input.json`, inconsistency with the init model will cause exception. I suggest a more consistent user experience for the two modes.

_No response_

iProzd:
Fix #3747. Fix #3455. 

- Consistent fine-tuning with init-model, now in pt, fine-tuning include three steps: 
1. Change model params (for multitask fine-tuning, random fitting and type-related params), 
2. Init-model, 
3. Change bias

- By default, input will use user input while fine-tuning, instead of being overwritten by that in the pre-trained model. When adding “--use-pretrain-script”, user can use that in the pre-trained model. 

- Now `type_map` will use that in the user input instead of overwritten by that in the pre-trained model. 

Note:
1. After discussed with @wanghan-iapcm, **behavior of fine-tuning in TF is kept as before**. If needed in the future, it can be implemented then.
2. Fine-tuning using DOSModel in PT need to be fixed. (an issue will be opened, maybe fixed in another PR, cc @anyangml )

## Summary by CodeRabbit

- **New Features**
  - Added support for using model parameters from a pretrained model script.
  - Introduced new methods to handle type-related parameters and fine-tuning configurations.

- **Documentation**
  - Updated documentation to clarify the model section requirements and the new `--use-pretrain-script` option for fine-tuning.

- **Refactor**
  - Simplified and improved the readability of key functions related to model training and fine-tuning.

- **Tests**
  - Added new test methods and utility functions to ensure consistency of type mapping and parameter updates.


Chengqian-Zhang:
There is an obvious problem with this design, when I don't use `--use-pretrain-script`, if the model parameters in the `input.json` I provide is not the same as the model parameters of the pre-trained model, then there must be a failure to load the parameter during `self.wrapper.load_state_dict(state_dict)`. Because the network parameters of `self.wrapper` are initialized based on the model parameters of `input.json`, while `state_dict` is the network parameters of the pre-trained model.

For example, if `sel = 120` in pretrained model while `sel = 80` in `input.json`:
Singletask finetuning from a single model:
`dp --pt train finetune_single.json --finetune single.pt`
`RuntimeError: Error(s) in loading state_dict for ModelWrapper:
        size mismatch for model.Default.atomic_model.descriptor.repinit.mean: copying a param with shape torch.Size([3, 120, 4]) from checkpoint, the shape in current model is torch.Size([3, 80, 4]).`

Singletask finetuning from a multitask model:
`dp --pt train finetune_single.json --finetune multi.pt`
`RuntimeError: Error(s) in loading state_dict for ModelWrapper:
        size mismatch for model.Default.atomic_model.descriptor.repinit.mean: copying a param with shape torch.Size([3, 120, 4]) from checkpoint, the shape in current model is torch.Size([3, 80, 4]).`

deepmd/pt/model/task/ener.py
```diff
@@ -181,6 +182,31 @@ def serialize(self) -> dict:
...
+            The mapping index of newly defined types to those in the pretrained model.
+        prefix : str
+            The prefix of the param keys.
+
+        Returns
+        -------
+        updated_dict: Dict[str, torch.Tensor]
+            Updated type related params.
+        """
+        raise NotImplementedError
```
njzjz, line 208: btw, it's unclear to me why this is not implemented.

wanghan-iapcm:
I will review the PR after we discuss it online @iProzd @njzjz

njzjz (changes requested):

source/tests/pt/test_finetune.py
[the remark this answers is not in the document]
iProzd: The e2e uts are now in `source/tests/pt/test_finetune.py:test_finetune_slim_type`, maybe it's better to add a universal test for it.

deepmd/dpmodel/utils/type_embed.py
```diff
@@ -159,3 +161,14 @@ def serialize(self) -> dict:
             "type_map": self.type_map,
             "embedding": self.embedding_net.serialize(),
         }
+
+    def slim_type_map(self, type_map: List[str]) -> None:
+        """Change the type related params to slimmed ones, according to slimmed `type_map` and the original one in the model."""
+        assert len(self.neuron) == 1, "Only one layer type embedding can be slimmed!"
```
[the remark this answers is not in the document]
iProzd: I was trying to solve the problem below but now I notice that the resnet problem still exists even when layer==1.

deepmd/dpmodel/utils/type_embed.py
```diff
@@ -159,3 +161,14 @@ def serialize(self) -> dict:
...
+        slim_index = get_index_between_two_maps(self.type_map, type_map)
+        self.type_map = type_map
+        self.ntypes = len(type_map)
+        self.embedding_net.layers[0].num_in = len(type_map)
+        self.embedding_net.layers[0].matrix = self.embedding_net.layers[0].matrix[
+            slim_index
+        ]
```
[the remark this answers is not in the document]
iProzd: We may need to add special case for `resnet` condition in the first layer:
```
first_matrix = self.embedding_net.layers[0].matrix
if self.neuron[0] == self.ntypes:
    first_matrix += np.eye(self.ntypes)
elif self.neuron[0] == self.ntypes * 2:
    first_matrix += np.concat([np.eye(self.ntypes), np.eye(self.ntypes)], dim=-1)

first_matrix = first_matrix[slim_index]
new_ntypes = len(type_map)

if self.neuron[0] == new_ntypes:
    first_matrix -= np.eye(new_ntypes)
elif self.neuron[0] == new_ntypes * 2:
    first_matrix -= np.concat([np.eye(new_ntypes), np.eye(new_ntypes)], dim=-1)

self.embedding_net.layers[0].matrix = first_matrix
```

Or we need to add a constraint on resnet option on type_embed module, which may affect more.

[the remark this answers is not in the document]
iProzd: Do you have any idea? @njzjz @wanghan-iapcm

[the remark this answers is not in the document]
njzjz: I think the first situation is okay but needs test for all 3x3 situations.

deepmd/dpmodel/fitting/general_fitting.py
```diff
@@ -185,6 +193,22 @@ def get_sel_type(self) -> List[int]:
         """
         return [ii for ii in range(self.ntypes) if ii not in self.exclude_types]
 
+    def get_type_map(self) -> List[str]:
+        """Get the name to each type of atoms."""
+        return self.type_map
+
+    def slim_type_map(self, type_map: List[str]) -> None:
```
wanghan-iapcm, line 200: the UT for change type_map of fitting is necessary.

wanghan-iapcm (changes requested):

deepmd/main.py
```diff
@@ -255,6 +255,11 @@ def main_parser() -> argparse.ArgumentParser:
         default=None,
         help="Finetune the frozen pretrained model.",
     )
+    parser_train.add_argument(
+        "--use-pretrain-script",
+        action="store_true",
+        help="(Supported Backend: PyTorch) Use model params in the script of the pretrained model instead of user input when doing finetuning.",
```
njzjz, line 261: Could you explain what the current behavior is in TF?

source/tests/universal/common/cases/descriptor/utils.py
njzjz, line 1: It seems that you didn't test the type embedding itself?

deepmd/utils/finetune.py
wanghan-iapcm, line 1: UTs are needed for `get_index_between_two_maps`, `map_atom_exclude_types` and `map_pair_exclude_types`.

source/tests/universal/common/cases/descriptor/utils.py
```diff
@@ -93,3 +98,133 @@ def test_exclude_types(
...
+            "Ar",
+        ]  # 18 elements
+        for old_tm, new_tm, em, econf in itertools.product(
+            [
+                deepcopy(full_type_map_test),  # 18 elements
+                deepcopy(
+                    full_type_map_test[:16]
+                ),  # 16 elements, double of tebd default first dim
+                deepcopy(full_type_map_test[:8]),  # 8 elements, tebd default first dim
+                ["H", "O"],  # slimmed types
```
wanghan-iapcm, line 142: i do not think you need deep copies.

source/tests/universal/common/cases/descriptor/utils.py
```diff
@@ -93,3 +98,133 @@ def test_exclude_types(
...
+            ],  # new_type_map
+            [[], [[0, 1]], [[1, 1]]],  # exclude_types for original_type_map
+            [False, True],  # use_econf_tebd
+        ):
+            if len(old_tm) >= len(new_tm):
+                continue
+            # use shuffled type_map
+            shuffle(old_tm)
+            shuffle(new_tm)
```
wanghan-iapcm, line 159: operations are not deterministic.

deepmd/dpmodel/descriptor/descriptor.py
```diff
@@ -129,3 +129,29 @@ def call(
...
+    """
+    if des_with_stat is not None:
+        extend_davg = des_with_stat["davg"]
+        extend_dstd = des_with_stat["dstd"]
+    else:
+        extend_shape = [len(type_map), *list(des["davg"].shape[1:])]
+        extend_davg = np.zeros(extend_shape, dtype=des["davg"].dtype)
+        extend_dstd = np.ones(extend_shape, dtype=des["dstd"].dtype)
+    des["davg"] = np.concatenate([des["davg"], extend_davg], axis=0)
+    des["dstd"] = np.concatenate([des["dstd"], extend_dstd], axis=0)
```
wanghan-iapcm, line 157: you append the stats of the `des_with_stat` to the statistics of `self`. 
1. do you assume that the type map of `des_with_stat` is the new type map? 
2. is `type_map` argument assumed to be the new type map? 
3. please provide a better doc for the assumptions. 
4. please also not that the behavior should work with `get_index_between_two_maps`

source/tests/universal/common/cases/descriptor/utils.py
```diff
@@ -93,3 +98,133 @@ def test_exclude_types(
...
+                    full_type_map_test[:16]
+                ),  # 16 elements, double of tebd default first dim
+                deepcopy(full_type_map_test[:8]),  # 8 elements, tebd default first dim
+                ["H", "O"],  # slimmed types
+            ],  # new_type_map
+            [[], [[0, 1]], [[1, 1]]],  # exclude_types for original_type_map
+            [False, True],  # use_econf_tebd
+        ):
+            if len(old_tm) >= len(new_tm):
+                continue
```
wanghan-iapcm, line 156: why skipped? please provide the reason

source/tests/universal/common/cases/descriptor/utils.py
```diff
@@ -93,3 +98,133 @@ def test_exclude_types(
...
+            )
+            old_tm_module.change_type_map(new_tm)
+            new_tm_dd = self.forward_wrapper(old_tm_module)
+            rd_new_tm, _, _, _, sw_new_tm = new_tm_dd(
+                coord_ext_device,
+                new_tm_index[atype_ext_device],
+                nlist_device,
+                mapping=mapping_device,
+            )
+            np.testing.assert_allclose(rd_old_tm, rd_new_tm)
```
wanghan-iapcm, line 186: The ut does not consider the case if the type of the input atoms are belonging to the new type map but not belonging to the old type map. 
In this case we do not test the consistency between the old and the new descriptors, but check if the stats for the new types are setup correctly.

source/tests/universal/common/cases/fitting/utils.py
wanghan-iapcm, line 1: please check the comments for the descriptor tests, they apply in the fitting case.

source/tests/universal/common/cases/fitting/utils.py
```diff
@@ -0,0 +1,191 @@
...
+                            err_msg=f"compare {kk} output between 0 and {ii}",
+                        )
+
+    def test_exclude_types(
+        self,
+    ):
+        atype_device = self.atype_ext[:, : self.nloc]
+        serialize_dict = self.module.serialize()
+        # set random bias
+        rng = np.random.default_rng()
```
wanghan-iapcm, line 73: plz fix the seed.

deepmd/dpmodel/utils/type_embed.py
```diff
@@ -159,3 +143,76 @@ def serialize(self) -> dict:
             "type_map": self.type_map,
             "embedding": self.embedding_net.serialize(),
         }
+
+    def change_type_map(
```
wanghan-iapcm, line 147: a standalone UT for the method is necessary.

wanghan-iapcm (changes requested):

source/tests/universal/common/cases/descriptor/utils.py
iProzd: Deleted. It's a debug phrase when developing type_map extending and mistakenly kept.

deepmd/dpmodel/descriptor/descriptor.py
iProzd: 1.2. yes
3.4. docs updated

This behaivior is developed for extending new types unseen in old type_map. 

In this situation, some type related params in the model, such as statistics, need to be extended to have a length of `len(old_map) + len(new_map)` in the type related dimension. 

`get_index_between_two_maps` will generate negative indices `i - len(new_map)` if `new_map[i]` is not in the `old_map`.
Then positive indices from 0 to `len(old_map) - 1` will select old params of types in `old_map`, while negative indices from `-len(new_map)` to -1 will select new params of types in `new_map`.

deepmd/main.py
```diff
@@ -255,6 +255,11 @@ def main_parser() -> argparse.ArgumentParser:
         default=None,
         help="Finetune the frozen pretrained model.",
     )
+    parser_train.add_argument(
+        "--use-pretrain-script",
+        action="store_true",
+        help="Use model parameters from the script of the pretrained model instead of user input when doing finetuning. Note: This behavior is default and unchangeable in TensorFlow.",
```
njzjz, line 261: Do we plan to change it in TF? If so, we may have an issue to track it.

njzjz (approved):

wanghan-iapcm (approved):

deepmd/main.py
iProzd: I've opened an issue about this https://github.com/deepmodeling/deepmd-kit/issues/3869. I think we may change it in the future if there is a clear demand.

PR #3820, opened 2024-05-25
ci: fix test-python test_durations and its caches

njzjz:
The previous workflow file has several bugs.


wanghan-iapcm (approved):

wanghan-iapcm:
LGTM, sry i missed it for a few days.

Issue #3139, opened 2024-01-12
[Docs] document equations for `se_atten_v2`


njzjz:
### Summary

The current documentation has equations for `se_atten` but not for  `se_atten_v2`.

### Detailed Description

Add equations for `se_atten_v2` to documentation.

PR #3828, opened 2024-05-28
docs: add document equations for `se_atten_v2`

Chengqian-Zhang:
Solve issue #3139
`"se_atten_v2"` is inherited from `"se_atten"` with the following parameter modifications:

```json
      "tebd_input_mode": "strip",
      "smooth_type_embedding": true,
      "set_davg_zero": false
```

I add the equations for parameter `"tebd_input_mode"`.

## Summary by CodeRabbit

- **Documentation**
  - Detailed the default value and functionality of the `"tebd_input_mode"` parameter.
  - Highlighted the performance superiority of `"se_atten_v2"` over `"se_atten"`.
  - Specified a model compression requirement for `se_atten_v2`.

### DIFF
--- a/doc/model/train-se-atten.md
+++ b/doc/model/train-se-atten.md
@@ -174,7 +174,6 @@ When `"tebd_input_mode"` is set to `"strip"`, the embedding matrix $\mathcal{G}^
 ```math
    (\mathcal{G}^i)_j = \mathcal{N}_{e,2}(s(r_{ij})) + \mathcal{N}_{e,2}(s(r_{ij})) \odot ({N}_{e,2}(\{s(r_{ij}), \mathcal{A}^i, \mathcal{A}^j\}) \odot s(r_{ij})) \quad \mathrm{or}\quad (\mathcal{G}^i)_j = \mathcal{N}_{e,2}(s(r_{ij})) + \mathcal{N}_{e,2}(s(r_{ij})) \odot ({N}_{e,2}(\{s(r_{ij}), \mathcal{A}^j\}) \odot s(r_{ij}))
 ```
- 
 
 Practical evidence demonstrates that `"se_atten_v2"` offers better and more stable performance compared to `"se_atten"`.
 
--- a/doc/model/train-se-atten.md
+++ b/doc/model/train-se-atten.md
@@ -168,7 +168,11 @@ We highly recommend using the version 2.0 of the attention-based descriptor `"se
 The `tebd_input_mode` can take `"concat"` and `"strip"` as values. The default value in `"se_atten_v2"` is `"strip"`. When `tebd_input_mode` is set to `"strip"`, the embedding matrix $\mathcal{G}^i$ is constructed as:
 
 ```math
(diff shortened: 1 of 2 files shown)


iProzd (changes requested):
1. Using `tebd_input_mode` instead of `"tebd_input_mode"` for model params.
2. It's better to place `"concat"` formulation in the description of `se_atten` descriptor (e.g. "Note that we obtain G using the type embedding method by default in this descriptor. By default, we concat the type embeddings...") and keep doc for `tebd_input_mode` and `"strip"` formulation in `se_atten_v2`, by introducing `"strip"` as a default value for `tebd_input_mode` in `se_atten_v2` .

njzjz:
@iProzd I notice the documentation for se_atten is still old

iProzd (approved):

doc/model/train-se-atten.md
```diff
@@ -161,6 +165,16 @@ We highly recommend using the version 2.0 of the attention-based descriptor `"se
       "set_davg_zero": false
 ```
 
+The `tebd_input_mode` can take `"concat"` and `"strip"` as values. The default value in `"se_atten_v2"` is `"strip"`. When `tebd_input_mode` is set to `"strip"`, the embedding matrix $\mathcal{G}^i$ is constructed as:
```
njzjz, line 168: I don't think `se_atten_v2` has a `tebd_input_mode` argument and supports `concat`. FYI @iProzd

Chengqian-Zhang: I notice that the implementation of `:se_attn_v2"` is not the same in TensorFlow backend and PyTorch backend, I have modified the description, please review. @njzjz @iProzd 
image

doc/model/train-se-atten.md
```diff
@@ -161,6 +165,16 @@ We highly recommend using the version 2.0 of the attention-based descriptor `"se
       "set_davg_zero": false
 ```
 
+When using PyTorch backend, you must continue to use descriptor `"se_atten"` and specify `tebd_input_mode` as `"strip"` and `smooth_type_embedding` as `"true"`, which achieves the effect of `"se_atten_v2"`. The `tebd_input_mode` can take `"concat"` and `"strip"` as values. When using TensorFlow backend, you need to use descriptor `"se_atten_v2"` and do not need to set `tebd_input_mode` and `smooth_type_embedding` because the default value of `tebd_input_mode` is `"strip"`, and the default value of `smooth_type_embedding` is `"true"` in TensorFlow backend. When `tebd_input_mode` is set to `"strip"`, the embedding matrix $\mathcal{G}^i$ is constructed as:
```
wanghan-iapcm, line 168: I suggest we also provide the `"se_atten_v2"` implementation in pt/dp implementations.

wanghan-iapcm (approved):

doc/model/train-se-atten.md
njzjz: Tracked in #3831.

njzjz (approved):

Chengqian-Zhang:
I think it is better to update the documentation in #3840.

doc/model/train-se-atten.md
```diff
@@ -122,6 +126,16 @@ We highly recommend using the version 2.0 of the attention-based descriptor `"se
...
```
njzjz, line 129: When cherry-picking #3828 to `r2`, need to check if this sentence applies to `r2`. Need review from @iProzd

iProzd: In tensorflow, only type_one_side=false is supported when using se_atten_v2, see #3745 . So the formulation here is only the former one. Other sentences about tf is correct.

PR #3837, opened 2024-05-29
fix: bugs in uts for polar and dipole fit

iProzd:
Fix following trivial bugs in dipole and polar fit uts:
1. `box` was not used in `extend_input_and_build_neighbor_list` (which means they were all tested in nopbc mode, if shifted coord is outside the box (sometimes) and normalized explicitly, results are not the same.) Input for fitting also used extended_atype instead of atype. (Only same when nopbc.)
2. Using of `mixed_types` is disordered, mismatched with descriptor or sometimes with nlist. Now only use `mixed_types`==False since the descriptor output is not in mixed types.

## Summary by CodeRabbit

- **Tests**
  - Improved consistency in parameter handling for various test methods.
  - Updated `mixed_types` parameter to dynamically use `self.dd0.mixed_types()` across multiple test functions for better flexibility and accuracy.

### DIFF
--- a/source/tests/pt/model/test_dipole_fitting.py
+++ b/source/tests/pt/model/test_dipole_fitting.py
@@ -194,7 +194,7 @@ class TestEquivalence(unittest.TestCase):
                     _,
                     nlist,
                 ) = extend_input_and_build_neighbor_list(
-                    xyz, atype, self.rcut, self.sel, False, box=self.cell
+                    xyz + self.shift, atype, self.rcut, self.sel, False, box=self.cell
                 )
 
                 rd0, gr0, _, _, _ = self.dd0(
--- a/source/tests/pt/model/test_dipole_fitting.py
+++ b/source/tests/pt/model/test_dipole_fitting.py
@@ -83,7 +83,7 @@ class TestDipoleFitting(unittest.TestCase, TestCaseSingleFrameWithNlist):
                 embedding_width=self.dd0.get_dim_emb(),
                 numb_fparam=nfp,
                 numb_aparam=nap,
-                mixed_types=False,
+                mixed_types=self.dd0.mixed_types(),
             ).to(env.DEVICE)
             ft1 = DPDipoleFitting.deserialize(ft0.serialize())
             ft2 = DipoleFittingNet.deserialize(ft1.serialize())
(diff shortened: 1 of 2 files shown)


anyangml (approved):

source/tests/pt/model/test_dipole_fitting.py
```diff
@@ -84,7 +83,7 @@ def test_consistency(
                 embedding_width=self.dd0.get_dim_emb(),
                 numb_fparam=nfp,
                 numb_aparam=nap,
-                mixed_types=mixed_types,
+                mixed_types=False,
```
wanghan-iapcm, line 86:
```suggestion
                mixed_types=dd0.mixed_type(),
```

source/tests/pt/model/test_dipole_fitting.py
```diff
@@ -171,7 +169,7 @@ def test_rot(self):
...
```
wanghan-iapcm, line 172:
```suggestion
                mixed_types=self.dd0.mixed_type(),
```

source/tests/pt/model/test_dipole_fitting.py
```diff
@@ -196,7 +194,7 @@ def test_rot(self):
                     _,
                     nlist,
                 ) = extend_input_and_build_neighbor_list(
-                    xyz + self.shift, atype, self.rcut, self.sel, not mixed_types
+                    xyz + self.shift, atype, self.rcut, self.sel, False, box=self.cell
```
wanghan-iapcm, line 197:
```suggestion
                    xyz + self.shift, atype, self.rcut, self.sel, self.dd0.mixed_type(), box=self.cell
```

source/tests/pt/model/test_dipole_fitting.py
```diff
@@ -231,7 +229,7 @@ def test_permu(self):
                 _,
                 nlist,
             ) = extend_input_and_build_neighbor_list(
-                coord[idx_perm], atype, self.rcut, self.sel, True
+                coord[idx_perm], atype, self.rcut, self.sel, False, box=self.cell
```
wanghan-iapcm, line 232:
```suggestion
                coord[idx_perm], atype, self.rcut, self.sel, self.dd0.mixed_type(), box=self.cell
```

wanghan-iapcm (changes requested):
better to use the `mixed_type` attribute of the descriptor. 

i have provided a few examples, please revise all the mixed_types parameters accordingly.

njzjz (approved):

wanghan-iapcm (approved):

Issue #3831, opened 2024-05-28
[Feature Request] add `se_atten_v2` and redirect `dpa1` to `se_atten_v2`


njzjz:
### Summary

- add `se_atten_v2` to PyTorch and DP
- redirect `dpa1` to `se_atten_v2`

### Detailed Description

Per discussion.

iProzd:
Maybe no need to redirect `dpa1` to `se_atten_v2`?
`dpa1` is equivalent to `se_atten` from original definition, while `se_atten_v2` is a special case of `dpa1` to use `strip` structure.

PR #3840, opened 2024-05-30
Feat: add `se_atten_v2`to PyTorch and DP

Chengqian-Zhang:
Solve #3831 and #3139
- add `se_atten_v2` to PyTorch and DP
- add document equation for `se_attn_v2`

## Summary by CodeRabbit

- **New Features**
  - Introduced a new descriptor class with enhanced configuration options and methods for serialization and deserialization.
  - Added new configurable parameters to the descriptor setup for improved flexibility.

- **Documentation**
  - Updated function documentation to reflect new arguments and usage instructions.

- **Bug Fixes**
  - Refined serialization logic to handle new parameters and class types more accurately.
  - Improved error messages for better clarity during serialization processes.


Chengqian-Zhang:
@iProzd  thinks there is no need to redirect `dpa1` to `se_atten_v2`. `dpa1` is equivalent to `se_atten` from original definition, while `se_atten_v2` is a special case of `dpa1` to use strip structure. So I'm not redirecting it for now.

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -871,6 +871,154 @@ def call(
...
+            concat_output_tebd=concat_output_tebd,
+            spin=spin,
+            stripped_type_embedding=stripped_type_embedding,
+            use_econf_tebd=use_econf_tebd,
+            type_map=type_map,
+            # consistent with argcheck, not used though
+            seed=seed,
+        )
+
+    def serialize(self) -> dict:
```
njzjz, line 947: At this time, TF `DescrptSeAttenV2` doesn't have the `serialize` method.

We should add the same method.

njzjz:
Do you want to update the documentation in #3828 or in this PR??

Chengqian-Zhang:
> Do you want to update the documentation in #3828 or in this PR??

I think it is better to update the documentation in this PR.

wanghan-iapcm (commented):
could you please provide consistency UT like the se_atten descriptor?

doc/model/train-se-atten.md
```diff
@@ -126,7 +126,7 @@ We highly recommend using the version 2.0 of the attention-based descriptor `"se
...
 ```
 
-When using PyTorch backend, you must continue to use descriptor `"se_atten"` and specify `tebd_input_mode` as `"strip"` and `smooth_type_embedding` as `"true"`, which achieves the effect of `"se_atten_v2"`. The `tebd_input_mode` can take `"concat"` and `"strip"` as values. When using TensorFlow backend, you need to use descriptor `"se_atten_v2"` and do not need to set `tebd_input_mode` and `smooth_type_embedding` because the default value of `tebd_input_mode` is `"strip"`, and the default value of `smooth_type_embedding` is `"true"` in TensorFlow backend. When `tebd_input_mode` is set to `"strip"`, the embedding matrix $\mathcal{G}^i$ is constructed as:
+You can either continue to use descriptor `"se_atten"` and specify `tebd_input_mode` as `"strip"` and `smooth_type_embedding` as `"true"`, which achieves the effect of `"se_atten_v2"`, or use descriptor `"se_atten_v2"` and do not need to set `tebd_input_mode` and `smooth_type_embedding` because the default value of `tebd_input_mode` is `"strip"` and the default value of `smooth_type_embedding` is `"true"`. When `tebd_input_mode` is set to `"strip"`, the embedding matrix $\mathcal{G}^i$ is constructed as:
```
njzjz, line 129: > You can either continue to use descriptor `"se_atten"` and specify `tebd_input_mode` as `"strip"` and `smooth_type_embedding` as `"true"`, which achieves the effect of `"se_atten_v2"`,

This may not be required to tell users.

> the default value of `tebd_input_mode` is `"strip"` and the default value of `smooth_type_embedding` is `"true"`

It may not be "default" considering one can not set the parameter when using `se_atten_v2`.

njzjz:
@iProzd could you take a look at the unit test failure?

deepmd/tf/descriptor/se_atten.py
```diff
@@ -1954,8 +1950,8 @@ def serialize(self, suffix: str = "") -> dict:
         }
         if self.tebd_input_mode in ["strip"]:
             assert (
-                type(self) is DescrptDPA1Compat
-            ), "only DescrptDPA1Compat can serialize when tebd_input_mode=='strip'"
+                type(self) is not DescrptSeAtten
+            ), "only DescrptDPA1Compat ansd DescrptSeAttenV2 can serialize when tebd_input_mode=='strip'"
```
njzjz, line 1954:
```suggestion
            ), "only DescrptDPA1Compat and DescrptSeAttenV2 can serialize when tebd_input_mode=='strip'"
```

wanghan-iapcm (approved):

deepmd/utils/argcheck.py
```diff
@@ -631,14 +631,91 @@ def descrpt_se_atten_args():
...
         Argument(
-            "set_davg_zero", bool, optional=True, default=False, doc=doc_set_davg_zero
+            "stripped_type_embedding",
+            bool,
+            optional=True,
+            default=None,
+            doc=doc_stripped_type_embedding,
+        ),
+        Argument(
+            "set_davg_zero", bool, optional=True, default=True, doc=doc_set_davg_zero
```
iProzd, line 672:
```suggestion
            "set_davg_zero", bool, optional=True, default=False, doc=doc_set_davg_zero
```

iProzd: please notice this default value

iProzd (changes requested):

iProzd (approved):

deepmd/utils/argcheck.py
```diff
@@ -631,15 +631,92 @@ def descrpt_se_atten_args():
...
+        "Else, the scaling of attention weights is setting to `temperature`."
+    )
+    doc_normalize = (
+        "Whether to normalize the hidden vectors during attention calculation."
+    )
+    doc_concat_output_tebd = (
+        "Whether to concat type embedding at the output of the descriptor."
+    )
+    doc_stripped_type_embedding = (
+        "(Deprecated, kept only for compatibility.) Whether to strip the type embedding into a separate embedding network. "
```
njzjz, line 656: Why is a deprecated option added?

njzjz: Btw, I remember `tebd_input_mode` is forced to `strip`. Please correct me if I am wrong. @iProzd

Chengqian-Zhang: I have deleted the `doc_stripped_type_embedding`. In `"se_atten_v2"`, `tebd_input_mode` is forced to `strip`. So `tebd_input_mode` and `stripped_type_embedding`(and the latter is deprecated) are both useless here.

njzjz (approved):

PR #3841, opened 2024-05-31
fix: correct `exclude_types` in descriptors

iProzd:
1. make `exclude_types` consistent with mask in nlist for all descriptors. (bugs fixed in dpa1 and dpa2)
2. add universal tests for descriptor. (now only test_exclude_types)
3. `TestCaseSingleFrameWithNlist` in source/tests/pt/model/test_env_mat.py will be removed in a seperate PR.

## Summary by CodeRabbit

- **New Features**
  - Added new test case classes for simulation testing and common test cases on atomic models.

- **Bug Fixes**
  - Adjusted exclusion mask calculations and applications affecting neighbor lists and switch behavior in various descriptors.

- **Chores**
  - Updated GitHub Actions workflow to install CMake directly via pip, removing dependency on an external action.


source/tests/universal/common/cases/descriptor/utils.py
```diff
@@ -0,0 +1,95 @@
...
+    def setUp(self):
+        TestCaseSingleFrameWithNlist.setUp(self)
+        self.input_dict = {
+            "ntypes": self.nt,
+            "rcut": self.rcut,
+            "rcut_smth": self.rcut_smth,
+            "sel": self.sel,
+        }
+
+    def test_forward(self):
```
wanghan-iapcm, line 29: named by test_forward_consistency?

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -821,6 +823,8 @@ def call(
         atype_embd_nnei = np.tile(atype_embd[:, np.newaxis, :], (1, nnei, 1))
         # nfnl x nnei
         nlist_mask = nlist != -1
+        # nfnl x nnei
+        nlist_mask = nlist_mask * exclude_mask.astype(bool)
```
wanghan-iapcm, line 827:
```suggestion
        nlist_mask = np.where(exclude_mask, nlist_mask, 0)
```

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -810,6 +810,8 @@ def call(
         nf, nloc, nnei, _ = dmatrix.shape
         exclude_mask = self.emask.build_type_exclude_mask(nlist, atype_ext)
         # nfnl x nnei
+        exclude_mask = exclude_mask.reshape(nf * nloc, nnei)
+        # nfnl x nnei
         nlist = nlist.reshape(nf * nloc, nnei)
```
wanghan-iapcm, line 815: shall we exclude nlist here ? then we do not need to exclude the mask.

deepmd/pt/model/descriptor/se_atten.py
```diff
@@ -481,8 +481,6 @@ def forward(
         nlist_mask = nlist != -1
```
wanghan-iapcm, line 481: shall we exclude nlist before it is used ?

deepmd/tf/descriptor/se_atten.py
```diff
@@ -727,12 +739,9 @@ def _pass_filter(
                         tf.reshape(self.avg_looked_up, [-1, 1]), [1, self.ndescrpt]
                     ),
                 )
+                #  (nsamples, natoms, nnei)
```
njzjz, line 742: I suggest commenting with `nloc` or `nall` instead of `natoms`, considering `nloc != nall` in the simulation.

deepmd/tf/descriptor/se_atten.py
```diff
@@ -718,6 +718,18 @@ def _pass_filter(
...
+                tf.reshape(tf.cast(mask, self.filter_precision), [-1, 4]),
+                [0, 0],
+                [-1, 1],
+            )
+            #  (nsamples * natoms, 1,  nnei)
+            self.nmask *= tf.reshape(
+                nei_exclude_mask,
+                [-1, 1, self.sel_all_a[0]],
+            )
+            self.negative_mask = -(2 << 32) * (1.0 - self.nmask)
```
njzjz, line 732: To my understanding, `nmask` and `negative_mask` are only used when `smooth` is False, right?

iProzd: Yes you're correct. I've placed them in False condition.

wanghan-iapcm (approved):

PR #3851, opened 2024-06-04
feat: add has_message_passing API

### LINKED ISSUE #3713: Replace string search with API call in freeze function
In the  function of , the model type is currently determined by searching for a string in . This method is not robust and could lead to errors if the string format changes. It is suggested to replace this string search with a dedicated API call that can more reliably determine the model type.

Affected code:
```python
if '"type": "dpa2"' in model.model_def_script:
```

This issue was identified in PR https://github.com/deepmodeling/deepmd-kit/pull/3657 and discussed in the comments: https://github.com/deepmodeling/deepmd-kit/pull/3657#discussion_r1576909697

### LINKED ISSUE #3733: This should be fixed after implement api to decide whether need to
message passing and rename this metadata

Line: 58
https://github.com/deepmodeling/deepmd-kit/blob/d0fe13cbd8a305f08c23cef703c62d323313b2b8//source/api_cc/src/DeepPotPT.cc#L55-L65

njzjz:
Fix #3713. Fix #3733. This is a breaking change.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new method `has_message_passing` across various descriptor classes. This method returns a boolean indicating whether the descriptor has message passing capability.


CaRoLZhangxy (commented):
Maybe the return value of `has_message_passing` of dpa1 and dpa2 descriptor should be decided by the return value of the same method in se_atten and reformer descriptors. Then the definition of  `has_message_passing` of these descriptors should be added

wanghan-iapcm (approved):

CaRoLZhangxy (approved):

PR #3877, opened 2024-06-13
pt: add 4 tabulate_fusion op

cherryWangY:


source/op/pt/tabulate_multi_device.cc
```diff
@@ -0,0 +1,979 @@
...
+  // compute
+  if (device == "GPU") {
+#if GOOGLE_CUDA || TENSORFLOW_USE_ROCM
+    deepmd::tabulate_fusion_se_t_gpu(descriptor, table, table_info, em_x, em,
+                                     nloc, nnei_i, nnei_j, last_layer_size);
+#endif  // GOOGLE_CUDA || TENSORFLOW_USE_ROCM
+  } else if (device == "CPU") {
+    deepmd::tabulate_fusion_se_t_cpu(descriptor, table, table_info, em_x, em,
+                                     nloc, nnei_i, nnei_j, last_layer_size);
+  }
```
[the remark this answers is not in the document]
njzjz: Not related.

njzjz (changes requested):

source/tests/pt/test_tabulate_fusion_se_a.py
```diff
@@ -0,0 +1,1452 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+import unittest
+
+import torch
+
+
+class TestTabulateFusionSeAOp(unittest.TestCase):
+    def setUp(self):
+        self.table_tensor = torch.tensor(
```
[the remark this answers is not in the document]
njzjz: Btw, I also need to configure the CI environment for CUDA.

source/tests/pt/test_tabulate_fusion_se_a.py
```diff
@@ -0,0 +1,1461 @@
...
+        )
+
+        descriptor_tensor = forward_result[0]
+
+        # Check the shape
+        self.assertEqual(descriptor_tensor.shape, self.expected_descriptor_tensor.shape)
+
+        # Check the values
+        torch.testing.assert_close(
+            descriptor_tensor, self.expected_descriptor_tensor, atol=1e-5, rtol=1e-5
```
njzjz, line 1418: The current test is for `float32`. We may test **both** `float32` and `float64` and ensure `float64` has an accuracy of 1e-10.

njzjz: The dtype should be explicitly given, such as `dtype = torch.float64`.

njzjz: I have no idea. I expect @nahso could explain it.

cherryWangY: We think that the test samples in `tabulate_fusion_se_a.cc` cannot support the 1e-10 level test.  For example, the descriptor has a precision of only 7 decimal places, so the test only passes the test at the 1e-7 level and fails at 1e-8.
So I ran the results directly as the expected results and all tests passed now.

njzjz:
@coderabbitai resolve

njzjz (approved):

wanghan-iapcm (approved):

PR #3880, opened 2024-06-14
feat: add seeds to dpmodel and fix seeds in tf & pt

### LINKED ISSUE #3799: [BUG] Seeds of descriptor/fitting in dpmodel are not passed to network
### Bug summary

Seeds of descriptor/fitting in dpmodel are not passed to network

### DeePMD-kit Version

3.0.0a

### Backend and its version

 PyTorch v2.1.2

Built from source

See above

### Steps to Reproduce

See above

_No response_

njzjz:
Fix #3799.

## Summary by CodeRabbit

- **New Features**
  - Introduced flexibility in specifying seed values, allowing either an integer or a list of integers.
  - Enhanced seed parameter usage across various initialization methods and classes for more controlled randomization.

- **Improvements**
  - Updated seed initialization logic to include additional computations and dynamic adjustments.
  - Enhanced documentation for parameters in multiple classes, providing clearer usage guidelines.


deepmd/dpmodel/descriptor/dpa2.py
```diff
@@ -408,6 +408,7 @@ def init_subclass_params(sub_data, sub_class):
             resnet_dt=self.repinit_args.resnet_dt,
             smooth=smooth,
             type_one_side=self.repinit_args.type_one_side,
+            seed=seed,
```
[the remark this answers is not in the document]
njzjz: What's the difference between them?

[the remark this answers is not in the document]
njzjz: I think this is a wrong suggestion.

deepmd/dpmodel/fitting/general_fitting.py
```diff
@@ -166,6 +169,9 @@ def __init__(
                     self.resnet_dt,
                     self.precision,
                     bias_out=True,
+                    seed=seed + ii * (len(self.neuron) + 1)
```
wanghan-iapcm, line 172: shall we define a class method telling the number of random seeds used by one fitting net?

deepmd/dpmodel/descriptor/se_e2_a.py
```diff
@@ -189,15 +189,16 @@ def __init__(
...
+        for ii, embedding_idx in enumerate(
+            itertools.product(range(self.ntypes), repeat=self.embeddings.ndim)
         ):
             self.embeddings[embedding_idx] = EmbeddingNet(
                 in_dim,
                 self.neuron,
                 self.activation_function,
                 self.resnet_dt,
                 self.precision,
+                seed=seed + len(self.neuron) * ii if seed is not None else None,
```
wanghan-iapcm, line 201: shall we define a class method telling the number of random seeds used by one embedding net?

wanghan-iapcm (changes requested):
I find one has to spend much effort on maintaining the number of seeds used by each module. I would recommend the parallel random number generation feature by which we do not need to manually record the number of consumed seeds. see e.g. https://numpy.org/doc/stable/reference/random/parallel.html

njzjz:
> I find one has to spend much effort on maintaining the number of seeds used by each module. I would recommend the parallel random number generation feature by which we do not need to manually record the number of consumed seeds. see e.g. https://numpy.org/doc/stable/reference/random/parallel.html

It's unclear how to share arguments with PT when input is a `numpy.random.SeedSequence`, considering PT also uses several functions in dpmodel.

njzjz:
> > I find one has to spend much effort on maintaining the number of seeds used by each module. I would recommend the parallel random number generation feature by which we do not need to manually record the number of consumed seeds. see e.g. https://numpy.org/doc/stable/reference/random/parallel.html
> 
> It's unclear how to share arguments with PT when input is a `numpy.random.SeedSequence`, considering PT also uses several functions in dpmodel.

I just realized one can do this:

```py
>>> import numpy as np
>>> rng = np.random.default_rng([1,2,3,4,5])
```

So it will work if we change the type of seed from `int | None` to `int | list[int] | None`.

deepmd/dpmodel/fitting/general_fitting.py
njzjz: Now it accepts a `list[int]`, so this is not needed anymore.

deepmd/dpmodel/descriptor/se_e2_a.py
njzjz: Now it accepts a `list[int]`, so this is not needed anymore.

deepmd/pt/utils/utils.py
```diff
@@ -113,8 +115,13 @@ def dict_to_device(sample_dict):
...
 
 
-def get_generator(seed: Optional[int] = None) -> Optional[torch.Generator]:
+def get_generator(
+    seed: Optional[Union[int, List[int]]] = None,
+) -> Optional[torch.Generator]:
     if seed is not None:
+        if isinstance(seed, list):
+            # suggested by GitHub Copilot
+            seed = hash(tuple(seed))
```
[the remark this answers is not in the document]
njzjz: It seems you are right: https://stackoverflow.com/questions/16452252/does-pythons-hash-function-remain-identical-across-different-versions

I though we could use hashlib for consistent hashing.

wanghan-iapcm (commented):
Could you plz add a UT for mixed entropy to make sure the implementation is consistent with numpy.

njzjz:
> Could you plz add a UT for mixed entropy to make sure the implementation is consistent with numpy.

No, it's not consistent, as PyTorch limits the seed to (-0xffff_ffff_ffff_ffff, 0xffff_ffff_ffff_ffff). It is only ensured that the seed generated by `mix_entropy` has high entropy by using a similar way to NumPy.

iProzd (approved):

wanghan-iapcm (approved):

PR #3919, opened 2024-06-27
pt: fix compile error when use intel mpi

CaRoLZhangxy:
`MPIX_Query_cuda_support()` is not defined in intelMPI

## Summary by CodeRabbit

- **New Features**
  - Introduced conditional handling for CUDA awareness based on the MPI version to improve compatibility and performance for CUDA-enabled environments.


njzjz:
I can find `MPIX_Query_cuda_support` in `mpi.h`, so I don't know what you mean.

```sh
grep MPIX_Query_cuda_support ~/intel/oneapi/mpi/2021.13/include/mpi.h
int MPIX_Query_cuda_support(void) MPICH_API_PUBLIC;
int PMPIX_Query_cuda_support(void) MPICH_API_PUBLIC;
```

CaRoLZhangxy:
> ```shell
> grep MPIX_Query_cuda_support ~/intel/oneapi/mpi/2021.13/include/mpi.h
> ```

My intel MPI version is 2021.5.1 and there is no definition for `MPIX_Query_cuda_support`. What's more, it seems that there is no official api document for intel MPI.

njzjz:
> My intel MPI version is 2021.5.1 and there is no definition for `MPIX_Query_cuda_support`.

Perhaps use CMake's [`CheckFunctionExists`](https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html). It seems very hard to check whether this function exists according to other flags.

source/op/pt/CMakeLists.txt
```diff
@@ -20,6 +20,11 @@ endif()
 
 find_package(MPI)
 if(MPI_FOUND)
+  include(CheckCXXSymbolExists)
+  check_cxx_symbol_exists(MPIX_Query_cuda_support "mpi.h" CUDA_AWARE)
```
njzjz, line 24: I remember openmpi defines `MPIX_Query_cuda_support` in `mpi-ext.h`.

njzjz, line 24: I am wondering whether the MPI directory should be passed (does cmake know it). Do you have a test?

source/op/pt/comm.cc
```diff
@@ -100,7 +100,11 @@ class Border : public torch::autograd::Function<Border> {
       int version, subversion;
       MPI_Get_version(&version, &subversion);
       if (version >= 4) {
+#ifdef NO_CUDA_AWARE
+        cuda_aware = 0;
+#elif
```
njzjz, line 105: Isn't it `#else`?

njzjz (changes requested):

source/op/pt/CMakeLists.txt
```diff
@@ -20,6 +20,16 @@ endif()
 
 find_package(MPI)
 if(MPI_FOUND)
+  include(CheckCXXSymbolExists)
+  set(CMAKE_REQUIRED_INCLUDES ${MPI_INCLUDE_PATH})
+  set(CMAKE_REQUIRED_LIBRARIES ${MPI_LIBRARIES})
```
njzjz, line 25: Both `${MPI_INCLUDE_PATH}` and `${MPI_LIBRARIES}` are deprecated: https://cmake.org/cmake/help/latest/module/FindMPI.html#backward-compatibility

njzjz (approved):

PR #3925, opened 2024-06-28
feat(pt): support multitask argcheck

iProzd:
Note that:
1. docs for multitask args are not supported, may need help.
2. `trim_pattern="_*"` is not supported for repeat dict Argument, may need to update dargs.

## Summary by CodeRabbit

- **New Features**
  - Enhanced training configuration to support multi-task mode with additional arguments for data configuration.
  - Updated example configurations to reflect multi-task mode changes.

- **Bug Fixes**
  - Improved logic for updating and normalizing configuration during training regardless of multi-task mode.

- **Dependencies**
  - Upgraded `dargs` package requirement to version `>= 0.4.7`.

- **Tests**
  - Added new test cases for multi-task scenarios in `TestExamples` class.


wanghan-iapcm (approved):

njzjz (changes requested):
Need to bump dargs to >=0.4.7

njzjz:
> `trim_pattern="_*"` is not supported for repeat dict Argument, may need to update dargs.

I checked the dargs API and it is not easy to do so.

iProzd:
> > `trim_pattern="_*"` is not supported for repeat dict Argument, may need to update dargs.
> 
> I checked the dargs API and it is not easy to do so. It [uses](https://github.com/njzjz/dargs/blob/d93408596907c96e2909054498db07f6a078ad6c/dargs/dargs.py#L546-L554) a `sub_hook` API but this API is also used by [`check_strict`](https://github.com/njzjz/dargs/blob/d93408596907c96e2909054498db07f6a078ad6c/dargs/dargs.py#L430-L431), which we don't want to call.
> 
> Maybe raise an issue in dargs...

Tracked in https://github.com/deepmodeling/dargs/issues/70

njzjz (approved):

PR #3926, opened 2024-06-28
feat(pt/tf): init-(frz)-model use pretrain script

iProzd:
Support `--use-pretrain-script` for pt&tf when doing init-(frz)-model.

### DIFF
--- a/deepmd/main.py
+++ b/deepmd/main.py
@@ -258,7 +258,9 @@ def main_parser() -> argparse.ArgumentParser:
     parser_train.add_argument(
         "--use-pretrain-script",
         action="store_true",
-        help="Use model parameters from the script of the pretrained model instead of user input when doing finetuning. Note: This behavior is default and unchangeable in TensorFlow.",
+        help="When performing fine-tuning or init-model, "
+        "utilize the model parameters provided by the script of the pretrained model rather than relying on user input. "
+        "It is important to note that in TensorFlow, this behavior is the default and cannot be modified for fine-tuning. ",
     )
     parser_train.add_argument(
         "-o",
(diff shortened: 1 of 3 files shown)


wanghan-iapcm (commented):
shall we design a test case for this feature?

njzjz (approved):

njzjz (approved):

source/tests/pt/test_init_model.py
```diff
@@ -0,0 +1,132 @@
...
+                empty_config["model"]["descriptor"] = {}
+                empty_config["model"]["fitting_net"] = {}
+                empty_config["training"]["numb_steps"] = 0
+                empty_config["training"]["save_ckpt"] = ckpt_model
+                tmp_input = tempfile.NamedTemporaryFile(delete=False, suffix=".json")
+                with open(tmp_input.name, "w") as f:
+                    json.dump(empty_config, f, indent=4)
+                os.system(
+                    f"dp --pt train {tmp_input.name} --init-model {self.models[-1]} --use-pretrain-script --skip-neighbor-stat"
+                )
```
njzjz, line 63: I don't think `os.system` throws an error when the command fails.

You can try:
```py
>>> import os;os.system("svsdvsvsdv");print("still touch")
sh: line 1: svsdvsvsdv: command not found
32512
still touch
```

So you may use 
https://github.com/deepmodeling/deepmd-kit/blob/60927373ded5d61c4c9181e42f8f05cd0314cb04/source/tests/pt/common.py#L7-L29

njzjz (changes requested):

wanghan-iapcm (approved):

njzjz (approved):

PR #3933, opened 2024-06-30
feat(pt/tf): add bias changing param/interface

iProzd:
Add bias changing param/interface

For pt/tf, add `training/change_bias_after_training` to change out bias once after training.

For pt, add a separate command `change-bias` to change trained model(pt/pth, multi/single) out bias for specific data:

```
dp change-bias model.pt -s data -n 10 -m change
```

UTs for this feature are still in consideration.

## Summary by CodeRabbit

- **New Features**
  - Added a new subcommand `change-bias` to adjust model output bias in the PyTorch backend.
  - Introduced test cases for changing model biases via new test suite.
  
- **Documentation**
  - Added documentation for the new `change-bias` command, including usage and options.
  - Updated `index.rst` to include a new entry for `change-bias` under the `Model` section.

- **Bug Fixes**
  - Adjusted data handling in `make_stat_input` to limit processing to a specified number of batches.

- **Refactor**
  - Restructured training configuration to include the parameter `change_bias_after_training`.
  - Modularized data requirement handling and bias adjustment functions.

### DIFF
--- a/deepmd/main.py
+++ b/deepmd/main.py
@@ -659,6 +659,61 @@ def main_parser() -> argparse.ArgumentParser:
         help="treat all types as a single type. Used with se_atten descriptor.",
     )
 
+    # change_bias
+    parser_change_bias = subparsers.add_parser(
+        "change-bias",
+        parents=[parser_log],
+        help="(Supported backend: PyTorch) Change model out bias according to the input data.",
+        formatter_class=RawTextArgumentDefaultsHelpFormatter,
+        epilog=textwrap.dedent(
+            """\
+        examples:
+            dp change-bias model.pt -s data -n 10 -m change
+        """
+        ),
+    )
+    parser_change_bias.add_argument(
+        "INPUT", help="The input checkpoint file or frozen model file"
+    )
+    parser_change_bias.add_argument(
+        "-s",
+        "--system",
+        default=".",
+        type=str,
+        help="The system dir. Recursively detect systems in this directory",
+    )
+    parser_change_bias.add_argument(
+        "-n",
+        "--numb-batch",
+        default=0,
+        type=int,
+        help="The number of frames for bias changing in one data system. 0 means all data.",
+    )
+    parser_change_bias.add_argument(
+        "-m",
+        "--mode",
+        type=str,
+        default="change",
+        choices=["change", "set"],
+        help="The mode for changing energy bias: \n"
+        "change (default) : perform predictions using input model on target dataset, "
+        "and do least square on the errors to obtain the target shift as bias.\n"
+        "set : directly use the statistic bias in the target dataset.",
+    )
+    parser_change_bias.add_argument(
+        "-o",
+        "--output",
+        default=None,
+        type=str,
+        help="The model after changing bias.",
+    )
+    parser_change_bias.add_argument(
+        "--model-branch",
+        type=str,
+        default=None,
+        help="Model branch chosen for changing bias if multi-task model.",
+    )
+
     # --version
     parser.add_argument(
         "--version", action="version", version=f"DeePMD-kit v{__version__}"
@@ -831,6 +886,7 @@ def main():
         "convert-from",
         "train-nvnmd",
         "show",
+        "change-bias",
     ):
         deepmd_main = BACKENDS[args.backend]().entry_point_hook
     elif args.command is None:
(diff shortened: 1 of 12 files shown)


deepmd/pt/entrypoints/main.py
```diff
@@ -377,6 +387,111 @@ def show(FLAGS):
             log.info(f"The fitting_net parameter is {fitting_net}")
 
 
+def change_bias(FLAGS):
+    if FLAGS.INPUT.split(".")[-1] == "pt":
```
njzjz, line 391: This is unsafe is no dot in the filename. Could you use `endswith`?

deepmd/pt/entrypoints/main.py
```diff
@@ -377,6 +387,111 @@ def show(FLAGS):
             log.info(f"The fitting_net parameter is {fitting_net}")
 
 
+def change_bias(FLAGS):
```
njzjz, line 390: I suggest not passing `FLAGS` but expanding the dict.

njzjz: Other functions in this file have similar issues but can be handled in a separated PR.

wanghan-iapcm (changes requested):
1. please add unit test. 
2. along with stat from data, shall we allow the user explicitly setting the bias number?

deepmd/pt/entrypoints/main.py
iProzd: Tracked in https://github.com/deepmodeling/deepmd-kit/issues/3934.

iProzd:
@coderabbitai do you have any idea how to add a ut for command `dp --pt change-bias` proposed in this PR?

deepmd/main.py
```diff
@@ -51,6 +51,18 @@ def get_ll(log_level: str) -> int:
...
+def float_list(value_str):
+    """Convert string to a list of float values."""
+    values = value_str.split(",")
+    try:
+        return [float(v) for v in values]
+    except ValueError:
+        raise argparse.ArgumentTypeError(
+            f"'{value_str}' cannot be converted to a list of floats, "
+            f"please use ',' to separate each float value."
+        )
```
njzjz, line 63:
```suggestion
    except ValueError as e:
        raise argparse.ArgumentTypeError(
            f"'{value_str}' cannot be converted to a list of floats, "
            f"please use ',' to separate each float value."
        ) from e
```

deepmd/main.py
```diff
@@ -659,6 +671,70 @@ def main_parser() -> argparse.ArgumentParser:
...
+        "--system",
+        default=".",
+        type=str,
+        help="The system dir. Recursively detect systems in this directory",
+    )
+    parser_change_bias_source.add_argument(
+        "-b",
+        "--bias-value",
+        default=None,
+        type=float_list,
```
njzjz, line 702: I am wondering why not using `type=float, nargs='+'`. An example should be given, anyway.

deepmd/main.py
njzjz: We should turn on [`B904`](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/)

wanghan-iapcm (approved):

doc/model/change-bias.md
```diff
@@ -0,0 +1,42 @@
...
+
+:::
+
+:::{tab-item} Changing bias using user input for **energy model**:
+
+```sh
+dp --pt change-bias model.pt -b -92.523,-187.66 -o model_updated.pt
+```
+
+Here, `-b` specifies user-defined energy bias for each type, separated by `,`,
```
njzjz, line 37: separated by space?

doc/model/index.rst
njzjz, line 1: Should we put in the model section or the training section?

iProzd: It's more about changing an attribute of model without training. Maybe model section is better?

njzjz: At least, I don't understand why it is ahead of `train-energy`.

iProzd: Oh I got it, I've put `change-bias` the same place with `sel`. It seems that `sel` indeed follows the descriptor sections while `change-bias` should follow all the model sections. Updated.

njzjz (approved):

PR #3945, opened 2024-07-03
feat(pt): add datafile option for change-bias

iProzd:
## Summary by CodeRabbit

- **New Features**
  - Added optional `--datafile` argument to specify a file for system data processing.

- **Bug Fixes**
  - Improved `help` messages for `--datafile` argument to clarify its usage.

- **Tests**
  - Enhanced test coverage for changing bias with a new method that handles data from a system file.

PR #3948, opened 2024-07-03
fix: deeptensor output, add dipole stat UT

anyangml:
## Summary by CodeRabbit

- **New Features**
  - Introduced `model_dipole` for enhanced dipole model configurations.
  - Added support for "global_dipole" data in finetuning tests.

- **Improvements**
  - Updated return type of the `eval` method to ensure consistency and clarity in the output.
  - Improved energy calculation logic in finetuning tests by considering additional data types.

- **Testing**
  - Enhanced test coverage for dipole and DOS models with updated test cases and configurations.

### DIFF
--- a/deepmd/infer/deep_tensor.py
+++ b/deepmd/infer/deep_tensor.py
@@ -50,7 +50,7 @@ class DeepTensor(DeepEval):
         aparam: Optional[np.ndarray] = None,
         mixed_type: bool = False,
         **kwargs: dict,
-    ) -> np.ndarray:
+    ) -> Tuple[np.ndarray]:
         """Evaluate the model.
 
         Parameters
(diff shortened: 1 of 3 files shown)


anyangml:
I may close this PR if ndarry is the expected output type.

wanghan-iapcm (commented):
Could you please explain the reason of using `Tuple[np.ndarray]` instead of `np.ndarray` as returned type of `DeepTensor`?

anyangml:
> Could you please explain the reason of using `Tuple[np.ndarray]` instead of `np.ndarray` as returned type of `DeepTensor`?

It seems the other DeepModels all return a tuple object, I thought they should be consistent. When adding the new UT, DeepDipole eval needs special handling if a ndarray is returned. Although the UT is not as important, since dipole model does not apply bias, just want to check the changes made in #3945.

PR #3958, opened 2024-07-09
breaking(pt/tf/dp): disable bias in type embedding

iProzd:
This PR addresses an issue observed during training with DPA2 on complex datasets, such as `mptraj`. Specifically, the **learning curves of energy** from the **2024Q1-based branch** and the **devel branch** show significant differences at the very beginning when setting `tebd_dim` = 256 (and thus descriptor `dim_out` = 128 + 256). The issue is illustrated in the following image:

Example Image

After removing the bias in the type embedding, which affects the standard deviation of the descriptor when `tebd_dim` is very large, the learning curve improves significantly:

Example Image

Notably, this behavior is not prominent when using a `tebd_dim` that is relatively smaller than the descriptor itself, such as when using DPA2 with `tebd_dim` = 8 or using DPA1.

The same issue exists in econf of type embedding, which will be solved in a separated PR.

**NOTE**
**This PR disables bias in type embedding in all backends, which is a breaking change.**

## Summary by CodeRabbit

- **New Features**
  - Introduced `use_tebd_bias` and `bias` parameters across various components to control the use of bias in type embeddings and networks.
  
- **Updates**
  - Updated serialization and deserialization methods to include the new parameters and ensure version compatibility.

### DIFF
--- a/deepmd/dpmodel/descriptor/se_atten_v2.py
+++ b/deepmd/dpmodel/descriptor/se_atten_v2.py
@@ -161,7 +161,7 @@ class DescrptSeAttenV2(DescrptDPA1):
     def deserialize(cls, data: dict) -> "DescrptSeAttenV2":
         """Deserialize from dict."""
         data = data.copy()
-        check_version_compatibility(data.pop("@version"), 2, 2)
+        check_version_compatibility(data.pop("@version"), 2, 1)
         data.pop("@class")
         data.pop("type")
         variables = data.pop("@variables")
@@ -170,6 +170,9 @@ class DescrptSeAttenV2(DescrptDPA1):
         attention_layers = data.pop("attention_layers")
         data.pop("env_mat")
         embeddings_strip = data.pop("embeddings_strip")
+        # compat with version 1
+        if "use_tebd_bias" not in data:
+            data["use_tebd_bias"] = True
         obj = cls(**data)
 
         obj.se_atten["davg"] = variables["davg"]
--- a/deepmd/dpmodel/descriptor/se_atten_v2.py
+++ b/deepmd/dpmodel/descriptor/se_atten_v2.py
@@ -64,6 +64,7 @@ class DescrptSeAttenV2(DescrptDPA1):
         spin: Optional[Any] = None,
         stripped_type_embedding: Optional[bool] = None,
         use_econf_tebd: bool = False,
+        use_tebd_bias: bool = False,
         type_map: Optional[List[str]] = None,
         # consistent with argcheck, not used though
         seed: Optional[Union[int, List[int]]] = None,
@@ -100,6 +101,7 @@ class DescrptSeAttenV2(DescrptDPA1):
             spin=spin,
             stripped_type_embedding=stripped_type_embedding,
             use_econf_tebd=use_econf_tebd,
+            use_tebd_bias=use_tebd_bias,
             type_map=type_map,
             # consistent with argcheck, not used though
             seed=seed,
@@ -111,7 +113,7 @@ class DescrptSeAttenV2(DescrptDPA1):
         data = {
             "@class": "Descriptor",
             "type": "se_atten_v2",
-            "@version": 1,
+            "@version": 2,
             "rcut": obj.rcut,
             "rcut_smth": obj.rcut_smth,
             "sel": obj.sel,
@@ -134,6 +136,7 @@ class DescrptSeAttenV2(DescrptDPA1):
             "type_one_side": obj.type_one_side,
             "concat_output_tebd": self.concat_output_tebd,
             "use_econf_tebd": self.use_econf_tebd,
+            "use_tebd_bias": self.use_tebd_bias,
             "type_map": self.type_map,
             # make deterministic
             "precision": np.dtype(PRECISION_DICT[obj.precision]).name,
@@ -158,7 +161,7 @@ class DescrptSeAttenV2(DescrptDPA1):
     def deserialize(cls, data: dict) -> "DescrptSeAttenV2":
         """Deserialize from dict."""
         data = data.copy()
-        check_version_compatibility(data.pop("@version"), 1, 1)
+        check_version_compatibility(data.pop("@version"), 2, 2)
         data.pop("@class")
         data.pop("type")
         variables = data.pop("@variables")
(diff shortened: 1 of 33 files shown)


deepmd/utils/argcheck.py
```diff
@@ -104,6 +105,13 @@ def type_embedding_args():
         Argument(
             "use_econf_tebd", bool, optional=True, default=False, doc=doc_use_econf_tebd
         ),
+        Argument(
+            "use_tebd_bias",
+            bool,
+            optional=True,
+            default=False,
+            doc=doc_use_tebd_bias,
+        ),
```
njzjz, line 114: While the default value of `activation_function` is `tanh` and the default value of `use_tebd_bias` is `False`, isn't the output value limited to $(-1,1)$?

deepmd/dpmodel/utils/network.py
```diff
@@ -581,6 +581,7 @@ def __init__(
             resnet_dt: bool = False,
             precision: str = DEFAULT_PRECISION,
             seed: Optional[Union[int, List[int]]] = None,
+            bias: bool = True,
```
njzjz, line 584: `bias` should be added to the docstring.

deepmd/tf/utils/network.py
```diff
@@ -289,18 +292,26 @@ def embedding_net(
...
+            else None
         )
         if initial_variables is not None:
             scope = tf.get_variable_scope().name
             w_initializer = tf.constant_initializer(
                 initial_variables[scope + "/matrix_" + str(ii) + name_suffix]
             )
-            b_initializer = tf.constant_initializer(
-                initial_variables[scope + "/bias_" + str(ii) + name_suffix]
+            b_initializer = (
```
njzjz, line 309: My idea is this (take the example of TF)

```suggestion
            bias = (scope + "/bias_" + str(ii) + name_suffix) in initial_variables
            b_initializer = (
```

njzjz (changes requested):

deepmd/utils/argcheck.py
wanghan-iapcm: It should not be a big issue to eliminate the intermediate type representation (embedded types)  in range (-1, 1). 
In the subsequent usages, e.g. in embedding nets, the intermediate type rep. firstly passes linear operator in the first layer of the embedding net. Therefore, we do not need to add a linear layer after the activation.

wanghan-iapcm (commented):
I agree with the comments from @njzjz, and it will LGTM when the comments are addressed.

deepmd/dpmodel/utils/network.py
anyangml: should we name this `use_tebd_bias`

deepmd/utils/argcheck.py
iProzd: In fact, we use `none` as default value of `activation_function` when user **does not set the `type_embedding` key** in TF, see https://github.com/deepmodeling/deepmd-kit/blob/devel/deepmd/tf/model/model.py#L696 .

deepmd/dpmodel/utils/network.py
iProzd: > should we name this `use_tebd_bias`

It's in `make_embedding_network`, which is also used by other layers. So `bias` may be more suitable.

deepmd/dpmodel/descriptor/se_atten_v2.py
```diff
@@ -158,7 +161,7 @@ def serialize(self) -> dict:
     def deserialize(cls, data: dict) -> "DescrptSeAttenV2":
         """Deserialize from dict."""
         data = data.copy()
-        check_version_compatibility(data.pop("@version"), 1, 1)
+        check_version_compatibility(data.pop("@version"), 2, 2)
```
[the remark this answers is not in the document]
iProzd: For dp/tf, I make it compatible in https://github.com/deepmodeling/deepmd-kit/pull/3958/commits/27434e7167e1f7128427466b806e29ec96d5f788. But for old trained `model.pt`, I found making it compatible with new code may make code unnecessarily confusing. (check state_dict keys and add extra logics.) Maybe we leave it as a breaking change?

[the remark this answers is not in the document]
njzjz: You just need to use `self.register_parameter("bias", None)` instead of `self.bias = None`, so `load_state_dict` can correctly override it. See https://github.com/pytorch/pytorch/blob/cb2bce98de5bba1c5415a08d7b17c608f6c77b62/torch/nn/modules/activation.py#L1084. By the way, other parameters like `idt` should follow the same way to have the maximum compatibility.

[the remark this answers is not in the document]
njzjz: Oh I found this issue: https://github.com/pytorch/pytorch/issues/8104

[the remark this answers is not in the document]
njzjz: It seems this is the only way:
```py
import torch

class TestParameter(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.bias = None

    def load_state_dict(self, state):
        if self.bias is None and "bias" in state:
            self.bias = torch.nn.Parameter(data=state["bias"])
        super().load_state_dict(state)

m=TestParameter()
m.load_state_dict({"bias": torch.zeros(3)})
print(m.bias)
```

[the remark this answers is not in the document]
njzjz: It may be better to make a breaking change? One can convert it to the DP model using the old version, and then back to the PyTorch model using the new version.

anyangml (approved):

deepmd/dpmodel/descriptor/se_atten_v2.py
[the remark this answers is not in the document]
iProzd: > It may be better to make a breaking change? One can convert it to the DP model using the old version, and then back to the PyTorch model using the new version.

I agree.

wanghan-iapcm (approved):

deepmd/utils/argcheck.py
wanghan-iapcm: This is a strange behavior. I do no understand why the activation function is set to None if user does not set the type_embedding

PR #3967, opened 2024-07-11
fix(pt): add `finetune_head` to argcheck

iProzd:
Add `finetune_head`  to argcheck.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `finetune_head` argument for specifying the fitting net during multi-task fine-tuning, with optional random initialization if not set.
  
- **Bug Fixes**
  - Improved handling for specific conditions by automatically removing the "finetune_head" key from the configuration.

- **Tests**
  - Updated multitask training and finetuning tests to include new configuration manipulations.
  - Removed the `_comment` field from test configuration files to ensure cleaner test setups.

### DIFF
--- a/deepmd/pt/utils/finetune.py
+++ b/deepmd/pt/utils/finetune.py
@@ -155,7 +155,6 @@ def get_finetune_rules(
             change_model_params=change_model_params,
         )
         finetune_links["Default"] = finetune_rule
-        model_config.pop("finetune_head", None)  # for argcheck
     else:
         assert model_branch == "", (
             "Multi-task fine-tuning does not support command-line branches chosen!"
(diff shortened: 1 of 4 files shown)


njzjz:
Why isn't this option in the argcheck?

iProzd:
> Why isn't this option in the argcheck?

This option works only to get finetune rules before argcheck, see https://github.com/deepmodeling/deepmd-kit/blob/devel/deepmd/pt/entrypoints/main.py#L253 . After this, the option is useless.

We can deleted it from the model definition or keep a placeholder in argcheck, which one do we prefer? @njzjz @wanghan-iapcm

njzjz:
We use argcheck to generate the documentation, so all user-defined arguments should be in the argcheck, i.e. in the automatically generated documentation.

njzjz (approved):

njzjz (approved):

PR #3971, opened 2024-07-12
feat(tf): improve the activation setting in tebd

iProzd:
## Summary by CodeRabbit

- **New Features**
  - Added logging to notify users when multiple neurons are used with a "linear" activation function, equivalent to a single linear layer.

- **Configuration Updates**
  - Introduced `"activation_function": "linear"` in various neural network configuration files (`input.json`).
  - Added `"activation_function": "tanh"` in the `input.json` file for water model configuration.

### DIFF
--- a/deepmd/tf/utils/type_embed.py
+++ b/deepmd/tf/utils/type_embed.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
+import logging
 import re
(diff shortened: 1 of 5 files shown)


njzjz (approved):

njzjz:
We also need to improve `seed`.

wanghan-iapcm (approved):

PR #3976, opened 2024-07-14
fix(pt/tf/dp): normalize the econf

iProzd:
## Summary by CodeRabbit

- **Improvements**
  - Enhanced the accuracy of energy configuration embeddings by adding a normalization step.

### DIFF
--- a/deepmd/dpmodel/utils/type_embed.py
+++ b/deepmd/dpmodel/utils/type_embed.py
@@ -234,5 +234,6 @@ def get_econf_tebd(type_map, precision: str = "default"):
         [electronic_configuration_embedding[kk] for kk in type_map],
         dtype=PRECISION_DICT[precision],
     )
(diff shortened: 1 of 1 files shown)


deepmd/dpmodel/utils/type_embed.py
```diff
@@ -234,5 +234,6 @@ def get_econf_tebd(type_map, precision: str = "default"):
         [electronic_configuration_embedding[kk] for kk in type_map],
         dtype=PRECISION_DICT[precision],
     )
+    econf_tebd /= econf_tebd.sum(-1, keepdims=True)  # do normalization
```
njzjz, line 237: Isn't it `keepdim`?

https://pytorch.org/docs/stable/generated/torch.Tensor.sum.html

njzjz: Oh, I realize it is NumPy

njzjz (approved):

PR #3993, opened 2024-07-18
fix(pt): fix lammps nlist sort with large sel

iProzd:
`nlist` was not sorted and passed into `build_multiple_neighbor_list` when using dpa2 and large sel (larger than lammps max nnei).

Bug fixed on several error systems but a UT needed.

- [x] UT for special sel case in dpa2.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new parameter to enhance sorting flexibility in computation methods.
  - Added methods to check if sorted neighbor lists are required, providing improved control over model behavior.

- **Bug Fixes**
  - Enhanced functionality related to neighbor list formatting for improved performance across models.

- **Documentation**
  - Updated method signatures to reflect new parameters and functionalities, clarifying usage for end-users.


njzjz:
> when using dpa2 and large sel

Do you mean it only affects dpa2 but not other descriptors?

iProzd:
> > when using dpa2 and large sel
> 
> Do you mean it only affects dpa2 but not other descriptors?

yes, since only dpa2 uses `build_multiple_neighbor_list`, which requires nlist to be sorted: https://github.com/deepmodeling/deepmd-kit/blob/devel/deepmd/pt/utils/nlist.py#L220-L222

deepmd/pt/model/model/ener_model.py
```diff
@@ -105,6 +105,7 @@ def forward_lower(
         aparam: Optional[torch.Tensor] = None,
         do_atomic_virial: bool = False,
         comm_dict: Optional[Dict[str, torch.Tensor]] = None,
+        extra_nlist_sort: bool = True,
```
wanghan-iapcm, line 108: it seems that one needs to assign different values for dpa1 and dpa2.

njzjz:
> > > when using dpa2 and large sel
> > 
> > 
> > Do you mean it only affects dpa2 but not other descriptors?
> 
> yes, since only dpa2 (and linear_atomic_model) uses `build_multiple_neighbor_list`, which requires nlist to be sorted: https://github.com/deepmodeling/deepmd-kit/blob/devel/deepmd/pt/utils/nlist.py#L220-L222

In this case, why do you only modify the energy model?

source/tests/infer/deeppot_dpa_sel.pth
njzjz, line 1: What is the difference between this file and `deeppot_dpa.pth`?

deepmd/pt/model/model/make_model.py
```diff
@@ -215,6 +215,7 @@ def forward_common_lower(
             aparam: Optional[torch.Tensor] = None,
             do_atomic_virial: bool = False,
             comm_dict: Optional[Dict[str, torch.Tensor]] = None,
+            extra_nlist_sort: bool = False,
```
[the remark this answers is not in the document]
iProzd: Because the nlist passed to `forward` interface is always sorted, we only need to process the `forward_lower` case.

source/tests/infer/deeppot_dpa_sel.pth
iProzd: It has smaller sel for `repformer`, which will select wrong neighbors if unsorted nlist is given.

source/lmp/tests/test_lammps_dpa_sel_pt.py
```diff
@@ -0,0 +1,727 @@
...
+    os.remove(data_type_map_file)
+
+
+def _lammps(data_file, units="metal") -> PyLammps:
+    lammps = PyLammps()
+    lammps.units(units)
+    lammps.boundary("p p p")
+    lammps.atom_style("atomic")
+    lammps.plugin(
+        "load /home/data/zhangd/deepmd-kit-sources/deepmd-kit-refactor/dp/lib/libdeepmd_lmp.so"
```
njzjz, line 253: should be removed

njzjz (changes requested):

source/lmp/tests/test_lammps_dpa_sel_pt.py
iProzd: removed.

CaRoLZhangxy (approved):

njzjz (approved):

deepmd/dpmodel/descriptor/dpa2.py
```diff
@@ -553,6 +553,15 @@ def has_message_passing(self) -> bool:
...
         )
 
+    def need_sorted_nlist_for_lower(self) -> bool:
+        """Returns whether the descriptor needs sorted nlist when using `forward_lower`."""
+        return any(
+            [
+                self.repinit.need_sorted_nlist_for_lower(),
+                self.repformers.need_sorted_nlist_for_lower(),
+            ]
+        )
```
wanghan-iapcm, line 563: shouldn't be `True` ?

deepmd/dpmodel/descriptor/repformers.py
```diff
@@ -401,6 +401,10 @@ def has_message_passing(self) -> bool:
         """Returns whether the descriptor block has message passing."""
         return True
 
+    def need_sorted_nlist_for_lower(self) -> bool:
+        """Returns whether the descriptor block needs sorted nlist when using `forward_lower`."""
+        return True
```
wanghan-iapcm, line 406: Shouldn't be `False`? Suppose the `repformer` is used alone it does not need a sorted nlist.

wanghan-iapcm (changes requested):

deepmd/dpmodel/descriptor/repformers.py
iProzd: I see, exactly. Fixed in https://github.com/deepmodeling/deepmd-kit/pull/3993/commits/be56b5e431c13ea9cff32d1cd98933849bfea9c3 .

deepmd/pt/model/atomic_model/linear_atomic_model.py
```diff
@@ -108,6 +108,10 @@ def has_message_passing(self) -> bool:
         """Returns whether the atomic model has message passing."""
         return any(model.has_message_passing() for model in self.models)
 
+    def need_sorted_nlist_for_lower(self) -> bool:
+        """Returns whether the atomic model needs sorted nlist when using `forward_lower`."""
+        return any(model.need_sorted_nlist_for_lower() for model in self.models)
```
njzjz, line 113: `True`?

https://github.com/deepmodeling/deepmd-kit/blob/be56b5e431c13ea9cff32d1cd98933849bfea9c3/deepmd/pt/model/atomic_model/linear_atomic_model.py#L212

njzjz (changes requested):

njzjz (approved):

wanghan-iapcm (approved):

Issue #3969, opened 2024-07-12
[BUG] CUDA out of memory, when only 1600 atoms, using the pytorch model with spin


shiruosong:
### Bug summary

I trained a pytorch version of the BiFeO3 model with DPSPIN. When I use the model to do minimization, with only 1600 atoms, I get an error that CUDA out of memory. The machine type is c12_m92_1 * NVIDIA V100. 

I had previously run DPLR with 1-2w atoms normally, and even the normal DP-tf model with more atoms.  For DPSPIN-tf, 1,600 atoms are also far from the limit. But for DPSPIN-pytorch, you can't do that anymore.

The ERROR is listed below: 
Setting up Verlet run ...
  Unit style    : metal
  Current step  : 0
  Time step     : 0.0001
terminate called after throwing an instance of 'std::runtime_error'
  what():  The following operation failed in the TorchScript interpreter.
Traceback of TorchScript, serialized code (most recent call last):
  File "code/__torch__/deepmd/pt/model/model/transform_output.py", line 154, in forward_lower
    vvi = split_vv1[_44]
    svvi = split_svv1[_44]
    _45 = _36(vvi, svvi, coord_ext, do_virial, do_atomic_virial, )
          ~~~ <--- HERE
    ffi, aviri, = _45
    ffi0 = torch.unsqueeze(ffi, -2)
  File "code/__torch__/deepmd/pt/model/model/transform_output.py", line 201, in task_deriv_one
    extended_virial0 = torch.matmul(_53, torch.unsqueeze(extended_coord, -2))
    if do_atomic_virial:
      extended_virial_corr = _50(extended_coord, atom_energy, )
                             ~~~ <--- HERE
      extended_virial2 = torch.add(extended_virial0, extended_virial_corr)
      extended_virial1 = extended_virial2
  File "code/__torch__/deepmd/pt/model/model/transform_output.py", line 234, in atomic_virial_corr
    ops.prim.RaiseException("AssertionError: ")
    extended_virial_corr00 = _55
  _61 = torch.autograd.grad([sumce1], [extended_coord], lst, None, True)
        ~~~~~~~~~~~~~~~~~~~ <--- HERE
  extended_virial_corr1 = _61[0]
  _62 = torch.__isnot__(extended_virial_corr1, None)

Traceback of TorchScript, original code (most recent call last):
[...]

PR #3996, opened 2024-07-19
fix(cc): add `atomic` argument to `DeepPotBase::computew`

njzjz:
See https://github.com/deepmodeling/deepmd-kit/issues/3969 for the background.

## Summary by CodeRabbit

- **New Features**
  - Introduced an 'atomic' parameter in various compute functions to enable atomic energy and virial calculations, providing more granular control over computations.

### DIFF
--- a/source/api_cc/include/DeepPot.h
+++ b/source/api_cc/include/DeepPot.h
@@ -61,30 +61,31 @@ class DeepPotBase {
    * nframes x natoms x dim_aparam.
    * natoms x dim_aparam. Then all frames are assumed to be provided with the
    *same aparam.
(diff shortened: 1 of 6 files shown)


CaRoLZhangxy (approved):
LGTM

wanghan-iapcm (approved):

iProzd (commented):
It seems that the atomic option is only in C++ and not in C, while the default lammps compile option is C. We should better support this option in C.

njzjz:
> It seems that the atomic option is only in C++ and not in C, while the default lammps compile option is C. We should better support this option in C.

This may be completed in a separated PR.

iProzd (approved):

PR #4010, opened 2024-07-23
fix(c): call C++ interface without atomic properties when they are not requested

njzjz:
This is an internal change and doesn't affect the API.

## Summary by CodeRabbit

- **New Features**
	- Enhanced computation functions for better flexibility in processing parameters based on specific flags.
- **Bug Fixes**
	- Improved performance and resource management by adapting the compute method's parameters dynamically.

### DIFF
--- a/source/api_c/src/c_api.cc
+++ b/source/api_c/src/c_api.cc
@@ -199,8 +199,13 @@ inline void DP_DeepPotCompute_variant(DP_DeepPot* dp,
   std::vector<double> e;
   std::vector<VALUETYPE> f, v, ae, av;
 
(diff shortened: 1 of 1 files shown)


iProzd (approved):

wanghan-iapcm (changes requested):
The doc of the output variables `atomic_virial` and `atomic_energy` should be updated.

njzjz:
> The doc of the output variables `atomic_virial` and `atomic_energy` should be updated. 

https://github.com/deepmodeling/deepmd-kit/blob/5dd0062e281f2c8a40ba929e2a9a41a58071a2c7/source/api_c/include/c_api.h#L181-L182

The documentation is correct before this PR.

wanghan-iapcm (approved):

PR #4015, opened 2024-07-24
fix(pt): use user seed in `DpLoaderSet`

iProzd:
## Summary by CodeRabbit

- **New Features**
	- Introduced a `seed` parameter for enhanced control over randomness in the training process.
	- Improved randomness handling for distributed training setups to ensure unique batch sequences.

- **Bug Fixes**
	- Enhanced robustness in seed setup during initialization to prevent unnecessary function calls with a `None` value.

- **Chores**
	- Streamlined the random seed management by removing redundant seed-setting logic in training processes.


deepmd/pt/utils/dataloader.py
```diff
@@ -80,7 +80,8 @@ def __init__(
         seed=10,
```
njzjz, line 80: Should we change 10 to None?

njzjz: Or remove the default value

iProzd: Yes I think None is better.

CaRoLZhangxy (approved):

njzjz (approved):

wanghan-iapcm (approved):

Issue #3679, opened 2024-04-17
LAMMPS fix dplr issue


cesaremalosso:
### Summary

`ERROR: find a bonded pair that is not on the same processor, something should not happen (../fix_dplr.cpp:345)` when restart from a previous simulation run.

### DeePMD-kit Version

2.2.3

### Backend and its version

2.13.0

### Details

Hi,

I'm encountering the following error:
`ERROR: find a bonded pair that is not on the same processor, something should not happen (../fix_dplr.cpp:345)`

when I try to restart a LAMMPS run from the previous ended one. I'm using a DPLR model and I'm running it just with 1 MPI process (`mpirun -np 1`) for the LAMMPS `lmp_mpi`. I attach the input file for LAMMPS as well as the two restart file from the previous run

[files.zip](https://github.com/deepmodeling/deepmd-kit/files/15009961/files.zip)

cesaremalosso:
Hi, has this issue been fixed?

Yi-FanLi:
> Hi, has this issue been fixed?

Hi @cesaremalosso, I am sorry that I was so overwhelmed in the past few weeks. In the recent days I have more time. I am working on this issue. Thanks for your patience.

cesaremalosso:
Thank you very much! I have another related question. It seems that the `kspace_style pppm/dplr`, which is used to account for the long-range interactions, is quite slow in LAMMPS, significantly slowing down the MD simulation. Do you think it would be beneficial to implement OpenMP thread parallelization to speed this up? Perhaps using GPUs for both the short-range NNP and the Wannier NN, while using CPU processes for the particle-particle particle-mesh solver?

njzjz:
> Perhaps using GPUs for both the short-range NNP and the Wannier NN, while using CPU processes for the particle-particle particle-mesh solver

This is the current implementation. `pppm/dplr` is based on `pppm` and the pppm solver is implemented in LAMMPS.

I know `pppm` has some variants, such as `pppm/gpu`, `pppm/omp`, `pppm/intel`, but they need to be tested.

Could you raise a separate issue?

cesaremalosso:
Hi @njzjz , I raise a separate issue here: [#4078](https://github.com/deepmodeling/deepmd-kit/issues/4078)

Yi-FanLi:
[...]

PR #4084, opened 2024-08-28
fix dplr: support initializing from a restart file

Yi-FanLi:
Thie PR aims at fixing #3679.

The `get_valid_pairs` will raise an error when
1) reading a restart file written from previous dplr jobs
2) reading a data file written from previous dplr jobs
3) executing 'run 0' for twice

The origin is that the check in `get_valid_pairs` is too strict for the initialization process. The check requires that a pair of real and virtual atoms found by iterating over the bondlist to be the local atoms of the same processor. This requirement is automatically satisfied during the running of MD because the positions of virtual atoms are set equal to those of real atoms before rebuilding bondlists. However, during initialization (`Verlet::setup(int flag)`), the bondlist is built based on the initial configuration provided by the user. If the user provides a configuration in which the virtual atoms are placed at the Wannier centroids, it is possible that a bond formed by a pair of real and virtual atoms are between a local and a ghost atom, which triggers this error.

This PR loosens the requirement on the initial configuration provided by the user. In the `setup_post_neighbor` function, the `get_valid_pairs` function searches for bonded pairs but skips the aforementioned check. It sets the coordinates of the virtual atoms to the coordinates of the real atoms, and then rebuilds the bondlist. Afterwards, the check can be passed. 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method for setting up pre-exchange operations, enhancing simulation lifecycle management.
- **Changes**
	- Renamed the `post_integrate` method to `pre_exchange` to better reflect its role in the simulation process.
	- Updated mask settings to improve clarity and context within the simulation lifecycle.
- **Chores**
	- Updated ownership context for Docker image in the build workflow configuration.


source/lmp/fix_dplr.cpp
```diff
@@ -312,16 +313,68 @@ void FixDPLR::init() {
...
 /* ---------------------------------------------------------------------- */
 
 void FixDPLR::setup(int vflag) {
-  // if (strstr(update->integrate_style,"verlet"))
   post_force(vflag);
-  // else {
-  //   error->all(FLERR, "respa is not supported by this fix");
-  // }
+  printf("end of setup\n");
+  printf("nlocal = %d\n", atom->nlocal);
```
njzjz, line 377: should be cleaned up

source/lmp/fix_dplr.cpp
njzjz, line 1: could you add a test? For example, run 0 after run 0

Yi-FanLi: Test added.

njzjz (approved):

source/lmp/fix_dplr.cpp
```diff
@@ -312,25 +303,61 @@ void FixDPLR::init() {
...
+void FixDPLR::setup(int vflag) { post_force(vflag); }
 
 /* ---------------------------------------------------------------------- */
 
 void FixDPLR::min_setup(int vflag) { setup(vflag); }
 
 /* ---------------------------------------------------------------------- */
 
-void FixDPLR::get_valid_pairs(vector<pair<int, int> > &pairs) {
+void FixDPLR::get_valid_pairs(vector<pair<int, int> > &pairs, int setup) {
```
wanghan-iapcm, line 360: could you plz change to name to e.g. `is_setup` or `strict_check_valid_bond_pair`, and change the type to bool.

wanghan-iapcm (changes requested):

source/lmp/fix_dplr.cpp
Yi-FanLi: Done!

wanghan-iapcm (approved):

PR #4089, opened 2024-08-30
breaking(pt/dp): tune new sub-structures for DPA2

iProzd:
## Summary by CodeRabbit

- **New Features**
	- Introduced support for three-body interactions in the descriptor model, enhancing predictive capabilities.
	- Added new parameters for greater configurability in the `RepinitArgs`, `RepformerArgs`, and related classes.

- **Bug Fixes**
	- Improved handling of serialization and deserialization processes to include new parameters.

- **Tests**
	- Expanded test cases to validate new configurations and parameters related to three-body interactions and model flexibility.


iProzd:
One question: should we set default values for these new params (e.g. `g1_out_conv`, `g1_out_mlp`) to be `True`? 
If so, then this PR will be a **breaking change** but easier for users to get the best settings. 

Eamples will be added after this comment is resovled.

njzjz:
> One question: should we set default values for these new params (e.g. `g1_out_conv`, `g1_out_mlp`) to be `True`?
> If so, then this PR will be a **breaking change** but easier for users to get the best settings.

I don't think it is a problem before the stable version is released.

deepmd/pt/model/descriptor/dpa2.py
```diff
@@ -210,9 +234,41 @@ def init_subclass_params(sub_data, sub_class):
...
+                self.rcut_list = [
+                    self.repinit_three_body.get_rcut(),
+                    self.repformers.get_rcut(),
+                    self.repinit.get_rcut(),
+                ]
+                self.nsel_list = [
+                    self.repinit_three_body.get_nsel(),
+                    self.repformers.get_nsel(),
+                    self.repinit.get_nsel(),
+                ]
```
wanghan-iapcm, line 271:
```suggestion
        self.rcsl_list = [(self.repformers.get_rcut(), self.repformers.get_nsel()), 
                                  (self.repinit.get_rcut(), self.repinit.get_nsel())]
        if self.use_three_body:
            self.rcsl_list.append((self.repinit_three_body.get_rcut(), self.repinit_three_body.get_sel()))
        self.rcsl_list.sort()
        for ii in range(1, len(self.rcsl_list)):
            assert self.rcsl_list[ii-1][1] <= self.rcsl_list[ii][1], "rcut and sel are not in the same order"
        self.rcut_list = [ii[0] for ii in self.rcsl_list]
        self.nsel_list = [ii[1] for ii in self.rcsl_list]
```

deepmd/pt/model/descriptor/repformer_layer.py
```diff
@@ -908,14 +961,24 @@ def _cal_hg(
...
             g2 = _apply_switch(g2, sw)
-            invnnei = (1.0 / float(nnei)) * torch.ones(
-                (nb, nloc, 1, 1), dtype=g2.dtype, device=g2.device
-            )
+            if not use_sqrt_nnei:
+                invnnei = (1.0 / float(nnei)) * torch.ones(
+                    (nb, nloc, 1, 1), dtype=g2.dtype, device=g2.device
+                )
+            else:
+                invnnei = (1.0 / (float(nnei) ** 0.5)) * torch.ones(
```
njzjz, line 979: You can use [`torch.rsqrt`](https://pytorch.org/docs/stable/generated/torch.rsqrt.html).

deepmd/dpmodel/descriptor/dpa2.py
```diff
@@ -779,16 +900,43 @@ def serialize(self) -> dict:
...
+                    "repinit_three_body_variable": repinit_three_body_variable,
+                }
+            )
         return data
 
     @classmethod
     def deserialize(cls, data: dict) -> "DescrptDPA2":
         data = data.copy()
-        check_version_compatibility(data.pop("@version"), 2, 1)
+        check_version_compatibility(data.pop("@version"), 3, 3)
```
njzjz, line 930: I think we are still possible to support old versions just by writing some `if...else` statements.

```py
if version >= 3:
   ...
else:
   ...
```

njzjz: Do we have any breaking change here?

wanghan-iapcm (approved):

njzjz (approved):

PR #4116, opened 2024-09-10
chore: support preset bias of atomic model output

wanghan-iapcm:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new `preset_out_bias` parameter for enhanced model configuration, allowing users to define biases.
	- Added documentation for the `preset_out_bias` parameter in the model arguments for improved clarity.

- **Bug Fixes**
	- Implemented validation to ensure the `preset_out_bias` length matches the model's type map, preventing runtime errors.

- **Tests**
	- Added unit tests for the `get_model` function to validate model attributes and ensure proper error handling for the new bias parameter.

### DIFF
--- a/deepmd/pt/model/atomic_model/base_atomic_model.py
+++ b/deepmd/pt/model/atomic_model/base_atomic_model.py
@@ -67,9 +67,9 @@ class BaseAtomicModel(torch.nn.Module, BaseAtomicModel_):
         of the atomic model. Implemented by removing the pairs from the nlist.
     rcond : float, optional
         The condition number for the regression of atomic energy.
-    preset_out_bias : Dict[str, List[Optional[np.array]]], optional
+    preset_out_bias : Dict[str, List[Optional[np.ndarray]]], optional
         Specifying atomic energy contribution in vacuum. Given by key:value pairs.
-        The value is a list specifying the bias. the elements can be None or np.array of output shape.
+        The value is a list specifying the bias. the elements can be None or np.ndarray of output shape.
         For example: [None, [2.]] means type 0 is not set, type 1 is set to [2.]
         The `set_davg_zero` key in the descrptor should be set.
 
@@ -81,7 +81,7 @@ class BaseAtomicModel(torch.nn.Module, BaseAtomicModel_):
         atom_exclude_types: List[int] = [],
         pair_exclude_types: List[Tuple[int, int]] = [],
         rcond: Optional[float] = None,
-        preset_out_bias: Optional[Dict[str, np.array]] = None,
+        preset_out_bias: Optional[Dict[str, np.ndarray]] = None,
     ):
         torch.nn.Module.__init__(self)
         BaseAtomicModel_.__init__(self)
--- a/deepmd/pt/model/atomic_model/base_atomic_model.py
+++ b/deepmd/pt/model/atomic_model/base_atomic_model.py
@@ -11,6 +11,7 @@ from typing import (
     Union,
 )
 
+import numpy as np
 import torch
 
 from deepmd.dpmodel.atomic_model import (
@@ -66,7 +67,7 @@ class BaseAtomicModel(torch.nn.Module, BaseAtomicModel_):
         of the atomic model. Implemented by removing the pairs from the nlist.
     rcond : float, optional
         The condition number for the regression of atomic energy.
-    preset_out_bias : Dict[str, List[Optional[torch.Tensor]]], optional
+    preset_out_bias : Dict[str, List[Optional[np.array]]], optional
         Specifying atomic energy contribution in vacuum. Given by key:value pairs.
         The value is a list specifying the bias. the elements can be None or np.array of output shape.
         For example: [None, [2.]] means type 0 is not set, type 1 is set to [2.]
@@ -80,7 +81,7 @@ class BaseAtomicModel(torch.nn.Module, BaseAtomicModel_):
         atom_exclude_types: List[int] = [],
         pair_exclude_types: List[Tuple[int, int]] = [],
         rcond: Optional[float] = None,
-        preset_out_bias: Optional[Dict[str, torch.Tensor]] = None,
+        preset_out_bias: Optional[Dict[str, np.array]] = None,
     ):
         torch.nn.Module.__init__(self)
         BaseAtomicModel_.__init__(self)
--- a/deepmd/pt/model/model/__init__.py
+++ b/deepmd/pt/model/model/__init__.py
@@ -176,6 +176,15 @@ def get_standard_model(model_params):
     fitting = BaseFitting(**fitting_net)
     atom_exclude_types = model_params.get("atom_exclude_types", [])
     pair_exclude_types = model_params.get("pair_exclude_types", [])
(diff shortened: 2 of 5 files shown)


deepmd/pt/model/model/__init__.py
```diff
@@ -151,6 +154,22 @@ def get_zbl_model(model_params):
     )
 
 
+def _convert_preset_out_bias_to_torch_tensor(preset_out_bias, type_map):
+    if preset_out_bias is not None:
+    if preset_out_bias is not None:
```
njzjz, line 159:
```suggestion
```

deepmd/pt/model/model/__init__.py
```diff
@@ -196,6 +218,7 @@ def get_standard_model(model_params):
         type_map=model_params["type_map"],
         atom_exclude_types=atom_exclude_types,
         pair_exclude_types=pair_exclude_types,
+        preset_out_bias=preset_out_bias,
```
njzjz, line 221: Need to confirm here. The type hint shows it accepts a `Optional[Dict[str, np.ndarray]]`, but this PR passes a `Optional[Dict[str, List[Optional[torch.tensor]]]]`.

https://github.com/deepmodeling/deepmd-kit/blob/8394101c60f8418ee3727c512d051816c259c690/deepmd/dpmodel/atomic_model/base_atomic_model.py#L43

Also, `torch.tensor` should not be serialized; `numpy.ndarray` should be stored in `@variables`.

https://github.com/deepmodeling/deepmd-kit/blob/8394101c60f8418ee3727c512d051816c259c690/deepmd/dpmodel/atomic_model/base_atomic_model.py#L230-L241

deepmd/utils/argcheck.py
```diff
@@ -1833,6 +1834,13 @@ def model_args(exclude_hybrid=False):
                 default=[],
                 doc=doc_only_pt_supported + doc_atom_exclude_types,
             ),
+            Argument(
+                "preset_out_bias",
+                dict,
```
njzjz, line 1839: More specific types can be used like `Dict[str, Optional[float]]` which is supported by dargs.

deepmd/utils/argcheck.py
```diff
@@ -1771,6 +1771,7 @@ def model_args(exclude_hybrid=False):
     doc_spin = "The settings for systems with spin."
     doc_atom_exclude_types = "Exclude the atomic contribution of the listed atom types"
     doc_pair_exclude_types = "The atom pairs of the listed types are not treated to be neighbors, i.e. they do not see each other."
+    doc_preset_out_bias = "The preset bias of the atomic output. Is provided as a dict. Taking the energy model that has three atom types for example, the preset_out_bias may be given as `{ 'energy': [None, 0., 1.] }`. In this case the bias of type 1 and 2 are set to 0. and 1., respectively.The set_davg_zero should be set to true."
```
njzjz, line 1774: Note: `'xxx'`, `None`, `0.` are valid in Python but not valid in JSON. (should be `"xxx"`, `null`, `0.0`)

njzjz (changes requested):

njzjz:
```
2024-09-10T20:18:26.6301739Z =========================== short test summary info ============================
2024-09-10T20:18:26.6302282Z FAILED source/tests/pt/model/test_property_fitting.py::TestInvariance::test_trans - AssertionError: 
2024-09-10T20:18:26.6302507Z Not equal to tolerance rtol=1e-07, atol=0
2024-09-10T20:18:26.6302517Z 
2024-09-10T20:18:26.6302717Z Mismatched elements: 55 / 55 (100%)
2024-09-10T20:18:26.6302927Z Max absolute difference: 0.00115797
2024-09-10T20:18:26.6303063Z Max relative difference: 0.02320881
2024-09-10T20:18:26.6303396Z  x: array([[[ 1.013011, -0.031964,  0.467955, -1.055318,  0.08362 ,
2024-09-10T20:18:26.6303659Z           0.343627,  0.329521, -1.231861, -0.326444, -1.820717,
2024-09-10T20:18:26.6303812Z           1.033206],...
2024-09-10T20:18:26.6304088Z  y: array([[[ 1.01299 , -0.03189 ,  0.467994, -1.055362,  0.08366 ,
2024-09-10T20:18:26.6304419Z           0.343748,  0.329551, -1.231867, -0.326422, -1.82073 ,
2024-09-10T20:18:26.6304542Z           1.033258],...
2024-09-10T20:18:26.6304892Z = 1 failed, 8331 passed, 3631 skipped, 59809 deselected, 145 warnings in 1976.94s (0:32:56) =
```

deepmd/pt/model/model/__init__.py
wanghan-iapcm: actually np.array is enough. however, the type hint in https://github.com/deepmodeling/deepmd-kit/blob/devel/deepmd/pt/utils/stat.py is a mess. Let me fix.

wanghan-iapcm: @njzjz see https://github.com/deepmodeling/deepmd-kit/pull/4116/commits/836e6c2bfd21994a7e2f22443f1d3dc79682767c

wanghan-iapcm: most functions in https://github.com/deepmodeling/deepmd-kit/blob/devel/deepmd/pt/utils/stat.py can be moved to deepmd/utils .  @njzjz what do you think?

njzjz: Good to move

deepmd/pt/model/atomic_model/base_atomic_model.py
```diff
@@ -80,7 +81,7 @@ def __init__(
         atom_exclude_types: List[int] = [],
         pair_exclude_types: List[Tuple[int, int]] = [],
         rcond: Optional[float] = None,
-        preset_out_bias: Optional[Dict[str, torch.Tensor]] = None,
+        preset_out_bias: Optional[Dict[str, np.array]] = None,
```
njzjz, line 84: np.ndarray

deepmd/pt/utils/stat.py
```diff
@@ -242,7 +242,7 @@ def compute_output_stats(
     keys: Union[str, List[str]] = ["energy"],
     stat_file_path: Optional[DPPath] = None,
     rcond: Optional[float] = None,
-    preset_bias: Optional[Dict[str, List[Optional[torch.Tensor]]]] = None,
+    preset_bias: Optional[Dict[str, List[Optional[np.array]]]] = None,
```
njzjz, line 245: np.ndarray

deepmd/pt/utils/stat.py
```diff
@@ -405,7 +405,7 @@ def compute_output_stats_global(
     ntypes: int,
     keys: List[str],
     rcond: Optional[float] = None,
-    preset_bias: Optional[Dict[str, List[Optional[torch.Tensor]]]] = None,
+    preset_bias: Optional[Dict[str, List[Optional[np.array]]]] = None,
```
njzjz, line 408: np.ndarray

njzjz (approved):

PR #4117, opened 2024-09-10
[Fix] Wrap log_path with Path

HydrogenSulfate:
`FLAGS.log_path` need to be wrapped into `Path` type, or will raise error when call `.parent`.

## Summary by CodeRabbit

- **Improvements**
	- Enhanced handling of log file paths for better compatibility and functionality in file operations.

### DIFF
--- a/deepmd/pt/entrypoints/main.py
+++ b/deepmd/pt/entrypoints/main.py
@@ -555,7 +555,7 @@ def main(args: Optional[Union[List[str], argparse.Namespace]] = None):
     else:
         FLAGS = args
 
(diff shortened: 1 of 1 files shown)


njzjz:
It seems that the default value of `FLAGS.log_path` is `None`. Please consider this situation.

HydrogenSulfate:
> It seems that the default value of `FLAGS.log_path` is `None`. Please consider this situation.

You are right, it should be same as <https://github.com/deepmodeling/deepmd-kit/blob/a5346f216508d119fe456d496c2ee469ba9b88ea/deepmd/entrypoints/main.py#L52-L54>

njzjz (approved):

PR #4120, opened 2024-09-11
fix: bugs in uts for property fit

Chengqian-Zhang:
This bug is totally the same as PR #3837. 
Fix following trivial bugs in property fit uts:

- box was not used in extend_input_and_build_neighbor_list (which means they were all tested in nopbc mode, if shifted coord is outside the box (sometimes) and normalized explicitly, results are not the same.) Input for fitting also used extended_atype instead of atype. (Only same when nopbc.)
- Using of mixed_types is disordered, mismatched with descriptor or sometimes with nlist. Now only use mixed_types==False since the descriptor output is not in mixed types.
- Remove useless parameter `fit_diag` and `scale` test in property fitting. Add parameter `intensive` and `bias_method` test in property fitting.

## Summary by CodeRabbit

- **New Features**
	- Introduced new parameters `intensive` and `bias_method` for enhanced flexibility in property fitting tests.
	- Added a new test class `TestInvarianceOutCell` with a method `test_trans` to evaluate invariance under transformations.
	- Updated existing tests to improve clarity and maintainability by removing the `scale` variable.

- **Bug Fixes**
	- Refactored test methods to ensure correct parameter usage, enhancing the reliability of test outcomes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
         exclude: ^source/3rdparty
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.6.3
+    rev: v0.6.4
     hooks:
       - id: ruff
         args: ["--fix"]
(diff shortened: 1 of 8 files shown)


iProzd (approved):

njzjz:
These tests seems to be duplicate of those in `tests/universal`. Do we have a plan to remove them?

wanghan-iapcm (approved):

Issue #4108, opened 2024-09-06
[BUG] finetuning property fitting with multi-dimensional data causes error


theAfish:
### Bug summary

I have tested the new property fitting model in fine-tuning procedures with the pre-trained OpenLAM_2.2.0_27heads_beta3.pt.
The dataset I used is in the examples folder and has a dimension of 3. Raised errors about tensor size mismatch. See the Error Log below.

### DeePMD-kit Version

DeePMD-kit v3.0.0a1.dev320+g46632f90

### Backend and its version

torch 2.4.1+cu121

### How did you download the software?

Built from source

### Input Files, Running Commands, Error Log, etc.

Commands: `dp --pt train input_finetune.json --finetune OpenLAM_2.2.0_27heads_beta3.pt`

Input File:
[input_finetune.json](https://github.com/user-attachments/files/16905913/input_finetune.json)

The data files I used are in `examples/property/data`

Error Log:
```
To get the best performance, it is recommended to adjust the number of threads by setting the environment variables OMP_NUM_THREADS, DP_INTRA_OP_PARALLELISM_THREADS, and DP_INTER_OP_PARALLELISM_THREADS. See https://deepmd.rtfd.io/parallelism/ for more information.
[2024-09-06 17:33:46,629] DEEPMD INFO    DeePMD version: 3.0.0a1.dev320+g46632f90
[2024-09-06 17:33:46,629] DEEPMD INFO    Configuration path: input_finetune.json
[2024-09-06 17:33:46,672] DEEPMD INFO     _____               _____   __  __  _____           _     _  _
[2024-09-06 17:33:46,672] DEEPMD INFO    |  __ \             |  __ \ |  \/  ||  __ \         | |   (_)| |
[2024-09-06 17:33:46,672] DEEPMD INFO    | |  | |  ___   ___ | |__) || \  / || |  | | ______ | | __ _ | |_
[2024-09-06 17:33:46,672] DEEPMD INFO    | |  | | / _ \ / _ \|  ___/ | |\/| || |  | ||______|| |/ /| || __|
[2024-09-06 17:33:46,672] DEEPMD INFO    | |__| ||  __/|  __/| |     | |  | || |__| |        |   < | || |_
[2024-09-06 17:33:46,672] DEEPMD INFO    |_____/  \___| \___||_|     |_|  |_||_____/         |_|\_\|_| \__|
[2024-09-06 17:33:46,672] DEEPMD INFO    Please read and cite:
[...]

PR #4145, opened 2024-09-19
fix(pt): finetuning property/dipole/polar/dos fitting with multi-dimensional data causes error

Chengqian-Zhang:
Fix issue #4108 

If a pretrained model is labeled with energy and the `out_bias` is one dimension. If we want to finetune a dos/polar/dipole/property model using this pretrained model, the `out_bias` of finetuning model is multi-dimension(example: numb_dos = 250). An error occurs: 
`RuntimeError: Error(s) in loading state_dict for ModelWrapper:`
`       size mismatch for model.Default.atomic_model.out_bias: copying a param with shape torch.Size([1, 118, 1]) from checkpoint, the shape in current model is torch.Size([1, 118, 250]).`
`        size mismatch for model.Default.atomic_model.out_std: copying a param with shape torch.Size([1, 118, 1]) from checkpoint, the shape in current model is torch.Size([1, 118, 250]).`

When using new fitting, old out_bias is useless because we will recompute the new bias in later code. So we do not need to load old out_bias when using new fitting finetune.

## Summary by CodeRabbit

- **New Features**
	- Enhanced parameter collection for fine-tuning, refining criteria for parameter retention.
	- Introduced a model checkpoint file for saving and resuming training states, facilitating iterative development.

- **Tests**
	- Added a new test class to validate training and fine-tuning processes, ensuring model performance consistency across configurations.


deepmd/pt/train/training.py
```diff
@@ -484,7 +484,11 @@ def collect_single_finetune_params(
                             if i != "_extra_state" and f".{_model_key}." in i
                         ]
                         for item_key in target_keys:
-                            if _new_fitting and ".fitting_net." in item_key:
+                            if _new_fitting and (
+                                (".fitting_net." in item_key)
+                                or (".out_bias" in item_key)
+                                or (".out_std" in item_key)
```
njzjz, line 490: I think it should be `".descriptor" not in item_key`. When the model has other variables in the future, is it expected to keep it or replace it?

Chengqian-Zhang: In current version, the two ways of writing `".descriptor" not in item_key` and `(".fitting_net." in item_key) or (".out_bias" in item_key) or (".out_std" in item_key)` are fully equivalent. Please @iProzd  take a look at this.

iProzd: I think `".descriptor." not in item_key` is better. Yet @Chengqian-Zhang please check and ensure that they are equivalent.

Chengqian-Zhang: image

Chengqian-Zhang: I an sure that in current version, if "descriptor" and "fitting" do not in target_keys, only "out_bias" and "out_std" remain.

njzjz (approved):

wanghan-iapcm (approved):

iProzd (approved):

PR #4147, opened 2024-09-19
fix: set `weights_only=True` for `torch.load`

### LINKED ISSUE #4143: `torch.load` warnings
> /home/njzjz/codes/deepmd-kit/deepmd/pt/infer/inference.py:37: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
>  state_dict = torch.load(model_ckpt, map_location=DEVICE)

njzjz:
Fix #4143.

### DIFF
--- a/deepmd/pt/entrypoints/main.py
+++ b/deepmd/pt/entrypoints/main.py
@@ -275,7 +275,9 @@ def train(FLAGS):
         FLAGS.init_model is not None or FLAGS.init_frz_model is not None
     ) and FLAGS.use_pretrain_script:
         if FLAGS.init_model is not None:
-            init_state_dict = torch.load(FLAGS.init_model, map_location=DEVICE)
+            init_state_dict = torch.load(
+                FLAGS.init_model, map_location=DEVICE, weights_only=True
+            )
             if "model" in init_state_dict:
                 init_state_dict = init_state_dict["model"]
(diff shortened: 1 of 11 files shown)


njzjz:
```
WeightsUnpickler error: Unsupported class numpy.core.multiarray.scalar
```

Surprisingly, in some place, NumPy arrays are saved to the state dict. cc @iProzd @wanghan-iapcm

njzjz:
```py
>>> type(torch.load("model.ckpt.pt")["model"]["_extra_state"]["train_infos"]["lr"])
<class 'numpy.float64'>
```

njzjz:
It's almost one month past. Let's merge this PR.

wanghan-iapcm (approved):

iProzd (approved):

PR #4148, opened 2024-09-20
fix(pt): make `state_dict` safe for `weights_only`

iProzd:
See #4147 and #4143.
We can first make `state_dict` safe for `weights_only`, then make a breaking change when loading `state_dict` in the future.

## Summary by CodeRabbit

- **New Features**
  - Enhanced model saving functionality by ensuring learning rates are consistently stored as floats, improving type consistency.
  
- **Bug Fixes**
  - Updated model loading behavior in tests to focus solely on model weights, which may resolve issues related to state dictionary loading.

### DIFF
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -1030,10 +1030,13 @@ class Trainer:
             if dist.is_available() and dist.is_initialized()
             else self.wrapper
         )
-        module.train_infos["lr"] = lr
+        module.train_infos["lr"] = float(lr)
         module.train_infos["step"] = step
+        optim_state_dict = self.optimizer.state_dict()
+        for item in optim_state_dict["param_groups"]:
+            item["lr"] = float(item["lr"])
         torch.save(
-            {"model": module.state_dict(), "optimizer": self.optimizer.state_dict()},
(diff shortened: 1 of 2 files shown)


njzjz (approved):

deepmd/pt/train/training.py
```diff
@@ -1030,10 +1030,13 @@ def save_model(self, save_path, lr=0.0, step=0):
             if dist.is_available() and dist.is_initialized()
             else self.wrapper
         )
-        module.train_infos["lr"] = lr
+        module.train_infos["lr"] = float(lr)
         module.train_infos["step"] = step
+        optim_state_dict = self.optimizer.state_dict()
```
wanghan-iapcm, line 1035: need to deep copy?

iProzd: Not necessary. `module.state_dict()` is already a copy, which will not affect the innner params.

iProzd: Here we only modified the float values, where deep copy is not necessary. Anyway for safety tensors, we should indeed use deep copy.

wanghan-iapcm (approved):

PR #4169, opened 2024-09-29
feat(pt): train with energy Hessian

1azyking:
## Summary by CodeRabbit

- **New Features**
	- Introduced support for Hessian calculations across various components, enhancing the model's capabilities.
	- Added a new loss function for Hessian, allowing for more comprehensive training scenarios.
	- New JSON configuration files for multi-task and single-task learning models.
	- Enhanced output handling to include Hessian data in model evaluations.
	- Added new methods and properties to support Hessian in several classes and modules.

- **Bug Fixes**
	- Improved handling of output shapes and results related to Hessian data.

- **Documentation**
	- Updated documentation to include new Hessian properties and training guidelines.
	- Added sections detailing Hessian configurations and requirements in the training documentation.

- **Tests**
	- Added unit tests for the new Hessian-related functionalities to ensure consistency and correctness.
	- Enhanced existing test cases to incorporate Hessian data handling and validation.


wanghan-iapcm:
please fix all the conflicts  otherwise we cannot proceed. @1azyking

deepmd/pt/model/descriptor/env_mat.py
```diff
@@ -21,10 +21,11 @@ def _make_env_mat(
     nall = coord.shape[1]
     mask = nlist >= 0
     # nlist = nlist * mask  ## this impl will contribute nans in Hessian calculation.
-    nlist = torch.where(mask, nlist, nall - 1)
+    nlist = torch.where(mask, nlist, nall)
     coord_l = coord[:, :natoms].view(bsz, -1, 1, 3)
     index = nlist.view(bsz, -1).unsqueeze(-1).expand(-1, -1, 3)
-    coord_r = torch.gather(coord, 1, index)
+    coord_pad = torch.concat([coord, coord[:, -1:, :] + rcut], dim=1)
```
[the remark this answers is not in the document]
1azyking: Committed from [fix(pt): fix hess calculation for nopbc system](https://github.com/iProzd/deepmd-kit/commit/52f89fb5ffe5166f686b9cc101a5d976dc3d1e78)

deepmd/pt/model/model/ener_hess_model.py
```diff
@@ -0,0 +1,145 @@
...
+        }
+        if self.do_grad_r("energy"):
+            output_def["force"] = deepcopy(out_def_data["energy_derv_r"])
+            output_def["force"].squeeze(-2)
+        if self.do_grad_c("energy"):
+            output_def["virial"] = deepcopy(out_def_data["energy_derv_c_redu"])
+            output_def["virial"].squeeze(-2)
+            output_def["atom_virial"] = deepcopy(out_def_data["energy_derv_c"])
+            output_def["atom_virial"].squeeze(-3)
+        output_def["hessian"] = deepcopy(out_def_data["energy_derv_r_derv_r"])
```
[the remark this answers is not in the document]
1azyking: Such `squeeze` operation is identical to [deepmd/pt/model/model/ener_model.py](https://github.com/deepmodeling/deepmd-kit/blob/7ce5b032503d92abd104627f9d6b9b54967444cf/deepmd/pt/model/model/ener_model.py#L49)

deepmd/pt/model/model/ener_hess_model.py
```diff
@@ -0,0 +1,145 @@
...
+                if do_atomic_virial:
+                    model_predict["atom_virial"] = model_ret["energy_derv_c"].squeeze(
+                        -3
+                    )
+            else:
+                model_predict["force"] = model_ret["dforce"]
+            if "mask" in model_ret:
+                model_predict["mask"] = model_ret["mask"]
+            model_predict["hessian"] = model_ret["energy_derv_r_derv_r"]
+            model_predict["hessian"].squeeze(-2)
```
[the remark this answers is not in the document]
1azyking: Such squeeze operation is identical to [deepmd/pt/model/model/ener_model.py](https://github.com/deepmodeling/deepmd-kit/blob/7ce5b032503d92abd104627f9d6b9b54967444cf/deepmd/pt/model/model/ener_model.py#L49)

deepmd/entrypoints/test.py
```diff
@@ -573,6 +623,9 @@ def print_ener_sys_avg(avg: Dict[str, float]):
     log.info(f"Virial RMSE        : {avg['rmse_v']:e} eV")
     log.info(f"Virial MAE/Natoms  : {avg['mae_va']:e} eV")
     log.info(f"Virial RMSE/Natoms : {avg['rmse_va']:e} eV")
+    if "rmse_h" in avg.keys():
```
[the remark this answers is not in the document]
1azyking: Identical to the [previous version](https://github.com/deepmodeling/deepmd-kit/blob/7ce5b032503d92abd104627f9d6b9b54967444cf/deepmd/entrypoints/test.py#L579).

deepmd/pt/model/model/ener_hess_model.py
```diff
@@ -0,0 +1,145 @@
...
+                if do_atomic_virial:
+                    model_predict["extended_virial"] = model_ret[
+                        "energy_derv_c"
+                    ].squeeze(-3)
+            else:
+                assert model_ret["dforce"] is not None
+                model_predict["dforce"] = model_ret["dforce"]
+        else:
+            model_predict = model_ret
+        return model_predict
```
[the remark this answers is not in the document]
1azyking: Training with energy Hessian is not required for lower-level operations as of now.

examples/hessian/single-task/input.json
```diff
@@ -0,0 +1,103 @@
...
+                240,
+                240,
+                240
+            ],
+            "resnet_dt": true,
+            "seed": 1,
+            "_comment": " that's all"
+        },
+        "_comment": " that's all"
+    },
```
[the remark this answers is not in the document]
1azyking: The rcut settings are the same as [examples/water/dpa2/input_torch.json](https://github.com/deepmodeling/deepmd-kit/blob/0e0fc1a63e478d3e56285b520b34a9c58488d659/examples/water/dpa2/input_torch.json#L12).

deepmd/pt/train/training.py
```diff
@@ -273,8 +273,20 @@ def get_lr(lr_params):
...
             self.opt_type, self.opt_param = get_opt_param(training_params)
 
+        # loss_param_tmp for Hessian activation
+        loss_param_tmp = None
+        if not self.multi_task:
+            loss_param_tmp = config["loss"]
+        else:
+            for model_key in self.model_keys:
+                loss_param_tmp = config["loss_dict"][model_key]
+
```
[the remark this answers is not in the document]
1azyking: Followed the modifications of the forked version.

deepmd/utils/argcheck.py
```diff
@@ -1979,6 +1979,142 @@ def loss_ener():
...
+            doc=doc_limit_pref_gf,
+        ),
+        Argument(
+            "numb_generalized_coord",
+            int,
+            optional=True,
+            default=0,
+            doc=doc_numb_generalized_coord,
+        ),
+    ]
```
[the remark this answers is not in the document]
1azyking: The `loss_ener_hess` and `loss_ener` are expected to be separated to avoid confusion, despite the duplicated code.

deepmd/pt/loss/ener_hess.py
```diff
@@ -0,0 +1,453 @@
...
+                more_loss["mae_h"] = self.display_if_exist(mae_h.detach(), find_hessian)
+
+        if self.has_ae and "atom_energy" in model_pred and "atom_ener" in label:
+            atom_ener = model_pred["atom_energy"]
+            atom_ener_label = label["atom_ener"]
+            find_atom_ener = label.get("find_atom_ener", 0.0)
+            pref_ae = pref_ae * find_atom_ener
+            atom_ener_reshape = atom_ener.reshape(-1)
+            atom_ener_label_reshape = atom_ener_label.reshape(-1)
+            l2_atom_ener_loss = torch.square(
```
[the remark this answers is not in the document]
1azyking: The same as the forked version.

1azyking (commented):
My review is finished with all the potentail issues fixed.

njzjz (changes requested):
Please fix all checks so we can do further review.

image

1azyking (commented):
Conversations resolved

wanghan-iapcm:
please fix conflict and pre-commit.

1azyking (commented):
All the conversations resolved

deepmd/entrypoints/test.py
```diff
@@ -304,6 +304,10 @@ def test_ener(
     if dp.has_spin:
         data.add("spin", 3, atomic=True, must=True, high_prec=False)
         data.add("force_mag", 3, atomic=True, must=False, high_prec=False)
+    if hasattr(dp.deep_eval, "input_param") and dp.deep_eval.input_param.get(
```
njzjz, line 307: Please add a public API to `dp` instead of using `hasattr`.

deepmd/infer/deep_pot.py
```diff
@@ -239,6 +245,11 @@ def eval(
             force_mag = results["energy_derv_r_mag"].reshape(nframes, natoms, 3)
             mask_mag = results["mask_mag"].reshape(nframes, natoms, 1)
             result = (*list(result), force_mag, mask_mag)
+        if "energy_derv_r_derv_r" in list(results.keys()):
+            hessian = results["energy_derv_r_derv_r"].reshape(
+                nframes, 3 * natoms, 3 * natoms
+            )
+            result += (hessian,)
```
njzjz, line 252: I prefer the result to have a fixed number of return values regarding what the model is. The Hession can be obtained when a specific argument is set to True, or in another method.

deepmd/pt/infer/deep_eval.py
```diff
@@ -459,9 +461,7 @@ def _eval_model(
                 results.append(out)
             else:
                 shape = self._get_output_shape(odef, nframes, natoms)
-                results.append(
-                    np.full(np.abs(shape), np.nan)  # pylint: disable=no-explicit-dtype
-                )  # this is kinda hacky
+                results.append(np.full(np.abs(shape), np.nan))  # pylint: disable=no-explicit-dtype; this is kinda hack
```
njzjz, line 464: I don't get the reason to change this line.

deepmd/pt/infer/deep_eval.py
```diff
@@ -539,9 +539,7 @@ def _eval_model_spin(
                 results.append(out)
             else:
                 shape = self._get_output_shape(odef, nframes, natoms)
-                results.append(
-                    np.full(np.abs(shape), np.nan)  # pylint: disable=no-explicit-dtype
-                )  # this is kinda hacky
+                results.append(np.full(np.abs(shape), np.nan))  # pylint: disable=no-explicit-dtype; this is kinda hacky
```
njzjz, line 542: I don't get the reason to change this line.

deepmd/pt/loss/ener_hess.py
```diff
@@ -0,0 +1,454 @@
...
+)
+from deepmd.pt.utils.env import (
+    GLOBAL_PT_FLOAT_PRECISION,
+)
+from deepmd.utils.data import (
+    DataRequirementItem,
+)
+
+
+class EnergyHessianStdLoss(TaskLoss):
```
njzjz, line 24: It looks to me that the energy loss should be reused rather than being rewritten.

deepmd/pt/model/descriptor/env_mat.py
```diff
@@ -21,10 +21,11 @@ def _make_env_mat(
...
+    coord_r = torch.gather(coord_pad, 1, index)
```
njzjz, line 28: What is the reason for changing these lines?

deepmd/pt/train/training.py
```diff
@@ -1101,7 +1116,7 @@ def get_data(self, is_train=True, task_key="Default"):
                     batch_data = next(iter(self.validation_data[task_key]))
 
         for key in batch_data.keys():
-            if key == "sid" or key == "fid" or key == "box" or "find_" in key:
+            if key == "sid" or key == "fid" or key == "box":
```
njzjz, line 1119: What is the reason to change this line?

deepmd/utils/argcheck.py
```diff
@@ -212,17 +212,9 @@ def get_all_argument(self, exclude_hybrid: bool = False) -> list[Argument]:
...
-                arguments.append(
-                    Argument(
-                        name=name, dtype=dict, sub_fields=metd(), alias=alias, doc=doc
-                    )
-                )
-            else:
-                raise ValueError(f"Invalid return type {type(args)}")
+            arguments.append(
+                Argument(name=name, dtype=dict, sub_fields=metd(), alias=alias, doc=doc)
+            )
```
njzjz, line 217: What is the reason to change this?

deepmd/utils/argcheck.py
```diff
@@ -1884,7 +1876,9 @@ def model_args(exclude_hybrid=False):
             Variant(
                 "type",
                 [
-                    *model_args_plugin.get_all_argument(),
+                    standard_model_args(),
+                    frozen_model_args(),
+                    pairtab_model_args(),
```
njzjz, line 1881: What is the reason to change this?

deepmd/utils/argcheck.py
```diff
@@ -2921,7 +3077,7 @@ def gen_json_schema(multi_task: bool = False) -> str:
     return json.dumps(generate_json_schema(arg))
 
 
-def normalize(data, multi_task: bool = False):
+def normalize(data, multi_task=False):
```
njzjz, line 3080: What is the reason to change this?

doc/model/train-energy-hessian.md
njzjz, line 1: should be added to TOC

doc/model/train-energy-hessian.md
```diff
@@ -0,0 +1,48 @@
+# Fit energy Hessian
+
+To train a model that takes Hessian matrices, i.e., the second order derivatives of energies w.r.t coordinates as input, you only need to prepare full Hessian matrices and modify the `loss` section to define the Hessian-specific settings, keeping other sections the same as the normal energy model's input script.
+
+Note that fitting energy Hessians is only supported in the **PyTorch** backend as of now.
```
njzjz, line 5: We have a fixed format to express this, like below

https://github.com/deepmodeling/deepmd-kit/blob/c2944eb37a736ffd4c51b697fefe5518d3653c2a/doc/model/dpa2.md?plain=1#L1-L5

doc/model/train-energy-hessian.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Energy Hessian Loss
+
+If you want to train with Hessians, you are expected to add the start and limit prefactors of Hessians, i.e.,`start_pref_h` and `limit_pref_h` to the `loss` section in the `input.json`:
```
njzjz, line 9: Please add links for options like below.

https://github.com/deepmodeling/deepmd-kit/blob/c2944eb37a736ffd4c51b697fefe5518d3653c2a/doc/model/train-energy.md?plain=1#L92

doc/model/train-energy-hessian.md
```diff
@@ -0,0 +1,48 @@
...
+      "start_pref_h": 10,
+      "limit_pref_h": 1
+   },
+```
+
+The options `start_pref_e`, `limit_pref_e`, `start_pref_f`, `limit_pref_f`, `start_pref_v`, and `limit_pref_v`, determine the start and limit prefactors of energy, force, and virial, respectively. The calculation and definition of Hessian loss are the same as for the other terms.
+
+If one does not want to train with virial, they may set the virial prefactors `start_pref_v` and `limit_pref_v` to 0.
+
+## Hessian format in PyTorch
```
njzjz, line 29: should be addressed in the data session

examples/hessian/multi-task/input.json
njzjz, line 1: New examples should be added to `source/tests/common/test_examples.py`

examples/hessian/single-task/input.json
njzjz, line 1: New examples should be added to `source/tests/common/test_examples.py`

njzjz (changes requested):

deepmd/pt/model/descriptor/env_mat.py
1azyking: Changes are taken from [fix(pt): fix hess calculation for nopbc system](https://github.com/iProzd/deepmd-kit/commit/52f89fb5ffe5166f686b9cc101a5d976dc3d1e78). If not, elements in the last three columns of Hessian matrix derived from `torch.autograd.functional.hessian` would be nan.

deepmd/entrypoints/test.py
1azyking: Thanks for the update. Since `has_hessian` has been added to the `DeepEval` class, we can directly check `dp.has_hessian` instead of using `hasattr`.

examples/hessian/single-task/input.json
1azyking: It has been added as `p_examples / "hessian"/ "single_task" / "input.json"` to `source/tests/common/test_examples.py`.

examples/hessian/multi-task/input.json
1azyking: It has been added as `p_examples / "hessian"/ "multi_task" / "input.json"` to `source/tests/common/test_examples.py`.

deepmd/infer/deep_pot.py
1azyking: It has been modified in the same form as `self.deep_eval.get_has_spin` above.

deepmd/pt/loss/ener_hess.py
1azyking: You are right. I think this confusion is solved by merging `ener_hess.py` to `ener.py`, where class `EnergyHessianStdLoss`  inherits from `EnergyStdLoss`.

deepmd/pt/train/training.py
```diff
@@ -273,8 +274,18 @@ def get_lr(lr_params):
...
+            loss_param_tmp = config["loss"]
+        else:
+            loss_param_tmp = {
+                model_key: config["loss_dict"][model_key]
+                for model_key in self.model_keys
+            }
+
         # Model
-        self.model = get_model_for_wrapper(model_params)
+        self.model = get_model_for_wrapper(model_params, loss_param_tmp)
```
wanghan-iapcm, line 288: I do not understand why you need the loss parameters to get a model.

deepmd/pt/train/training.py
```diff
@@ -1271,15 +1290,22 @@ def get_single_model(
     return model
 
 
-def get_model_for_wrapper(_model_params):
+def get_model_for_wrapper(
+    _model_params,
+    _loss_params=None,
```
wanghan-iapcm, line 1295: very bad design to determine if a model is hessian from the loss function. 
you need to provide a new model type like "ener_hess" for your newly define class EnergyHessianModel

wanghan-iapcm (changes requested):

deepmd/pt/model/descriptor/env_mat.py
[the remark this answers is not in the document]
1azyking: Please @iProzd resolve this conversation.

deepmd/infer/deep_pot.py
```diff
@@ -21,6 +21,9 @@
 
 
 class DeepPot(DeepEval):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
```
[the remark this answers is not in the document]
njzjz: Actually, this can be removed. I don't see the reason to add it.

deepmd/pt/model/descriptor/env_mat.py
wanghan-iapcm, line 1: please check the following idea should fix
https://github.com/deepmodeling/deepmd-kit/blob/0517b593066b7810563c207e59b8e0e088289f9c/deepmd/dpmodel/utils/env_mat.py

deepmd/infer/deep_pot.py
```diff
@@ -179,6 +183,8 @@ def eval(
         atomic_virial
             The atomic virial of the system, in shape (nframes, natoms, 9). Only returned
             when atomic is True.
+        hessian
```
njzjz, line 186: The returned type annotations in lines 120 and 134 should be updated.

deepmd/entrypoints/test.py
```diff
@@ -531,8 +543,33 @@ def test_ener(
...
+                (
+                    data_h_triu,
+                    pred_h_triu,
+                ),
+                axis=1,
+            )
+            save_txt_file(
+                detail_path.with_suffix(".h.out"),
+                h,
+                header=f"{system}: data_h pred_h",
```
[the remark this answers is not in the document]
1azyking: Since the shape of Hessian matrices vary from frames with different numbers of atoms,  the Hessian matrices can only be reshaped to `(-1, )` before being printed to the `detail.h.out` file. Therefore, the header needs to discard the information of the order of elements and retain only two columns, namely `data_h` and `pred_h`. 
For example, testing two systems with numbers of atoms of 15 and 18 respectively, using `--numb-test 1`, the shape of `detail.h.out` is `(4941, 2)`, where 4941=(15\*3)^2+(18\*3)^2.

deepmd/entrypoints/test.py
```diff
@@ -531,8 +543,33 @@ def test_ener(
             "pred_vyy pred_vyz pred_vzx pred_vzy pred_vzz",
             append=append_detail,
         )
+        if dp.has_hessian:
+            _n_frames_, _n_hessian_ = test_data["hessian"][:numb_test].shape
+            _n_atoms_ = np.int32(np.sqrt(_n_hessian_) / 3)  # n_hessian = 3na*3na
+            triu_indices = np.triu_indices(
```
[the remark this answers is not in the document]
1azyking: Good suggestion. The full Hessian matrices would be returned.

deepmd/pt/train/training.py
1azyking: `EnergyHessianStdLoss` takes the similar form of `EnergyStdLoss`. Both of them utilize `ener` as the type of `fitting_net`. As described in the corresponding [documentation](https://github.com/1azyking/deepmd-kit/blob/enerhess/doc/model/train-energy-hessian.md), one who wants to train with Hessian can simply add the parameters `start_pref_h` and `limit_pref_h` to the `loss` section of the input file, without modifying the `type` of `fitting_net` or `loss` section. Therefore, in the function `get_model_for_wrapper`, `_loss_params` is expected to be passed to check for the valid `pref_h` parameters, and it is not possible to distinguish between `EnergyModel` and `EnergyHessianModel` solely from `_model_params`. In my view, to pass the `_loss_params` might be the simplest method to get `EnergyHessianModel`.

deepmd/pt/train/training.py
1azyking: plz refer to my reply in another conversation in this script.

deepmd/infer/deep_pot.py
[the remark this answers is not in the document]
1azyking: These two lines have been removed.

deepmd/pt/loss/ener.py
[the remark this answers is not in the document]
1azyking: UT of `EnergyHessianStdLoss` has been added in [test_loss.py](https://github.com/1azyking/deepmd-kit/blob/0256c0b33ce3ef38eda2a97d3b80f26d31c7054b/source/tests/pt/test_loss.py#L605).

deepmd/pt/model/descriptor/env_mat.py
1azyking: The modification from the above link has been applied to [deepmd/dpmodel/utils/env_mat.py](https://github.com/1azyking/deepmd-kit/blob/90d2468042dfaf0adeadc87142d8fcbf254d386b/deepmd/dpmodel/utils/env_mat.py#L65).

deepmd/infer/deep_pot.py
1azyking: `Optional[np.ndarray]` has been added to the tail of returned tuple.

deepmd/pt/infer/deep_eval.py
```diff
@@ -130,7 +130,8 @@ def __init__(
                         ] = state_dict[item].clone()
                 state_dict = state_dict_head
             model = get_model(self.input_param).to(DEVICE)
-            model = torch.jit.script(model)
+            if "Hessian" not in str(type(model)):
+                model = torch.jit.script(model)
```
njzjz, line 134: Please try [`torch.jit.ignore`](https://pytorch.org/docs/stable/generated/torch.jit.ignore.html) and see if it works rather than making a special usage here.

deepmd/pt/infer/deep_eval.py
```diff
@@ -159,6 +160,11 @@ def __init__(
         self._has_spin = getattr(self.dp.model["Default"], "has_spin", False)
         if callable(self._has_spin):
             self._has_spin = self._has_spin()
+        self._has_hessian = hasattr(self, "input_param") and self.input_param.get(
+            "hessian_mode", False
+        )
+        if callable(self._has_hessian):
+            self._has_hessian = self._has_hessian()
```
njzjz, line 167: I don't understand this

deepmd/infer/deep_pot.py
```diff
@@ -99,7 +100,9 @@ def eval(
         aparam: Optional[np.ndarray],
         mixed_type: bool,
         **kwargs: Any,
-    ) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+    ) -> tuple[
+        np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray, Optional[np.ndarray]
+    ]:
```
njzjz, line 105: It should be

```suggestion
    ) -> Union[tuple[
        np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray
    ], tuple[
        np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray
    ]]:
```

They have different meanings. `Optional[x]` is equivalent to `x | None`.

deepmd/infer/deep_pot.py
```diff
@@ -113,7 +116,7 @@ def eval(
         aparam: Optional[np.ndarray],
         mixed_type: bool,
         **kwargs: Any,
-    ) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+    ) -> tuple[np.ndarray, np.ndarray, np.ndarray, Optional[np.ndarray]]:
```
njzjz, line 119:
```suggestion
    ) -> Union[
        tuple[np.ndarray, np.ndarray, np.ndarray],
        tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray],
    ]:
```

deepmd/utils/argcheck.py
```diff
@@ -2284,6 +2440,20 @@ def loss_ener_spin():
...
+        #     default=0.00,
+        #     doc=doc_start_pref_h,
+        # ),
+        # Argument(
+        #     "limit_pref_h",
+        #     [float, int],
+        #     optional=True,
+        #     default=0.00,
+        #     doc=doc_limit_pref_h,
+        # ),
```
njzjz, line 2456: Unused codes should not be added

deepmd/pt/loss/ener.py
```diff
@@ -411,3 +411,75 @@ def label_requirement(self) -> list[DataRequirementItem]:
                 )
             )
         return label_requirement
+
+
+class EnergyHessianStdLoss(EnergyStdLoss):
```
njzjz, line 416: I am wondering why we cannot use the same loss class for `EnergyStdLoss` and `EnergyHessianStdLoss`, considering hession is derived from energy. @wanghan-iapcm what do you think?

deepmd/dpmodel/utils/env_mat.py
njzjz, line 1: This file has been updated by #4251.

deepmd/pt/model/descriptor/env_mat.py
```diff
@@ -21,12 +21,15 @@ def _make_env_mat(
...
     coord_r = coord_r.view(bsz, natoms, nnei, 3)
     diff = coord_r - coord_l
+    # avoid the possibility that coord[:, -1:, :] + rcut is the same as the coordinate of a real atom
+    diff = torch.where(torch.abs(diff) < 1e-30, torch.full_like(diff, 1e-30), diff)
     length = torch.linalg.norm(diff, dim=-1, keepdim=True)
```
njzjz, line 33: No need to hack the returned `diff`.

```suggestion
    diff_ = torch.where(torch.abs(diff) < 1e-30, torch.full_like(diff, 1e-30), diff)
    length = torch.linalg.norm(diff_, dim=-1, keepdim=True)
```

deepmd/pt/model/descriptor/env_mat.py
```diff
@@ -21,12 +21,15 @@ def _make_env_mat(
     nall = coord.shape[1]
     mask = nlist >= 0
     # nlist = nlist * mask  ## this impl will contribute nans in Hessian calculation.
-    nlist = torch.where(mask, nlist, nall - 1)
+    nlist = torch.where(mask, nlist, nall)
     coord_l = coord[:, :natoms].view(bsz, -1, 1, 3)
     index = nlist.view(bsz, -1).unsqueeze(-1).expand(-1, -1, 3)
-    coord_r = torch.gather(coord, 1, index)
+    coord_pad = torch.concat([coord, coord[:, -1:, :] + rcut], dim=1)
+    coord_r = torch.gather(coord_pad, 1, index)
```
njzjz, line 28: Are these changes still needed?

doc/model/train-energy-hessian.md
```diff
@@ -0,0 +1,50 @@
...
+      "start_pref_v": 0,
+      "limit_pref_v": 0,
+      "start_pref_h": 10,
+      "limit_pref_h": 1
+   },
+```
+
+The options {ref}`start_pref_e <loss[ener]/start_pref_e>`, {ref}`limit_pref_e <loss[ener]/limit_pref_e>`, {ref}`start_pref_f <loss[ener]/start_pref_f>`, {ref}`limit_pref_f <loss[ener]/limit_pref_f>`, {ref}`start_pref_v <loss[ener]/start_pref_v>` and {ref}`limit_pref_v <loss[ener]/limit_pref_v>` determine the start and limit prefactors of energy, force, and virial, respectively. The calculation and definition of Hessian loss are the same as for the other terms.
+
+If one does not want to train with virial, then he/she may set the virial prefactors {ref}`start_pref_v <loss[ener]/start_pref_v>` and {ref}`limit_pref_v <loss[ener]/limit_pref_v>` to 0.
```
njzjz, line 29: This doesn't match the actual code. Do we use `ener` or `ener_hess`?

doc/model/train-energy-hessian.md
njzjz, line 1: Please add a warning that the Hessian model cannot be frozen.

deepmd/pt/infer/deep_eval.py
```diff
@@ -159,6 +160,11 @@ def __init__(
         self._has_spin = getattr(self.dp.model["Default"], "has_spin", False)
         if callable(self._has_spin):
             self._has_spin = self._has_spin()
+        self._has_hessian = hasattr(self, "input_param") and self.input_param.get(
```
njzjz, line 163: please use `self.model_def_script`

njzjz (changes requested):

deepmd/pt/model/descriptor/env_mat.py
1azyking: Yes, or `nan` would occur in the second-order derivatives of nopbc system.

deepmd/pt/infer/deep_eval.py
1azyking: It has been removed.

doc/model/train-energy-hessian.md
1azyking: It is `ener_hess` that should be utilized here.

deepmd/pt/infer/deep_eval.py
1azyking: I tried using `torch.jit.ignore`, but it doesn't work as expected. The method `_cal_hessian_all` cannot be compiled by `torch.jit.script`. While applying `@torch.jit.ignore` to `_cal_hessian_all` allows `EnergyHessianModel` to pass the JIT script compilation, the issue arises when performing the forward pass. Specifically, when calculating the batch output, the `RecursiveScriptModule` throws a `RuntimeError` because it cannot simultaneously compute JIT-compiled energy and force outputs along with the JIT-ignored Hessian.

As a result, I did not apply `@torch.jit.ignore` to `_cal_hessian_all`. Instead, I opted for a `try-except` block when calling `torch.jit.script(model)`, which serves as a workaround for this issue and allows for the special usage scenario.

deepmd/pt/model/descriptor/env_mat.py
```diff
@@ -21,13 +21,16 @@ def _make_env_mat(
...
     coord_r = coord_r.view(bsz, natoms, nnei, 3)
     diff = coord_r - coord_l
-    length = torch.linalg.norm(diff, dim=-1, keepdim=True)
+    # avoid the possibility that coord[:, -1:, :] + rcut is the same as the coordinate of a real atom
+    diff_ = torch.where(torch.abs(diff) < 1e-30, torch.full_like(diff, 1e-30), diff)
+    length = torch.linalg.norm(diff_, dim=-1, keepdim=True)
```
iProzd, line 33: It's not necessary to do this, see comments above.

deepmd/pt/model/descriptor/env_mat.py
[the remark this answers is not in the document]
iProzd: It's not necessary to consider this because:
1. Whatever `coord[:, -1:, :] + rcut` is, the computed weight will be masked using nlist `mask`.
2. If very small possibility `coord[:, -1:, :] + rcut` is the same as the coordinate of a real atom, the computed length will then be added a 1.0 protection to avoid being zeros, see `length = length + ~mask.unsqueeze(-1)` below.

deepmd/pt/model/model/make_hessian_model.py
```diff
@@ -115,8 +115,8 @@ def forward_common(
 
         def _cal_hessian_all(
             self,
-            coord: torch.Tensor,
-            atype: torch.Tensor,
+            coord,
+            atype,
```
iProzd, line 119: why remove this

deepmd/pt/model/model/make_hessian_model.py
```diff
@@ -185,7 +184,7 @@ def __init__(
             self,
             obj: CM,
             ci: int,
-            atype: torch.Tensor,
+            atype,
```
iProzd, line 187: why remove this

iProzd (changes requested):

deepmd/pt/model/descriptor/env_mat.py
1azyking: It is returned to the former state.

deepmd/pt/model/model/make_hessian_model.py
1azyking: For consistency with lines in `make_model.py`; now recovered.

deepmd/pt/model/model/make_hessian_model.py
1azyking: For consistency with lines in `make_model.py`; now recovered.

iProzd (approved):

deepmd/entrypoints/test.py
```diff
@@ -303,6 +303,8 @@ def test_ener(
     if dp.has_spin:
         data.add("spin", 3, atomic=True, must=True, high_prec=False)
         data.add("force_mag", 3, atomic=True, must=False, high_prec=False)
+    if dp.has_hessian:
+        data.add("hessian", 1, atomic=True, must=True, high_prec=False)
```
njzjz, line 307: Why is the dimension `1` here? Does it really work? (I see it's not covered)

deepmd/pt/model/model/make_hessian_model.py
```diff
@@ -172,11 +172,10 @@ def _cal_hessian_one_component(
             # fparam: Optional[torch.Tensor] = None,  # nfp
             # aparam: Optional[torch.Tensor] = None,  # (nloc x nap)
             wc = wrapper_class_forward_energy(self, ci, atype, box, fparam, aparam)
-
             hess = torch.autograd.functional.hessian(
                 wc,
                 coord,
-                create_graph=False,
+                create_graph=True,
```
njzjz, line 178: Should it be `self.training`?

deepmd/utils/argcheck.py
```diff
@@ -2230,6 +2232,158 @@ def loss_ener():
...
+            "numb_generalized_coord",
+            int,
+            optional=True,
+            default=0,
+            doc=doc_numb_generalized_coord,
+        ),
+    ]
+
+
+@loss_args_plugin.register("ener_hess")
```
njzjz, line 2309: I don't get what the difference between `ener` and `ener_hess` is. It seems they are exactly the same? Is `loss_ener_hess` still used? If yes, Please add clear documentation to this argument via `doc="..."`,

deepmd/pt/model/model/ener_hess_model.py
```diff
@@ -0,0 +1,102 @@
...
+from .make_model import (
+    make_model,
+)
+
+DPEnergyModel_ = make_model(DPEnergyAtomicModel)
+DPEnergyModel_ = make_hessian_model(DPEnergyModel_)
+
+
+@BaseModel.register("ener_hess")
+class EnergyHessianModel(DPModelCommon, DPEnergyModel_):
```
wanghan-iapcm, line 33: I recommend we merging the `EnergyHessianModel` into `EnergyModel`, i.e. making a energy model by `make_hessian_model`, and control if the model outputs hessian by the method `requires_hessian`. 
- if `requires_hessian` is not set, the behavior of the returned object from `make_hessian_model` should be exactly the same as that from `make_model`.
- if  `requires_hessian` is set, the model gives hessian. 

The core idea is to control the model's output at **runtime**.

The advantage of the design is that we can easily switch from a standard energy model to hessian energy model and vice versa (imagine that we want to firstly train a model without hessian, and then finetune it with hessian), and do not need to maintain two implementations.

wanghan-iapcm (changes requested):

doc/model/train-energy-hessian.md
```diff
@@ -0,0 +1,54 @@
+# Fit energy Hessian {{ pytorch_icon }}
+
+:::{note}
+**Supported backends**: PyTorch {{ pytorch_icon }}
+:::
+
+:::{warning}
+The model trained with Hessian cannot be frozen.
```
wanghan-iapcm, line 8: The model parameters should be able to transferred to a standard energy model, which can be frozen. 

It would be of a great convenient if a user is allow to transfer a standard model to hessian model and vice versa.

wanghan-iapcm (changes requested):

deepmd/pt/model/model/ener_hess_model.py
wanghan-iapcm: I realized that the main difficulty is the jit. The hessian model is not jitable.

QuantumMisaka:
I'm much curious about this Hessian DP model
- where are the Hessian labels from? I've heard that they are from Phonopy by ABACUS/VASP, is this true？
- Is this Hessian model compatible with DPA-1/DPA-2 ?
- The most applicable scenario for the Hessian model is TS search, can the Hessian from DP model used by TS search method like Sella ? (which is used by NewtonNet and developed by same developers)

Thanks！

doc/model/train-energy-hessian.md
1azyking: Thanks for your suggestion. In the new version, it has been stated that "If freezing is forced, the model will be treated as a standard energy model, and the frozen model will no longer be able to output Hessian predictions."

deepmd/pt/model/model/ener_hess_model.py
1azyking: Good suggestions for simplifying and optimizing the code structure. Now the `ener_hess_model.py` is removed and class `EnergyHessianModel` is merged to class `EnergyModel` in `ener_model.py` with Hessian calculation activated by an additional attribute.

1azyking: Since the Hessian model is not specified separately, the `jit` cannot be optionally declared during the model definition, so the `torch.script.jit` in `deep_eval.py` will look less elegant.

deepmd/utils/argcheck.py
1azyking: Thanks for the heads-up. The `ener_hess` is now unused and removed.

deepmd/pt/model/model/make_hessian_model.py
1azyking: Sure. It has been changed into `create_graph=self.training`.

deepmd/entrypoints/test.py
1azyking: The Hessian matrix has a shape of `3N×3N`. If we set `ndof` to `3N`, the number of elements in each row would vary, which could complicate the conversion process later. On the other hand, if we output it as a `9×N^2` matrix, where `ndof` would be 9, the mapping between the matrix elements and atoms might become unclear. To simplify both the output data and the conversion between the Hessian matrix and atomic data, I decided to output the Hessian matrix in a flattened form with shape `9N^2×1` and set `ndof` to 1. This allows for a consistent left-to-right, top-to-bottom order in the matrix, ensuring a more straightforward relationship between the matrix elements and atoms.

1azyking:
> I'm much curious about this Hessian DP model
> 
> * where are the Hessian labels from? I've heard that they are from Phonopy by ABACUS/VASP, is this true？
> * Is this Hessian model compatible with DPA-1/DPA-2 ?
> * The most applicable scenario for the Hessian model is TS search, can the Hessian from DP model used by TS search method like Sella ? (which is used by NewtonNet and developed by same developers)
> 
> Thanks！


Thank you for your comment! Here's a response to your questions:

1. The Hessian labels can come from various sources. To the best of my knowledge:
* In VASP, you can calculate the second derivative matrix by setting the `IBRION` parameter.
* In Gaussian, adding the `freq` keyword enables the calculation of the Hessian matrix.
* In ABACUS, the Hessian matrix can be output by setting the `FULL_FORCE_CONSTANTS` parameter.
* In CP2K, the Hessian matrix can also be obtained by configuring the input file appropriately.

  Note that the Hessian-related output files from these programs often need to be processed using tools like `Phonopy` or `cclib` to extract the Hessian matrix.

2. The Hessian model is compatible with both DPA-1 and DPA-2. However, since we compute the Hessian using `torch.autograd.functional.hessian` method, the Hessian model currently supports only the **pytorch backend**.

3. The Hessian matrix has many applications, such as simulating vibrational-rotational spectra and performing transition state (TS) searches. The branch under development focuses on enabling the model to fit and predict Hessian matrices. Most of our adjustments are concentrated on training and testing functionalities (please refer to the [documentation](https://github.com/1azyking/deepmd-kit/blob/enerhess/doc/model/train-energy-hessian.md) for more details). The method you mentioned, such as Sella, may require further development for integration with the Hessian model.

If you have any additional questions, feel free to leave another comment!

deepmd/pt/train/training.py
```diff
@@ -1257,8 +1280,14 @@ def get_single_model(
...
 
-def get_model_for_wrapper(_model_params, resuming=False):
+def get_model_for_wrapper(
+    _model_params,
+    resuming=False,
+    _loss_params=None,
+):
     if "model_dict" not in _model_params:
+        if _loss_params is not None and whether_hessian(_loss_params):
+            _model_params["hessian_mode"] = True
```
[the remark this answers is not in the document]
1azyking: Now involved in [source/tests/pt/test_change_bias.py](https://github.com/1azyking/deepmd-kit/blob/enerhess/source/tests/pt/test_change_bias.py#L100).

doc/model/train-energy-hessian.md
iProzd, line 1: It's better to add tutorials with commands in the following situations:
1. users want to train hessian from scratch and freeze an energy model.
2. users want to train hessian from a pretrained energy model and freeze an energy model.

1azyking: Thx for your advice. The related contents have been added to the documentation.

njzjz (approved):

examples/hessian/single_task/input.json
```diff
@@ -0,0 +1,103 @@
...
+  "_comment": "that's all",
+  "model": {
+    "type_map": [
+      "C",
+      "H",
+      "N",
+      "O"
+    ],
+    "descriptor": {
+      "type": "dpa2",
```
wanghan-iapcm, line 11: the configuration of dpa2 looks outdated. please update with `example/water/dpa2`. 
I would recommend the `small` dpa2 configuration for a fast testing.

wanghan-iapcm (changes requested):

examples/hessian/single_task/input.json
1azyking: Thanks for reminding. It has been updated.

wanghan-iapcm (approved):

PR #4240, opened 2024-10-22
docs: document the floating-point precision of the model

njzjz:
## Summary by CodeRabbit

- **New Features**
	- Added a new section on `precision` in the documentation, enhancing navigation.
	- Introduced detailed guidelines on floating-point precision settings for the model.
	- Included structured instructions for creating models with the PyTorch backend.

- **Documentation**
	- Expanded troubleshooting documentation related to model precision issues, including data accuracy and training recommendations.
	- Enhanced guidelines for integrating new components into user configurations and ensuring model integrity across different backends.

### DIFF
--- a/doc/development/create-a-model-pt.md
+++ b/doc/development/create-a-model-pt.md
@@ -137,6 +137,15 @@ class SomeAtomicModel(BaseAtomicModel, torch.nn.Module):
         pass
 ```
 
+### Floating-point precision
+
+When creating a new component, the floating-point precision should obey the [Floating-point precision of the model](../model/precision.md) section.
+In implementation, the component should
+
+- store parameters in the component precision, except those for input and output normalization;
+- store normalization parameters in {py:data}`deepmd.pt.utils.env.GLOBAL_PT_FLOAT_PRECISION`;
+- after input normalization, cast the input tensor to the component precision; before output normalization, cast the output tensor to the {py:data}`deepmd.pt.utils.env.GLOBAL_PT_FLOAT_PRECISION`.
+
 ## Register new arguments
 
 To let someone uses your new component in their input file, you need to create a new method that returns some `Argument` of your new component, and then register new arguments. For example, the code below
--- a/doc/development/create-a-model-pt.md
+++ b/doc/development/create-a-model-pt.md
@@ -142,9 +142,9 @@ class SomeAtomicModel(BaseAtomicModel, torch.nn.Module):
 When creating a new component, the floating-point precision should obey the [Floating-point precision of the model](../model/precision.md) section.
 In implementation, the component should
 
-- store parameters in the component precision, except those for input and output normalization;
-- store normalization parameters in {py:data}`deepmd.pt.utils.env.GLOBAL_PT_FLOAT_PRECISION`;
-- after input normalization, cast the input tensor to the component precision; before output normalization, cast the output tensor to the {py:data}`deepmd.pt.utils.env.GLOBAL_PT_FLOAT_PRECISION`.
+- store parameters in the component precision, except those for output normalization;
+- store output normalization parameters in {py:data}`deepmd.pt.utils.env.GLOBAL_PT_FLOAT_PRECISION`;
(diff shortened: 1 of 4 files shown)


wanghan-iapcm (commented):
could you please also write a precision guide line for developers? 
for developers, shall we change all GLOBAL_XX_FLOAT_PRECISION to XX_INTERFACE_FLOAT_PRECISION ?

doc/model/precision.md
```diff
@@ -0,0 +1,15 @@
+# Floating-point precision of the model
+
+The following options control the precision of the model:
+
+- The environment variable {envvar}`DP_INTERFACE_PREC` controls the interface precision of the model, the descriptor, and the fitting, the precision of the environmental matrix, and the precision of the normalized parameters for the environmental matrix and the fitting output.
+- The training parameter {ref}`model[standard]/fitting_net[ener]/precision <precision>` controls the precision of neural networks in the descriptor and the fitting, and the subsequent operations after the output of neural networks.
```
iProzd, line 6: This ref seems not working. BTW, precisions in descriptor & type embedding(tf) should also be listed here?

njzjz: It seems it should be ```{ref}`precision <model[standard]/fitting_net[ener]/precision>````.

wanghan-iapcm:
> > could you please also write a precision guide line for developers?
> 
> What should we discuss for developers in addition to the users?

1. model parameters is stored in model's precision, except the bias and std. 
2. each forward method should be convert the input to the model's internal precision before any operation. 
3. one converts the output to interface precision, apply std and bias and then return.

doc/development/create-a-model-pt.md
```diff
@@ -137,6 +137,15 @@ class SomeAtomicModel(BaseAtomicModel, torch.nn.Module):
         pass
 ```
 
+### Floating-point precision
+
+When creating a new component, the floating-point precision should obey the [Floating-point precision of the model](../model/precision.md) section.
+In implementation, the component should
+
+- store parameters in the component precision, except those for input and output normalization;
```
wanghan-iapcm, line 145: do we need to store the input normalization in interface precision? current implementation does not:
https://github.com/wanghan-iapcm/deepmd-kit/blob/0c6b42a51c809141ed757639ebd82c781c9eda0e/deepmd/pt/model/descriptor/se_a.py#L423-L424
the inputs are converted to internal precision anyway and we do not expect large bias/std for inputs.

njzjz: In TF, it does:

https://github.com/deepmodeling/deepmd-kit/blob/02580c2fd9e1f35b7fb9b3e6320a67deb553f163/deepmd/tf/descriptor/se_a.py#L627-L641

wanghan-iapcm: these parameters will any way be cast to internal precision, which is potentially lower than global precision.

njzjz: I don't think it is cast in TF, so it's questionable what is the best way

https://github.com/deepmodeling/deepmd-kit/blob/02580c2fd9e1f35b7fb9b3e6320a67deb553f163/deepmd/tf/descriptor/se_a.py#L651-L664

https://github.com/deepmodeling/deepmd-kit/blob/02580c2fd9e1f35b7fb9b3e6320a67deb553f163/deepmd/tf/fit/ener.py#L549-L579
https://github.com/deepmodeling/deepmd-kit/blob/02580c2fd9e1f35b7fb9b3e6320a67deb553f163/deepmd/tf/fit/ener.py#L598-L602

In PT, it's confusing that fparam and aparam are not cast, although the normalization parameters are stored in the fitting precision

https://github.com/deepmodeling/deepmd-kit/blob/02580c2fd9e1f35b7fb9b3e6320a67deb553f163/deepmd/pt/model/task/fitting.py#L419-L442

njzjz: https://github.com/deepmodeling/deepmd-kit/pull/4240/commits/5958e8f135b787e61ec4754080068025fa5fadfd

wanghan-iapcm (approved):

PR #3930, opened 2024-06-29
fix(tf): prevent `fitting_attr` variable scope from becoming `fitting_attr_1`

njzjz:
Fix #3928. Prevent `fitting_attr` from becoming `fitting_attr_1`.

## Summary by CodeRabbit

- **Refactor**
  - Improved TensorFlow variable scope management by switching to `tf.AUTO_REUSE` to streamline code and reduce the likelihood of variable reuse conflicts.

Issue #3928, opened 2024-06-29
Bug: Error when loading a model with the se_a_mask descriptor


wanghan-iapcm:
### Discussed in https://github.com/deepmodeling/deepmd-kit/discussions/3924

<div type='discussions-op-text'>

<sup>Originally posted by **lukasbaldauf** June 28, 2024</sup>
Hi All, 
I want to evaluate a trained model using the `se_a_mask` descriptor, but I'm encountering an error. The training goes smoothly and I get good accuracies for my system, I'm just having trouble loading the model. I get the same error when evaluating the trained zinc_protein example system (see the error message below). It seems like something related to "dfparam" and "daparam", where the Tensors are missing.

I get the same errors for deepmd versions 2.2.7 and 2.2.10, and tensorflow versions 2.9.0 and 2.15.0.

For the zinc example, I train and freeze the model as such:
```shell
dp train zinc_se_a_mask.json --skip-neighbor-stat
dp freeze -o graph.mask.pb
```
The problem occurs when I want to load the model:
```python
from deepmd.infer import DeepPot
model = DeepPot("graph.mask.pb")
```
---
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/lukasb/miniforge3/envs/deepmd_gpu_2.2.7/lib/python3.10/site-packages/deepmd/infer/deep_pot.py", line 156, in __init__
    self._get_tensor(tensor_name, attr_name)
  File "/home/lukasb/miniforge3/envs/deepmd_gpu_2.2.7/lib/python3.10/site-packages/deepmd/infer/deep_eval.py", line 165, in _get_tensor
    tensor = self.graph.get_tensor_by_name(tensor_path)
  File "/home/lukasb/miniforge3/envs/deepmd_gpu_2.2.7/lib/python3.10/site-packages/tensorflow/python/framework/ops.py", line 4128, in get_tensor_by_name
    return self.as_graph_element(name, allow_tensor=True, allow_operation=False)
  File "/home/lukasb/miniforge3/envs/deepmd_gpu_2.2.7/lib/python3.10/site-packages/tensorflow/python/framework/ops.py", line 3952, in as_graph_element
    return self._as_graph_element_locked(obj, allow_tensor, allow_operation)
  File "/home/lukasb/miniforge3/envs/deepmd_gpu_2.2.7/lib/python3.10/site-packages/tensorflow/python/framework/ops.py", line 3992, in _as_graph_element_locked
[...]

PR #4243, opened 2024-10-23
fix(tf): fix compress suffix in DescrptDPA1Compat

### LINKED ISSUE #4114: [BUG] fix `enable_compression` in TF `DescrptDPA1Compat`
As discussed with @cherryWangY, the `enable_compression` in TF `DescrptDPA1Compat` (directly inherited from `DescrptSeAtten`) doesn't consider the suffix of the type embedding. The type embedding in the original `DescrptSeAtten` doesn't contain a suffix, but that in `DescrptDPA1Compat` does.

https://github.com/deepmodeling/deepmd-kit/blob/b6e27e4fbdd27ef63951d95729a75d638062d0ef/deepmd/tf/descriptor/se_atten.py#L2223

Users may not touch this issue, as a descriptor usually has no suffix.

iProzd:
Fix #4114 .

## Summary by CodeRabbit

- **New Features**
	- Enhanced compression capabilities in descriptor models with new optional parameters for improved flexibility.
	- Improved serialization processes for attention layers, allowing for better handling of scaling factors and normalization.
	- Dynamic tensor name construction in utility functions to accommodate varying suffixes.

- **Bug Fixes**
	- Adjusted method parameters to ensure compatibility and functionality with new suffix options.

- **Tests**
	- Introduced a new test suite to validate the functionality of the TensorFlow-based descriptor model, ensuring consistent output with the updated features.

### DIFF
--- a/deepmd/tf/descriptor/se_atten.py
+++ b/deepmd/tf/descriptor/se_atten.py
@@ -423,6 +423,7 @@ class DescrptSeAtten(DescrptSeA):
         table_stride_2: float = 0.1,
         check_frequency: int = -1,
         suffix: str = "",
+        tebd_suffix: str = "",
     ) -> None:
         """Reveive the statisitcs (distance, max_nbor_size and env_mat_range) of the training data.
 
@@ -444,6 +445,8 @@ class DescrptSeAtten(DescrptSeA):
             The overflow check frequency
         suffix : str, optional
             The suffix of the scope
+        tebd_suffix : str, optional
+            The suffix of the type embedding scope, only for DescrptDPA1Compat
         """
         # do some checks before the mocel compression process
         assert (
@@ -496,7 +499,9 @@ class DescrptSeAtten(DescrptSeA):
             min_nbor_dist, table_extrapolate, table_stride_1, table_stride_2
         )
 
-        self.final_type_embedding = get_two_side_type_embedding(self, graph)
+        self.final_type_embedding = get_two_side_type_embedding(
+            self, graph, suffix=tebd_suffix
+        )
         type_side_suffix = get_extra_embedding_net_suffix(type_one_side=False)
         self.matrix = get_extra_side_embedding_net_variable(
             self, graph_def, type_side_suffix, "matrix", suffix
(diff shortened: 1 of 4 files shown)


deepmd/tf/utils/tabulate.py
```diff
@@ -126,7 +126,9 @@ def __init__(
         self.dstd = get_tensor_by_name_from_graph(
             self.graph, f"descrpt_attr{self.suffix}/t_std"
         )
-        self.ntypes = get_tensor_by_name_from_graph(self.graph, "descrpt_attr/ntypes")
+        self.ntypes = get_tensor_by_name_from_graph(
+            self.graph, f"descrpt_attr{self.suffix}/ntypes"
+        )
```
iProzd, line 131: It seems that we assumed all frozen models for compression has no `suffix` here before? @njzjz @wanghan-iapcm  
Also see https://github.com/deepmodeling/deepmd-kit/blob/devel/deepmd/tf/entrypoints/freeze.py#L109

wanghan-iapcm (approved):

deepmd/tf/descriptor/se_atten.py
```diff
@@ -558,17 +563,18 @@ def build(
...
-                dstd = np.ones([self.ntypes, self.ndescrpt])  # pylint: disable=no-explicit-dtype
+        with tf.variable_scope("descrpt_attr", reuse=reuse):
+            t_ntypes = tf.constant(self.ntypes, name="ntypes", dtype=tf.int32)
             t_rcut = tf.constant(
                 np.max([self.rcut_r, self.rcut_a]),
                 name="rcut",
                 dtype=GLOBAL_TF_FLOAT_PRECISION,
             )
-            t_ntypes = tf.constant(self.ntypes, name="ntypes", dtype=tf.int32)
+        with tf.variable_scope("descrpt_attr" + suffix, reuse=reuse):
```
njzjz, line 573: I am not sure the reason for changing the scope name, but I believe this is not what you expect: when a suffix is empty, the second variable_scope will create `descrpt_attr_1/xxx` tensor instead of `descrpt_attr/xxx`, which I believe is what you expect. See TensorFlow documentation: https://www.tensorflow.org/api_docs/python/tf/compat/v1/variable_scope
We had a similar issue in #3928, fixed in #3930.

njzjz (changes requested):

deepmd/tf/utils/tabulate.py
njzjz: We assumed the top descriptor has no suffix. This is true but this is not apply to the sub descriptor in a hybrid descriptor.

njzjz (approved):

PR #4246, opened 2024-10-23
feat(pt): support `use_aparam_as_mask` for pt backend

ChiahsinChu:
support `use_aparam_as_mask` for pt backend


source/tests/consistent/fitting/test_ener.py
```diff
@@ -88,7 +95,9 @@ def skip_pt(self) -> bool:
             precision,
             mixed_types,
             numb_fparam,
+            numb_aparam,
             atom_ener,
+            use_aparam_as_mask,
         ) = self.param
         return CommonTest.skip_pt
```
njzjz, line 102: Let's skip `use_aparam_as_mask and numb_aparam == 0` directly to reduce the test time.

```suggestion
        return CommonTest.skip_pt or (use_aparam_as_mask and numb_aparam == 0)
```

Please also apply to other methods.

njzjz: Alternatively, you can merge `numb_aparam` and `use_aparam_as_mask` into one matrix parameter, like below...

```py
    ("float64", "float32", "bfloat16"),  # precision
    (True, False),  # mixed_types
    (0, 1),  # numb_fparam
    ((0, False), (1, False), (1,True)),  # (numb_aparam, use_aparam_as_mask)
    ([], [-12345.6, None]),  # atom_ener
)
```

and unpack it 

```py
(
            resnet_dt,
            precision,
            mixed_types,
            numb_fparam,
            (numb_aparam, use_aparam_as_mask),
            atom_ener,
        ) = self.param
```

njzjz (changes requested):

njzjz (approved):

njzjz:
It seems the CI failed due to the test added previously on the certain GPU machine...

Issue #4249, opened 2024-10-24
[BUG] Wrong dptest results except for energy head


Chengqian-Zhang:
### Bug summary

I find property dptest's results to be very different from the learning curve. And I have found the reason. This is a general bug for all heads except for energy head.
In `/deepmd/entrypoints/test.py`  line 127 `tmap = dp.get_type_map() if isinstance(dp, DeepPot) else None`. If we use `DeepProperty` or `DeepPolar` or `DeepDOS`..... `tmap` is None. So  `type_map` is not modified again according to `type_map` in `input.json`. So `atype` is wrong in the model forward process. The model prediction value is wrong.

### DeePMD-kit Version

All versions

### Backend and its version

pytorch

### How did you download the software?

Offline packages

### Input Files, Running Commands, Error Log, etc.

See above.

### Steps to Reproduce

See above.

### Further Information, Files, and Links

See above.

njzjz:
It seems that in the `r2` branch, `get_type_map()` is only available in `DeepPot`. After we refactor `DeepEval` in v3, this should not be a problem anymore.

PR #4280, opened 2024-10-30
fix(dptest): Wrong dptest results except for energy head

Chengqian-Zhang:
Solve issue #4249

In `/deepmd/entrypoints/test.py`  line 127 `tmap = dp.get_type_map() if isinstance(dp, DeepPot) else None`. If we use `DeepProperty` or `DeepPolar` or `DeepDOS`..... `tmap` is None. So  `type_map` is not modified again according to `type_map` in `input.json`. So `atype` is wrong in the model forward process. The model prediction value is wrong.

According to @njzjz , It seems that in the `r2` branch, `get_type_map()` is only available in `DeepPot`. After we refactor `DeepEval` in v3, this should not be a problem anymore.

I also change the order of `type_map` in UT to ensure that when `type_map` of `input.json` doesn't match the `type_map ` of data, the result of the dptest is still correct.

## Summary by CodeRabbit

- **New Features**
	- Introduced warning logs for unsupported `DeepGlobalPolar` model usage, recommending the `DeepPolar` model instead.
  
- **Bug Fixes**
	- Simplified logic for obtaining the type map, ensuring consistent retrieval from the updated source.
	- Adjusted model configuration in tests to influence type interpretation.

- **Documentation**
	- Improved clarity of comments and logging statements for better understanding.

### DIFF
--- a/deepmd/entrypoints/test.py
+++ b/deepmd/entrypoints/test.py
@@ -124,7 +124,7 @@ def test(
         log.info(f"# testing system : {system}")
 
         # create data class
-        tmap = dp.get_type_map() if isinstance(dp, DeepPot) else None
+        tmap = dp.get_type_map()
         data = DeepmdData(
             system,
             set_prefix="set",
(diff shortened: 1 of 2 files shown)


Chengqian-Zhang:
image

njzjz (approved):

njzjz:
> Maybe who can help me rerun the failed UT? I do not have the permission. I don't think it is a code problem.

Fixed.

iProzd (approved):

wanghan-iapcm (approved):

PR #4283, opened 2024-10-30
Print the reminder for the illegal memory error in the AutoBatchSize under tf

Yi-FanLi:
#3822 added a reminder for the illegal memory error. However, this reminder is only needed for tf. This PR moves the illegal memory reminder from base class AutoBatchSize to the inherited class under tf.

## Summary by CodeRabbit

- **New Features**
	- Enhanced `AutoBatchSize` class to initialize batch size from an environment variable, improving user guidance on memory management with TensorFlow.
- **Bug Fixes**
	- Removed redundant logging during initialization to streamline the process when GPU resources are available.

### DIFF
--- a/deepmd/tf/utils/batch_size.py
+++ b/deepmd/tf/utils/batch_size.py
@@ -1,4 +1,6 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
+import os
+import logging
 from packaging.version import (
     Version,
 )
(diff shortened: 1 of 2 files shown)


deepmd/tf/utils/batch_size.py
```diff
@@ -11,9 +13,22 @@
...
 )
 from deepmd.utils.batch_size import AutoBatchSize as AutoBatchSizeBase
+from deepmd.utils.batch_size import (
+    log,
+)
 
 
 class AutoBatchSize(AutoBatchSizeBase):
+    def __init__(self, initial_batch_size: int = 1024, factor: float = 2.0) -> None:
+        super().__init__(initial_batch_size, factor)
```
njzjz, line 23: only show on gpus?

Yi-FanLi: done

njzjz: Thanks.

njzjz (approved):

PR #4298, opened 2024-11-01
Add compression API to BaseModel and AtomicModel

cherryWangY:
## Summary by CodeRabbit

- **New Features**
	- Introduced `enable_compression` method across multiple classes to allow configuration of compression settings for descriptors.
	- Enhanced robustness of output definitions and serialization processes in the `DPAtomicModel` class.
	- Added `enable_compression` method to `LinearEnergyAtomicModel` for improved model compression capabilities.

- **Bug Fixes**
	- Improved error handling in the `fitting_output_def` method to ensure fallback functionality when the fitting network is unavailable.

These updates enhance the functionality and reliability of the model management features.

PR #4300, opened 2024-11-01
Add pt compress commad line

cherryWangY:
## Summary by CodeRabbit

- **New Features**
	- Introduced model compression functionality, allowing users to compress models directly via the command-line interface.
	- Added a new command option `"compress"` to trigger model compression.
	- Enhanced help messages and examples for the `"compress"` command to clarify usage with different backends.
	- Added a comprehensive JSON configuration file for model compression parameters.
	- Improved handling of compression parameters within descriptor classes for better organization and efficiency.

- **Bug Fixes**
	- Improved error handling for unsupported file formats during model loading.

- **Tests**
	- Introduced a suite of unit tests to evaluate the functionality of model compression, ensuring accuracy and performance across different configurations.
	- Enhanced tests for loading model parameters to ensure all required attributes are correctly handled.


deepmd/pt/entrypoints/main.py
njzjz, line 1: The command line `Supproted backend: TensorFlow` can be removed.

https://github.com/deepmodeling/deepmd-kit/blob/8355947ef052fd3806a361e25bbcff18f3e2f627/deepmd/main.py#L427

njzjz, line 1: Add an example for `dp --pt`:

https://github.com/deepmodeling/deepmd-kit/blob/8355947ef052fd3806a361e25bbcff18f3e2f627/deepmd/main.py#L431-L433

njzjz, line 1: The default value for `input` and `output` can remove `.pb` (making the argument has a filename prefix). However, the suffix should be added in the backend.

https://github.com/deepmodeling/deepmd-kit/blob/8355947ef052fd3806a361e25bbcff18f3e2f627/deepmd/main.py#L437-L450

An example is `dp freeze`

https://github.com/deepmodeling/deepmd-kit/blob/8355947ef052fd3806a361e25bbcff18f3e2f627/deepmd/main.py#L309-L315

In TF, PT backends, the suffix is added manually
https://github.com/deepmodeling/deepmd-kit/blob/8355947ef052fd3806a361e25bbcff18f3e2f627/deepmd/tf/entrypoints/main.py#L73-L75
https://github.com/deepmodeling/deepmd-kit/blob/8355947ef052fd3806a361e25bbcff18f3e2f627/deepmd/pt/entrypoints/main.py#L539

njzjz, line 1: I suggest moving the `compress` method into a new file. The current file becomes larger and larger.

njzjz, line 1: We need an end-to-end test for the command line interface after #4298 is merged.

deepmd/pt/entrypoints/main.py
```diff
@@ -549,6 +552,16 @@ def main(args: Optional[Union[list[str], argparse.Namespace]] = None):
             model_branch=FLAGS.model_branch,
             output=FLAGS.output,
         )
+    elif FLAGS.command == "compress":
+        FLAGS.input = str(Path(FLAGS.input).with_suffix(".pth"))
+        FLAGS.output = str(Path(FLAGS.output).with_suffix(".pth"))
```
njzjz, line 557: Please do the same thing for TF.

deepmd/pt/entrypoints/main.py
cherryWangY: How should we perform end-to-end test for the command line interface. And we have removed the suffix of input and output, do we need change the old test for tf compression command line?

njzjz: Simulate the user behavior, i.e. calling the dp --pt compress will get the compressed model.
(I am not sure how to check whether it's compressed)

> do we need change the old test for tf compression command line?

It should not a breaking change for the TF backend.

njzjz: Simulate the user behavior, i.e. calling the dp --pt compress will get the compressed model.
(I am not sure how to check whether it's compressed)

> do we need change the old test for tf compression command line?

It should not a breaking change for the TF backend.

njzjz: Example:

https://github.com/deepmodeling/deepmd-kit/blob/bfbe2edfba582864029008bef2e3908470cff142/source/tests/pt/test_dp_show.py#L52

cherryWangY: I roughly understand the logic of the current test code, and it seems unlikely to use `dp --pt show` to test whether it's compressed. A naive idea is to still compare the forward results before and after compression, but I don't know how to do this.

njzjz: > A naive idea is to still compare the forward results before and after compression, but I don't know how to do this.

I guess you can copy `source/tests/tf/test_model_compression_se_a.py`, and change all TF-related things to PT. (for example, `dp` --> `dp --pt`, `.pb` --> `.pth`)

cherryWangY: My server is currently being reinstalled. Sorry I can't try this until tomorrow.

cherryWangY:
> File "/workspaces/deepmd-kit/source/tests/pt/test_model_compression_se_a.py", line 51, in _init_models
    ret = run_dp("dp --pt train " + INPUT)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AssertionError: Torch not compiled with CUDA enabled

It seems that `dp --pt train` need torch compiled with CUDA enable?
@njzjz

njzjz:
> > File "/workspaces/deepmd-kit/source/tests/pt/test_model_compression_se_a.py", line 51, in _init_models
> 
>     ret = run_dp("dp --pt train " + INPUT)
> 
>           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
>     AssertionError: Torch not compiled with CUDA enabled
> 
> 
> 
> It seems that `dp --pt train` need torch compiled with CUDA enable?
> 
> @njzjz 

We set the default device to a virtual one. It should be never used.

https://github.com/deepmodeling/deepmd-kit/blob/dabedd230cd4541750707a9accbf729c50325d86/source/tests/pt/__init__.py#L7

But it seems that this case is not touched before. I'll take a look.

cherryWangY:
Because we just need to test wheather `dp --pt compress` is worked, I just use `se_a` as an example to test the end-to-end flow. I think we don't need to repeat end-to-end testing for each descriptor, as their correctness has been proven in previous tests.

source/tests/pt/model_compression/input.json
```diff
@@ -0,0 +1,85 @@
...
+      "_comment7": "that's all"
+    },
+    "validation_data": {
+      "systems": [
+        "model_compression/data"
+      ],
+      "batch_size": 1,
+      "numb_btch": 3,
+      "_comment8": "that's all"
+    },
```
njzjz, line 75: I think it's also okay to just remove the validation data. This may cause the error.

source/tests/pt/test_model_compression_se_a.py
```diff
@@ -0,0 +1,579 @@
...
+import unittest
+
+import numpy as np
+
+from deepmd.env import (
+    GLOBAL_NP_FLOAT_PRECISION,
+)
+from deepmd.pt.infer.deep_eval import (
+    DeepEval,
+)
```
njzjz, line 13: Use deepmd.infer.deep_eval.DeepEval

source/tests/pt/model_compression/input.json
njzjz: It's okay for this PR, but I'll check whether it is not covered in other UTs.

cherryWangY:
`dp --pt compress` successfully called `model.enable_compression`, but the model type read from `freeze` is `EnergyModel`
> AttributeError: 'EnergyModel' object has no attribute 'enable_compression'

> print(type(model))
    <class 'deepmd.pt.model.model.ener_model.EnergyModel'>

@njzjz

njzjz:
> `dp --pt compress` successfully called `model.enable_compression`, but the model type read from `freeze` is `EnergyModel`
> 
> > AttributeError: 'EnergyModel' object has no attribute 'enable_compression'
> 
> 
> 
> > print(type(model))
> 
>     <class 'deepmd.pt.model.model.ener_model.EnergyModel'>
> 
> 
> 
> @njzjz 

Isn't EnergyModel a subclass of BaseModel?

cherryWangY:
> > `dp --pt compress` successfully called `model.enable_compression`, but the model type read from `freeze` is `EnergyModel`
> > > AttributeError: 'EnergyModel' object has no attribute 'enable_compression'
> > 
> > 
> > > print(type(model))
> > 
> > 
> > ```
> > <class 'deepmd.pt.model.model.ener_model.EnergyModel'>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @njzjz
> 
> Isn't EnergyModel a subclass of BaseModel?

Sorry I don't seem to have updated the latest version of the code

njzjz:
Does your branch contain #4298?

cherryWangY:
> /workspaces/deepmd-kit/deepmd/pt/model/model/make_model.py
    self.atomic_model.enable_compression(
                self.get_min_nbor_dist(),
                table_extrapolate,
                table_stride_1,
                table_stride_2,
                check_frequency,
            )
           
The `self.get_min_nbor_dist()` seems to get `NoneType` instead of correct distance. I remember u say `self.min_nbor_dist` will be calculated automatically
@njzjz

njzjz:
> The `self.get_min_nbor_dist()` seems to get `NoneType` instead of correct distance. I remember u say `self.min_nbor_dist` will be calculated automatically

Below are the related codes:
https://github.com/deepmodeling/deepmd-kit/blob/dabedd230cd4541750707a9accbf729c50325d86/deepmd/pt/entrypoints/main.py#L302-L352

I will check why it does not work. But a situation for `dp compress` is that if `min_nbor_dist` is not stored in the model (for example, the user may use `--skip-neighbor-stat` during training), then calculate it according to the data.

njzjz:
It looks to me PyTorch state_dict doesn't store everything...

njzjz:
I expect b186980cdcf32a05dd001de122cd41c0315d6725 to fix the issue.

njzjz:
> I expect [b186980](https://github.com/deepmodeling/deepmd-kit/commit/b186980cdcf32a05dd001de122cd41c0315d6725) to fix the issue.

I create a PR https://github.com/cherryWangY/deepmd-kit/pull/1 to merge this commit to your branch

cherryWangY:
> File "/workspaces/deepmd-kit/source/tests/pt/test_model_compression_se_a.py", line 53, in _init_models
    ret = run_dp("dp --pt freeze -o " + frozen_model)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/deepmd-kit/source/tests/pt/common.py", line 48, in run_dp
    main(cmds)
  File "/workspaces/deepmd-kit/deepmd/main.py", line 927, in main
    deepmd_main(args)
  File "/workspaces/deepmd-kit/.venv/lib/python3.12/site-packages/torch/distributed/elastic/multiprocessing/errors/__init__.py", line 355, in wrapper
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/workspaces/deepmd-kit/deepmd/pt/entrypoints/main.py", line 547, in main
    freeze(model=FLAGS.model, output=FLAGS.output, head=FLAGS.head)
  File "/workspaces/deepmd-kit/deepmd/pt/entrypoints/main.py", line 368, in freeze
    model = inference.Tester(model, head=head).model
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/deepmd-kit/deepmd/pt/infer/inference.py", line 65, in __init__
    self.wrapper.load_state_dict(state_dict)
  File "/workspaces/deepmd-kit/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 2584, in load_state_dict
    raise RuntimeError(
RuntimeError: Error(s) in loading state_dict for ModelWrapper:
        Unexpected key(s) in state_dict: "model.Default.min_nbor_dist".

@njzjz

njzjz:
I see https://github.com/pytorch/pytorch/issues/16675. Maybe change `self.register_buffer("min_nbor_dist", None)` to `self.register_buffer("min_nbor_dist", torch.tensor(-1.0, dtype=torch.float64, device=env.DEVICE))`? Then `-1.0` is a placeholder for no min_nbor_dist was calculated.

njzjz:
There seems to be other methods that need to change. I'll take a look tomorrow.

njzjz:
Could you push your changes? I can continue debugging based on it.

> Btw, the test take a long time to run on codespaces, nearly 20 minutes before the error appeared. I'm worried that this will prolong the entire test time.

You can only test your file like

```sh
pytest xx.py
```

njzjz:
Or do you mean a single test needs 20 min?

cherryWangY:
> Or do you mean a single test needs 20 min?

Yeah. I just run `python test_model_compression_se_a.py`.

> Could you push your changes? I can continue debugging based on it.

Sorry, I went to bed right after commenting last night.

cherryWangY:
```
Element info[0] - , Value: -1
Element info[1] - , Value: 15
Element info[2] - , Value: 75
Element info[3] - , Value: 0.01
Element info[4] - , Value: 0.10000000000000001
Element info[5] - , Value: -1
Element info[0] - , Value: -1
Element info[1] - , Value: 15
Element info[2] - , Value: 75
Element info[3] - , Value: 0.01
Element info[4] - , Value: 0.10000000000000001
Element info[5] - , Value: -1
~~~
```
This is the content of `info` I printed out. It looks like info is a list. 

If you want to output some debugging information, I suggest saving the output in a file, otherwise too much error information may exceed your terminal buffer and you will not be able to see your debugging information

njzjz:
> This is the content of `info` I printed out. It looks like info is a list.

I see, you need to store it as a tensor buffer

cherryWangY:
> > This is the content of `info` I printed out. It looks like info is a list.
> 
> I see, you need to store it as a tensor buffer

I don't understand what you mean, do you mean set `register_buffer` like `min_nbor_dist`?

njzjz:
> I don't understand what you mean, do you mean set `register_buffer` like `min_nbor_dist`?

Yes, only this can be saved in the frozen torchscript model

cherryWangY:
Where do I need to register `info`, `deepmd/dpmodel/descriptor/make_base_descriptor.py`?

And why do we save `info`? In `enable_compresssion()`, we get the following variables.
```python
  self.compress = True
  self.table_data = table_data
  self.table_config = table_config
  self.lower = lower
  self.upper = upper
```
If this variables are saved in the frozen torchscript model, I think `compress_forward()` can be excuted correctly. 
Moreover, `info` is not a fixed list, so I think it is not feasible to register and save it. What needs to be saved should be the variables in `enable_compression()`.

njzjz:
But you need to save the compressed model, right?

cherryWangY:
> But you need to save the compressed model, right?

yes

cherryWangY:
But `dp --pt compress` only calls `enable_compression()`, and does not call `forward()`. How can we save `info`?

njzjz:
> does not call `forward()`

`forward()` is saved in the TorchScript model. This is the so-called "script", which can be executed in C++.

Any variables required by `forward` should be saved in the model.

I think `register_buffer` can be used in `enable_compression`.

cherryWangY:
> Any variables required by `forward` should be saved in the model.

That's right.

cherryWangY:
```python
  lower: dict[str, int]
  upper: dict[str, int]
  table_data: dict[str, torch.Tensor]
  table_config: list[Union[int, float]]
```
If we register variables for these, we need convert them into tensors before `register_buffer`, which will lose the key value.

njzjz:
We can also use [`ParameterDict`](https://pytorch.org/docs/stable/generated/torch.nn.ParameterDict.html) if you feel it's more convenient. There's no difference, considering there's no gradient defined.

cherryWangY:
[`ParameterDict`](https://pytorch.org/docs/stable/generated/torch.nn.ParameterDict.html) can manage parameters more efficient. But my question is how to save the `dict`'s key, like `self.lower`. 

Moreover, the number of elements in `self.lower` is not fixed. It will change dynamically according to the descriptor, and its key value will also change. I think some of the above methods are difficult to save it.

njzjz:
I think dynamic keys are not a problem, since we don't need to initlize them in `__init__` in advance. Just add them in `enable_compression` and then save them.

cherryWangY:
> I think dynamic keys are not a problem, since we don't need to initlize them in `__init__` in advance. Just add them in `enable_compression` and then save them.

Oh, I see. We can use "lower + key" as new key.

cherryWangY:
> RuntimeError: 
Unable to extract string literal index. ModuleDict indexing is only supported with string literals. For example, 'i = "a"; self.layers[i](x)' will fail because i is not a literal. Enumeration of ModuleDict is supported, e.g. 'for k, v in self.items(): out = v(inp)':
  File "/workspaces/deepmd-kit/deepmd/pt/model/descriptor/se_a.py", line 777
                # ]
                info = [
                    self.compress_params["lower_"+net].item(),
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE

This is why I want to use `get`.

njzjz:
It seems to me `ParameterDict.get` is not supported by torchscript...

cherryWangY:
> deepmd/pt/utils/serialization.py:76: in deserialize_to_file
    model.min_nbor_dist = float(data["@variables"]["min_nbor_dist"])

cherryWangY:
> source/tests/infer/case.py:172: in get_model
    convert_backend(INPUT=self.filename, OUTPUT=out_file)
deepmd/entrypoints/convert_backend.py:27: in convert_backend
    out_hook(OUTPUT, data)
deepmd/pt/utils/serialization.py:76: in deserialize_to_file
    model.min_nbor_dist = float(data["@variables"]["min_nbor_dist"])
    TypeError: cannot assign 'float' as buffer 'min_nbor_dist' (torch.nn.Buffer, torch.Tensor or None expected)

change to
```python
model.min_nbor_dist = torch.tensor(float(data["@variables"]["min_nbor_dist"]), dtype=env.GLOBAL_PT_FLOAT_PRECISION, device=env.DEVICE)
```

njzjz:
> Btw, the test take a long time to run on codespaces, nearly 20 minutes before the error appeared. I'm worried that this will prolong the entire test time.

I see that the CI takes too much time, so you must resolve the tabluate time-consuming problem first.

deepmd/main.py
```diff
@@ -424,29 +424,30 @@ def main_parser() -> argparse.ArgumentParser:
...
+        default="frozen_model",
         type=str,
-        help="The original frozen model, which will be compressed by the code",
+        help="The original frozen model, which will be compressed by the code. Filename (prefix) of the input model file. TensorFlow backend: suffix is .pb; PyTorch backend: suffix is .pth",
     )
     parser_compress.add_argument(
         "-o",
         "--output",
-        default="frozen_model_compressed.pb",
+        default="compressed_model",
```
njzjz, line 448: I am going to revert it to `frozen_model_compressed`. The main reason is that DP-GEN has relied on the default value of `--output`...

https://github.com/deepmodeling/dpgen/blob/0f7e15c99d05c4bb83db7797956d611cbc1dc689/dpgen/generator/run.py#L853

```suggestion
        default="frozen_model_compressed",
```

deepmd/main.py
```diff
@@ -424,29 +424,30 @@ def main_parser() -> argparse.ArgumentParser:
...
-        help="(Supported backend: TensorFlow) compress a model",
+        help="Compress a model",
         formatter_class=RawTextArgumentDefaultsHelpFormatter,
         epilog=textwrap.dedent(
             """\
         examples:
             dp compress
-            dp compress -i graph.pb -o compressed.pb
+            dp compress -i graph -o compressed
+            dp --pt compress -i frozen_model -o compressed_model
```
njzjz, line 434: I prefer to give the whole filename. Note the input accepts either filename or filename prefix, not filename prefix only.

```suggestion
            dp --tf compress -i frozen_model.pb -o compressed_model.pb
            dp --pt compress -i frozen_model.pth -o compressed_model.pth
```

njzjz (changes requested):

njzjz (approved):
I am satisfied with the current state of this PR.

njzjz:
Wait for #4326 to be merged.

cherryWangY:
> Wait for #4326 to be merged.

OK, I'm troubled by the current error.

cherryWangY:
I think we need to update [dpa1.pth](https://github.com/deepmodeling/deepmd-kit/blob/devel/source/tests/pt/model/models/dpa1.pth)

PR #4301, opened 2024-11-01
feat(jax): zbl

njzjz:
## Release Notes

- **New Features**
	- Introduced new classes: `DPZBLLinearEnergyAtomicModel` and `PairTabAtomicModel`, enhancing atomic model functionalities.
	- Added `get_zbl_model` function for constructing `DPZBLModel` from input data.
	- Improved error handling in vector normalization with `safe_for_vector_norm` and `safe_for_sqrt`.

- **Bug Fixes**
	- Enhanced distance calculations in `format_nlist` to prevent NaN errors.

- **Documentation**
	- Updated comments and docstrings for clarity on recent changes.

- **Tests**
	- Enhanced test support for JAX backend in `test_zbl_ener.py`.

### DIFF
--- a/deepmd/dpmodel/model/make_model.py
+++ b/deepmd/dpmodel/model/make_model.py
@@ -503,14 +503,11 @@ def make_model(T_AtomicModel: type[BaseAtomicModel]):
                 index = ret.reshape(n_nf, n_nloc * n_nnei, 1).repeat(3, axis=2)
                 coord1 = xp.take_along_axis(extended_coord, index, axis=1)
                 coord1 = coord1.reshape(n_nf, n_nloc, n_nnei, 3)
-                # jax raises NaN error using norm
-                # but note: we don't actually need to sqrt here; the squared value is enough
-                # rr = xp.linalg.norm(coord0[:, :, None, :] - coord1, axis=-1)
-                rr2 = xp.sum(xp.square(coord0[:, :, None, :] - coord1), axis=-1)
-                rr2 = xp.where(m_real_nei, rr2, float("inf"))
-                rr2, ret_mapping = xp.sort(rr2, axis=-1), xp.argsort(rr2, axis=-1)
+                rr = xp.linalg.norm(coord0[:, :, None, :] - coord1, axis=-1)
+                rr = xp.where(m_real_nei, rr, float("inf"))
+                rr, ret_mapping = xp.sort(rr, axis=-1), xp.argsort(rr, axis=-1)
                 ret = xp.take_along_axis(ret, ret_mapping, axis=2)
-                ret = xp.where(rr2 > rcut * rcut, -1, ret)
+                ret = xp.where(rr > rcut, -1, ret)
                 ret = ret[..., :nnei]
             # not extra_nlist_sort and n_nnei <= nnei:
             elif n_nnei == nnei:


deepmd/dpmodel/model/make_model.py
```diff
@@ -475,11 +475,14 @@ def _format_nlist(
...
-                rr, ret_mapping = xp.sort(rr, axis=-1), xp.argsort(rr, axis=-1)
+                # jax raises NaN error using norm
+                # but note: we don't actually need to sqrt here; the squared value is enough
+                # rr = xp.linalg.norm(coord0[:, :, None, :] - coord1, axis=-1)
+                rr2 = xp.sum(xp.square(coord0[:, :, None, :] - coord1), axis=-1)
+                rr2 = xp.where(m_real_nei, rr2, float("inf"))
+                rr2, ret_mapping = xp.sort(rr2, axis=-1), xp.argsort(rr2, axis=-1)
                 ret = xp.take_along_axis(ret, ret_mapping, axis=2)
-                ret = xp.where(rr > rcut, -1, ret)
+                ret = xp.where(rr2 > rcut * rcut, -1, ret)
```
njzjz, line 485: I have to modify here to fix NaN forces. However, I don't quite understand why the energy has a gradient to the neighbor llist. Did I miss something?

deepmd/dpmodel/utils/nlist.py
```diff
@@ -215,30 +215,36 @@ def build_multiple_neighbor_list(
...
+    coord2 = xp_take_along_axis(coord1, index, axis=1)
+    coord2 = xp.reshape(coord2, (nb, nloc, nsel, 3))
     diff = coord2 - coord0[:, :, None, :]
-    rr = np.linalg.norm(diff, axis=-1)
-    rr = np.where(nlist_mask, float("inf"), rr)
+    # jax raises NaN error using norm
+    # but note: we don't actually need to sqrt here; the squared value is enough
+    # rr = xp.linalg.vector_norm(diff, axis=-1)
+    rr2 = xp.sum(xp.square(diff), axis=-1)
+    rr2 = xp.where(nlist_mask, xp.full_like(rr2, float("inf")), rr2)
```
njzjz, line 240: This is another place that raises NaN, but I don't understand why nlist matters.

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -943,7 +946,7 @@ def call(
         else:
             raise NotImplementedError
 
-        normed = xp.linalg.vector_norm(
+        normed = safe_for_vector_norm(
```
njzjz, line 949: I have to hack here

deepmd/dpmodel/model/make_model.py
njzjz: Anyway, I think avoiding sqrt also improves performance.

wanghan-iapcm: Interesting. I agree with you.

wanghan-iapcm: stop gradient at the nlist? https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.stop_gradient.html

wanghan-iapcm (commented):
the autodiff should not pass the nlist, see my comment on using the jax.lax.stop_gradient that may fix the issue.

deepmd/dpmodel/model/make_model.py
njzjz: I found I couldn't reproduce the original error after reverting this piece of code. Not sure the reason...

njzjz: But I think the reason of NaN is that the neighbor list is rebuilt in the linear atomic model.

njzjz: Ok, 84803b847e641cf5a688bf2b6b9c16f3632b25c1 should fix it. I am wondering whether nlist in pytorch also has the gradient.

wanghan-iapcm:
are you planning to revert the modification on the nlist?

njzjz:
> are you planning to revert the modification on the nlist?

Reverted.

anyangml (approved):

wanghan-iapcm (approved):

PR #4343, opened 2024-11-12
fix(dpmodel): fix precision

njzjz:
## Release Notes

- **New Features**
	- Introduced a new environment variable `DP_DTYPE_PROMOTION_STRICT` to enhance precision handling in TensorFlow tests.
	- Added a decorator `@cast_precision` to several descriptor classes, improving precision management during computations.
	- Updated JAX configuration to enable strict dtype promotion based on the new environment variable.
	- Enhanced serialization and deserialization processes to include precision attributes across multiple classes.

- **Bug Fixes**
	- Enhanced type handling and input processing in the `GeneralFitting` class for better output predictions.
	- Improved handling of atomic contributions and exclusions in the `BaseAtomicModel` class.
	- Addressed potential type mismatches during matrix operations in the `NativeLayer` class.

- **Chores**
	- Updated caching mechanisms in the testing workflow to ensure unique keys based on run parameters.

### DIFF
--- a/deepmd/dpmodel/common.py
+++ b/deepmd/dpmodel/common.py
@@ -125,7 +125,7 @@ def cast_precision(func: Callable[..., Any]) -> Callable[..., Any]:
     """A decorator that casts and casts back the input
     and output tensor of a method.
 
-    The decorator should be used in a classmethod.
+    The decorator should be used on an instance method.
 
     The decorator will do the following thing:
     (1) It casts input arrays from the global precision
@@ -201,8 +201,8 @@ def safe_cast_array(
 
     Parameters
     ----------
-    input : tf.Tensor
-        Input tensor
+    input : np.ndarray or None
+        Input array
     from_precision : str
         Array data type that is casted from
     to_precision : str
@@ -210,8 +210,8 @@ def safe_cast_array(
 
     Returns
     -------
-    tf.Tensor
-        casted Tensor
+    np.ndarray or None
+        casted array
     """
     if array_api_compat.is_array_api_obj(input):
         xp = array_api_compat.array_namespace(input)
--- a/deepmd/dpmodel/common.py
+++ b/deepmd/dpmodel/common.py
@@ -10,6 +10,7 @@ from typing import (
     Any,
     Callable,
     Optional,
+    overload,
 )
 
 import array_api_compat
@@ -177,9 +178,15 @@ def cast_precision(func: Callable[..., Any]) -> Callable[..., Any]:
     return wrapper
 
 
+@overload
 def safe_cast_array(
     input: np.ndarray, from_precision: str, to_precision: str
-) -> np.ndarray:
+) -> np.ndarray: ...
+@overload
+def safe_cast_array(input: None, from_precision: str, to_precision: str) -> None: ...
+def safe_cast_array(
+    input: Optional[np.ndarray], from_precision: str, to_precision: str
+) -> Optional[np.ndarray]:
     """Convert an array from a precision to another precision.
 
     If input is not an array or without the specific precision, the method will not
--- a/deepmd/dpmodel/common.py
+++ b/deepmd/dpmodel/common.py
@@ -172,6 +172,11 @@ def cast_precision(func: Callable[..., Any]) -> Callable[..., Any]:
             return tuple(
                 safe_cast_array(vv, self.precision, "global") for vv in returned_tensor
             )
+        elif isinstance(returned_tensor, dict):
+            return {
+                kk: safe_cast_array(vv, self.precision, "global")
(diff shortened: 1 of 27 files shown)


njzjz (commented):
I revert "skip FP32 for NP". The issue is that if dpmodel is skipped, the reference backend is fallback to TF, which is not expected...

deepmd/dpmodel/fitting/general_fitting.py
wanghan-iapcm, line 1: better to implement the type cast in the `call` method in `invar_fitting`, `dipole_fitting` and so on. 

the reason is that dipole fitting has extra operations on the output of low level `_call_common`, which are supposed to be operated in internal precision.

deepmd/dpmodel/common.py
```diff
@@ -116,6 +120,94 @@ def to_numpy_array(x: Any) -> Optional[np.ndarray]:
...
+                safe_cast_array(vv, self.precision, "global") for vv in returned_tensor
+            )
+        else:
+            return safe_cast_array(returned_tensor, self.precision, "global")
+
+    return wrapper
+
+
+def safe_cast_array(
+    input: np.ndarray, from_precision: str, to_precision: str
```
wanghan-iapcm, line 181: input may be `None`, thus the type hint should be `Optional[np.ndarray]`

deepmd/dpmodel/common.py
```diff
@@ -116,6 +120,94 @@ def to_numpy_array(x: Any) -> Optional[np.ndarray]:
...
+        # only convert tensors
+        returned_tensor = func(
+            self,
+            *[safe_cast_array(vv, "global", self.precision) for vv in args],
+            **{
+                kk: safe_cast_array(vv, "global", self.precision)
+                for kk, vv in kwargs.items()
+            },
+        )
+        if isinstance(returned_tensor, tuple):
```
wanghan-iapcm, line 170: handle the case of returning a `dict`, then the fitting modules can use the wrapper.

wanghan-iapcm (changes requested):

deepmd/dpmodel/common.py
njzjz: I don't think the fitting module can use the wrapper since the normalization is performed inside the fitting.

deepmd/dpmodel/fitting/general_fitting.py
njzjz: Not possible for the output. The normalization is inside `_call_common`.

njzjz: In the dipole fitting, `out @ gr` is after the output normalization.

deepmd/dpmodel/descriptor/dpa2.py
```diff
@@ -594,6 +595,7 @@ def init_subclass_params(sub_data, sub_class):
         self.rcut = self.repinit.get_rcut()
         self.ntypes = ntypes
         self.sel = self.repinit.sel
+        self.precision = precision
```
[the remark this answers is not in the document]
njzjz: Where is it assigned?

deepmd/dpmodel/fitting/general_fitting.py
```diff
@@ -364,6 +364,11 @@ def _call_common(
 
         """
         xp = array_api_compat.array_namespace(descriptor, atype)
+        descriptor = xp.astype(descriptor, get_xp_precision(xp, self.precision))
```
wanghan-iapcm, line 367: it seems that gr, g2 and h2 are not cast.

njzjz: gr, g2 and h2 are not used below. I don't know why they are input.

deepmd/dpmodel/fitting/general_fitting.py
```diff
@@ -452,13 +447,14 @@ def _call_common(
...
+                xp.astype(self.bias_atom_e, outs.dtype), xp.reshape(atype, [-1]), axis=0
+            ),
             [nf, nloc, net_dim_out],
         )
         # nf x nloc
         exclude_mask = self.emask.build_type_exclude_mask(atype)
         # nf x nloc x nod
         outs = xp.where(exclude_mask[:, :, None], outs, xp.zeros_like(outs))
-        return {self.var_name: xp.astype(outs, get_xp_precision(xp, "global"))}
+        return {self.var_name: xp.astype(outs, descriptor.dtype)}
```
wanghan-iapcm, line 460:
```suggestion
        return {self.var_name: outs)}
```
not cast should work

wanghan-iapcm (approved):

iProzd (approved):

PR #4354, opened 2024-11-13
feat(pt): add universal test for loss

iProzd:
## Release Notes

- **New Features**
	- Introduced a new `LossTest` class for enhanced testing of loss functions.
	- Added multiple parameterized test functions for various loss functions in the new `test_loss.py` file.
  
- **Bug Fixes**
	- Corrected tensor operations in the `DOSLoss` class to ensure accurate cumulative sum calculations.

- **Documentation**
	- Added SPDX license identifiers to multiple files for clarity on licensing terms.

- **Chores**
	- Refactored data conversion methods in the `PTTestCase` class for improved handling of tensors and arrays.

### DIFF
--- a/source/tests/universal/common/cases/loss/utils.py
+++ b/source/tests/universal/common/cases/loss/utils.py
@@ -1,0 +1,73 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+import numpy as np
+
+from deepmd.utils.data import (
+    DataRequirementItem,
+)
+
+from .....seed import (
+    GLOBAL_SEED,
+)
+
+
+class LossTestCase:
+    """Common test case for loss function."""
+
+    def setUp(self):
+        pass
+
+    def test_label_keys(self):
+        module = self.forward_wrapper(self.module)
+        label_requirement = self.module.label_requirement
+        label_dict = {item.key: item for item in label_requirement}
+        label_keys = sorted(label_dict.keys())
+        label_keys_expected = sorted(
+            [key for key in self.key_to_pref_map if self.key_to_pref_map[key] > 0]
+        )
+        np.testing.assert_equal(label_keys_expected, label_keys)
+
+    def test_forward(self):
+        module = self.forward_wrapper(self.module)
+        label_requirement = self.module.label_requirement
+        label_dict = {item.key: item for item in label_requirement}
+        label_keys = sorted(label_dict.keys())
+        natoms = 5
+
+        def fake_model():
+            model_predict = {
+                data_key: fake_input_one_frame(label_dict[data_key], natoms)
+                for data_key in label_keys
+            }
+            if "atom_ener" in model_predict:
+                model_predict["atom_energy"] = model_predict.pop("atom_ener")
+            return model_predict
+
+        labels = {
+            data_key: fake_input_one_frame(label_dict[data_key], natoms)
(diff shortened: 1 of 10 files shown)


source/tests/universal/common/cases/loss/utils.py
```diff
@@ -0,0 +1,75 @@
...
+        natoms = 5
+
+        def fake_model():
+            model_predict = {
+                data_key: fake_input_one_frame(label_dict[data_key], natoms)
+                for data_key in label_keys
+            }
+            if "atom_ener" in model_predict:
+                model_predict["atom_energy"] = model_predict.pop("atom_ener")
+            model_predict.update({"mask_mag": np.ones([1, natoms, 1], dtype=np.bool_)})
```
wanghan-iapcm, line 44: we are using int32 for masks. 
 https://github.com/deepmodeling/deepmd-kit/blob/bb6b06b12955d3bf2b126e11e54f30fe2dd8bbb2/deepmd/dpmodel/atomic_model/base_atomic_model.py#L204

njzjz: Indeed I feel `bool` is more reasonable for 0-1 values. I have forgotten the reason why int32 is used...

source/tests/universal/pt/loss/test_loss.py
```diff
@@ -0,0 +1,47 @@
...
+@parameterized(
+    (
+        *[(param_func, EnergyStdLoss) for param_func in LossParamEnergyList],
+        *[(param_func, EnergySpinLoss) for param_func in LossParamEnergySpinList],
+        *[(param_func, DOSLoss) for param_func in LossParamDosList],
+        *[(param_func, TensorLoss) for param_func in LossParamTensorList],
+        *[(param_func, PropertyLoss) for param_func in LossParamPropertyList],
+    )  # class_param & class
+)
+class TestFittingPT(unittest.TestCase, LossTest, PTTestCase):
```
njzjz, line 39:
```suggestion
class TestLossPT(unittest.TestCase, LossTest, PTTestCase):
```

source/tests/universal/common/cases/loss/utils.py
```diff
@@ -0,0 +1,75 @@
...
+            return model_predict
+
+        labels = {
+            data_key: fake_input_one_frame(label_dict[data_key], natoms)
```
njzjz, line 48: Testing at least two frames is better.

source/tests/universal/common/cases/loss/utils.py
wanghan-iapcm: I agree. We may need a PR to clean up.

iProzd: The output mask `mask_mag` for virtual atoms in spin model is `bool`:
https://github.com/deepmodeling/deepmd-kit/blob/bb6b06b12955d3bf2b126e11e54f30fe2dd8bbb2/deepmd/pt/model/model/spin_model.py#L128

Maybe we keep this `bool` and clean up other int32 mask in another PR ?

njzjz (approved):

source/tests/universal/common/cases/loss/utils.py
njzjz: I agree

Issue #4358, opened 2024-11-14
[BUG] multi-task example does not work


ChiahsinChu:
### Bug summary

The example for multi-task does not work and gets an errors when calculating neighbour statistics. 

### DeePMD-kit Version

DeePMD-kit v2.2.11

### Backend and its version

TensorFlow v2.10.0

### How did you download the software?

Built from source

### Input Files, Running Commands, Error Log, etc.

Input files: `deepmd-kit/examples/water_multi_task/ener_dipole`

Running command:
```bash
dp train input.json
```
Error log:
```bash
DEEPMD INFO    Calculate neighbor statistics... (add --skip-neighbor-stat to skip this step)
Traceback (most recent call last):
  File "/home/jxzhu/apps/miniconda3/envs/deepmd-v2.2.11/bin/dp", line 8, in <module>
    sys.exit(main())
  File "/home/jxzhu/apps/deepmd/2.2.11/deepmd_utils/main.py", line 657, in main
    deepmd_main(args)
  File "/home/jxzhu/apps/deepmd/2.2.11/deepmd/entrypoints/main.py", line 74, in main
    train_dp(**dict_args)
  File "/home/jxzhu/apps/deepmd/2.2.11/deepmd/entrypoints/train.py", line 149, in train
    jdata = update_sel(jdata)
  File "/home/jxzhu/apps/deepmd/2.2.11/deepmd/entrypoints/train.py", line 512, in update_sel
    jdata_cpy["model"] = Model.update_sel(jdata, jdata["model"])
  File "/home/jxzhu/apps/deepmd/2.2.11/deepmd/model/model.py", line 566, in update_sel
    return cls.update_sel(global_jdata, local_jdata)
  File "/home/jxzhu/apps/deepmd/2.2.11/deepmd/model/multi.py", line 663, in update_sel
    local_jdata_cpy["descriptor"] = Descriptor.update_sel(
  File "/home/jxzhu/apps/deepmd/2.2.11/deepmd/descriptor/descriptor.py", line 510, in update_sel
    return cls.update_sel(global_jdata, local_jdata)
  File "/home/jxzhu/apps/deepmd/2.2.11/deepmd/descriptor/se.py", line 158, in update_sel
    return update_one_sel(global_jdata, local_jdata_cpy, False)
  File "/home/jxzhu/apps/deepmd/2.2.11/deepmd/entrypoints/train.py", line 479, in update_one_sel
    tmp_sel = get_sel(
[...]

PR #4360, opened 2024-11-15
fix bug: cannot set property `default_mesh` of `DeepmdDataSystem`

ChiahsinChu:
Currently, the multi-task cannot work correctly, and an error is thrown. (See #4358 for more details)
To fix this bug, a minor change in `DeepmdDataSystem` is made to make the property `default_mesh` editable. 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new attribute for caching default mesh values, improving efficiency.
	- Added functionality to set test size as a percentage of total structures.
	- New method for automated test size calculation based on specified percentages.

- **Deprecations**
	- Marked the `get_test` method as deprecated, shifting to a new approach for accessing test data.

### DIFF
--- a/deepmd_utils/utils/data_system.py
+++ b/deepmd_utils/utils/data_system.py
@@ -95,6 +95,7 @@ class DeepmdDataSystem:
         self.system_dirs = systems
         self.nsystems = len(self.system_dirs)
         self.data_systems = []
+        self._default_mesh = None
         for ii in self.system_dirs:
             self.data_systems.append(
                 DeepmdData(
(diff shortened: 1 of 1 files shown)


deepmd_utils/utils/data_system.py
```diff
@@ -233,12 +234,18 @@ def _load_test(self, ntests=-1):
     @lru_cache(maxsize=None)
```
njzjz, line 234: It's cached, so I am wondering whether the returned value can be changed.

ChiahsinChu: I think it works as expected. I have a simple test case:
```python
from typing import List
import numpy as np
from functools import lru_cache


class Tester:
    def __init__(self):
        self._default_mesh = None

    @property
    @lru_cache(maxsize=None)
    def default_mesh(self) -> List[np.ndarray]:
        """Mesh for each system."""
        if self._default_mesh is None:
            self._default_mesh = [np.ones(1), np.ones(1)]
        return self._default_mesh

    @default_mesh.setter
    def default_mesh(self, value: List[np.ndarray]):
        self._default_mesh = value


obj = Tester()
print(obj.default_mesh)
tmp_mesh = [np.zeros(1), np.zeros(1)]
obj.default_mesh += tmp_mesh
print(obj.default_mesh)
obj.default_mesh += tmp_mesh
print(obj.default_mesh)
```
which returns:
```text
[array([1.]), array([1.])]
[array([1.]), array([1.]), array([0.]), array([0.])]
[array([1.]), array([1.]), array([0.]), array([0.]), array([0.]), array([0.])]
```

njzjz: It seems that `+=` will not change the pointer to an object, so it's reasonable.

```py
>>> a=[]
>>> id(a)
140386313046464
>>> a+=[5]
>>> id(a)
140386313046464
```

njzjz (approved):

PR #4370, opened 2024-11-17
Fix: Atomic stat with multi-system

anyangml:
## Summary by CodeRabbit

- **New Features**
	- Enhanced model prediction process with improved batch handling.
	- Restructured output data processing for better compatibility during concatenation.
	- Added new test data files for water tensor dipole simulations.

- **Bug Fixes**
	- Improved error handling for file existence checks and bias restoration in statistics computation.

- **Documentation**
	- Updated function signatures to reflect internal logic modifications.
	- Expanded training data systems in configuration for water tensor training.


njzjz (approved):

wanghan-iapcm (approved):

wanghan-iapcm (commented):
Could you please add a UT for detecting the bug?

wanghan-iapcm (changes requested):
The data files are too large (.5M). 
The bug should be triggered with only one or two frames, thus please make it as small as possible to save the size of the package.

anyangml:
> The data files are too large (.5M). The bug should be triggered with only one or two frames, thus please make it as small as possible to save the size of the package.

Kept 2frames for each system.

PR #4400, opened 2024-11-21
fix(pt/dp): make strip more efficient

iProzd:
The strip methods are different between se_atten and se_t_tebd, it's not necessary to merge them.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new optional parameter `type_embedding` in various methods across descriptor classes to enhance handling of atomic types and embeddings.
	- Added a method `get_full_embedding` in the `TypeEmbedNet` class for easier access to complete embeddings.

- **Bug Fixes**
	- Improved error handling and assertions for the new `type_embedding` parameter in multiple classes to prevent runtime errors.

- **Documentation**
	- Updated method signatures and docstrings to reflect the addition of `type_embedding`.

### DIFF
--- a/deepmd/pt/model/network/network.py
+++ b/deepmd/pt/model/network/network.py
@@ -297,6 +297,20 @@ class TypeEmbedNet(nn.Module):
         return self.embedding(atype.device)[atype]
 
     def get_full_embedding(self, device: torch.device):
+        """
+        Get the type embeddings of all types.
+
+        Parameters
+        ----------
+        device : torch.device
+            The device on which to perform the computation.
+
+        Returns
+        -------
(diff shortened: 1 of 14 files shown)


njzjz (approved):

deepmd/pt/model/network/network.py
```diff
@@ -296,6 +296,9 @@ def forward(self, atype):
         """
         return self.embedding(atype.device)[atype]
 
+    def get_full_embedding(self, device: torch.device):
```
wanghan-iapcm, line 299: please add doc for the method.

wanghan-iapcm (approved):

PR #4417, opened 2024-11-25
feat(pt/dp): support case embedding and sharable fitting

iProzd:
## Summary by CodeRabbit

- **New Features**
	- Introduced a `set_case_embd` method across multiple atomic model classes to enhance case embedding functionality.
	- Added a `dim_case_embd` parameter in various fitting classes to support case-specific embedding dimensions.
	- Updated serialization methods to include `dim_case_embd` in the output.
	- Added a comprehensive JSON configuration for multitask models in water simulations.
	- Introduced a new function to validate case embedding dimensions in multi-task training.
	- Updated the `share_params` method in the `DescrptDPA2` class to streamline parameter sharing logic.

- **Bug Fixes**
	- Improved version compatibility checks in deserialization methods across several classes.

- **Documentation**
	- Enhanced documentation for multi-task training, emphasizing the transition to PyTorch and detailing configuration changes.

- **Tests**
	- Updated test cases to incorporate new parameters and configurations related to case embeddings.
	- Introduced new tests for multitask learning configurations.


deepmd/pt/model/atomic_model/linear_atomic_model.py
```diff
@@ -158,6 +158,14 @@ def get_model_rcuts(self) -> list[float]:
     def get_sel(self) -> list[int]:
         return [max([model.get_nsel() for model in self.models])]
 
+    def set_dataid(self, data_idx):
+        """
+        Set the data identification of this atomic model by the given data_idx,
+        typically concatenated with the output of the descriptor and fed into the fitting net.
+        """
+        for model in self.models:
+            model.set_dataid(data_idx)
```
anyangml, line 167: check if model has `set_dataid` method? seems pairtab model will raise `NotImplementedError`

examples/water_multi_task/pytorch_example/input_torch_sharefit.json
```diff
@@ -0,0 +1,155 @@
...
+        "numb_dataid": 2,
+        "_comment": " that's all"
+      },
+      "_comment": "that's all"
+    },
+    "model_dict": {
+      "water_1": {
+        "type_map": "type_map_all",
+        "descriptor": "dpa2_descriptor",
+        "fitting_net": "shared_fit_with_id:1"
```
wanghan-iapcm, line 76: confused by `"shared_fit_with_id:1"`
what does the `1` mean?
can one use a different id like `"fitting_net": "shared_fit_with_id:0"`? 
How is it related to the training dataset?

wanghan-iapcm (changes requested):
I find `dataid` is not self-explanatory. I would suggest something like `caseid` .

`numb_dataid` should be named as `numb_cases`

source/tests/pt/model/water/multitask_sharefit.json
```diff
@@ -0,0 +1,133 @@
...
+        "O",
+        "H",
+        "B"
+      ],
+      "my_descriptor": {
+        "type": "se_e2_a",
+        "sel": [
+          46,
+          92
+        ],
```
njzjz, line 14: Why is the size of `my_type_map` 3 and the size of `sel` 2?

deepmd/dpmodel/atomic_model/make_base_atomic_model.py
```diff
@@ -68,6 +68,14 @@ def get_sel(self) -> list[int]:
             """Returns the number of selected atoms for each type."""
             pass
 
+        @abstractmethod
+        def set_dataid(self, data_idx) -> None:
```
njzjz, line 72: please provide type annotation for `data_idx`

njzjz (changes requested):

deepmd/pt/model/atomic_model/linear_atomic_model.py
iProzd: Maybe that's more clear to rasie `NotImplementedError` inside pairtab model, rather than specifically check here?

source/tests/pt/model/water/multitask_sharefit.json
iProzd: Oh I copied from `pt/model/water/multitask.json` and it's a typo. And I found that the descriptor part will be replaced by other templates, so the settings here has no effect. https://github.com/iProzd/deepmd-kit/blob/share_fit_prod/source/tests/pt/test_multitask.py#L203-L205
Fixed.

deepmd/pt/model/task/dipole.py
```diff
@@ -103,6 +104,7 @@ def __init__(
             resnet_dt=resnet_dt,
             numb_fparam=numb_fparam,
             numb_aparam=numb_aparam,
+            numb_caseid=numb_caseid,
```
wanghan-iapcm, line 107: `numb_case` sounds more reasonable

wanghan-iapcm (changes requested):

deepmd/pt/model/task/dipole.py
iProzd: `numb_caseid` means the dimension of `caseid`, similar as `numb_fparam`.  `numb_caseid` == `numb_case` only when we use one-hot embedding, maybe we will use other methods in teh future. So I suggest to use `numb_caseid` here.

anyangml (approved):

deepmd/pt/model/task/dipole.py
wanghan-iapcm: Then it is more reasonable to directly set the encoded case, not the id of the case.

wanghan-iapcm: By your currently design, you set the index of a case, number of cases is more explanatory than number of caseid

iProzd: Renamed to `dim_case_embd`.

doc/train/multi-task-training.md
```diff
@@ -53,9 +53,21 @@ Specifically, there are several parts that need to be modified:
...
+      - Level 1: Shares type embedding only
+    - Not all descriptors support all levels (e.g., se_a only supports level 0)
+- - For fitting nets, we only support the default `shared_level`=0, where all parameters will be shared except for `bias_atom_e` and `case_embd`.
+- - To conduct multitask training, there are two typical approaches:
+    1. **Descriptor sharing only**: Share the descriptor with `shared_level`=0. See [here](../../examples/water_multi_task/pytorch_example/input_torch.json) for an example.
+    2. **Descriptor and fitting network sharing with data identification**:
+       - Share the descriptor and the fitting network with `shared_level`=0.
+       - {ref}`dim_case_embd <model[standard]/fitting_net[ener]/dim_case_embd>` must be set to the number of model branches, which will distinguish different data tasks using a one-hot embedding.
+       - See [here](../../examples/water_multi_task/pytorch_example/input_torch_sharefit.json) for an example.
+- - The parts that are exclusive to each model can be written following the previous definition.
```
njzjz, line 70: image

What's the reason to add two `-`?

njzjz (approved):

wanghan-iapcm (approved):

PR #4423, opened 2024-11-26
Fix: add model_def_script to ZBL

anyangml:
## Summary by CodeRabbit

- **New Features**
	- Enhanced model definition handling for improved encapsulation and consistency across different model types.

- **Bug Fixes**
	- Ensured that model definition scripts are correctly set to a JSON string representation for all model instances.


njzjz:
@iProzd is it possible to add some tests in the universal directory?

anyangml:
Pairtab model does not support compression. @njzjz

njzjz (approved):

njzjz:
> Pairtab model does not support compression. @njzjz 

This seems to be a different topic. My insight is that an empty method is enough.

PR #4426, opened 2024-11-26
perf: optimize training loop

caic99:
Improvements to the training process:

* [`deepmd/pt/train/training.py`](diffhunk://#diff-a90c90dc0e6a17fbe2e930f91182805b83260484c9dc1cfac3331378ffa34935R659): Added a check to skip setting the model to training mode if it already is. The profiling result shows it takes some time to recursively set it to all models.

* [`deepmd/pt/train/training.py`](diffhunk://#diff-a90c90dc0e6a17fbe2e930f91182805b83260484c9dc1cfac3331378ffa34935L686-L690): Modified the gradient clipping function to include the `error_if_nonfinite` parameter, and removed the manual check for non-finite gradients and the associated exception raising.

## Summary by CodeRabbit

- **New Features**
	- Improved training loop with enhanced error handling and control flow.
	- Updated gradient clipping logic for better error detection.
	- Refined logging functionality for training and validation results.

- **Bug Fixes**
	- Prevented redundant training calls by adding conditional checks.

- **Documentation**
	- Clarified method logic in the `Trainer` class without changing method signatures.

### DIFF
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -764,7 +764,7 @@ class Trainer:
             if self.display_in_training and (
                 display_step_id % self.disp_freq == 0 or display_step_id == 1
             ):
-                self.wrapper.eval() # Will set to train mode before fininshing validation
+                self.wrapper.eval()  # Will set to train mode before fininshing validation
 
                 def log_loss_train(_loss, _more_loss, _task_key="Default"):
                     results = {}
@@ -925,6 +925,7 @@ class Trainer:
                     writer.add_scalar(
                         f"{task_key}/{item}", more_loss[item], display_step_id
                     )
+
         self.wrapper.train()
         self.t0 = time.time()
         self.total_train_time = 0.0
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -684,7 +684,9 @@ class Trainer:
                 loss.backward()
                 if self.gradient_max_norm > 0.0:
                     grad_norm = torch.nn.utils.clip_grad_norm_(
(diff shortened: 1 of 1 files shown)


deepmd/pt/train/training.py
```diff
@@ -656,7 +656,8 @@ def step(_step_id, task_key="Default") -> None:
             # PyTorch Profiler
             if self.enable_profiler or self.profiling:
                 prof.step()
-            self.wrapper.train()
+            if not self.wrapper.training:
```
njzjz, line 659: I think it just checks the wrapper itself. Is there a situation that the wrapper is in the training mode but the model is in the eval mode?

njzjz: Btw, I don't think it should be put in the loop

caic99: > I think it just checks the wrapper itself. Is there a situation that the wrapper is in the training mode but the model is in the eval mode?

It seems that the status change only happens here
https://github.com/deepmodeling/deepmd-kit/blob/53bbbdbb890e14af79fc8e8a8de778ad3cc06750/deepmd/pt/train/training.py#L769
So I guess the answer should be no.

caic99: > Btw, I don't think it should be put in the loop

Agreed - maybe move it to the end of validation block?

njzjz: > move it to the end of validation block

Reasonable. Validation is not performed every step.

njzjz (approved):

caic99:
Before:
image

After:

image

PR #4432, opened 2024-11-27
Feat: add pairtab compression

anyangml:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new method `enable_compression` in the PairTabAtomicModel class, indicating that the model does not support compression settings.
  
- **Documentation**
	- Added docstring for the `enable_compression` method to clarify its purpose.


deepmd/pt/model/atomic_model/linear_atomic_model.py
```diff
@@ -574,6 +574,17 @@ def deserialize(cls, data) -> "DPZBLLinearEnergyAtomicModel":
...
+    def enable_compression(
+        self,
+        min_nbor_dist: float,
+        table_extrapolate: float = 5,
+        table_stride_1: float = 0.01,
+        table_stride_2: float = 0.1,
+        check_frequency: int = -1,
+    ) -> None:
+        """Pairtab model does not support compression."""
+        pass
```
njzjz, line 586: DPAtomicModel should be compressed. Only PairTab can be skipped.

njzjz (changes requested):

PR #4433, opened 2024-11-27
Perf: remove redundant checks on data integrity

caic99:
Systems are aggregated here
https://github.com/deepmodeling/deepmd-kit/blob/f343a3b212edab5525502e0261f3068c0b6fb1f6/deepmd/utils/data_system.py#L802
and later initialized here
https://github.com/deepmodeling/deepmd-kit/blob/f343a3b212edab5525502e0261f3068c0b6fb1f6/deepmd/utils/data_system.py#L809-L810

This process will instantiate `DeepmdData` class, and it will perform data integrity checks
https://github.com/deepmodeling/deepmd-kit/blob/e695a91ca6f7a1c9c830ab1c58b7b7a05db3da23/deepmd/utils/data.py#L80-L82

Besides, the checking process enumerates all items for all ranks, which is unnecessary and quite slow. So this PR removes this check.

## Summary by CodeRabbit

- **New Features**
	- Enhanced flexibility in defining test sizes by allowing percentage input for the `test_size` parameter.
	- Introduced a new method to automatically compute test sizes based on the specified percentage of total data.
	- Improved path handling to accept both string and Path inputs, enhancing usability.

- **Bug Fixes**
	- Improved error handling for invalid paths, ensuring users receive clear feedback when files are not found.

- **Deprecation Notice**
	- The `get_test` method is now deprecated, with new logic implemented for loading test data when necessary.

### DIFF
--- a/deepmd/utils/data_system.py
+++ b/deepmd/utils/data_system.py
@@ -28,9 +28,6 @@ from deepmd.utils.data import (
 from deepmd.utils.out_stat import (
     compute_stats_from_redu,
 )
-from deepmd.utils.path import (
-    DPPath,
-)
 
 log = logging.getLogger(__name__)
 
--- a/deepmd/utils/data_system.py
+++ b/deepmd/utils/data_system.py
@@ -753,7 +753,7 @@ def process_systems(systems: Union[str, list[str]]) -> list[str]:
         systems = expand_sys_str(systems)
     elif isinstance(systems, list):
         systems = systems.copy()
-    assert systems, "No systems provided"
+    return systems
     return systems
 
 
--- a/deepmd/utils/data_system.py
+++ b/deepmd/utils/data_system.py
@@ -100,6 +100,7 @@ class DeepmdDataSystem:
         del rcut
         self.system_dirs = systems
         self.nsystems = len(self.system_dirs)
+        assert self.nsystems > 0, "No systems provided"
         self.data_systems = []
         for ii in self.system_dirs:
             self.data_systems.append(
@@ -752,12 +753,7 @@ def process_systems(systems: Union[str, list[str]]) -> list[str]:
         systems = expand_sys_str(systems)
     elif isinstance(systems, list):
         systems = systems.copy()
-    help_msg = "Please check your setting for data systems"
-    # check length of systems
-    if len(systems) == 0:
-        msg = "Cannot find any valid data systems"
-        log.fatal(msg)
-        raise OSError(msg, help_msg)
+    assert systems, "No systems provided"
     return systems
 
 
--- a/deepmd/utils/data_system.py
+++ b/deepmd/utils/data_system.py
@@ -100,7 +100,8 @@ class DeepmdDataSystem:
         del rcut
         self.system_dirs = systems
         self.nsystems = len(self.system_dirs)
-        assert self.nsystems > 0, "No systems provided"
+        if self.nsystems <= 0:
+            raise ValueError("No systems provided")
         self.data_systems = []
         for ii in self.system_dirs:
             self.data_systems.append(
--- a/deepmd/utils/data_system.py
+++ b/deepmd/utils/data_system.py
@@ -754,7 +754,6 @@ def process_systems(systems: Union[str, list[str]]) -> list[str]:
     elif isinstance(systems, list):
         systems = systems.copy()
     return systems
-    return systems
 
 
 def get_data(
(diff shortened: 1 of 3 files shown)


deepmd/utils/data_system.py
```diff
@@ -758,20 +755,9 @@ def process_systems(systems: Union[str, list[str]]) -> list[str]:
     help_msg = "Please check your setting for data systems"
     # check length of systems
     if len(systems) == 0:
-        msg = "cannot find valid a data system"
+        msg = "Can not find any valid data systems"
```
njzjz, line 758:
```suggestion
        msg = "Cannot find any valid data systems"
```

njzjz:
> This process will instantiate `DeepmdData` class, and it will perform data integrity checks

I don't get your point. It seems one checks `type.raw` and the other one checks `type_map.raw`.

caic99:
> It seems one checks `type.raw` and the other one checks `type_map.raw`.

@njzjz What I mean is that those files are read in the init process of `DeepmdData`, including `type.raw` as shown above, and other files required like `type_map.raw`.
https://github.com/deepmodeling/deepmd-kit/blob/e695a91ca6f7a1c9c830ab1c58b7b7a05db3da23/deepmd/utils/data.py#L72-L79
Thus, the data integrity has been checked in this process.

caic99:
> Please also check the usage of `process_systems` below

Nice catch. 

The systems are instantiated at https://github.com/deepmodeling/deepmd-kit/blob/a852aa9f4611356b89880e4a5a06b15b29e4a3fb/deepmd/pt/entrypoints/main.py#L152-L153
https://github.com/deepmodeling/deepmd-kit/blob/a852aa9f4611356b89880e4a5a06b15b29e4a3fb/deepmd/pt/utils/dataloader.py#L90-L91
https://github.com/deepmodeling/deepmd-kit/blob/a852aa9f4611356b89880e4a5a06b15b29e4a3fb/deepmd/pt/utils/dataset.py#L28

So it'll be OK.
Merging those two functions into one might me a better choice, but maybe in another PR.

njzjz:
> > It seems one checks `type.raw` and the other one checks `type_map.raw`.
> 
> 
> 
> @njzjz What I mean is that those files are read in the init process of `DeepmdData`, including `type.raw` as shown above, and other files required like `type_map.raw`.
> 
> https://github.com/deepmodeling/deepmd-kit/blob/e695a91ca6f7a1c9c830ab1c58b7b7a05db3da23/deepmd/utils/data.py#L72-L79
> 
> Thus, the data integrity has been checked in this process.

I read data.py and I don't think it checks whether the given path exists or not.

caic99:
> I read data.py and I don't think it checks whether the given path exists or not.

https://github.com/deepmodeling/deepmd-kit/blob/e695a91ca6f7a1c9c830ab1c58b7b7a05db3da23/deepmd/utils/data.py#L668-L675

It will raise an exception here if some path does not exist.

njzjz:
> @njzjz It will raise an exception here if some path does not exist.

The error message seems to be less clear, which needs to be improved.

```pycon
>>> import h5py
>>> with h5py.File("abc.hdf5", "w") as f:
...     pass
...
>>> from deepmd.utils.data import DeepmdData
>>> DeepmdData("abc.hdf5#azsffd")
  File "<stdin>", line 1, in <module>
  File "/home/njzjz/codes/deepmd-kit/deepmd/utils/data.py", line 65, in __init__
    raise FileNotFoundError(f"No {set_prefix}.* is found in {sys_path}")
FileNotFoundError: No set.* is found in abc.hdf5#azsffd
>>> from deepmd.utils.data_system import process_systems
>>> process_systems(["abc.hdf5#azsffd"])
dir abc.hdf5#azsffd is not a valid dir
  File "<stdin>", line 1, in <module>
  File "/home/njzjz/codes/deepmd-kit/deepmd/utils/data_system.py", line 770, in process_systems
    raise OSError(msg, help_msg)
OSError: [Errno dir abc.hdf5#azsffd is not a valid dir] Please check your setting for data systems
```

njzjz:
> > The error message seems to be less clear, which needs to be improved.
> 
> 
> 
> @njzjz  You mean the error message from `process_systems`? Any suggestions?

I mean the error message from `DeepmdData`, when you remove the error messages from `process_systems`. It does not report the input path is missing.

deepmd/utils/data_system.py
```diff
@@ -103,6 +100,7 @@ def __init__(
         del rcut
         self.system_dirs = systems
         self.nsystems = len(self.system_dirs)
+        assert self.nsystems > 0, "No systems provided"
```
njzjz, line 103: `assert` should not be used; see https://docs.astral.sh/ruff/rules/assert/

njzjz (changes requested):

deepmd/utils/data_system.py
caic99: Make sense; I was not aware of it before.

njzjz (approved):

PR #4435, opened 2024-11-27
chore: refactor training loop

caic99:
## Summary by CodeRabbit

- **New Features**
	- Enhanced training loop to support multi-task training, allowing for more flexible model selection.

- **Improvements**
	- Streamlined `step` function to accept only the step ID, simplifying its usage.
	- Adjusted logging and model saving mechanisms for consistency with the new training flow.
	- Improved random seed management for enhanced reproducibility in data processing.
	- Enhanced error handling in data retrieval to ensure seamless operation during data loading.
	- Added type hints for better clarity in data loader attributes.

### DIFF
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -655,7 +655,7 @@ class Trainer:
         def step(_step_id, task_key="Default") -> None:
             if self.multi_task:
                 model_index = dp_random.choice(
-                    np.arange(self.num_model),
+                    np.arange(self.num_model, dtype=np.int_),
                     p=self.model_prob,
                 )
                 task_key = self.model_keys[model_index]
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -653,6 +653,12 @@ class Trainer:
             prof.start()
 
         def step(_step_id, task_key="Default") -> None:
+            if self.multi_task:
+                model_index = dp_random.choice(
+                    np.arange(self.num_model),
+                    p=self.model_prob,
+                )
+                task_key = self.model_keys[model_index]
             # PyTorch Profiler
             if self.enable_profiler or self.profiling:
                 prof.step()
@@ -929,24 +935,8 @@ class Trainer:
 
         self.t0 = time.time()
         self.total_train_time = 0.0
-        for step_id in range(self.num_steps):
-            if step_id < self.start_step:
-                continue
-            if self.multi_task:
-                chosen_index_list = dp_random.choice(
-                    np.arange(
-                        self.num_model, dtype=np.int32
-                    ),  # int32 should be enough for # models...
-                    p=np.array(self.model_prob),
-                    size=self.world_size,
-                    replace=True,
-                )
-                assert chosen_index_list.size == self.world_size
-                model_index = chosen_index_list[self.rank]
-                model_key = self.model_keys[model_index]
-            else:
-                model_key = "Default"
-            step(step_id, model_key)
+        for step_id in range(self.start_step, self.num_steps):
+            step(step_id)
             if JIT:
                 break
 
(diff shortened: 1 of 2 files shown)


deepmd/pt/train/training.py
```diff
@@ -653,6 +653,12 @@ def run(self) -> None:
             prof.start()
 
         def step(_step_id, task_key="Default") -> None:
+            if self.multi_task:
+                model_index = dp_random.choice(
+                    np.arange(self.num_model),
```
caic99, line 658:
```suggestion
                    np.arange(self.num_model, dtype=np.int_),
```

deepmd/pt/train/training.py
```diff
@@ -653,6 +653,12 @@ def run(self) -> None:
             prof.start()
 
         def step(_step_id, task_key="Default") -> None:
+            if self.multi_task:
+                model_index = dp_random.choice(
+                    np.arange(self.num_model, dtype=np.int_),
+                    p=self.model_prob,
+                )
```
njzjz, line 660: I see the hack for multiprocessing is removed. Have we ensured random seeds are different in different processes?

njzjz: See also https://numpy.org/doc/stable/reference/random/parallel.html

caic99: > Have we ensured random seeds are different in different processes?

Yes, by not setting the seed explicitly.

https://github.com/deepmodeling/deepmd-kit/blob/e695a91ca6f7a1c9c830ab1c58b7b7a05db3da23/deepmd/utils/random.py#L9
This initializes the generator with a random state.
For torch, the `seed()` method is never called.
For tensorflow it goes here
https://github.com/deepmodeling/deepmd-kit/blob/e695a91ca6f7a1c9c830ab1c58b7b7a05db3da23/deepmd/tf/entrypoints/train.py#L220-L225

njzjz: If the seed is not set explicitly, different processes may have the same seed.

njzjz: TF seed is also wrong.

caic99: > If the seed is not set explicitly, different processes may have the same seed.

`[np.random.RandomState](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState)` initializes seed from random generators if the seed is not provided. So I think it would be less likely for different processes having the same seed.

caic99: @njzjz Do you think we should mandatorily set a seed?

njzjz: > @njzjz Do you think we should mandatorily set a seed?

Indeed, it's surprising that the random seed set by the user is not used, meaning the training may not be reproducible.

deepmd/pt/utils/dataloader.py
```diff
@@ -50,6 +51,7 @@ def setup_seed(seed) -> None:
     torch.manual_seed(seed)
     torch.cuda.manual_seed_all(seed)
     torch.backends.cudnn.deterministic = True
+    dp_random.seed(seed)
```
njzjz, line 54: should also seed with rank, i.e. [rank, seed]; see the example at https://numpy.org/doc/stable/reference/random/parallel.html#sequence-of-integer-seeds.

caic99: Maybe in another PR fixing #4440?

caic99: Currently the seed has been set to _seed from config file_ + rank, which is OK for the basic need.

njzjz: > Currently the seed has been set to seed from config file + rank

I don't get this point. It's done by TF but not PT.

caic99: > Currently the seed has been set to _seed from config file_ + rank, which is OK for the basic need.

@njzjz here
https://github.com/deepmodeling/deepmd-kit/blob/e695a91ca6f7a1c9c830ab1c58b7b7a05db3da23/deepmd/pt/entrypoints/main.py#L141

njzjz (approved):

PR #4437, opened 2024-11-27
refactor: simplify dataset construction

caic99:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new function for dataset construction, enhancing data loading processes.
	- Added a method to improve pickling and unpickling capabilities for path handling classes.

- **Bug Fixes**
	- Updated summary printing to prevent redundant output during distributed training.

- **Refactor**
	- Simplified initialization of the BackgroundConsumer class.
	- Streamlined consumer thread and queue handling in the BufferedIterator class.


caic99:
@anyangml This PR adds a more detailed warning output in the case of dataset reading is throttled.

deepmd/pt/train/training.py
```diff
@@ -1053,47 +1053,22 @@ def save_model(self, save_path, lr=0.0, step=0) -> None:
...
+        data, dataloader = (
+            (self.training_data, self.training_dataloader)
+            if is_train
+            else (self.validation_data, self.validation_dataloader)
+        )
+        if data is None and not is_train:
+            return {}, {}, {}
+        if self.multi_task:
+            data = data[task_key]
+            dataloader = dataloader[task_key]
```
wanghan-iapcm, line 1065: It seems that your change is not equivalent to the original code (see original code line 1088-1089)
```suggestion
        if self.multi_task:
            data = data[task_key]
            dataloader = dataloader[task_key]
        if data is None and not is_train:
            return {}, {}, {}
```

deepmd/pt/train/training.py
```diff
@@ -1053,47 +1053,22 @@ def save_model(self, save_path, lr=0.0, step=0) -> None:
             checkpoint_files[0].unlink()
 
     def get_data(self, is_train=True, task_key="Default"):
-        if not self.multi_task:
-            if is_train:
-                try:
-                    batch_data = next(iter(self.training_data))
-                except StopIteration:
-                    # Refresh the status of the dataloader to start from a new epoch
-                    with torch.device("cpu"):
```
njzjz, line 1062: You may not pass the tests if you remove this line.

deepmd/utils/path.py
```diff
@@ -44,6 +44,9 @@ def __new__(cls, path: str, mode: str = "r"):
             raise FileNotFoundError(f"{path} not found")
         return super().__new__(cls)
 
+    def __getnewargs__(self):
+        return (self.path, self.mode)
```
njzjz, line 48: I don't see any promise in this parent class that `self.path` and `self.mode` will be defined, so it's not self-consistent.

njzjz: Related: https://stackoverflow.com/a/54814371/9567349

njzjz (changes requested):

deepmd/pt/train/training.py
caic99: Will fix. But why this is only required for single task training data, not for multi task, not for valid data?

njzjz: Those lines might not be covered.

njzjz: I am not 100% sure about this comment. Please check the UT result...

deepmd/pt/train/training.py
```diff
@@ -1053,47 +1053,22 @@ def save_model(self, save_path, lr=0.0, step=0) -> None:
...
+        if self.multi_task:
+            data = data[task_key]
+            dataloader = dataloader[task_key]
+        if data is None and not is_train:
+            return {}, {}, {}
+        try:
+            batch_data = next(iter(data))
+        except StopIteration:
+            # Refresh the status of the dataloader to start from a new epoch
+            data = BufferedIterator(iter(dataloader))
```
caic99, line 1070:
```suggestion
            with torch.device("cpu"):
                data = BufferedIterator(iter(dataloader))
```

deepmd/pt/train/training.py
caic99: > I am not 100% sure about this comment. Please check the UT result...

haha you are right

caic99:
I'm reverting changes to `training.py` for there are too many changes in one PR, making it hard to test.

njzjz (approved):

CaRoLZhangxy (approved):

deepmd/pt/utils/dataloader.py
```diff
@@ -202,70 +210,48 @@ def print_summary(
...
-                    self.warning_time = time.time()
-
-        # Get next example
+        start_wait = time.time()
         item = self._queue.get()
+        wait_time = time.time() - start_wait
+        if (
+            wait_time > 1.0
+        ):  # Even for Multi-Task training, each step usually takes < 1s
+            log.warning(f"Data loading is slow, waited {wait_time:.2f} seconds.")
```
wanghan-iapcm, line 252: It may happen that to many warning messages flushes to the screen. do you consider restrict the number of warnings like the original implementation?

caic99: make sense; patched.

caic99:
I've trained 100k steps from scratch. Since this PR only changes the warning mechanism of DataLoader, the loss curve does not change significantly. The deviation is not introduced in this PR - I've attached the curve of two runs on devel branch, and they are not exactly the same.

<details><summary>Run 1 on devel branch</summary>
<p>

image

</p>
</details> 

<details><summary>Run 2 on devel branch</summary>
<p>

image


</p>
</details> 

<details><summary>This PR</summary>
<p>

image


</p>
</details>

wanghan-iapcm (approved):

PR #4438, opened 2024-11-27
fix: unmark `eval_pd` as `abstractmethod`

njzjz:
Many classes don't have this method and become abstract classes. It looks to me like these tests are skipped silently and don't throw errors.

## Summary by CodeRabbit

- **Bug Fixes**
	- Updated the `eval_pd` method to provide a clearer indication of its implementation status by raising a `NotImplementedError`.
	- Changed the `skip_pd` class variable to always skip Paddle-related tests.

- **Tests**
	- Adjusted the `skip_pd` property in the `TestSeA` and `TestEner` classes to directly reflect the installation status of Paddle.

### DIFF
--- a/source/tests/consistent/common.py
+++ b/source/tests/consistent/common.py
@@ -185,7 +185,6 @@ class CommonTest(ABC):
         """
         raise NotImplementedError("Not implemented")
 
-    @abstractmethod
     def eval_pd(self, pd_obj: Any) -> Any:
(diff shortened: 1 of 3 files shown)


njzjz:
cc @HydrogenSulfate

iProzd (approved):

wanghan-iapcm (approved):

PR #4463, opened 2024-12-09
Perf: use fused Adam optimizer

caic99:
This PR sets the Adam optimizer to use the `fused=True` parameter. 
For the profiling result shown below, this modification brings an 2.75x improvement on optimizer update (22ms vs. 8ms) and ~3% improvement for total speed up (922ms vs. 892ms). The benchmark case is training a DPA-2 Q3 release model. Please note that the absolute time may differs between steps.

<details><summary>Before</summary>
<p>

image

</p>
</details> 

<details><summary>After</summary>
<p>

image

</p>
</details> 

[Ref](https://pytorch.org/docs/stable/generated/torch.optim.Adam.html):
> The foreach and fused implementations are typically faster than the for-loop, single-tensor implementation, with **fused being theoretically fastest** with both vertical and horizontal fusion. As such, if the user has not specified either flag (i.e., when foreach = fused = None), we will attempt defaulting to the foreach implementation when the tensors are all on CUDA. Why not fused? Since the fused implementation is relatively new, we want to give it sufficient bake-in time.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved optimizer performance during training by modifying the initialization of the Adam optimizer.

- **Documentation**
	- Updated method signature for clarity in the `Trainer` class.

### DIFF
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -579,7 +579,7 @@ class Trainer:
         # author: iProzd
         if self.opt_type == "Adam":
             self.optimizer = torch.optim.Adam(
-                self.wrapper.parameters(), lr=self.lr_exp.start_lr
+                self.wrapper.parameters(), lr=self.lr_exp.start_lr, fused=True
             )
             if optimizer_state_dict is not None and self.restart_training:
                 self.optimizer.load_state_dict(optimizer_state_dict)


caic99:
Running with CPU is still possible with this parameter set.

njzjz:
> For the profiling result shown below, this modification brings an 2.75x improvement on optimizer update (22ms vs. 8ms) and ~3% improvement for total speed up (922ms vs. 892ms). The benchmark case is training a DPA-2 Q3 release model. Please note that the absolute time may differs between steps.

It might be wrong to benchmark the time on the CPU since the GPU is asynchronous and is the bottleneck.

caic99:
> which is the bottleneck

@njzjz What you've mentioned is the general case for NN itself. However, for optimizer update in deepmd-kit, the bottleneck is at the CPU side. You can refer to the "GPU SM efficiency" row from the profiling plot.

njzjz (approved):

Issue #4105, opened 2024-09-05
Add universal/consistent tests for loss modules


njzjz:
> > > > ```diff
> > > > + Hits        42870    43087     +217     
> > > > - Misses       7822     7934     +112     
> > > > - Partials      946      948       +2     
> > > > ```
> > > 
> > > 
> > > I see that 1/3 of the new codes have not been tested.
> > 
> > 
> > Where can I see these datas about missing UT? I will add UT based on these datas.
> 
> You can click the link sent by @codecov or click the codecov checks.
> 
> It looks like the loss module is not tested. @iProzd Do we have a universal test fixture for loss functions?

Not yet, maybe we need a discussion to design a universal test for loss modules.

_Originally posted by @iProzd in https://github.com/deepmodeling/deepmd-kit/issues/3867#issuecomment-2331399624_

njzjz:
Now we have the universal tests but not the consistent tests.

PR #4466, opened 2024-12-10
feat (tf/pt): add atomic weights to tensor loss

ChiahsinChu:
Interfaces are of particular interest in many studies. However, the configurations in the training set to represent the interface normally also include large parts of the bulk material. As a result, the final model would prefer the bulk information while the interfacial information is less learnt. It is difficult to simply improve the proportion of interfaces in the configurations since the electronic structures of the interface might only be reasonable with a certain thickness of bulk materials. Therefore, I wonder whether it is possible to define weights for atomic quantities in loss functions. This allows us to add higher weights for the atomic information for the regions of interest and probably makes the model "more focused" on the region of interest. 
In this PR, I add the keyword `enable_atomic_weight` to the loss function of the tensor model. In principle, it could be generalised to any atomic quantity, e.g., atomic forces. 
I would like to know the developers' comments/suggestions about this feature. I can add support for other loss functions and finish unit tests once we agree on this feature. 

Best. 

## Summary by CodeRabbit

- **New Features**
	- Introduced an optional parameter for atomic weights in loss calculations, enhancing flexibility in the `TensorLoss` class.
	- Added a suite of unit tests for the `TensorLoss` functionality, ensuring consistency between TensorFlow and PyTorch implementations.

- **Bug Fixes**
	- Updated logic for local loss calculations to ensure correct application of atomic weights based on user input.

- **Documentation**
	- Improved clarity of documentation for several function arguments, including the addition of a new argument related to atomic weights.


wanghan-iapcm (changes requested):
Please provide the unit test for the feature. 

do you consider implementing it to the pt backend?

ChiahsinChu:
> Please provide the unit test for the feature.
> 
> do you consider implementing it to the pt backend?

In this PR, this feature is implemented for both tf and pt backend.

njzjz:
I am not sure if we need to wait for #4105.

njzjz (approved):
I approve the current PR, but I am not sure when #4105 can be ready.

wanghan-iapcm:
> I approve the current PR, but I am not sure when #4105 can be ready.

I think #4105 should not block the merge of this pr.

wanghan-iapcm (approved):

PR #4478, opened 2024-12-19
Perf: load data systems on rank 0

caic99:
The current implementation loads data on each rank. This will stress the file system.
In this PR, only rank 0 will load data systems, and it will be broadcasted to each rank.
The data sampler initialized later will still use the exclusive seed of each rank.

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of distributed data loading for improved synchronization across processes.
	- Added broadcasting of the constructed dataset to ensure consistency in all processes.

- **Bug Fixes**
	- Implemented safeguards to prevent incomplete data distribution by asserting the integrity of the dataset.

### DIFF
--- a/deepmd/pt/utils/dataloader.py
+++ b/deepmd/pt/utils/dataloader.py
@@ -97,7 +97,7 @@ class DpLoaderSet(Dataset):
         global_rank = dist.get_rank() if dist.is_initialized() else 0
         if global_rank == 0:
             log.info(f"Constructing DataLoaders from {len(systems)} systems")
-            with Pool(os.cpu_count()) as pool:
+            with Pool(env.NUM_WORKERS) as pool:
                 self.systems = pool.map(construct_dataset, systems)
         else:
             self.systems = [None] * len(systems)  # type: ignore
--- a/deepmd/pt/utils/dataloader.py
+++ b/deepmd/pt/utils/dataloader.py
@@ -2,7 +2,7 @@
 import logging
 import os
 import time
-from multiprocessing.dummy import (
+from multiprocessing import (
     Pool,
 )
 from queue import (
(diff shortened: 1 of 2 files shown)


deepmd/pt/utils/dataloader.py
```diff
@@ -87,26 +87,24 @@ def __init__(
...
-                if dist.is_available() and dist.is_initialized()
-                else 1
-            )
-        ) as pool:
-            self.systems = pool.map(construct_dataset, systems)
-
+        self.systems: list[DeepmdDataSetForLoader] = []
+        global_rank = dist.get_rank() if dist.is_initialized() else 0
+        if global_rank == 0:
+            if len(systems) >= 100:
```
wanghan-iapcm, line 99: why do you need this condition?  maybe better to report the number of system no matter how many system you have

wanghan-iapcm (changes requested):

njzjz:
> In this PR, only rank 0 will load data systems, and it will be broadcasted to each rank

Why do you expect each rank loads the same data?

caic99:
> > In this PR, only rank 0 will load data systems, and it will be broadcasted to each rank
> 
> Why do you expect each rank loads the same data?

Good question. Currently every rank reads the config from the same `input.json` file, and initializes datasets there.

njzjz:
> > > In this PR, only rank 0 will load data systems, and it will be broadcasted to each rank
> 
> > 
> 
> > Why do you expect each rank loads the same data?
> 
> 
> 
> Good question. Currently every rank reads the config from the same `input.json` file, and initializes datasets there.

I don't get your point. I believe the data is loaded asynchronously.

caic99:
> > > Why do you expect each rank loads the same data?

> I don't get your point. I believe the data is loaded asynchronously.

I see. What you mean is not all ranks are accessing the same data files, and this is correct. The case is when the data is stored in a shared file system with a bandwidth limit.

njzjz:
So it just affects the initialization of the data class? Well, I am not sure how much data is loaded in the initialization step.

wanghan-iapcm (approved):

deepmd/pt/utils/dataloader.py
```diff
@@ -87,26 +87,23 @@ def __init__(
...
-                else 1
-            )
-        ) as pool:
-            self.systems = pool.map(construct_dataset, systems)
-
+        self.systems: list[DeepmdDataSetForLoader] = []
+        global_rank = dist.get_rank() if dist.is_initialized() else 0
+        if global_rank == 0:
+            log.info(f"Constructing DataLoaders from {len(systems)} systems")
+            with Pool(os.cpu_count()) as pool:
```
njzjz, line 100: Why not using `env.NUM_WORKERS` here?

caic99:
```suggestion
            with Pool(env.NUM_WORKERS) as pool:
```

njzjz (approved):

deepmd/pt/utils/dataloader.py
```diff
@@ -87,26 +87,23 @@ def __init__(
...
-                else 1
-            )
-        ) as pool:
-            self.systems = pool.map(construct_dataset, systems)
-
+        self.systems: list[DeepmdDataSetForLoader] = []
+        global_rank = dist.get_rank() if dist.is_initialized() else 0
+        if global_rank == 0:
+            log.info(f"Constructing DataLoaders from {len(systems)} systems")
+            with Pool(env.NUM_WORKERS) as pool:
```
njzjz, line 100:
```suggestion
            with Pool(max(1, env.NUM_WORKERS)) as pool:
```

PR #4488, opened 2024-12-23
pd: add missing `dp.eval()` in pd backend

HydrogenSulfate:
Switch to eval mode when evaluating model, otherwise `self.training` will be `True`, backward graph will be created and cause OOM

## Summary by CodeRabbit

- **New Features**
	- Enhanced model evaluation state management to ensure correct behavior during evaluation.

- **Bug Fixes**
	- Improved type consistency in the `normalize_coord` function for better computational accuracy.

### DIFF
--- a/deepmd/pd/infer/deep_eval.py
+++ b/deepmd/pd/infer/deep_eval.py
@@ -237,6 +237,11 @@ class DeepEval(DeepEvalBackend):
             The output of the evaluation. The keys are the names of the output
             variables, and the values are the corresponding output arrays.
         """
+        # switch to eval mode
+        is_training = self.dp.training
+        if is_training:
+            self.dp.eval()
+
         # convert all of the input to numpy array
         atom_types = np.array(atom_types, dtype=np.int32)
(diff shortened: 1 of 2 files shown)


njzjz (commented):
Please take a look at @caic99's comment in #4426. Calling `eval()` or `to()` might be expensive and should be considered moving out of the loop.

HydrogenSulfate:
> Please take a look at @caic99's comment in #4426. Calling `eval()` or `to()` might be expensive and should be considered moving out of the loop.

#4426 seems only correct the training state when evaluating during training, but there are still issues in the dp test.

caic99:
> > Please take a look at @caic99's comment in #4426. Calling `eval()` or `to()` might be expensive and should be considered moving out of the loop.
> 
> #4426 seems only correct the training state when evaluating during training, but there are still issues in the dp test.

I did not check if the model has been set to eval state for other entrypoints and backends besides training on PyTorch. Maybe we should fix it for all backends?

njzjz (approved):

PR #4499, opened 2024-12-24
Perf: allow tf32 datatype for matmul

caic99:
Pytorch by default disables using TF32 data type starting from A100 GPU. Enabling TF32 utilizes tensor core on A100 GPUs, expecting better performance.

I will later attach some test results on the speed-up and accuracy of this PR.

Ref:
https://pytorch.org/docs/stable/notes/cuda.html#tf32-on-ampere
https://pytorch.org/docs/stable/generated/torch.set_float32_matmul_precision.html

## Summary by CodeRabbit

- **New Features**
	- Introduced support for TensorFloat-32 (TF32) in PyTorch to enhance matrix multiplication performance on compatible hardware.

### DIFF
--- a/deepmd/pt/entrypoints/main.py
+++ b/deepmd/pt/entrypoints/main.py
@@ -103,7 +103,8 @@ def get_trainer(
     finetune_links=None,
 ):
     multi_task = "model_dict" in config.get("model", {})
-
+    # https://pytorch.org/docs/stable/notes/cuda.html#tf32-on-ampere
+    torch.backends.cuda.matmul.allow_tf32 = True
     # Initialize DDP
     local_rank = os.environ.get("LOCAL_RANK")
     if local_rank is not None:


wanghan-iapcm (changes requested):
It is better to provide an option in the input script to control the behavior of using tensorcore.

caic99:
> It is better to provide an option in the input script to control the behavior of using tensorcore.

@njzjz Do you have any suggestions on adding this option? Maybe as an option under "training", or somehow integrating with precision controlling flags?

wanghan-iapcm:
it seems that the precision during the training and inference time should be consistent.
have you tested training at tf32 and inference at fp32? does it work or any loss of precision?

njzjz:
> > It is better to provide an option in the input script to control the behavior of using tensorcore.
> 
> @njzjz Do you have any suggestions on adding this option? Maybe as an option under "training", or somehow integrating with precision controlling flags?

I think we can make the decision after we see the benchmark result.

caic99:
I tested training the DPA-3 alpha model from scratch on AlMgCu dataset with different matmul precision on A800 GPU, and here is the result:

metric | highest (FP32) | high (TF32) | medium (BF16)
-- | -- | -- | --
Energy MAE  eV | 5.20E-01 | 4.93E-01 | 5.00E-01
Energy RMSE eV | 8.59E-01 | 9.14E-01 | 8.47E-01
Energy MAE/Natoms eV | 2.61E-02 | 2.47E-02 | 2.49E-02
Energy RMSE/Natoms eV | 4.45E-02 | 4.95E-02 | 4.64E-02
Force MAE  eV/A | 5.88E-02 | 6.18E-02 | 6.05E-02
Force RMSE eV/A | 8.37E-02 | 8.91E-02 | 8.67E-02
Virial MAE  eV | 9.85E-01 | 1.12E+00 | 1.02E+00
Virial RMSE eV | 2.03E+00 | 3.19E+00 | 1.99E+00
Virial MAE/Natoms eV | 4.84E-02 | 5.50E-02 | 5.09E-02
Virial RMSE/Natoms eV | 9.69E-02 | 1.40E-01 | 9.88E-02

> Note: the definition of highest, high, and medium comes [here](https://pytorch.org/docs/stable/generated/torch.set_float32_matmul_precision.html). 
**Please note that this precision setting only applies for the internal instructions used for MatMul.**

---

Same configuration as above, with 8 GPUs instead of one:

  - FP32 (Highest): batch   10000: total wall time = 1466.63 s
  - TF32 (High): batch   10000: total wall time = 1271.87 s
  - BF16 (Medium): batch   10000: total wall time = 1272.94 s

Using TF32 improves ~15% training speed.

---

> have you tested training at tf32 and inference at fp32? does it work or any loss of precision?

I used FP32 and TF32 for `dp test`, and the results are identical (the first 5 significant digits for `dp test` metrics are the same), regardless of the model trained in FP32 or TF32.

caic99:
I'll further test the impact of TF32 using OMat dataset.

caic99:
> I'll further test the impact of TF32 using OMat dataset.

Here is the `dp test` result on models trained on OMat with 1,000,000 steps:

| Metric                      | Highest                    | High                      | Medium                    |
|-----------------------------|----------------------------|---------------------------|---------------------------|
| Energy MAE (eV)             | 4.108578e-01               | 4.071145e-01              | 4.090323e-01              |
| Energy RMSE (eV)            | 1.070856e+00               | 1.063639e+00              | 1.074205e+00              |
| Energy MAE per Atom (eV)    | 2.659915e-02               | 2.638256e-02              | 2.646975e-02              |
| Energy RMSE per Atom (eV)   | 7.000563e-02               | 7.024324e-02              | 7.010443e-02              |
| Force MAE (eV/A)            | 1.082807e-01               | 1.077952e-01              | 1.075284e-01              |
| Force RMSE (eV/A)           | 1.753274e-01               | 1.745146e-01              | 1.741189e-01              |
| Virial MAE (eV)             | 1.316554e+00               | 1.311814e+00              | 1.313077e+00              |
| Virial RMSE (eV)            | 2.139529e+01               | 2.138941e+01              | 2.140013e+01              |
| Virial MAE per Atom (eV)    | 8.632328e-02               | 8.597799e-02              | 8.593321e-02              |
| Virial RMSE per Atom (eV)   | 1.258486e+00               | 1.257990e+00              | 1.258309e+00              |

njzjz:
It's a bit confused to me that you only enable tf32 during training but not inference?

caic99:
> I used FP32 and TF32 for `dp test`, and the results are identical (the first 5 significant digits for `dp test` metrics are the same), regardless of the model trained in FP32 or TF32.

> It's a bit confusing to me that you only enable tf32 during training but not inference.

@njzjz I've tested using `dp test`. Do you think these tests are enough?

njzjz:
> > I used FP32 and TF32 for `dp test`, and the results are identical (the first 5 significant digits for `dp test` metrics are the same), regardless of the model trained in FP32 or TF32.
> 
> > It's a bit confusing to me that you only enable tf32 during training but not inference.
> 
> @njzjz I've tested using `dp test`. Do you think these tests are enough?

I don't think `dp test` will call `get_trainer` that you modifies in this PR.

caic99:
> It's a bit confusing to me that you only enable tf32 during training but not inference.

@njzjz I executed `dp test` with different accuracy over a model trained with `highest` config:

| Metric                        | highest           | high              | medium            |
|-------------------------------|-------------------|-------------------|-------------------|
| **Energy MAE (eV)**           | 4.108578e-01      | 4.111513e-01      | 4.111513e-01      |
| **Energy RMSE (eV)**          | 1.070856e+00      | 1.071176e+00      | 1.071176e+00      |
| **Energy MAE/Natoms (eV)**    | 2.659915e-02      | 2.661123e-02      | 2.661123e-02      |
| **Energy RMSE/Natoms (eV)**   | 7.000563e-02      | 7.001191e-02      | 7.001191e-02      |
| **Force MAE (eV/Å)**          | 1.082807e-01      | 1.093453e-01      | 1.093453e-01      |
| **Force RMSE (eV/Å)**         | 1.753274e-01      | 1.767162e-01      | 1.767162e-01      |
| **Virial MAE (eV)**           | 1.316554e+00      | 1.317480e+00      | 1.317480e+00      |
| **Virial RMSE (eV)**          | 2.139529e+01      | 2.139539e+01      | 2.139539e+01      |
| **Virial MAE/Natoms (eV)**    | 8.632328e-02      | 8.635347e-02      | 8.635347e-02      |
| **Virial RMSE/Natoms (eV)**   | 1.258486e+00      | 1.258491e+00      | 1.258491e+00      |

caic99:
For using this feature without code modification, set environment variable `TORCH_ALLOW_TF32_CUBLAS_OVERRIDE=1`. 
[Ref](https://pytorch.org/docs/stable/cuda_environment_variables.html#:~:text=TORCH_ALLOW_TF32_CUBLAS_OVERRIDE)

PR #4501, opened 2024-12-24
(dpa3 alpha) add skip stat 

iProzd:


deepmd/utils/argcheck.py
```diff
@@ -1490,6 +1490,12 @@ def dpa3_repflow_args():
             default=4,
             doc=doc_axis_neuron,
         ),
+        Argument(
+            "skip_stat",
+            bool,
+            optional=True,
+            default=False,
+        ),
```
wanghan-iapcm, line 1498: better to use `fix_stat_std = 0.3`

PR #4505, opened 2024-12-25
Perf: replace unnecessary `torch.split` with indexing

caic99:
Some operations only use the first segment of the result tensor of `torch.split`. In this case, all the other segments are created and discarded. This slightly adds an overhead to the training process.

## Summary by CodeRabbit

- **Bug Fixes**
	- Simplified tensor slicing operations in the `RepformerLayer` class and the `nlist_distinguish_types` function, enhancing readability and performance.
  
- **Documentation**
	- Updated comments for clarity regarding tensor shapes in the `RepformerLayer` class.


caic99:
image
`torch.split` is an expensive function consisting of kernels on creating and copying data.

PR #4511, opened 2024-12-26
Chore: prevent from printing additional header in `lcurve.out` on resuming

caic99:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved logging control during training by adjusting header printing when resuming training.
	- Enhanced data handling to ensure seamless continuation of training and validation processes when data is exhausted.


njzjz (changes requested):
1. I don't suggest so, as the checkpoint is not saved every steps. You may see 10000 10100 (restart) 10000 10100. It's not clear if the title is removed.

2. All backends should have the same behavior.

PR #4513, opened 2024-12-26
Perf: use F.linear for MLP

caic99:
It brings <1% speedup.

## Summary by CodeRabbit

- **Refactor**
	- Simplified linear transformation implementation in the neural network layer
	- Improved code readability and efficiency in matrix operations

### DIFF
--- a/deepmd/pt/model/network/mlp.py
+++ b/deepmd/pt/model/network/mlp.py
@@ -204,15 +204,15 @@ class MLPLayer(nn.Module):
         if not env.DP_DTYPE_PROMOTION_STRICT:
             xx = xx.to(self.prec)
         yy = F.linear(xx, self.matrix.t(), self.bias)
-        # some activation functions are in-place, prevent further modification on `yy`; needs to be cloned
-        yy = self.activate(yy).clone()
-        if self.idt is not None:
-            yy *= self.idt
+        yy = self.activate(yy)
+        yy = yy * self.idt if self.idt is not None else yy
         if self.resnet:
             if xx.shape[-1] == yy.shape[-1]:
-                yy += xx
+                yy = yy + xx
             elif 2 * xx.shape[-1] == yy.shape[-1]:
-                yy += torch.concat([xx, xx], dim=-1)
+                yy = yy + torch.concat([xx, xx], dim=-1)
+            else:
+                yy = yy
         if not env.DP_DTYPE_PROMOTION_STRICT:
             yy = yy.to(ori_prec)
         return yy
--- a/deepmd/pt/model/network/mlp.py
+++ b/deepmd/pt/model/network/mlp.py
@@ -8,6 +8,7 @@ from typing import (
 import numpy as np
 import torch
 import torch.nn as nn
+import torch.nn.functional as F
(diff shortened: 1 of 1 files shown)


wanghan-iapcm (commented):
why introducing the change? 
do you observe explicit improvement in the efficiency of the linear transformations?

caic99:
> why introducing the change? 

I found this possibility when investigating on the matmuls not using Tensor Cores. This op fuses matmul and add when bias is present.

>do you observe explicit improvement in the efficiency of the linear transformations?

No, since currently no models are using MLP layer with bias. 
However, with bias=True and large input/output dimensions, the improvement is considerable. I set bias=True with g1_dim=3840 (10x larger for demonstration), and the end-to-end speedup of this modification is ~1.5x. 

Maybe we could just keep this change for future use?

njzjz:
I remember that torchscript could do the optimization automatically.

caic99:
> I remember that torchscript could do the optimization automatically.

Yes, but we are not using torchscript in the training loop by now. I'll test torchscript later.

caic99:
> I'll test torchscript later.

@njzjz Setting `JIT=True` in `env.py` results in `torch.jit.Error: Unable to infer type of dictionary: Cannot infer concrete type of torch.nn.Module` for DPA-2 models. I'm not sure whether it is a bug.

njzjz (approved):

deepmd/pt/model/network/mlp.py
```diff
@@ -202,20 +203,16 @@ def forward(
...
-        yy = yy * self.idt if self.idt is not None else yy
+        if self.idt is not None:
+            yy *= self.idt
         if self.resnet:
             if xx.shape[-1] == yy.shape[-1]:
                 yy += xx
             elif 2 * xx.shape[-1] == yy.shape[-1]:
                 yy += torch.concat([xx, xx], dim=-1)
-            else:
-                yy = yy
```
wanghan-iapcm, line 218: no reason to remove the else. explicit written here to complete all possible cases.

deepmd/pt/model/network/mlp.py
```diff
@@ -202,20 +203,16 @@ def forward(
...
-            torch.matmul(xx, self.matrix) + self.bias
-            if self.bias is not None
-            else torch.matmul(xx, self.matrix)
-        )
+        yy = F.linear(xx, self.matrix.t(), self.bias)
+        # some activation functions are in-place, prevent further modification on `yy`; needs to be cloned
         yy = self.activate(yy).clone()
-        yy = yy * self.idt if self.idt is not None else yy
+        if self.idt is not None:
+            yy *= self.idt
```
wanghan-iapcm, line 210: why changing in this way?

wanghan-iapcm (changes requested):

deepmd/pt/model/network/mlp.py
caic99: I think it is more readable. 
But you are right, the code style does not match this function. I'm reverting this change.

wanghan-iapcm (approved):

PR #4535, opened 2025-01-06
fix: lower `num_workers` to 4

caic99:
For multi-task training in pytorch, each data source will have their own dataloader. If the number of workers of dataloaders is large, there will be many (number of tasks * num_workers) worker processes stressing CPU.

## Summary by CodeRabbit

- **Performance Optimization**
  - Adjusted default maximum worker configuration from 8 to 4 CPUs
  - Reduced potential parallel processing resources for the environment
- **Documentation**
  - Updated documentation to reflect the change in default value for `NUM_WORKERS` from 8 to 4

### DIFF
--- a/deepmd/pt/utils/env.py
+++ b/deepmd/pt/utils/env.py
@@ -21,7 +21,7 @@ try:
     ncpus = len(os.sched_getaffinity(0))
 except AttributeError:
     ncpus = os.cpu_count()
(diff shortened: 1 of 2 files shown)


njzjz (changes requested):
Please update the documentation: https://docs.deepmodeling.com/projects/deepmd/en/stable/env.html#envvar-NUM_WORKERS

njzjz (approved):

PR #4577, opened 2025-02-03
Fix version in DeePMDConfigVersion.cmake

RMeli:
The double quotes end up as follows in `DeePMDConfigVersion.cmake`:
```cmake
set(PACKAGE_VERSION "$<IF:?:"0.0.0">")
```
This result in the following warning:
```
CMake Error (dev) at /opt/cp2k-toolchain/install/deepmd-kit-3.0.1/lib/cmake/DeePMD/DeePMDConfigVersion.cmake:10:                                             
  Syntax Warning in cmake code at column 30                                                                                                                                                                                                                                                                                
  Argument not separated from preceding token by whitespace.                                                                                                 
Call Stack (most recent call first):                                                                                                                         
  CMakeLists.txt:703 (find_package)                                                                                                                          
This error is for project developers. Use -Wno-error=dev to suppress it.      
```


njzjz (approved):
LGTM, but need to rebase.

We only accept new PRs on the `devel` branch.

RMeli:
Sorry, my bad. I'll fix the conflicts.

> We only accept new PRs on the `devel` branch.

It would be help to avoid these issues if the `devel` branch was the default branch on GitHub.

njzjz:
> It would help to avoid these issues if the `devel` branch was the default branch on GitHub.

This can be discussed, @wanghan-iapcm

PR #4583, opened 2025-02-07
fix: throw exception in consumer thread

caic99:
This pull request includes changes to improve error handling in the `dataloader.py` files for both `deepmd/pd/utils` and `deepmd/pt/utils`. The most important changes are the addition of a try-except block to handle exceptions during the data loading process and signaling the end of iteration properly.

Error handling improvements:

* [`deepmd/pd/utils/dataloader.py`](diffhunk://#diff-9739b359f76568c738a7f8df9a0235879c2b9dd166975323e328747bb065022eR261-R267): Added a try-except block to catch exceptions during the iteration over the data source, and modified the signaling of the end of iteration to use `StopIteration()`.
* [`deepmd/pt/utils/dataloader.py`](diffhunk://#diff-0f565156af2bed1d158fe99e5b06ec50ec97ed7de45fb8e0f1144dc40f74f17dR221-R226): Added a try-except block to catch exceptions during the iteration over the data source, and ensured the end of iteration is signaled with `StopIteration()`.

## Summary by CodeRabbit

- Bug Fixes
	- Improved robustness in background data processing. Errors encountered during data loading are now explicitly reported, and process completion is clearly signaled, ensuring more reliable operation.

### DIFF
--- a/deepmd/pt/utils/dataloader.py
+++ b/deepmd/pt/utils/dataloader.py
@@ -218,11 +218,13 @@ class BackgroundConsumer(Thread):
         self._source = source  # Main DL iterator
 
     def run(self) -> None:
-        for item in self._source:
-            self._queue.put(item)  # Blocking if the queue is full
-
-        # Signal the consumer we are done; this should not happen for DataLoader
-        self._queue.put(StopIteration())
+        try:
+            for item in self._source:
+                self._queue.put(item)  # Blocking if the queue is full
+        except Exception as e:
+            self._queue.put(e)
+        else:
+            self._queue.put(StopIteration())
 
 
 QUEUESIZE = 32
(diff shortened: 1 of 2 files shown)


deepmd/pt/utils/dataloader.py
```diff
@@ -218,11 +218,13 @@ def __init__(self, queue, source) -> None:
...
-        for item in self._source:
-            self._queue.put(item)  # Blocking if the queue is full
-
-        # Signal the consumer we are done; this should not happen for DataLoader
-        self._queue.put(StopIteration())
+        try:
+            for item in self._source:
+                self._queue.put(item)  # Blocking if the queue is full
+        except Exception as e:
+            self._queue.put(e)
```
iProzd, line 225: Note that the exception is not covered by ut. And what possible exception will be raised instead of `StopIteration` here?

iProzd (changes requested):

deepmd/pt/utils/dataloader.py
caic99: In case of the pytorch dataloader throws unexpected exceptions. Generally this code path is not used.

Issue #4586, opened 2025-02-07
[BUG] `polar` model with `pt` backend raises error


Yi-FanLi:
### Bug summary

I am training a `polar` model with the `pytorch` backend. The rmse is `nan` in the `lcurve.out` file and after training for some steps, an error is raised and the training stops. 

### DeePMD-kit Version

DeePMD-kit v3.0.1.dev89+gc9baf668

### Backend and its version

PyTorch 2.6.0+cu126

### How did you download the software?

Built from source

### Input Files, Running Commands, Error Log, etc.

I am running the training task in [examples](https://github.com/deepmodeling/deepmd-kit/tree/devel/examples/water_tensor/polar) and using the following input file:

[polar_input.json](https://github.com/user-attachments/files/18708942/polar_input.json)

The `lcurve.out` file looks like this:
```
# step    rmse_global_polar_val rmse_global_polar_trn   lr      
# If there is no available reference data, rmse_*_{val,trn} will print nan
      1           nan         nan    1.0e-03
    100           nan         nan    1.0e-03
    200           nan         nan    1.0e-03
    300           nan         nan    1.0e-03
    400           nan         nan    1.0e-03
    500           nan         nan    1.0e-03
    600           nan         nan    1.0e-03
    700           nan         nan    1.0e-03
    800           nan         nan    1.0e-03
    900           nan         nan    1.0e-03
```

The error message is:
```
[2025-02-07 10:45:35,058] DEEPMD INFO    batch     900: trn: rmse_global_polar =      nan, lr = 1.00e-03
[2025-02-07 10:45:35,059] DEEPMD INFO    batch     900: val: rmse_global_polar =      nan
[2025-02-07 10:45:35,059] DEEPMD INFO    batch     900: total wall time = 1.09 s
terminate called after throwing an instance of 'c10::Error'
  what():  CUDA error: initialization error
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

Exception raised from c10_cuda_check_implementation at /pytorch/c10/cuda/CUDAException.cpp:43 (most recent call first):
[...]

PR #4587, opened 2025-02-08
Fix: pt tensor loss label name

anyangml:
To address polar nan loss mentioned in #4586 

## Summary by CodeRabbit

- **Refactor**
	- Adjusted the internal processing order in computation routines to enhance consistency while maintaining the same overall user experience.
	- Updated model prediction handling to ensure compatibility in shape during statistical computations, reducing potential runtime errors.

### DIFF
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -1253,13 +1253,11 @@ def get_loss(loss_params, start_lr, _ntypes, _model):
         if "mask" in model_output_type:
             model_output_type.pop(model_output_type.index("mask"))
         tensor_name = model_output_type[0]
-        loss_params["tensor_name"] = tensor_name
         loss_params["tensor_size"] = _model.model_output_def()[tensor_name].output_size
-        label_name = tensor_name
-        loss_params["label_name"] = label_name
-        if label_name == "polarizability":
-            label_name = "polar"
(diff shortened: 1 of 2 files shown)


wanghan-iapcm (approved):

deepmd/pt/train/training.py
```diff
@@ -1256,9 +1256,9 @@ def get_loss(loss_params, start_lr, _ntypes, _model):
         loss_params["tensor_name"] = tensor_name
         loss_params["tensor_size"] = _model.model_output_def()[tensor_name].output_size
         label_name = tensor_name
+        loss_params["label_name"] = label_name
         if label_name == "polarizability":
             label_name = "polar"
-        loss_params["label_name"] = label_name
         loss_params["tensor_name"] = label_name
```
njzjz, line 1262: I am a bit confused here - line 1256 and line 1262 define `loss_params["tensor_name"]` twice with two different variables.

anyangml: for polar model, the tensor_name has to be overwrite to `polar`. I think we can remove line 1256 though

njzjz (approved):

Yi-FanLi:
I tested this PR and it fixes the `nan` issue. Thanks a lot! @anyangml

PR #4628, opened 2025-03-03
feat(pt): add atomic_weight to dipole model inference

ChiahsinChu:
Currently, the pytorch dipole model returns dipole (atomic quantities), global dipole (sum of atomic quantities) and force/virial operation on global dipoles. I am interested in not only the atomic dipole but also the derivative of total dipole w.r.t. coord, which is relevant to the dielectric response of the systems and has been used to train field-response MLPs (https://www.nature.com/articles/s41467-024-52491-3). However, the "global dipole" is not always the real total dipole of systems due to the missing information on charges. If the charges associated with the winner centroids are not identical, there is no way for users to get the derivative of (real) total dipole w.r.t. coord from DW models. (I tried to calculate the gradient of output dipole w.r.t. the input coord, but the gradient chain seems to be cut in the model.) Therefore, I add the `atomic_weight` parameter to the `model.forward`, which allows a user-defined atomic weight. 

There are some potential advantages of this implementation:
- allow users to calculate the derivative of (real) total dipole w.r.t. coord 
- allow further implementation of DPLR method

This implementation has been only activated for dipole models yet, but making it work for other models should be straightforword. 

## Summary by CodeRabbit

- **New Features**
  - Introduced optional atomic weight input to enhance prediction accuracy.
  - Added support for the SiLU activation function.
  - Launched an advanced DPA3 descriptor and repflow block for flexible simulation setups.

- **Enhancements**
  - Improved type consistency and error handling to boost stability.
  - Streamlined distributed training initialization and workflow configurations.

- **Tests & Tooling**
  - Expanded unit tests for the DPA3 descriptor and atomic weight parameter.
  - Updated build/pre-commit tools for increased reliability.


wanghan-iapcm (commented):
I do not get what do you mean by " If the charges associated with the winner centroids are not identical, there is no way for users to get the derivative of (real) total dipole w.r.t. coord from DW models. (I tried to calculate the gradient of output dipole w.r.t. the input coord, but the gradient chain seems to be cut in the model.)"

Does the type exclusion mechanism solves your issue? you may check 
https://github.com/deepmodeling/deepmd-kit/blob/70bc6d89d057d348558b87865ff2f225cae677aa/deepmd/dpmodel/atomic_model/base_atomic_model.py#L42

ChiahsinChu:
> Does the type exclusion mechanism solves your issue?

No. It is not what I mean. The `atom_exclude_types `, as far as my understanding, is the equivalent of `sel_type` in the tf model. 

An example for my case:
I want to train a DW model for LiOH aqueous solution, in which the charges associated with the WCs of O and Li should be -8 and -2, respectively. (No WC for H.) In this case, the `model_pred["force"]`, which is the negative derivative of the sum of atomic dipoles w.r.t. atomic coordinates, does not correspond to the derivative of total dipole moment w.r.t. coord.

deepmd/pt/model/model/make_model.py
```diff
@@ -293,6 +296,15 @@ def forward_common_lower(
                 aparam=ap,
                 comm_dict=comm_dict,
             )
+            # add weight to atomic_output
+            if hasattr(self.atomic_model, "fitting_net"):
+                if hasattr(self.atomic_model.fitting_net, "var_name"):
+                    kw = self.atomic_model.fitting_net.var_name
```
njzjz, line 302: I feel that `self.atomic_model.fitting_output_def` could be used here

njzjz, line 302: I feel that `self.atomic_model.fitting_output_def` could be used here

deepmd/pt/model/model/dipole_model.py
njzjz, line 1: It's a bit confusing since it breaks the consistency of APIs between different models and backends.

ChiahsinChu: Would it be better to add this keyword for all models and backends?

njzjz: Not sure what is the best way

wanghan-iapcm: I would suggest incorporating the atomic weight into the base atomic model, following a similar approach to how atom masks are managed. This would ensure that atomic weight is consistently supported across all atomic modes.

PR #4633, opened 2025-03-05
fix(ase): aviod duplicate stress calculation for ase calculator

iProzd:
## Summary by CodeRabbit

- **Refactor**
  - Revised the logic behind stress evaluation to ensure results are produced reliably when complete input is provided, and error notifications are triggered appropriately when essential data is missing.

### DIFF
--- a/deepmd/calculator.py
+++ b/deepmd/calculator.py
@@ -138,12 +138,13 @@ class DP(Calculator):
         self.results["virial"] = v[0].reshape(3, 3)
 
         # convert virial into stress for lattice relaxation
-        if "stress" in properties:
-            if sum(atoms.get_pbc()) > 0:
-                # the usual convention (tensile stress is positive)
-                # stress = -virial / volume
-                stress = -0.5 * (v[0].copy() + v[0].copy().T) / atoms.get_volume()
-                # Voigt notation
-                self.results["stress"] = stress.flat[[0, 4, 8, 5, 2, 1]]
-            else:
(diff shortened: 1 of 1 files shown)


wanghan-iapcm (approved):

njzjz (commented):
I don't get the point of this PR... Why is `stress` calculated when it is not in `properties`? And what is the relationship between the title and the change?

iProzd:
> I don't get the point of this PR... Why is `stress` calculated when it is not in `properties`? And what is the relationship between the title and the change?

@njzjz It's may be some tricky point in ase calling.
When performing 

```
energy = atoms.get_potential_energy()
forces = atoms.get_forces()
stress = atoms.get_stress()
```

during one-step calculation, **the calculator will magically be called twice**, even if you explictly add 'stress' key words in `properties: list[str] = ["energy", "forces", "virial", "stress"]`.

So I guess the ase will first call calculator with `properties=None` and store all data in `self.results`, and when finding some missing data, it will call calculator again with `properties` being missing keys.

This PR fixes this and meets the efficiency test.

njzjz:
The behavior is related to this piece of code:
https://gitlab.com/ase/ase/-/blob/master/ase/calculators/calculator.py#L533-540

`get_potential_energy` will set `properties` to `["energy"]`.
`get_forces` will not calculated again as `force` is already in `results`.

njzjz (approved):

PR #4647, opened 2025-03-11
feat: add SiLUT activation

iProzd:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new activation function called "SiLUT" that combines benefits from sigmoid and tanh behaviors.
	- Expanded the range of valid activation functions recognized by the module.
	- Added a configuration option for just-in-time compilation for custom operations.

- **Refactor**
	- Standardized the naming of activation functions across all supported frameworks for improved consistency.

These improvements offer users a streamlined experience with enhanced activation handling and performance optimizations.

### DIFF
--- a/deepmd/pt/utils/env.py
+++ b/deepmd/pt/utils/env.py
@@ -77,6 +77,7 @@ if intra_nthreads > 0:
 
 __all__ = [
     "CACHE_PER_SYS",
+    "CUSTOM_OP_USE_JIT",
     "DEFAULT_PRECISION",
     "DEVICE",
     "ENERGY_BIAS_TRAINABLE",
--- a/deepmd/pt/utils/env.py
+++ b/deepmd/pt/utils/env.py
@@ -34,6 +34,7 @@ else:
 JIT = False
 CACHE_PER_SYS = 5  # keep at most so many sets per sys in memory
 ENERGY_BIAS_TRAINABLE = True
+CUSTOM_OP_USE_JIT = False
 
 PRECISION_DICT = {
     "float16": torch.float16,
--- a/deepmd/pt/utils/utils.py
+++ b/deepmd/pt/utils/utils.py
@@ -21,7 +21,6 @@ from .env import (
 from .env import PRECISION_DICT as PT_PRECISION_DICT
 
 
-@torch.jit.script
 def silut_forward(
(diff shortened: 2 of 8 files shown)


deepmd/pt/utils/utils.py
```diff
@@ -11,17 +11,162 @@
...
+            return sig + x * sig * (1 - sig)
+
+        self.threshold = threshold
+        self.slope = float(silu_grad(threshold))
+        self.const = float(silu(threshold))
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        silu_part = F.silu(x)
+        mask = x >= self.threshold
+        if torch.any(mask):
```
njzjz, line 147: Are you sure it is faster after adding the if-statement? This needs a device-to-host communication, so will block the CPU.

iProzd: When training 9 layers DPA3 on example water systems with single GPU, using if-statement takes **23s** per 100 steps versus **29s** without. It seems that this device-to-host communication is much cheaper than additional calculation.

deepmd/pt/utils/utils.py
```diff
@@ -11,17 +11,164 @@
...
+)
 
 from .env import (
     DEVICE,
 )
 from .env import PRECISION_DICT as PT_PRECISION_DICT
 
 
+@torch.jit.script
+def silut_forward(
```
njzjz, line 25: I am unsure if the decorator increases the import timing. If so, we can compile the function in the runtime.

https://github.com/deepmodeling/deepmd-kit/blob/70bc6d89d057d348558b87865ff2f225cae677aa/deepmd/pt/utils/neighbor_stat.py#L138

njzjz (approved):

caic99 (approved):

wanghan-iapcm (approved):

deepmd/pt/utils/env.py
```diff
@@ -34,6 +34,7 @@
 JIT = False
 CACHE_PER_SYS = 5  # keep at most so many sets per sys in memory
 ENERGY_BIAS_TRAINABLE = True
+CUSTOM_OP_USE_JIT = False
```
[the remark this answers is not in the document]
njzjz: This variable should be added to `__all__` @iProzd

PR #4649, opened 2025-03-11
feat(jax): Hessian

njzjz:
## Summary by CodeRabbit

- **New Features**
	- Introduced enhanced Hessian fitting capabilities that extend model outputs to include second-order derivative information.
	- Integrated Hessian computations into the output transformation workflow for more detailed analytical results.

- **Tests**
	- Updated test suites to conditionally import modules based on Python version, ensuring compatibility with the JAX library.
	- Adjusted precision level in tests for finite differences to improve accuracy of comparisons.
	- Added a new environment variable for memory allocation handling in test configurations.
	- Introduced a new function for scatter summation specifically for JAX arrays.

### DIFF
--- a/deepmd/dpmodel/model/transform_output.py
+++ b/deepmd/dpmodel/model/transform_output.py
@@ -11,6 +11,7 @@ from deepmd.dpmodel.output_def import (
     ModelOutputDef,
     OutputVariableDef,
     get_deriv_name,
+    get_hessian_name,
     get_reduce_name,
 )
 
@@ -81,6 +82,7 @@ def communicate_extended_output(
 
     """
     xp = array_api_compat.get_namespace(mapping)
+    mapping_ = mapping
     new_ret = {}
     for kk in model_output_def.keys_outp():
         vv = model_ret[kk]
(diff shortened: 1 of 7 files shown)


deepmd/dpmodel/model/transform_output.py
```diff
@@ -116,6 +118,96 @@ def communicate_extended_output(
                 else:
                     # name holders
                     new_ret[kk_derv_r] = None
+                if vdef.r_hessian:
+                    kk_hess = get_hessian_name(kk)
+                    if model_ret[kk_hess] is not None:
+                        # jax only
+                        if array_api_compat.is_jax_array(force):
```
wanghan-iapcm, line 125: would it be better to implement the hessian scatter sum to somewhere in the jax module and import it here. 
otherwise the code would become difficult to read when more array api compatible backends are added. 

The same comment for the force part.

njzjz: I think I could try to convert it to a common one.

njzjz: > I think I could try to convert it to a common one.

#4654 tries to convert scatter_sum to a common one, but it needs Python 3.10.

njzjz: I've done an independent refactor in this PR, which doesn't conflict with #4654.

wanghan-iapcm (approved):

PR #4654, opened 2025-03-13
chore(array-api): implement scatter_sum

njzjz:
Address https://github.com/deepmodeling/deepmd-kit/pull/4649#discussion_r1992962447. This PR depends on [array_api_extra.at](https://data-apis.org/array-api-extra/generated/array_api_extra.at.html).


njzjz:
It seems that `array_api_extra` only supports Python 3.10...

njzjz:
Addressed in #5202.

PR #4661, opened 2025-03-20
feat(pt): add plugin for data modifier

ChiahsinChu:
- Add plugin for data modifier in pytorch backend
- Modify data in training 
- todo: add wrapper for frozen model


wanghan-iapcm (commented):
please add UT for the implementation.

deepmd/pt/train/training.py
```diff
@@ -134,6 +137,16 @@ def __init__(
         )
         self.num_model = len(self.model_keys)
 
+        # modifier for the training data
+        modifier_params = model_params.get("modifier", None)
+        if modifier_params is not None:
+            assert self.multi_task is False, (
```
njzjz, line 143: `NotImplementedError` is preferred to `assert`.

deepmd/pt/train/training.py
```diff
@@ -1043,10 +1072,13 @@ def save_model(self, save_path, lr=0.0, step=0) -> None:
...
-        torch.save(
-            {"model": module.state_dict(), "optimizer": optim_state_dict},
-            save_path,
-        )
+        save_dict = {
+            "model": module.state_dict(),
+            "optimizer": optim_state_dict,
+        }
+        if self.modifier is not None:
+            save_dict["data_modifier"] = self.modifier.state_dict()
```
njzjz, line 1080: I see this is saved, but how is it recovered?

deepmd/pt/train/training.py
```diff
@@ -215,12 +228,26 @@ def single_model_stat(
...
+                log.info(f"Using {modifier.modifier_type} as data modifier")
+                for _data in [_training_data, _validation_data]:
+                    if _data is not None:
+                        all_sampled = make_stat_input(
+                            _data.systems,
+                            _data.dataloaders,
+                            -1,
+                        )
+                        for sampled in all_sampled:
+                            modifier.modify_data(sampled)
```
iProzd, line 249: I noticed that this modification only affects the data statistics. It still uses the original data in `get_data` for actual training.

ChiahsinChu: I don't think so. `modifier.modify_data(sampled)` in line 267 has directly modify the training data `sampled`.

deepmd/pt/train/training.py
ChiahsinChu: I plan to restore it by adapting wrapper. Do you have any other idea for this?

deepmd/pt/train/training.py
iProzd: The `sampled` data is only used for data statistics, while it uses the original data from `get_data` function in Line 1092.

deepmd/pt/entrypoints/main.py
```diff
@@ -372,10 +384,24 @@ def freeze(
...
+        dm = tester.modifier
+        dm.eval()
+        with tempfile.NamedTemporaryFile(suffix=".pth", delete=False) as tmp_file:
+            torch.jit.save(
+                torch.jit.script(dm),
+                tmp_file,
+            )
+        with open(tmp_file.name, "rb") as f:
+            extra_files = {dm_output: f.read()}
+        os.unlink(tmp_file.name)  # Clean up the temporary file
```
njzjz: Implementation without writing to the disk:

```suggestion
        buffer = io.BytesIO()
        torch.jit.save(
            torch.jit.script(dm),
            buffer,
        )
        extra_files = {dm_output: buffer.getvalue()}
```

and `import io` at the top of the file.

njzjz: This can be applied to other files.

njzjz (changes requested):
Some copilot suggestions are useful, while others are not. Please check them.

deepmd/pt/train/training.py
```diff
@@ -388,6 +391,9 @@ def get_lr(lr_params: dict[str, Any]) -> LearningRateExp:
                     if self.finetune_links is not None
                     else False,
                 )
+                training_data[model_key].preload_and_modify_all_data()
+                if validation_data[model_key] is not None:
+                    validation_data[model_key].preload_and_modify_all_data()
```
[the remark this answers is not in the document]
ChiahsinChu: single_model_stat at line 381 only returns the function used for data stat. The data used for stat will be loaded in get_data_loader.

deepmd/utils/data.py
```diff
@@ -139,6 +140,13 @@ def __init__(
         # The prefix sum stores the range of indices contained in each directory, which is needed by get_item method
         self.prefix_sum = np.cumsum(frames_list).tolist()
 
+        self.apply_modifier_at_load = True
```
wanghan-iapcm: The cache may not work for large dataset due to memory limitation. I would suggest providing users with option to control this feature.

ChiahsinChu: I make this flag of using cache as user-defined. However, there is still an issue about CUDA re-initializiation if the modifier involve torch models. Currently, I skip the UT for this case.

ChiahsinChu: The CUDA re-init issue has been fixed.

deepmd/pt/train/wrapper.py
```diff
@@ -185,6 +188,10 @@ def forward(
 
         if self.inference_only or inference_only:
             model_pred = self.model[task_key](**input_dict)
+            if self.modifier is not None:
+                modifier_pred = self.modifier(**input_dict)
+                for k, v in modifier_pred.items():
+                    model_pred[k] = model_pred[k] - v
```
iProzd: Is it a typo? When inference, should it be `+` rather than `-` ?
```suggestion
            if self.modifier is not None:
                modifier_pred = self.modifier(**input_dict)
                for k, v in modifier_pred.items():
                    model_pred[k] = model_pred[k] + v
```

deepmd/pt/infer/deep_eval.py
```diff
@@ -171,8 +172,21 @@ def __init__(
...
+            # Load modifier if it exists in extra_files
+            if len(extra_files["data_modifier.pth"]) > 0:
+                # Create a file-like object from the in-memory data
+                modifier_data = extra_files["data_modifier.pth"]
+                if isinstance(modifier_data, bytes):
+                    modifier_data = io.BytesIO(modifier_data)
+                # Load the modifier directly from the file-like object
+                modifier = torch.jit.load(modifier_data, map_location=env.DEVICE)
+            self.dp = ModelWrapper(model, modifier=modifier)
+            self.modifier = modifier
```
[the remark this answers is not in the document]
njzjz: I don't find it used anywhere.

PR #4741, opened 2025-05-15
feat: use bfloat16 with `torch.autocast` on training

caic99:
This PR adds feature using `torch.autocast` with the accuracy of `torch.bfloat16` when training on nvidia GPUs.
By adding the decorator to `ModelWrapper.forward`, torch downcasts the data type for tensor computing and storage. This change expects a **37% peak memory reduction** compared with training under float32.

Todo:
- [x] Evaluate the impact on accuracy.
- [x] Make it configurable.

Note: 
- The element-wise operations of pytorch automatically upcast to the highest accuracy of the operands. Changes in `repflow_layer.py` makes the function output matches with the input if the input tensor is downcasted, even if tensors of higher accuracy like the parameters are involved.
- For unknown reasons, the `silut_forward_script` function fails to keep the data type of the output tensor as `bfloat16` when input tensor is of that type. Using raw silut function or using `torch.compile` can avoid the problem, so I believe the problem is related to `torch.jit.script`. So I have to manually cast the output data type to match with the input tensor.

Test results:
I trained 1 million steps on mptrj dataset, and the result shows that training under BF16 somehow affects the accuracy. This observation is consistent with other MLIP models: when training under BF16 accuracy, it might requires more training steps, and followed by fine-tuning in FP32 to keep the accuracy.

| Metric                        | BF16                     | FP32                     |
|-------------------------------|--------------------------|--------------------------|
| Energy MAE (eV)               | 8.649451 × 10^-1         | 8.084421 × 10^-1         |
| Energy RMSE (eV)              | 1.931963 × 10^0          | 1.765754 × 10^0          |
| Energy MAE/Natoms (eV)        | 3.495069 × 10^-2         | 3.300770 × 10^-2         |
| Energy RMSE/Natoms (eV)       | 8.879076 × 10^-2         | 8.778768 × 10^-2         |
| Force MAE (eV/Å)              | 1.158567 × 10^-1         | 1.103375 × 10^-1         |
| Force RMSE (eV/Å)             | 3.550471 × 10^-1         | 3.164068 × 10^-1         |
| Virial MAE (eV)               | 1.111046 × 10^0          | 1.051787 × 10^0          |
| Virial RMSE (eV)              | 3.856514 × 10^0          | 3.535781 × 10^0          |
| Virial MAE/Natoms (eV)        | 4.698042 × 10^-2         | 4.510412 × 10^-2         |
| Virial RMSE/Natoms (eV)       | 1.490477 × 10^-1         | 1.467897 × 10^-1         |

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved consistency of tensor data types during computations to prevent potential errors and ensure reliable results.
	- Ensured output tensors maintain the same data type as inputs for consistent processing.

- **New Features**
	- Enabled automatic mixed precision with bfloat16 precision on CUDA devices for enhanced performance during model execution.

### DIFF
--- a/deepmd/pt/model/descriptor/repflow_layer.py
+++ b/deepmd/pt/model/descriptor/repflow_layer.py
@@ -682,7 +682,6 @@ class RepFlowLayer(torch.nn.Module):
                     )
                 )
 
-            a_sw.to(edge_angle_update.dtype)
             # nb x nloc x a_nnei x a_nnei x e_dim
             weighted_edge_angle_update = (
                 a_sw[..., None, None] * a_sw[..., None, :, None] * edge_angle_update
--- a/deepmd/pt/model/descriptor/repflow_layer.py
+++ b/deepmd/pt/model/descriptor/repflow_layer.py
@@ -424,7 +424,7 @@ class RepFlowLayer(torch.nn.Module):
         sub_edge_update_ik = torch.matmul(edge_ebd, sub_edge_ik)
 
         result_update = (
-            bias
+            bias.to(sub_angle_update.dtype)
             + sub_node_update.unsqueeze(2).unsqueeze(3)
             + sub_edge_update_ij.unsqueeze(2)
             + sub_edge_update_ik.unsqueeze(3)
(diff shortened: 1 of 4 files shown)


caic99:
By setting model.descriptor.precision: "bfloat16", all the calculation process in descriptor is performed in bfloat16. The memory reduction is about 50%.
I'll test the performance gain with a larger batch size, which is available through the modifications of  https://github.com/deepmodeling/deepmd-kit/pull/4744.

PR #4744, opened 2025-05-16
perf: gather node embedding before matmul

caic99:
`node_ebd_ext` contains embedding on expanded atoms, which might be large for a large cut-off. Current implementation do matmul first, then gather tensor by the neighbors. This introduces saving `sub_node_ext_update` of size `nf * nall * ndim` in each repflow layer, where `nall` might be multiple times larger than `nloc`.
This PR do gathering first, then compute matmul. After this PR, the peak memory size is of `O(nlayer * nf * nloc + 1* nf*nall)`. I've tested that the `sub_node_ext_update` calculated by those two methods are numerically equal.

### DIFF
--- a/deepmd/pt/model/descriptor/repflow_layer.py
+++ b/deepmd/pt/model/descriptor/repflow_layer.py
@@ -455,10 +455,10 @@ class RepFlowLayer(torch.nn.Module):
 
         # nf * nloc * node/edge_dim
         sub_node_update = torch.matmul(node_ebd, node)
-        # nf * nall * node/edge_dim
-        sub_node_ext_update = torch.matmul(node_ebd_ext, node_ext)
         # nf * nloc * nnei * node/edge_dim
-        sub_node_ext_update = _make_nei_g1(sub_node_ext_update, nlist)
+        gathered_node_ebd_ext = _make_nei_g1(node_ebd_ext, nlist)
+        # nf * nloc * nnei * node/edge_dim
+        sub_node_ext_update = torch.matmul(gathered_node_ebd_ext, node_ext)
         # nf * nloc * nnei * node/edge_dim
         sub_edge_update = torch.matmul(edge_ebd, edge)
 


iProzd:
https://github.com/iProzd/deepmd-kit/commit/28803d926ecf5a9517264aa8ad6628f6e946eb6b maybe you can refer to and be consistent with this?

caic99:
> [iProzd@28803d9](https://github.com/iProzd/deepmd-kit/commit/28803d926ecf5a9517264aa8ad6628f6e946eb6b) maybe you can refer to and be consistent with this?

@iProzd I've modified the codes in alignment with your commit, and adopt your UTs on this PR and pass them. However other UTs are failing. I suspect that the numerical error while scattering and gathering features should be blamed. Any ideas?

deepmd/pt/model/descriptor/repflow_layer.py
iProzd, line 1: Maybe you need to abort modifications in this file, which means to keep exact modifications in commit https://github.com/iProzd/deepmd-kit/commit/28803d926ecf5a9517264aa8ad6628f6e946eb6b and I passed all the uts.

deepmd/pt/model/descriptor/repflow_layer.py
```diff
@@ -455,10 +454,8 @@ def optim_edge_update(
 
         # nf * nloc * node/edge_dim
         sub_node_update = torch.matmul(node_ebd, node)
-        # nf * nall * node/edge_dim
-        sub_node_ext_update = torch.matmul(node_ebd_ext, node_ext)
-        # nf * nloc * nnei * node/edge_dim
-        sub_node_ext_update = _make_nei_g1(sub_node_ext_update, nlist)
+        # nf * nloc * node/edge_dim
+        sub_node_ext_update = torch.matmul(nei_node_ebd, node_ext)
```
iProzd, line 458: You missed a _make_nei_g1 here. But when you use locak mapping, it's more efficient to keep the old implementation.

iProzd (changes requested):

deepmd/utils/argcheck.py
```diff
@@ -1421,6 +1421,12 @@ def descrpt_dpa3_args():
             default=False,
             doc=doc_use_tebd_bias,
         ),
+        Argument(
+            "use_loc_mapping",
+            bool,
+            optional=True,
+            default=True,
```
iProzd, line 1428: add a docstr for use_loc_mapping

iProzd (changes requested):

deepmd/pt/model/descriptor/repflow_layer.py
caic99: > Maybe you need to abort modifications in this file

Did you mean you'll submit another PR to this branch?

PR #4751, opened 2025-05-20
perf: use torch.topk to construct nlist

caic99:
This pull request updates the `build_neighbor_list` function in `deepmd/pt/utils/nlist.py` to improve its performance and correctness when selecting neighbors. The key change involves replacing `torch.sort` with `torch.topk` to optimize the selection of the nearest neighbors.

<details><summary>Details</summary>
<p>
Before: 16.4ms
image

After: 3.4ms
image

Step time goes from 212.5ms to 200.5ms, ~5% speed-up.

</p>
</details>


caic99:
@iProzd Since [`torch.topk`](https://docs.pytorch.org/docs/stable/generated/torch.topk.html) is not guaranteed to be stable, the neighbor atoms with the same distance might have different indices. This fails the UT.

iProzd (changes requested):
remove old codes from comments and make it shorter?

PR #4756, opened 2025-05-21
feat(pt/dp): add exponential switch function

iProzd:


deepmd/utils/argcheck.py
```diff
@@ -1497,6 +1497,17 @@ def dpa3_repflow_args():
         "Whether to make edge update smooth. "
         "If True, the edge update from angle message will not use self as padding."
     )
+    doc_use_exp_switch = (
+        "Whether to use an exponential switch function instead of a polynomial one in the neighbor update. "
+        "The exponential switch function ensures neighbor contributions smoothly diminish as the interatomic distance "
+        "`r` approaches the cutoff radius `rcut`. Specifically, the function is defined as: "
+        "s(r) = \\exp(-\\exp(C (r - rcut_smth) / rcut_smth)) for 0 < r \\leq rcut, and s(r) = 0 for r > rcut. "
+        "Here, `rcut_smth` is an adjustable smoothing factor, "
+        "while `C` is a tunable parameter controlling the decay rate (default: 20). "
```
wanghan-iapcm, line 1506: I find the users cannot really tune the parameter C here.

wanghan-iapcm (changes requested):

deepmd/utils/argcheck.py
iProzd: Perhaps we fix `C` here and let users only adjust the `rcut_smth`, as this covers most cases. I'll change the docstr.

wanghan-iapcm (approved):

PR #4761, opened 2025-05-22
chore(pt): use more warmup steps

caic99:
The SiLUT activation function is using JIT compile. Hence, we require more warm-up steps to avoid JIT compiling time interfering the profiling result. This commit also adds logging info of the saved profiling results.

### DIFF
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -658,7 +658,7 @@ class Trainer:
             writer = SummaryWriter(log_dir=self.tensorboard_log_dir)
         if self.enable_profiler or self.profiling:
             prof = torch.profiler.profile(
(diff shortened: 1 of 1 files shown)


deepmd/pt/train/training.py
```diff
@@ -1032,10 +1032,13 @@ def log_loss_valid(_task_key="Default"):
             writer.close()
         if self.enable_profiler or self.profiling:
             prof.stop()
+            log.info(
+                f"The profiling trace has been saved under {self.tensorboard_log_dir}"
+            )
```
njzjz, line 1037: This message is for `enable_profiler`, but one may set `enable_profiler` to `False` and `profiling` to True.

njzjz (approved):

PR #4763, opened 2025-05-26
Feat:support customized rglob

anyangml:
Currently, when passing a single str as the root data directory, the `expand_sys_str` function will automatically perform `rglob` to grab all systems. However, this depends on the structure of the data folder. There are scenarios where train/val folders are nested, i.e. "root/dataset\_\*/trn" & "root/dataset\_\*/val".  

A customizable rglob function is needed to provide more flexibility when constructing datasets, and to remove unnecessarily long data lists in the input file.

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying custom glob patterns to filter training and validation datasets, allowing more flexible and targeted data selection (PyTorch backend only).
  - Introduced recursive pattern matching to improve system directory selection based on user-defined criteria.
- **Tests**
  - Added new test cases to validate the customized glob pattern functionality for training and validation datasets.


deepmd/common.py
```diff
@@ -207,6 +207,30 @@ def expand_sys_str(root_dir: Union[str, Path]) -> list[str]:
...
+    Returns
+    -------
+    list[str]
+        list of string pointing to system directories
+    """
+    root_dir = Path(root_dir)
+    matches = []
+    for pattern in patterns:
+        matches.extend(
+            [str(d) for d in root_dir.rglob(pattern) if (d / "type.raw").is_file()]
```
caic99, line 229: Is it possible to do something like this to avoid repeatedly accessing a large directory?
```suggestion
            [str(d) for d in root_dir.rglob(pattern / "type.raw") if (d.is_file()]
```

deepmd/pt/entrypoints/main.py
```diff
@@ -114,9 +114,11 @@ def prepare_trainer_input_single(
             validation_dataset_params["systems"] if validation_dataset_params else None
         )
         training_systems = training_dataset_params["systems"]
-        training_systems = process_systems(training_systems)
+        trn_patterns = training_dataset_params.get("rglob_patterns", None)
+        training_systems = process_systems(training_systems, patterns=trn_patterns)
```
njzjz, line 118: `training_data/systems` is also used below
https://github.com/deepmodeling/deepmd-kit/blob/9b704878fa360069f9d04b2378044c51f855ab25/deepmd/pt/entrypoints/main.py#L313-L315

njzjz (commented):
This PR lacks UT

anyangml:
> This PR lacks UT

What do you think would be the best way to handle the tests? A separate UT for the individual function or an end-to-end integration test in PT test training?

PR #4768, opened 2025-05-27
pd: revert einsum to matmul for paddle backend

HydrogenSulfate:
revert einsum to matmul for paddle backend as einsum may not trigger infermeta error.

## Summary by CodeRabbit

- **Refactor**
  - Improved the internal computation method for certain tensor operations to enhance performance and maintainability. No changes to user-facing features or outputs.

### DIFF
--- a/deepmd/pd/model/model/transform_output.py
+++ b/deepmd/pd/model/model/transform_output.py
@@ -82,9 +82,7 @@ def task_deriv_one(
     assert extended_force is not None
     extended_force = -extended_force
     if do_virial:
-        extended_virial = paddle.einsum(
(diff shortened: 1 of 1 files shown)


caic99 (approved):
LGTM. @HydrogenSulfate Would you explain more about einsum "may not trigger infermeta error"?

caic99:
> There seems a bug in infermeta of einsum or operators it invoked when use `CINN` compiler mode. so it is better to use matmul instead.

@HydrogenSulfate Maybe you can try keep the first two dimensions explicitly in `einsum`, like "nmik,nmij->nmikj"

PR #4772, opened 2025-05-28
feat: add use_loc_mapping

caic99:
`node_ebd_ext` contains embedding on expanded atoms, which might be large for a large cut-off. Current implementation do matmul first, then gather tensor by the neighbors. This introduces saving `sub_node_ext_update` of size nf * nall * ndim in each repflow layer, where nall might be multiple times larger than nloc.
This PR do gathering first, then compute matmul. After this PR, the peak memory size is of O(nlayer * nf * nloc), unrelated to `nall`.

## Summary by CodeRabbit

- **New Features**
  - Introduced an optional setting to enable or disable local atom index mapping during descriptor computation, providing more flexibility in non-parallel inference scenarios.

- **Bug Fixes**
  - Enhanced test coverage and consistency checks to ensure descriptor outputs remain equivalent whether local mapping is enabled or not.

- **Documentation**
  - Improved parameter descriptions and comments to clarify the behavior and shape of key variables related to local mapping.

- **Tests**
  - Added new tests to validate the functional equivalence of descriptor outputs with and without local mapping.
  - Updated existing tests to support and verify the new local mapping option.

- **Style**
  - Improved assertion handling in tests by specifying absolute tolerance for numerical comparisons.


iProzd (changes requested):
A consistent test also need to be added in source/tests/consistent/descriptor/test_dpa3.py (and skip pd), see https://github.com/deepmodeling/deepmd-kit/pull/4760/files#diff-2175c5f9ca14ee18e75a5fe46c7ecd66d02fae4612599a7763a2e29acdf12943 for example.

deepmd/dpmodel/descriptor/repflows.py
```diff
@@ -145,6 +145,8 @@ class DescrptBlockRepflows(NativeOP, DescriptorBlock):
         In the dynamic selection case, neighbor-scale normalization will use `e_sel / sel_reduce_factor`
         or `a_sel / sel_reduce_factor` instead of the raw `e_sel` or `a_sel` values,
         accommodating larger selection numbers.
+    use_loc_mapping : bool, optional
+        Whether to use local atom index mapping in non-parallel inference.
```
iProzd, line 149:
```suggestion
    use_loc_mapping : bool, optional
        Whether to use local atom index mapping in training or non-parallel inference.
        When True, local indexing and mapping are applied to neighbor lists and embeddings during descriptor computation.
```

deepmd/dpmodel/utils/network.py
```diff
@@ -1020,6 +1021,8 @@ def get_graph_index(
         Masks of the neighbor list for angle. real nei 1 otherwise 0
     nall
         The number of extended atoms.
+    use_loc_mapping
+        Whether to use local atom index mapping in non-parallel inference.
```
iProzd, line 1025:
```suggestion
    use_loc_mapping
        Whether to use local atom index mapping in training or non-parallel inference.
        When True, local indexing and mapping are applied to neighbor lists and embeddings during descriptor computation.
```

deepmd/pt/model/descriptor/dpa3.py
```diff
@@ -89,6 +89,8 @@ class DescrptDPA3(BaseDescriptor, torch.nn.Module):
         Whether to use electronic configuration type embedding.
     use_tebd_bias : bool, Optional
         Whether to use bias in the type embedding layer.
+    use_loc_mapping : bool, Optional
+        Whether to use local atom index mapping in non-parallel inference.
```
iProzd, line 93:
```suggestion
    use_loc_mapping : bool, Optional
        Whether to use local atom index mapping in training or non-parallel inference.
        When True, local indexing and mapping are applied to neighbor lists and embeddings during descriptor computation.
```

deepmd/utils/argcheck.py
```diff
@@ -1421,6 +1421,14 @@ def descrpt_dpa3_args():
             default=False,
             doc=doc_use_tebd_bias,
         ),
+        Argument(
+            "use_loc_mapping",
+            bool,
+            optional=True,
+            default=True,
+            doc="Whether to use local atom index mapping in non-parallel inference. "
+            "When True, local indexing and mapping are applied to neighbor lists and embeddings during descriptor computation.",
```
iProzd, line 1430:
```suggestion
            doc="Whether to use local atom index mapping in training or non-parallel inference. "
            "When True, local indexing and mapping are applied to neighbor lists and embeddings during descriptor computation.",
```

PR #4783, opened 2025-06-05
feat(pt/pd): add size option to dp show

iProzd:
## Summary by CodeRabbit

- **New Features**
	- Added the ability to display model parameter counts using the "size" attribute in the "show" command.
	- Users can now view detailed parameter counts for descriptor, fitting-net, and total parameters for both single-task and multi-task models.

- **Tests**
	- Updated tests to verify correct output of model parameter counts when using the new "size" attribute.

- **Documentation**
	- Added new documentation for the "show" command detailing usage and output for various model attributes, including the new "size" attribute.

### DIFF
--- a/deepmd/pd/infer/deep_eval.py
+++ b/deepmd/pd/infer/deep_eval.py
@@ -482,6 +482,32 @@ class DeepEval(DeepEvalBackend):
         """Get model definition script."""
         return self.model_def_script
 
+    def get_model_size(self) -> dict:
+        """Get model parameter count.
+
+        Returns
+        -------
+        dict
+            A dictionary containing the number of parameters in the model.
+            The keys are 'descriptor', 'fitting_net', and 'total'.
+        """
+        params = self.dp.state_dict()
+        sum_param_des = sum(
+            params[k].numel()
+            for k in params.keys()
+            if "descriptor" in k and "mean" not in k and "stddev" not in k
+        )
+        sum_param_fit = sum(
+            params[k].numel()
+            for k in params.keys()
+            if "fitting" in k and "_networks" not in k
+        )
+        return {
+            "descriptor": sum_param_des,
+            "fitting-net": sum_param_fit,
+            "total": sum_param_des + sum_param_fit,
+        }
+
     def eval_descriptor(
         self,
         coords: np.ndarray,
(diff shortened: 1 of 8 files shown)


deepmd/pd/infer/deep_eval.py
```diff
@@ -482,6 +482,32 @@ def get_model_def_script(self) -> str:
...
+    def get_model_size(self) -> dict:
+        """Get model parameter count.
+
+        Returns
+        -------
+        dict
+            A dictionary containing the number of parameters in the model.
+            The keys are 'descriptor', 'fitting_net', and 'total'.
+        """
+        params = self.dp.state_dict()
```
njzjz, line 494: What about `self.dp.parameters()`? It will exclude buffers

deepmd/pt/infer/deep_eval.py
```diff
@@ -624,6 +624,32 @@ def get_model_def_script(self) -> str:
...
+    def get_model_size(self) -> dict:
+        """Get model parameter count.
+
+        Returns
+        -------
+        dict
+            A dictionary containing the number of parameters in the model.
+            The keys are 'descriptor', 'fitting_net', and 'total'.
+        """
+        params = self.dp.state_dict()
```
njzjz: What about self.dp.parameters()? It will exclude buffers

njzjz: or `named_parameters()`

deepmd/pt/infer/deep_eval.py
```diff
@@ -624,6 +624,30 @@ def get_model_def_script(self) -> str:
...
+        """
+        params_dict = dict(self.dp.named_parameters())
+        sum_param_des = sum(
+            params_dict[k].numel() for k in params_dict.keys() if "descriptor" in k
+        )
+        sum_param_fit = sum(
+            params_dict[k].numel()
+            for k in params_dict.keys()
+            if "fitting" in k and "_networks" not in k
+        )
```
Chengqian-Zhang, line 644: If we have model parameters which don't include "descriptor" and "_networks"(maybe in the future), we will neglect them when counting parameters. But I think it is ok now because we don't have this kind of model parameters.

Chengqian-Zhang (approved):

deepmd/pt/infer/deep_eval.py
iProzd: Maybe we can add other variants in the future if necessary.

PR #4793, opened 2025-06-10
fix(dpmodel/pt/pd/jax): pass trainable to layer & support JAX trainable & support TF tensor fitting trainable

njzjz:
1. For dpmodel, pt, and pd, pass the trainable parameter to the layer (not actually used in this PR).
2. For JAX, support the `trainable` parameter in the layer.
3. "trainable" is now serialized in dpmodel, tf, pt, pd, etc.
4. Support trainable in TF dipole & polar fitting.

## Summary by CodeRabbit

- **New Features**
  - Added a `trainable` parameter to numerous model components, descriptors, embedding and fitting networks, and attention layers, allowing users to enable or disable parameter trainability.
  - Ensured consistent propagation of the `trainable` flag across submodules and layers in PyTorch, TensorFlow, and JAX backends.
  - Enhanced serialization and deserialization processes to include the `trainable` attribute, maintaining trainability settings during model save and load.
  - Implemented conditional parameter wrapping in the JAX backend to differentiate trainable parameters from fixed variables.

- **Tests**
  - Updated test configurations to include the `trainable` flag, validating behavior with non-trainable components.

### DIFF
--- a/deepmd/dpmodel/utils/network.py
+++ b/deepmd/dpmodel/utils/network.py
@@ -690,10 +690,12 @@ def make_embedding_network(T_Network, T_NetworkLayer):
             precision: str = DEFAULT_PRECISION,
             seed: Optional[Union[int, list[int]]] = None,
             bias: bool = True,
-            trainable: bool = True,
+            trainable: Union[bool, list[bool]] = True,
         ) -> None:
             layers = []
             i_in = in_dim
+            if isinstance(trainable, bool):
+                trainable = [trainable] * len(neuron)
             for idx, ii in enumerate(neuron):
                 i_ot = ii
                 layers.append(
@@ -706,7 +708,7 @@ def make_embedding_network(T_Network, T_NetworkLayer):
                         resnet=True,
                         precision=precision,
                         seed=child_seed(seed, idx),
-                        trainable=trainable,
+                        trainable=trainable[idx],
                     ).serialize()
                 )
                 i_in = i_ot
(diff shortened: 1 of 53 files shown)


deepmd/pd/model/descriptor/dpa1.py
iProzd, line 1: Will the new trainable settings be overwritten by the previous setting in the high-level __init__?
```
# set trainable
for param in self.parameters():
    param.stop_gradient = not trainable
```

deepmd/pt/model/descriptor/dpa1.py
iProzd, line 1: Same as above.

deepmd/dpmodel/utils/network.py
```diff
@@ -786,14 +799,20 @@ def __init__(
             precision: str = DEFAULT_PRECISION,
             bias_out: bool = True,
             seed: Optional[Union[int, list[int]]] = None,
+            trainable: Union[bool, list[bool]] = True,
         ) -> None:
+            if trainable is None:
+                trainable = [True] * (len(neuron) + 1)
+            elif isinstance(trainable, bool):
+                trainable = [trainable] * (len(neuron) + 1)
```
iProzd, line 807:
```suggestion
            if trainable is None:
                trainable = [True] * (len(neuron) + 1)
            elif isinstance(trainable, bool):
                trainable = [trainable] * (len(neuron) + 1)
            else:
                trainable = trainable
```

iProzd (changes requested):

deepmd/pd/model/descriptor/dpa1.py
njzjz: In pytorch & paddle backends, the new argument is not actually used in this PR. It exists to be consistent with dpmodel and future PRs can do the modifications.

deepmd/dpmodel/utils/network.py
```diff
@@ -786,14 +799,22 @@
...
+            else:
+                trainable = trainable
```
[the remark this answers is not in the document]
njzjz: @iProzd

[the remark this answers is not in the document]
njzjz: Revert after discussion.
```suggestion
                pass
```

PR #4795, opened 2025-06-10
pd: support different label_dict in CINN

HydrogenSulfate:
pop unnecessary item when wrapping model with `jit.to_static`, so we can support se_e2_a/dpa2/dpa3 without extra modification.

@njzjz can you give some suggestions for better code improvements? The current approach of fetching data via `self.get_data` isn't very concise.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility by dynamically matching label input specifications to available label keys during model compilation when CINN is enabled. This prevents errors caused by mismatched label keys at runtime.

### DIFF
--- a/deepmd/pd/train/training.py
+++ b/deepmd/pd/train/training.py
@@ -607,6 +607,35 @@ class Trainer:
             )
 
             backend = "CINN" if CINN else None
+
+            # NOTE: This is a trick to decide the right input_spec for wrapper.forward
+            _, label_dict, _ = self.get_data(is_train=True, task_key="Default")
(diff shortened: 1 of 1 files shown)


njzjz:
> @njzjz can you give some suggestions for better code improvements? The current approach of fetching data via self.get_data isn't very concise.

Maybe @iProzd or @caic99 is more familiar with the get_data method.

caic99 (approved):
LGTM. @HydrogenSulfate The suggested changes by AI might help.

njzjz:
@coderabbitai resolve

PR #4800, opened 2025-06-11
update NvNMD training code

jiongwalai:
Update NvNMD training code:
1. support different FPGA device: vu9p (Bohrium and NvNMD web) and vu13p (coming soon)
2. set `"stripped_type_embedding": true, "smooth_type_embdding": true, "set_davg_zero": false` in nvnmd-v1
3. support setting seed in nvnmd dict

## Summary by CodeRabbit

- **New Features**
  - CLI option to initialize training from a frozen model; device selection support (vu9p/vu13p).

- **New/Changed Behavior**
  - Descriptor recovery pipeline updated with an additional descriptor component and propagated recovery mask; mapping mode adjusted.

- **Improvements**
  - Deep-copying of configs/data, explicit numeric precision, seed propagation, finer mapping resolution, version-aware data handling and packing.

- **Bug Fixes**
  - Updated tests and expected outputs to match revised behavior.

- **Chores**
  - Logging, validation and minor cleanups.


wanghan-iapcm:
@jiongwalai please fix the precommit issues.

deepmd/tf/nvnmd/data/data.py
```diff
@@ -118,9 +122,9 @@
     "end": "",
 }
 
-# change the configuration according to the max_nnei
-jdata_config_v0_ni128 = jdata_config_v0.copy()
-jdata_config_v0_ni256 = jdata_config_v0.copy()
+# change the configuration accordng to the max_nnei
```
njzjz: There seems to be some meaningless changes like this

njzjz (changes requested):

deepmd/tf/nvnmd/data/data.py
jiongwalai: There may be differences between the parameters `jdata_config_v0_ni128 ` and `jdata_config_v0_ni256`, which are used for different training configurations. I think they should be retained.

njzjz: Why did you change `according` to `accordng`?

njzjz:
I close this PR since it seems to be stale.

PR #4806, opened 2025-06-16
fix: training speed might be incorrect

caic99:
fix #4212

## Summary by CodeRabbit

- **Refactor**
	- Improved the logic for accumulating and reporting training time during model training, resulting in clearer and more consistent step counting and average time logging.

### DIFF
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -1032,7 +1032,7 @@ class Trainer:
                 with open("checkpoint", "w") as f:
                     f.write(str(self.latest_model))
 
-            if self.timing_in_training:
+            if self.timing_in_training and self.timed_steps:
                 msg = f"average training time: {self.total_train_time / self.timed_steps:.4f} s/batch"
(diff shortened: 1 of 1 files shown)


deepmd/pt/train/training.py
```diff
@@ -935,12 +935,19 @@ def log_loss_valid(_task_key="Default"):
                             eta=eta,
                         )
                     )
-                # the first training time is not accurate
                 if (
-                    (_step_id + 1 - self.start_step) > self.disp_freq
-                    or self.num_steps - self.start_step < 2 * self.disp_freq
+                    (self.num_steps - self.start_step)
+                    <= 2 * self.disp_freq  # not enough steps
```
njzjz, line 940: Why is `<` changed to `<=`?

caic99: I think both is OK

njzjz: okay, I think I got your point here...

njzjz (approved):

PR #4820, opened 2025-06-25
feat(pt): add `observed-type` option for dp show

iProzd:
## Summary by CodeRabbit

- **New Features**
  - Added support for displaying "observed-type" information in the model reporting tool, showing element types observed during training for single-task and multi-task models.
- **Tests**
  - Updated tests to verify correct reporting of observed types for both single-task and multi-task models.
- **Documentation**
  - Updated documentation to include the new "observed-type" attribute in model information display, with examples and explanations.


deepmd/pt/infer/deep_eval.py
```diff
@@ -648,6 +652,35 @@ def get_model_size(self) -> dict:
...
+        dict
+            A dictionary containing the information of type coverage in the model:
+            - 'type_num': the total number of covered types in this model.
+            - 'covered_type': a list of the covered types in this model.
+        """
+        buffers_dict = dict(self.dp.named_buffers())
+        type_map = np.array(self.type_map)
+        out_bias = None
+        for k in buffers_dict:
+            if ".out_bias" in k:
```
Chengqian-Zhang: In property fitting, all values in "xx.out_bias" are the same(the mean value of the label). So "covered_type" got by this method will always equal to "type_map" in property fitting. It is not correct. But I do not find a good solution to calculate the correct covered type in property fitting yet.

doc/model/show-model-info.md
```diff
@@ -60,6 +61,12 @@ Depending on the provided attributes and the model type, the output includes:
 
   - Prints the number of parameters for each component (`descriptor`, `fitting-net`, etc.), as well as the total parameter count.
 
+- **type-coverage**
```
njzjz: What is the difference between the type map and the type coverage? I think this word is a bit confusing.

iProzd: `type map` means the index mapping while `type coverage` means the real existing types covered by the model (assume that all existing types are included during statistics.)

deepmd/pt/infer/deep_eval.py
iProzd: Now only energy models are supported.

doc/model/show-model-info.md
iProzd: Renamed to `observed type`.

deepmd/pt/model/model/ener_model.py
```diff
@@ -44,6 +44,32 @@ def enable_hessian(self):
...
+        out_bias = self.atomic_model.get_out_bias()[0]
+
+        assert out_bias is not None, "No out_bias found in the model."
+        assert out_bias.dim() == 2, "The supported out_bias should be a 2D tensor."
+        assert out_bias.size(0) == len(type_map), (
+            "The out_bias shape does not match the type_map length."
+        )
+        bias_mask = torch.gt(torch.abs(out_bias), 1e-6).any(
+            dim=-1
+        )  # 1e-6 for stability
```
njzjz, line 65:
```suggestion
        ).detach().cpu()  # 1e-6 for stability
```

It's read below multiple times, so it's better to copy the tensor to the CPU first.

deepmd/pt/model/model/ener_model.py
```diff
@@ -44,6 +44,32 @@ def enable_hessian(self):
...
+
+        observed_type_list: list[str] = []
+        for i in range(len(type_map)):
+            if bias_mask[i]:
```
njzjz, line 69: Note: this needs to read the tensor from the CPU

PR #4826, opened 2025-06-30
pd: Add option for dynamic shape

HydrogenSulfate:
add flag for allow compiling dynamic input shape, if set to `0` or `False`, performance might be slightly enhanced for fixed shape. We set it to `True` for compativility.


HydrogenSulfate:
@njzjz hello, this PR is already for review.

njzjz (approved):

PR #4831, opened 2025-07-03
Update input_torch.json of DPA3 to fit DPA-3.1-3M.pt

Fillianore:
the e_sel is different from the size of 'DPA-3.1-3M.pt', while training, it causes error： 

`size mismatch for model.Default.atomic_model.descriptor.repflows.mean: copying a param with shape torch.Size([118, 1200, 4]) from checkpoint, the shape in current model is torch.Size([118, 120, 4]).`

now the e_sel in sample is change to 1200 to fit DPA-3.1-3M.pt

## Summary by CodeRabbit

* **Chores**
  * Updated configuration to increase the value of the "e_sel" parameter in the "repflow" section for improved performance or results.

### DIFF
--- a/examples/water/dpa3/input_torch.json
+++ b/examples/water/dpa3/input_torch.json
@@ -14,7 +14,7 @@
         "nlayers": 6,
         "e_rcut": 6.0,
         "e_rcut_smth": 5.3,
-        "e_sel": 120,
+        "e_sel": 1200,
         "a_rcut": 4.0,
         "a_rcut_smth": 3.5,
         "a_sel": 30,


iProzd:
Hi @Fillianore , thank you for submitting this PR. DPA3 operates in two primary modes controlled by the [`use_dynamic_sel`](https://docs.deepmodeling.com/projects/deepmd/en/stable/train/train-input.html#argument:model[standard]/descriptor[dpa3]/repflow/use_dynamic_sel) parameter:
- When set to `False`: See configuration example in [`examples/water/dpa3/input_torch.json`](https://github.com/deepmodeling/deepmd-kit/blob/devel/examples/water/dpa3/input_torch.json)
- When set to `True`: Refer to [`examples/water/dpa3/input_torch_dynamic.json`](https://github.com/deepmodeling/deepmd-kit/blob/devel/examples/water/dpa3/input_torch_dynamic.json).

The DPA-3.1-3M.pt model uses the dynamic selection mode (`use_dynamic_sel=True`), while there are still differences between its configuration and the water example, such as the switch function, activation, etc.

To obtain the exact configuration of the pretrained model for finetuning purposes, please consult the Bohrium notebook (https://www.bohrium.com/notebooks/57111746135) or AIS-Square model-card (https://www.aissquare.com/models/detail?pageType=models&name=DPA-3.1-3M&id=343) to see how to get the `input_finetune.json` for finetuning.

PR #4849, opened 2025-07-28
feat(pt): support zbl finetune

iProzd:
## Summary by CodeRabbit

* **New Features**
  * Added an option to control whether output statistics are computed or loaded across atomic models.

* **Bug Fixes**
  * More robust parameter transfer during fine‑tuning to handle renamed branches and missing pretrained keys.

* **Refactor**
  * Revised output-statistics workflow and refined per‑type output bias application in composite models.

* **Tests**
  * Simplified linear-model bias checks and added a ZBL finetuning test path.


deepmd/pt/model/atomic_model/base_atomic_model.py
```diff
@@ -363,6 +363,7 @@ def compute_or_load_stat(
         self,
         merged: Union[Callable[[], list[dict]], list[dict]],
         stat_file_path: Optional[DPPath] = None,
+        compute_out_stat: bool = True,
```
njzjz: Please note that the docs for this method needs to be changed. (Other  methods have the same issue)

njzjz (approved):

PR #4850, opened 2025-07-28
refactor(training): Average training loss for smoother and more representative logging

OutisLi:
This pull request modifies the training loop to improve the quality and readability of the reported training loss.

## Summary of Changes

Previously, the training loss and associated metrics (e.g., rmse_e_trn) reported in lcurve.out and the console log at each disp_freq step represented the instantaneous value from that single training batch. This could be quite noisy and subject to high variance depending on the specific batch sampled.

This PR introduces an accumulator for the training loss. The key changes are:

During each training step, the loss values are accumulated.

When a display step is reached, the accumulated values are averaged over the number of steps in that interval.

This averaged loss is then reported in the log and lcurve.out.

The accumulators are reset for the next interval.

The validation logic remains unchanged, continuing to provide a periodic snapshot of model performance, which is the standard and efficient approach.

## Significance and Benefits

Reporting the averaged training loss provides a much smoother and more representative training curve. The benefits include:

Reduced Noise: Eliminates high-frequency fluctuations, making it easier to see the true learning trend.

Improved Readability: Plotted learning curves from lcurve.out are cleaner and more interpretable.

Better Comparability: Simplifies the comparison of model performance across different training runs, as the impact of single-batch anomalies is minimized.

## A Note on Formatting

Please note that due to automatic code formatters (e.g., black, isort), some minor, purely stylistic changes may appear in the diff that are not directly related to the core logic.

## Summary by CodeRabbit

* **New Features**
  * Training loss values displayed during training are now averaged over the display interval, providing more stable and representative loss metrics for both single-task and multi-task modes.
  * Added an option to enable or disable averaging of training loss display via a new configuration setting.

* **Improvements**
  * Enhanced training loss reporting for improved monitoring and analysis during model training.

### DIFF
--- a/deepmd/utils/argcheck.py
+++ b/deepmd/utils/argcheck.py
@@ -3137,7 +3137,9 @@ def training_args(
     )
     doc_disp_training = "Displaying verbose information during training."
     doc_time_training = "Timing during training."
-    doc_disp_avg = "Display the average loss over the display interval for training sets."
+    doc_disp_avg = (
+        "Display the average loss over the display interval for training sets."
+    )
     doc_profiling = "Export the profiling results to the Chrome JSON file for performance analysis, driven by the legacy TensorFlow profiling API or PyTorch Profiler. The output file will be saved to `profiling_file`."
     doc_profiling_file = "Output file for profiling."
     doc_enable_profiler = "Export the profiling results to the TensorBoard log for performance analysis, driven by TensorFlow Profiler (available in TensorFlow 2.3) or PyTorch Profiler. The log will be saved to `tensorboard_log_dir`."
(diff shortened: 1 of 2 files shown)


deepmd/utils/argcheck.py
```diff
@@ -3213,6 +3216,7 @@ def training_args(
         Argument(
             "time_training", bool, optional=True, default=True, doc=doc_time_training
         ),
+        Argument("disp_avg", bool, optional=True, default=False, doc=doc_disp_avg),
```
njzjz:
```suggestion
        Argument("disp_avg", bool, optional=True, default=False, doc=doc_only_pt_supported + doc_disp_avg),
```

iProzd (approved):

PR #4855, opened 2025-07-31
Profile bug fix when both enable_profiler and profiling are set to true.

OutisLi:
When enable_profiler has been set to true, it will save the profiling results into json format in the tensorboard_log_dir. Currently, when both enable_profiler and profiling are set to true, it will throw an error `RuntimeError: Trace is already saved`. As a result, just ignore the profiling option when enable_profiler is true.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined profiling trace export logic to prevent duplicate or unintended trace exports when the new profiler is enabled. Profiling traces are now only exported if profiling is enabled and the new profiler is not active.
* **Documentation**
  * Updated profiling documentation to clarify that when the new profiler is enabled, profiling results are saved to TensorBoard logs instead of a Chrome JSON file.


caic99 (approved):

iProzd (approved):

deepmd/utils/argcheck.py
```diff
@@ -3137,7 +3137,7 @@ def training_args(
     )
     doc_disp_training = "Displaying verbose information during training."
     doc_time_training = "Timing during training."
-    doc_profiling = "Export the profiling results to the Chrome JSON file for performance analysis, driven by the legacy TensorFlow profiling API or PyTorch Profiler. The output file will be saved to `profiling_file`."
+    doc_profiling = "Export the profiling results to the Chrome JSON file for performance analysis, driven by the legacy TensorFlow profiling API or PyTorch Profiler. The output file will be saved to `profiling_file`. When enable_profiler is True, this option is ignored, since the profiling results will be saved to the TensorBoard log."
```
njzjz:
```suggestion
    doc_profiling = "Export the profiling results to the Chrome JSON file for performance analysis, driven by the legacy TensorFlow profiling API or PyTorch Profiler. The output file will be saved to `profiling_file`. In the PyTorch backend, when enable_profiler is True, this option is ignored, since the profiling results will be saved to the TensorBoard log."
```

PR #4858, opened 2025-08-01
feat: for user-input systems, it will recursively search for subsystems now.

OutisLi:
Previously, the systems in the input.json can either be a str containing a directory that has multiple systems to be searched or a list of direct path of systems.

Now it will also search for subsystems for each str in the list.

### DIFF
--- a/deepmd/utils/data_system.py
+++ b/deepmd/utils/data_system.py
@@ -737,6 +737,7 @@ def process_systems(
     """Process the user-input systems.
 
     If it is a single directory, search for all the systems in the directory.
+    If it is a list, each item can be either a system path or a directory to search.
     Check if the systems are valid.
 
     Parameters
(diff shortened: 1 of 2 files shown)


deepmd/utils/data_system.py
```diff
@@ -757,7 +758,15 @@ def process_systems(
         else:
             systems = rglob_sys_str(systems, patterns)
     elif isinstance(systems, list):
-        systems = systems.copy()
+        result_systems = []
+        for system in systems:
+            if isinstance(system, str):
```
njzjz, line 763: What's the other situations?

OutisLi: > What's the other situations?

The original code did not contain else, but I added that in the last commit. Thanks.

deepmd/utils/data_system.py
```diff
@@ -757,7 +758,17 @@ def process_systems(
...
+                # Try to expand as directory
+                expanded = expand_sys_str(system)
+                result_systems.extend(expanded)
```
njzjz, line 766: seems that it has the different logic from lines 756-759

deepmd/utils/data_system.py
```diff
@@ -757,7 +758,17 @@ def process_systems(
...
+            else:
```
njzjz, line 767: I don't get what situation will touch this case

PR #4869, opened 2025-08-05
fix: get correct intensive property prediction when using virtual atoms

Chengqian-Zhang:
When using virtual atoms, the property output of virtual atom is `0`.
- If predicting energy or other extensive properties, it works well, that's because the virtual atom property `0` do not contribute to the total energy or other extensive properties.
- However, if predicting intensive properties, there is some error. For example, a frame has two real atoms and two virtual atoms, the atomic property contribution is [2, 2, 0, 0](the atomic property of virtual atoms are always 0), the final property should be `(2+2)/real_atoms = 2`, not be `(2+2)/total_atoms =1`.

This PR is used to solve this bug mentioned above.

## Summary by CodeRabbit

* **New Features**
  * Models now provide accessors to retrieve property names and their fitting network; property fitting nets expose output definitions.

* **Bug Fixes**
  * Intensive property reduction respects atom masks so padded/dummy atoms are ignored, keeping results invariant to padding.

* **Tests**
  * Added PyTorch, JAX, and core tests validating consistent behavior with padded atoms.


njzjz (changes requested):
Please make the same change to other backends

Chengqian-Zhang:
I have change codes in `pt`, `dpmodel` and `jax` backend. However, `pd` and 'tf' are not changed because property fitting is not supported in the two backend.

iProzd (approved):

caic99 (approved):

deepmd/pt/model/model/transform_output.py
```diff
@@ -172,7 +173,20 @@ def fit_output_to_model_output(
         if vdef.reducible:
             kk_redu = get_reduce_name(kk)
             if vdef.intensive:
-                model_ret[kk_redu] = torch.mean(vv.to(redu_prec), dim=atom_axis)
+                if (mask is not None) and (
+                    mask == 0.0
+                ).any():  # containing padding atoms
```
njzjz: This will block the CPU to wait for the results of the GPU. Are you sure you want this?

deepmd/jax/model/base_model.py
```diff
@@ -46,7 +46,20 @@ def forward_common_atomic(
         atom_axis = -(len(shap) + 1)
         if vdef.reducible:
             kk_redu = get_reduce_name(kk)
-            model_predict[kk_redu] = jnp.sum(vv, axis=atom_axis)
+            if vdef.intensive:
+                mask = atomic_ret["mask"] if "mask" in atomic_ret else None
+                if (mask is not None) and (mask == 0.0).any():
```
njzjz: Also, due to the same reason, this cannot use JIT.

deepmd/pt/model/model/transform_output.py
Chengqian-Zhang: There is actually some error. I have changed the related code in each backend, do you think it can solve the problem you mentioned above?

Chengqian-Zhang: I find that all elements in `mask` is `1`(real atom) or `0`(virtual atom). So I make the code more simple.

njzjz (approved):

PR #4878, opened 2025-08-11
[WIP] pd: add unitest `test_deeppot_dpa_pd.cc`

HydrogenSulfate:
add unitest `test_deeppot_dpa_pd.cc`

## Summary by CodeRabbit

* **New Features**
  * Expanded inference interface to accept mapping and communication metadata; standardized full-graph execution flag handling.

* **Bug Fixes**
  * Improved runtime error reporting for missing output buffers with clearer messages.

* **Documentation**
  * Clarified the atomic-virial option via inline notes.

* **Tests**
  * Added CPU test suite validating DeepPot DPA (with and without periodic boundaries) across neighbor-list usage.

* **Chores**
  * Broadened pre-commit exclusions and updated CPU inference download source.


HydrogenSulfate:
@njzjz I found that the model exported from `deeppot_dpa.yaml` is different from the existing `deeppot_dpa.pth`. This causes Paddle to fail the `test_deeppot_dpa_pt.cc` unit test. Is there any way to obtain an inference model that is equivalent to `deeppot_dpa.pth`?

PR #4888, opened 2025-08-14
feat(dp/pt): add default_fparam

iProzd:
## Summary by CodeRabbit

* **New Features**
  * Optional default frame parameter (default_fparam) across fitting nets and model wrappers; new has_default_fparam() query exposed to users.

* **Behavior Changes**
  * fparam is no longer always required when a default exists — data loading, training, inference, and stats adapt to omit/use fparam; requirements are conditional on has_default_fparam().
  * TensorFlow paths accept/serialize default_fparam but disallow its runtime use.

* **Documentation**
  * Docstrings and serializers updated; serialization versions bumped.

* **Tests**
  * Tests extended to cover explicit vs default fparam flows.


deepmd/pt/infer/deep_eval.py
```diff
@@ -187,6 +187,10 @@ def get_dim_aparam(self) -> int:
         """Get the number (dimension) of atomic parameters of this DP."""
         return self.dp.model["Default"].get_dim_aparam()
 
+    def has_default_fparam(self) -> bool:
+        """Check if the model has default frame parameters."""
+        return self.dp.model["Default"].has_default_fparam()
```
njzjz: It seems to me this is a breaking change? Do you consider supporting old models?

deepmd/pt/model/task/fitting.py
```diff
@@ -375,6 +394,7 @@ def serialize(self) -> dict:
             "numb_fparam": self.numb_fparam,
             "numb_aparam": self.numb_aparam,
             "dim_case_embd": self.dim_case_embd,
+            "default_fparam": self.default_fparam,
```
njzjz, line 402: need to bump versions since a new parameter is introduced.

deepmd/pt/infer/deep_eval.py
iProzd: All pt/dp models are supported, other models returns False in default.

njzjz: Old models don't export this method (`has_default_fparam`).

deepmd/pt/utils/stat.py
```diff
@@ -62,6 +62,14 @@ def make_stat_input(
...
                     iterator = iter(dataloaders[i])
                     stat_data = next(iterator)
+                if (
+                    "find_fparam" in stat_data
+                    and "fparam" in stat_data
+                    and stat_data["find_fparam"] == 0.0
+                ):
+                    # for model using default fparam
+                    stat_data.pop("fparam")
+                    stat_data.pop("find_fparam")
```
Chengqian-Zhang, line 72: Why pop `fparam` and `find_fparam` here?

iProzd: When using model with default fparam to perform `change-bias`, stat sample should pop out zero fparam and using default fparam in fitting net. Another PR needed for implement the fparam stat pipeline, maybe after this PR.

PR #4893, opened 2025-08-15
feat: handle masked forces in test

caic99:


anyangml (approved):
LGTM

deepmd/entrypoints/test.py
```diff
@@ -417,8 +419,20 @@ def test_ener(
     mae_e = mae(diff_e)
     rmse_e = rmse(diff_e)
     diff_f = force - test_data["force"][:numb_test]
-    mae_f = mae(diff_f)
-    rmse_f = rmse(diff_f)
+    if find_atom_pref == 1:
+        atom_pref = test_data["atom_pref"][:numb_test]
```
caic99:
```suggestion
        atom_pref = test_data["atom_pref"][:numb_test].astype(bool)
```
@njzjz Is this what you mean?

njzjz: No, I mean `atom_pref` was not designed (when it was proposed in #41) as the mask. One may use it for other purposes and doesn't expect to see a masked result.

njzjz: perhaps the solution is providing both original results and the results multiply `atom_pref`

anyangml: I don’t see any reason why the current implementation cannot handle arbitrary weights. The only question is whether the weights should be considered when calculating the error; my intuition is that they should not.

anyangml: And in that case, we can simply convert them to bool type.

deepmd/entrypoints/test.py
```diff
@@ -453,8 +475,13 @@ def test_ener(
     if not out_put_spin and find_force == 1:
         log.info(f"Force  MAE         : {mae_f:e} eV/A")
         log.info(f"Force  RMSE        : {rmse_f:e} eV/A")
-        dict_to_return["mae_f"] = (mae_f, force.size)
-        dict_to_return["rmse_f"] = (rmse_f, force.size)
+        dict_to_return["mae_f"] = (mae_f, size_f)
+        dict_to_return["rmse_f"] = (rmse_f, size_f)
```
njzjz, line 469: Is it expected that the regular error `size_f` is affected by whether `atom_pref` is zero?

caic99: > Is it expected that the regular error `size_f` is affected by whether `atom_pref` is zero?

As discussed with @anyangml , `atom_pref` masks out elements with weight=0 when calculating the regular metrics, not the weighted one. @njzjz Is this expected?

anyangml: > Is it expected that the regular error `size_f` is affected by whether `atom_pref` is zero?

I think that's the expected behavior. In case where the force label is incorrect, and users use atom_pref=0 to mask out those atomic force labels, we should exclude those in calculating error.

njzjz: > In case where the force label is incorrect, and users use atom_pref=0 to mask out those atomic force labels, we should exclude those in calculating error.

This might be a historical issue. `atom_pref` was, at least in https://github.com/deepmodeling/dpgen/pull/52, not used as a mask, but just used to let some atoms have larger weights. We should not break the compatibility.

anyangml: i am fine with that. For users using atom_pref as a mask, they can simply use the weighted error.

njzjz (approved):

anyangml (approved):

iProzd (approved):

PR #4921, opened 2025-08-26
chore: Clarify CPU-specific thread warning

caic99:
clarify that thread configuration environment variables only apply to CPU

## Summary by CodeRabbit

* **New Features**
  * Device-aware default threading initialization across backends (CPU vs. GPU) for more sensible defaults.
  * Option to suppress CPU threading configuration warnings when running in CPU mode.
  * Clearer log messaging highlighting that certain thread settings apply only to CPU for best performance.
* **Documentation**
  * Updated parameter documentation to reflect the new CPU-mode option and clarify threading behavior.

### DIFF
--- a/deepmd/env.py
+++ b/deepmd/env.py
@@ -68,15 +68,21 @@ def set_env_if_empty(key: str, value: str, verbose: bool = True) -> None:
             )
 
 
-def set_default_nthreads() -> None:
+def set_default_nthreads(use_cpu: bool = False) -> None:
     """Set internal number of threads to default=automatic selection.
+
+    Parameters
+    ----------
+    use_cpu : bool, optional
+        If ``True``, suppress warnings about thread configuration,
+        by default ``False``.
 
     Notes
     -----
     `DP_INTRA_OP_PARALLELISM_THREADS` and `DP_INTER_OP_PARALLELISM_THREADS`
     control configuration of multithreading.
     """
-    if (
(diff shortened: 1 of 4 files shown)


deepmd/env.py
```diff
@@ -68,15 +68,21 @@ def set_env_if_empty(key: str, value: str, verbose: bool = True) -> None:
...
+        If ``True``, suppress warnings about thread configuration,
+        by default ``False``.
+
     Notes
     -----
     `DP_INTRA_OP_PARALLELISM_THREADS` and `DP_INTER_OP_PARALLELISM_THREADS`
     control configuration of multithreading.
     """
-    if (
+    if not use_cpu and (
```
njzjz, line 85:
```suggestion
    if use_cpu and (
```

deepmd/tf/env.py
```diff
@@ -263,7 +263,7 @@ def get_tf_session_config() -> Any:
     Any
         session configure object
     """
-    set_tf_default_nthreads()
+    set_tf_default_nthreads(use_cpu=os.environ.get("DEVICE") == "cpu")
```
njzjz, line 266: I don't think this environment variable is used in the TensorFlow backend.

njzjz (changes requested):
#1284 shows that setting `OMP_NUM_THREADS` is still meaningful in the GPU training. I am wondering why we dismiss the warning.

PR #4928, opened 2025-08-27
fix(finetune): calculate fitting stat when using random fitting in finetuning process

Chengqian-Zhang:
In finetuing process, the computation of fitting stat is skipped in previous code. There are two situations:
1. Finetuning from pretrained model's branch: it means pretrained model also has `fparam` or `aparam` which has the same meaning of finetuning task. The key `fparam_avg`/`fparam_inv_std`/ `aparam_avg`/`aparam_inv_std` load from the pretrained model. It is correct.
2. Finetuning using RANDOM fitting. The fitting stat should be calculated in this situation. But the computation of fitting stat is skipped now. There is some error.

### DIFF
--- a/deepmd/pt/model/atomic_model/base_atomic_model.py
+++ b/deepmd/pt/model/atomic_model/base_atomic_model.py
@@ -487,6 +487,26 @@ class BaseAtomicModel(torch.nn.Module, BaseAtomicModel_):
         else:
             raise RuntimeError("Unknown bias_adjust_mode mode: " + bias_adjust_mode)
 
+    def compute_fitting_stat(
+        self,
+        sample_merged
+    ) -> None:
+        """Compute the input statistics (e.g. mean and stddev) for the fittings from packed data..
+
+        Parameters
+        ----------
+        sample_merged : Union[Callable[[], list[dict]], list[dict]]
+            - list[dict]: A list of data samples from various data systems.
+                Each element, `merged[i]`, is a data dictionary containing `keys`: `torch.Tensor`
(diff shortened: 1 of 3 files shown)


deepmd/pt/model/atomic_model/base_atomic_model.py
```diff
@@ -493,6 +493,24 @@ def change_out_bias(
         else:
             raise RuntimeError("Unknown bias_adjust_mode mode: " + bias_adjust_mode)
 
+    def compute_fitting_stat(
```
wanghan-iapcm: I would recommend changing the method name to `compute_fitting_input_stat`.

wanghan-iapcm:
Limitation: the imput stat is not implemented in the python backend.

deepmd/pt/model/atomic_model/base_atomic_model.py
```diff
@@ -493,6 +493,24 @@ def change_out_bias(
...
+        self,
+        sample_merged: Union[Callable[[], list[dict]], list[dict]],
+    ) -> None:
+        """Compute the input statistics (e.g. mean and stddev) for the fittings from packed data.
```
njzjz:
```suggestion
        """Compute the input statistics (e.g. mean and stddev) for the atomic model from packed data.
```

deepmd/pt/model/atomic_model/base_atomic_model.py
Chengqian-Zhang: Have changed.

iProzd (approved):

PR #5038, opened 2025-11-08
fix(stat): Caculate correct fitting stat when using default fparam and using share fitting.

Chengqian-Zhang:
In this PR:
1. Support writing fitting stat to `stat_file` and loading fitting stat from `stat_file`
2. Ensure the fitting stat calculate is correct when using `default_fparam`
3. Support sharing fitting stat when using `share_fitting` in multitask mode.
4. Print the process of calculating fitting stat to the board via `log.info`.

## Summary by CodeRabbit

* **New Features**
  * Default frame parameters auto-fill missing samples and are exposed via a new accessor.
  * Compute/load/save per-parameter statistics to/from optional stat files.
  * Multitask training adds probability-weighted parameter sharing and propagates default fparam into data requirements.
  * Statistic items support scalar scaling for aggregation.

* **Refactor**
  * Parameter-sharing and statistic-propagation flows reorganized for consistent buffering and persistence.

* **Tests**
  * New tests and test data for stats computation, persistence, and multitask sharing.

### DIFF
--- a/deepmd/pd/model/task/fitting.py
+++ b/deepmd/pd/model/task/fitting.py
@@ -40,6 +40,9 @@ from deepmd.utils.finetune import (
     get_index_between_two_maps,
     map_atom_exclude_types,
 )
+from deepmd.utils.path import (
+    DPPath,
+)
 
 dtype = env.GLOBAL_PD_FLOAT_PRECISION
 device = env.DEVICE
@@ -76,6 +79,7 @@ class Fitting(paddle.nn.Layer, BaseFitting):
         self,
         merged: Union[Callable[[], list[dict]], list[dict]],
         protection: float = 1e-2,
+        stat_file_path: Optional[DPPath] = None,
     ) -> None:
         """
         Compute the input statistics (e.g. mean and stddev) for the fittings from packed data.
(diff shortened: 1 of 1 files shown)


wanghan-iapcm (approved):

deepmd/utils/env_mat_stat.py
```diff
@@ -48,6 +48,13 @@ def __add__(self, other: "StatItem") -> "StatItem":
             squared_sum=self.squared_sum + other.squared_sum,
         )
 
+    def __mul__(self, scalar: float) -> "StatItem":
+        return StatItem(
+            number=self.number * scalar,
+            sum=self.sum * scalar,
+            squared_sum=self.squared_sum * scalar,
+        )
```
njzjz, line 58: There are some type issues here:
`number` is `int` and `scalar` is `float`, int * float = float, so it cannot be assigned to number (expected an int),

Chengqian-Zhang: When using `share_fitting`, the fitting input stat will be weighted by `model_prob`. So `StatItem.number` may become `float`.
I change the type of `number` from `int` to `float`. Do you think it is resonable?

iProzd (approved):

PR #5050, opened 2025-11-13
test(common): add regression for atom type remap

OutisLi:
## Summary
  - add a regression test that reproduces the original atom-type remapping failure when the training system contains fewer atom types than the provided type_map
  - ensure the fix from commit 4ec04370fd5c7c5bd4bbdbd6f7a30de8e7c3fff0 remains covered going forward

## Summary by CodeRabbit

* **Tests**
  * Added unit tests for DeepmdData that validate atom-type remapping, handling of unused types, and correct loading/formatting of type arrays.

### DIFF
--- a/source/tests/common/test_deepmd_data.py
+++ b/source/tests/common/test_deepmd_data.py
@@ -20,15 +20,15 @@ class TestDeepmdDataTypeMap(unittest.TestCase):
         self.set_dir.mkdir()
 
         # minimal required dataset
-        self._atom_types = np.array([0, 1, 0, 1], dtype=np.int32)
-        np.savetxt(self.root / "type.raw", self._atom_types, fmt="%d")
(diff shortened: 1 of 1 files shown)


source/tests/common/test_deepmd_data.py
```diff
@@ -0,0 +1,48 @@
...
+
+class TestDeepmdDataTypeMap(unittest.TestCase):
+    def setUp(self) -> None:
+        self.tmpdir = tempfile.TemporaryDirectory()
+        self.root = Path(self.tmpdir.name)
+        self.set_dir = self.root / "set.000"
+        self.set_dir.mkdir()
+
+        # minimal required dataset
+        self._atom_types = np.array([0, 1, 0, 1], dtype=np.int32)
```
njzjz: why set self

OutisLi: removed

iProzd (approved):

PR #5055, opened 2025-11-15
feat(pt): Add support for SiLU activation function in gradient calculations

OutisLi:
- Introduced the SiLU (Sigmoid Linear Unit) activation function with corresponding gradient and second derivative calculations.
- Updated the activation function mapping to include SiLU, enhancing the flexibility of activation functions available in the DPTabulate class.

## Summary by CodeRabbit

* **New Features**
  * Added support for the SiLU (Swish) activation across runtime computations and gradients.
* **Tests**
  * Expanded test coverage to validate all supported activations (tanh, gelu, relu, relu6, softplus, sigmoid, silu) and their first/second derivatives across execution paths.
* **Documentation**
  * Updated activation references and lists to include SiLU among supported options.

### DIFF
--- a/source/tests/pt/test_tabulate.py
+++ b/source/tests/pt/test_tabulate.py
@@ -4,6 +4,9 @@ import unittest
 import numpy as np
 import torch
 
+from deepmd.dpmodel.utils.network import (
+    get_activation_fn,
+)
 from deepmd.pt.utils import (
     env,
 )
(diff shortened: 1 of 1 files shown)


deepmd/pt/utils/tabulate.py
```diff
@@ -84,6 +84,7 @@ def __init__(
             "relu6": 4,
             "softplus": 5,
             "sigmoid": 6,
+            "silu": 7,
```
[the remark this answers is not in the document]
OutisLi: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/deepmodeling/deepmd-kit/pull/5055#discussion_r2529720144)

source/tests/pt/test_tabulate.py
```diff
@@ -19,6 +19,41 @@
 )
 
 
+def get_activation_function(functype: int):
```
njzjz: Why not using the existing implementation?

OutisLi: The original test use np.tanh, so I implemented other ac funcs using numpy

njzjz: What is the difference between this method and `deepmd.dpmodel.utils.network.get_activation_fn`?

OutisLi: fixed

njzjz (approved):

PR #5082, opened 2025-12-04
Fix: dptest label mismatch

anyangml:
Currently, when running dptest on atomic polar. The order of atoms for `coord` and `label` are misaligned. 
This is a result of type_sel = None for coord, but not None for label. 
I tried to set it to None for label as well, but it broke TF compatibility.
So the easiest hack I can think of is to add a check while loading data.

## Summary by CodeRabbit

* **Tests**
  * Enhanced output handling in test suite to improve per-atom count tracking in selection outputs.


deepmd/utils/data.py
njzjz, line 1: We have already had an option in this function, `output_natoms_for_type_sel`. You could set it to true for your case.

anyangml: @copilot code review[agent] address this comment

iProzd (approved):

njzjz (approved):

Issue #5090, opened 2025-12-11
[BUG] pt model converted from tf model cannot be used to init training


ChiahsinChu:
### Bug summary

When I try to use `dp --pt train input.json --init-frz-model pt_model.pth` to init training with the pt model converted from tf model (i.e., by `dp convert-backend tf_model.pb pt_model.pth`), I get the error of missing keys in the stat_dict. Based on the error log, it should be a general issue for all kind of models rather than the dipole model I tried.

### DeePMD-kit Version

v3.1.1-52-gd774edea

### Backend and its version

TensorFlow v2.18.0-rc2-4-g6550e4bd802; PyTorch v2.5.1+cu121-ga8d6afb511a

### How did you download the software?

Built from source

### Input Files, Running Commands, Error Log, etc.

Input files:
Input files are adapted from `examples/water_tensor/dipole`

Running command:
```bash
dp --pt train input.json --init-frz-model ../00.tf_training/dw_model.pth 1>dp_train.stdout 2>dp_train.stderr
```

Error Log:
```bash
Traceback (most recent call last):
  File "/home/jxzhu/apps/miniconda3/envs/deepmd-devel/bin/dp", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/jxzhu/apps/deepmd/devel/deepmd/main.py", line 1020, in main
    deepmd_main(args)
  File "/home/jxzhu/apps/miniconda3/envs/deepmd-devel/lib/python3.11/site-packages/torch/distributed/elastic/multiprocessing/errors/__init__.py", line 355, in wrapper
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/home/jxzhu/apps/deepmd/devel/deepmd/pt/entrypoints/main.py", line 536, in main
    train(
  File "/home/jxzhu/apps/deepmd/devel/deepmd/pt/entrypoints/main.py", line 346, in train
    trainer = get_trainer(
              ^^^^^^^^^^^^
  File "/home/jxzhu/apps/deepmd/devel/deepmd/pt/entrypoints/main.py", line 193, in get_trainer
    trainer = training.Trainer(
              ^^^^^^^^^^^^^^^^^
  File "/home/jxzhu/apps/deepmd/devel/deepmd/pt/train/training.py", line 617, in __init__
    self.model.load_state_dict(frz_model.state_dict())
[...]

PR #5091, opened 2025-12-11
feat(pt): allow `--init-frz-model` for pt model converted from tf model

ChiahsinChu:
- Add try-catch block for loading frozen models
- Use strict=False when state_dict keys don't match
- Log warnings for model state_dict mismatches
- Prevent crashes when loading models with different architectures"

Fix bug in #5090 

## Summary by CodeRabbit

* **Bug Fixes**
  * Frozen model loading now uses non‑strict mode; missing or unexpected parameters are logged as warnings instead of causing failure.

* **Tests**
  * Added an end-to-end test covering conversion and loading of frozen models from external formats.
  * Test I/O standardized to UTF‑8; teardown extended to remove generated artifacts.

* **New Files**
  * Added a JSON test model configuration for validation and training workflows.

### DIFF
--- a/source/tests/pt/test_init_frz_model.py
+++ b/source/tests/pt/test_init_frz_model.py
@@ -159,7 +159,18 @@ class TestInitFrzModel(unittest.TestCase):
         config["training"]["numb_steps"] = 0
 
         trainer = get_trainer(config, init_frz_model=frozen_model)
+        # Explicit assertions to make test success criteria clear
+        self.assertIsNotNone(trainer, "Trainer should be successfully initialized")
+        self.assertTrue(
+            hasattr(trainer, "model"), "Trainer should have a model attribute"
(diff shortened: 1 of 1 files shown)


deepmd/pt/train/training.py
wanghan-iapcm, line 1: Please consider the following alternative suggested by AI:

```python
state = frz_model.state_dict()
missing, unexpected = self.model.load_state_dict(state, strict=False)

if missing or unexpected:
    log.warning(
        "Checkpoint loaded non-strictly. "
        f"Missing keys: {missing}, Unexpected keys: {unexpected}"
    )
```

ChiahsinChu: Solved.

iProzd (approved):

PR #5120, opened 2026-01-03
feat(tf): optimize data modifier calling in deepeval

ChiahsinChu:
# feat(tf): optimize data modifier calling in deepeval

This PR includes refactoring changes to improve the modifier architecture in the DeepMD TensorFlow implementation.

### 1. deepmd/tf/infer/deep_eval.py
- Refactored the modifier initialization logic to use a more generic approach
- Replaced hardcoded DipoleChargeModifier initialization with a dynamic class resolution using BaseModifier.get_class_by_type()
- Simplified the modifier parameter extraction by moving it to the modifier class itself
- Fixed atom_types reshaping issue in the _get_natoms_and_nframes method

### 2. deepmd/tf/modifier/dipole_charge.py
- Added a new static method get_params() to handle parameter extraction from the model
- This method encapsulates the logic for retrieving modifier-specific parameters from the model's tensors
- Improved charge map parsing with proper type conversion

## Summary

These changes make the modifier system more extensible by:
- Introducing a plugin-like architecture for modifiers
- Moving modifier-specific initialization logic to the respective modifier classes
- Making it easier to add new modifier types in the future without modifying the core DeepEval class
- Fixing a bug in atom type handling

The refactoring follows the DRY principle and improves code maintainability while preserving existing functionality.

## Summary by CodeRabbit

* **Refactor**
  * Switched to a unified, dynamic modifier system for modular handling.
  * Normalized atom-count derivation to support multi-frame and varied input shapes.
  * Added module-level logging and stricter error handling that prevents silent state mutation on modifier load failure.
* **New Features**
  * Added a standard API for extracting modifier parameters from a frozen model for consistent parameter retrieval across modifiers.

### DIFF
--- a/deepmd/tf/infer/deep_eval.py
+++ b/deepmd/tf/infer/deep_eval.py
@@ -149,9 +149,16 @@ class DeepEval(DeepEvalBackend):
 
         self.dm = None
         if self.modifier_type is not None:
-            modifier = BaseModifier.get_class_by_type(self.modifier_type)
-            modifier_params = modifier.get_params_from_frozen_model(self)
-            self.dm = modifier.get_modifier(modifier_params)
+            try:
+                modifier = BaseModifier.get_class_by_type(self.modifier_type)
+                modifier_params = modifier.get_params_from_frozen_model(self)
+                self.dm = modifier.get_modifier(modifier_params)
+            except Exception as exc:
+                raise RuntimeError(
+                    f"Failed to load data modifier '{self.modifier_type}'. "
+                    f"Use skip_modifier=True to load the model without the modifier. "
+                    f"Error: {exc}"
+                ) from exc
 
     def _init_tensors(self) -> None:
         tensor_names = {


deepmd/tf/infer/deep_eval.py
```diff
@@ -137,37 +140,23 @@ def __init__(
...
+                modifier = BaseModifier.get_class_by_type(self.modifier_type)
+                modifier_params = modifier.get_params_from_frozen_model(self)
+                self.dm = modifier.get_modifier(modifier_params)
+            except Exception as exc:
+                log.warning(
+                    f"Failed to load data modifier '{self.modifier_type}'. "
+                    "The model will be loaded without the data modifier. "
+                    f"Error details: {exc}"
+                )
+                self.modifier_type = None
```
njzjz: Why are errors ignored here without an exception?

ChiahsinChu: For the modifier `dipole_charge`, the errors mainly come from the loading of DW model. The DPLR model finds the DW model based on the training setup, which might fail when the DW model is an absolute path. With this exception, the users can wrap up the data modifier by themselves. 
In fact, the DW model (or more precisely, the inference of long-range correction) can be loaded from the DPLR energy model (https://gist.github.com/ChiahsinChu/ef273e024d8f020c65d21f6761285e5a), as in the lammps interface. However, there is another feature and I prefer another PR.

njzjz: I am worried that users may expect the DW model to be loaded, and suppressing the error is unsafe. The result is different with and without the DW model.

At least, it should not be just a warning. It may be better to add an option.

ChiahsinChu: > It may be better to add an option.

Good suggestion. I have made the revision, and now the user can skip the modifier by turn on the `skip_modifier` when init DeepEval.

PR #5147, opened 2026-01-11
feat(pt): add parameter numbers output

OutisLi:
## Summary by CodeRabbit

* **Chores**
  * Enhanced runtime logging during training initialization to report total and trainable parameter counts. Includes per-model breakdowns for multi-task configurations so model-size details are visible in logs for easier diagnostics and monitoring.

### DIFF
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -765,13 +765,29 @@ class Trainer:
         if self.rank == 0:
             self._log_parameter_count()
 
+    @staticmethod
+    def _count_parameters(model: torch.nn.Module) -> tuple[int, int]:
+        """
+        Count model parameters.
+
+        Parameters
+        ----------
+        model : torch.nn.Module
+            The model to count parameters for.
+
+        Returns
+        -------
+        tuple[int, int]
+            A tuple of (trainable, total) parameter counts.
+        """
+        trainable = sum(p.numel() for p in model.parameters() if p.requires_grad)
+        total = sum(p.numel() for p in model.parameters())
+        return trainable, total
+
     def _log_parameter_count(self) -> None:
         """Log model parameter count."""
         if not self.multi_task:
-            trainable = sum(
-                p.numel() for p in self.model.parameters() if p.requires_grad
-            )
-            total = sum(p.numel() for p in self.model.parameters())
+            trainable, total = self._count_parameters(self.model)
             log.info(
                 f"Model Params:  {total / 1e6:.3f} M   (Trainable: {trainable / 1e6:.3f} M)"
             )
@@ -781,11 +797,7 @@ class Trainer:
                 "The following per-task counts may include duplicates."
             )
             for model_key in self.model_keys:
-                model = self.model[model_key]
-                trainable = sum(
-                    p.numel() for p in model.parameters() if p.requires_grad
-                )
-                total = sum(p.numel() for p in model.parameters())
+                trainable, total = self._count_parameters(self.model[model_key])
                 log.info(
                     f"Model Params [{model_key}]: {total / 1e6:.3f} M   (Trainable: {trainable / 1e6:.3f} M)"
                 )


deepmd/pt/train/training.py
```diff
@@ -743,6 +743,55 @@ def warm_up_linear(step: int, warmup_steps: int) -> float:
...
+                    if model_type == "zbl":
+                        # ZBL model: get descriptor type from the DP sub-model
+                        models_data = serialized.get("models", [])
+                        if models_data and isinstance(models_data[0], dict):
+                            descriptor_data = models_data[0].get("descriptor", {})
+                            if isinstance(descriptor_data, dict):
+                                desc_type = descriptor_data.get("type", "UNKNOWN")
+                                return f"{str(desc_type).upper()} (with ZBL)"
+                        return "UNKNOWN (with ZBL)"
+            return "UNKNOWN"
```
njzjz: To print the model information, `print(model)` will print all modules inside the model.

deepmd/pt/train/training.py
```diff
@@ -743,6 +743,55 @@ def warm_up_linear(step: int, warmup_steps: int) -> float:
...
+
+        def count_parameters(model: torch.nn.Module) -> int:
+            """Count the total number of trainable parameters."""
+            return sum(p.numel() for p in model.parameters() if p.requires_grad)
```
iProzd: This only counts all the **trainable** params in the model, not ALL params. Either use "Model Trainable Params" or count ALL params.

iProzd: Could this be merged with function `get_model_size` [here](https://github.com/deepmodeling/deepmd-kit/blob/master/deepmd/pt/infer/deep_eval.py#L705) ?

iProzd (changes requested):

deepmd/pt/train/training.py
OutisLi: I want to print the registered name, but print can only get the class name, which may not be suitable to viewers

deepmd/pt/train/training.py
OutisLi: get_model_size is in DeepEval, not in EnergyModel (BaseModel). However, get_model_size can be implemented in BaseModel in the future

OutisLi: now display both

deepmd/pt/train/training.py
```diff
@@ -761,6 +761,31 @@ def warm_up_linear(step: int, warmup_steps: int) -> float:
...
+        if not self.multi_task:
+            trainable = sum(
+                p.numel() for p in self.model.parameters() if p.requires_grad
+            )
+            total = sum(p.numel() for p in self.model.parameters())
+            log.info(
+                f"Model Params:  {total / 1e6:.3f} M   (Trainable: {trainable / 1e6:.3f} M)"
+            )
+        else:
+            for model_key in self.model_keys:
```
iProzd, line 799: When in multitask, params will be shared across different tasks. While they are counted several times here. If it's not easy to count the actually total params here, at least a warning should be added in case of miss leading for users.

deepmd/pt/train/training.py
```diff
@@ -761,6 +761,35 @@ def warm_up_linear(step: int, warmup_steps: int) -> float:
...
+            log.warning(
+                "In multitask mode, parameters may be shared across tasks. "
+                "The following per-task counts may include duplicates."
+            )
+            for model_key in self.model_keys:
+                model = self.model[model_key]
+                trainable = sum(
+                    p.numel() for p in model.parameters() if p.requires_grad
+                )
+                total = sum(p.numel() for p in model.parameters())
```
njzjz: Lines 785-788 are duplicated as lines 771-774. They should be considered as a function.

OutisLi: done

iProzd (approved):

PR #5148, opened 2026-01-11
feat: use num_epoch to set num_steps

OutisLi:
## Summary by CodeRabbit

* **New Features**
  * Per-model epoch targets (num_epoch, num_epoch_dict) with automatic stop-step derivation and automatic per-task probability/step allocation for multi-task training.

* **Bug Fixes**
  * Stronger validation and clearer errors/logging for epoch/step resolution and sampler totals; safer defaults when configs are missing.

* **Documentation**
  * Updated training argument docs and aliases; clarified mutual-exclusion rules for num_steps/num_epoch/num_epoch_dict/model_prob.

* **Tests**
  * Added deterministic single- and multi-task sampling tests validating derived vs explicit stepping.


deepmd/utils/argcheck.py
```diff
@@ -3286,7 +3306,17 @@ def training_args(
...
         Argument(
-            "numb_steps", int, optional=False, doc=doc_numb_steps, alias=["stop_batch"]
+            "numb_steps",
+            int,
+            optional=True,
+            doc=doc_numb_steps,
+            alias=["stop_batch", "num_steps"],
+        ),
+        Argument(
+            "num_epoch",
```
Chengqian-Zhang: Why not "num**b**_epoch(s)", which is consistent to "numb_steps"?

deepmd/pt/train/training.py
```diff
@@ -430,6 +491,56 @@ def get_lr(lr_params: dict[str, Any]) -> BaseLR:
...
+                        sampler_weights,
+                    )
+                )
+            self.model_prob = resolve_model_prob(
+                self.model_keys,
+                training_params.get("model_prob"),
+                training_data,
+            )
+            total_numb_batch = int(
+                np.ceil(np.sum(np.asarray(per_task_total) * self.model_prob))
```
Chengqian-Zhang: What is the definition of **an epoch** in multi-task mode? To be honest, I do not understand `np.ceil(np.sum(np.asarray(per_task_total) * self.model_prob))`. If we want to run **an epoch** for **every** model branch, maybe `np.ceil(np.max(per_task_total / self.model_prob))` is correct?

Chengqian-Zhang (changes requested):

deepmd/utils/argcheck.py
OutisLi: numb is too strange

deepmd/pt/train/training.py
OutisLi: For multitask, per-task total_numb_batch values are computed as single task, and the final total_numb_batch is their model_prob-weighted sum as described in argcheck.py. np.ceil(np.max(per_task_total / self.model_prob)) can be used if the definition of an epoch is as you said. Since I haven't trained any multitask job, I do not know which is better

iProzd (changes requested):
There are common methods in pt and pd which can be merged.

Chengqian-Zhang (approved):

njzjz (approved):

deepmd/pt/train/training.py
```diff
@@ -436,6 +444,74 @@ def get_lr(lr_params: dict[str, Any]) -> BaseLR:
...
                     )
 
+        # Resolve training steps
+        per_task_total = []
+        if not self.multi_task:
+            sampler_weights = to_numpy_array(self.training_dataloader.sampler.weights)
+            total_numb_batch = compute_total_numb_batch(
+                training_data.index,
+                sampler_weights,
+            )
```
iProzd: This should only be calculated when `num_steps` is None

deepmd/pt/train/training.py
```diff
@@ -436,6 +444,74 @@ def get_lr(lr_params: dict[str, Any]) -> BaseLR:
...
+            for model_key in self.model_keys:
+                sampler_weights = to_numpy_array(
+                    self.training_dataloader[model_key].sampler.weights
+                )
+                per_task_total.append(
+                    compute_total_numb_batch(
+                        training_data[model_key].index,
+                        sampler_weights,
+                    )
+                )
```
iProzd: Same as above.

deepmd/pt/train/training.py
```diff
@@ -658,19 +734,13 @@ def single_model_finetune(
...
-            sum_prob = np.sum(self.model_prob)
-            assert sum_prob > 0.0, "Sum of model prob must be larger than 0!"
-            self.model_prob = self.model_prob / sum_prob
+        if self.multi_task and self.model_prob is None:
+            self.model_prob = resolve_model_prob(
+                self.model_keys,
+                training_params.get("model_prob"),
+                training_data,
+                rank=self.rank,
+            )
```
iProzd: Is L737-L743 necessary?

iProzd (changes requested):
Example inputs with `num_epoch` / `num_epoch_dict` should be added in `examples`, also with test in `source/tests/common/test_examples.py`.

iProzd (approved):

PR #5149, opened 2026-01-12
feat(pt): add HybridMuonOptimizer

OutisLi:
## Summary by CodeRabbit

* **New Features**
  * Added HybridMuon optimizer option with configurable hyperparameters (momentum alias, lr_adjust, lr_adjust_coeff, min_2d_dim, muon_2d_only, weight decay) and integrated training support (initialization, LR tracking, scheduler compatibility, checkpoint resume).
* **Tests**
  * Added comprehensive tests covering orthogonalization, optimizer step behavior, weight-decay effects, fallback cases, LR-adjust modes, and state save/load.

### DIFF
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -740,29 +740,29 @@ class Trainer:
             self.optimizer = AdaMuonOptimizer(
                 self.wrapper.parameters(),
                 lr=self.lr_exp.start_lr,
-                momentum=float(self.opt_param.get("momentum", 0.95)),
-                weight_decay=float(self.opt_param.get("weight_decay", 0.001)),
+                momentum=float(self.opt_param["momentum"]),
+                weight_decay=float(self.opt_param["weight_decay"]),
                 adam_betas=(
-                    float(self.opt_param.get("adam_beta1", 0.9)),
-                    float(self.opt_param.get("adam_beta2", 0.95)),
+                    float(self.opt_param["adam_beta1"]),
+                    float(self.opt_param["adam_beta2"]),
                 ),
-                lr_adjust=float(self.opt_param.get("lr_adjust", 10.0)),
-                lr_adjust_coeff=float(self.opt_param.get("lr_adjust_coeff", 0.2)),
+                lr_adjust=float(self.opt_param["lr_adjust"]),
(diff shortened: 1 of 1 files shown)


deepmd/pt/train/training.py
```diff
@@ -726,6 +730,20 @@ def warm_up_linear(step: int, warmup_steps: int) -> float:
                 lr_adjust=float(self.opt_param.get("lr_adjust", 10.0)),
                 lr_adjust_coeff=float(self.opt_param.get("lr_adjust_coeff", 0.2)),
             )
+        elif self.opt_type == "Muon":
+            self.optimizer = MuonOptimizer(
```
njzjz: Why not use [`torch.optim.Muon`](https://docs.pytorch.org/docs/stable/generated/torch.optim.Muon.html), just like `Adam`?

wanghan-iapcm: If the current implementation deviates from the standard Muon optimizer architecture, I recommend renaming it to avoid confusion and clearly distinguish its unique functionality.

OutisLi: I coupled Muon with Adam, the native Muon do not have this. After my experiments, my realization is better in efficiency and performance.

deepmd/pt/train/training.py
```diff
@@ -169,6 +171,10 @@ def get_opt_param(params: dict[str, Any]) -> tuple[str, dict[str, Any]]:
                 "momentum": params.get("momentum", 0.95),
                 "adam_beta1": params.get("adam_beta1", 0.9),
                 "adam_beta2": params.get("adam_beta2", 0.95),
+                "lr_adjust": params.get("lr_adjust", 10.0),
+                "lr_adjust_coeff": params.get("lr_adjust_coeff", 0.2),
+                "muon_2d_only": params.get("muon_2d_only", True),
+                "min_2d_dim": params.get("min_2d_dim", 1),
             }
```
njzjz, line 178: I don't find the meaning of this dict. It seems that `opt_param` is just a copy of `params`.

OutisLi: This requires removing the whole opt_param

wanghan-iapcm (approved):

PR #5154, opened 2026-01-15
refactor: unify learning rate schedulers with array API

OutisLi:
- Refactor BaseLR in dpmodel to use array_api_compat for backend-agnostic implementation
- Consolidate learning rate logic from TF/PT/PD backends into unified dpmodel layer
- Use array API operations (xp.where, xp.clip, etc.) for JIT compatibility across backends
- Add warmup support (warmup_steps, warmup_ratio, warmup_start_factor) during refactoring
- Add stop_ratio parameter as alternative to stop_lr for flexible configuration
- Implement mutual exclusion validation for stop_lr/stop_ratio and warmup_steps/warmup_ratio
- Update all backends to use unified BaseLR implementation
- Add comprehensive consistency tests across NumPy/PyTorch/JAX/array_api_strict backends

## Release Notes

* **New Features**
  * Added comprehensive warmup support for learning rate schedules with configurable warmup steps, ratios, and start factors.
  * Enhanced learning rate scheduling with unified configuration across TensorFlow, PyTorch, and Paddle backends.
  * Introduced flexible stop learning rate configuration using either absolute values or ratios.

* **Improvements**
  * Moved warmup configuration from training to learning rate settings for consistency.
  * Added automatic migration of legacy warmup settings for backward compatibility.
  * Expanded cosine annealing schedule support with proper warmup integration.

* **Documentation**
  * Added comprehensive learning rate scheduling documentation with examples.

### DIFF
--- a/deepmd/dpmodel/utils/learning_rate.py
+++ b/deepmd/dpmodel/utils/learning_rate.py
@@ -16,6 +16,9 @@ from deepmd.common import (
 from deepmd.dpmodel.array_api import (
     Array,
 )
+from deepmd.env import (
+    GLOBAL_NP_FLOAT_PRECISION,
+)
 from deepmd.utils.plugin import (
     PluginVariant,
     make_plugin_registry,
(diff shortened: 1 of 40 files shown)


deepmd/dpmodel/utils/learning_rate.py
```diff
@@ -118,29 +283,74 @@ class LearningRateCosine(BaseLR):
...
+            start_lr=start_lr,
+            stop_lr=stop_lr,
+            stop_ratio=stop_ratio,
+            stop_steps=stop_steps,
+            warmup_steps=warmup_steps,
+            warmup_ratio=warmup_ratio,
+            warmup_start_factor=warmup_start_factor,
+            **kwargs,
+        )
+        self.lr_min_factor = self.stop_lr / self.start_lr
```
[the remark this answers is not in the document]
OutisLi: start_lr cannot be 0

deepmd/utils/argcheck.py
```diff
@@ -2480,14 +2480,159 @@ def linear_ener_model_args() -> Argument:
...
+        "Mutually exclusive with warmup_steps."
+    )
+    doc_warmup_start_factor = (
+        "The factor of start_lr for the initial warmup learning rate. "
+        "The warmup learning rate starts from warmup_start_factor * start_lr. "
+        "Default is 0.0, meaning the learning rate starts from zero."
+    )
+
+    args = [
+        Argument("start_lr", float, optional=False, doc=doc_start_lr),
```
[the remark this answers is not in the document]
OutisLi: @iProzd How do you think about this. I think start_lr should be set explicitly

deepmd/pt/train/training.py
```diff
@@ -725,7 +697,8 @@ def warm_up_linear(step: int, warmup_steps: int) -> float:
                 self.optimizer.load_state_dict(optimizer_state_dict)
             self.scheduler = torch.optim.lr_scheduler.LambdaLR(
                 self.optimizer,
-                lambda step: warm_up_linear(step + self.start_step, self.warmup_steps),
+                lambda step: self.lr_exp.value(step + self.start_step)
+                / self.lr_exp.start_lr,
```
wanghan-iapcm: I would suggest providing a method for accessing the start_lr, rather than directly reads the data of the object.

wanghan-iapcm (changes requested):

wanghan-iapcm (approved):

deepmd/pt/utils/utils.py
```diff
@@ -227,18 +228,22 @@ def to_numpy_array(xx: None) -> None: ...
 
 
 def to_numpy_array(
-    xx: torch.Tensor | None,
+    xx: torch.Tensor | np.ndarray | float | None,
```
njzjz, line 239: Need to update the overload method in line 223.

deepmd/tf/utils/learning_rate.py
```diff
@@ -1,102 +1,136 @@
...
 
-        return tf.train.exponential_decay(
-            self.start_lr_,
-            global_step,
-            self.decay_steps_,
-            self.decay_rate_,
-            staircase=True,
+        lr = tf.numpy_function(
+            _lr_value, [global_step], Tout=GLOBAL_TF_FLOAT_PRECISION, name="lr_schedule"
         )
```
njzjz, line 100: I am wondering will it be more expensive. Previously the learning rate calculation is on the GPU. This PR puts it on the CPU and requires an extra host-to-device comunication every step.

examples/hessian/multi_task/input.json
njzjz, line 1: The formatting should be reverted, otherwise we are hard to review changes.

doc/train/training-advanced.md
njzjz, line 1: May we move the learning rate into a new section. It's much longer than before.

deepmd/tf/utils/learning_rate.py
njzjz: This may be minimal, but a benchmark result should be given.

source/tests/pt/test_lr.py
```diff
@@ -10,37 +10,46 @@
...
 
 class TestLearningRate(unittest.TestCase):
     def setUp(self) -> None:
         self.start_lr = 0.001
         self.stop_lr = 3.51e-8
-        self.decay_steps = np.arange(400, 601, 100)
-        self.stop_steps = np.arange(500, 1600, 500)
+        # decay_steps will be auto-adjusted if >= num_steps
+        self.decay_steps = np.arange(400, 501, 100)
+        self.num_steps = np.arange(500, 1600, 500)
```
njzjz: What is the reason to change the variable name.. I think don't change it if it is not broken.

doc/train/training-advanced.md
```diff
@@ -6,70 +6,284 @@ In this section, we will take `$deepmd_source_dir/examples/water/se_e2_a/input.j
...
-	},
-
-	"numb_steps":	1000000,
-	"seed":		1,
-	"disp_file":	"lcurve.out",
-	"disp_freq":	100,
-	"save_freq":	1000
+        "training_data": {
+            "systems":    ["../data_water/data_0/", "../data_water/data_1/", "../data_water/data_2/"],
+            "batch_size": "auto"
```
njzjz, line 17: I think the changes like this should be conducted in a seperate PR, using the formatting tools and applying to all files.

deepmd/utils/compat.py
```diff
@@ -370,6 +370,69 @@ def deprecate_numb_test(
...
+            continue
+        lr[key] = value
+        moved_keys.append(key)
+
+    if warning:
+        if skipped_keys:
+            warnings.warn(
+                "Legacy warmup settings under training were ignored because "
+                f"learning_rate already defines them: {', '.join(skipped_keys)}."
+            )
```
njzjz: I think an error should be raised if there are two conflict definations.

deepmd/utils/argcheck.py
```diff
@@ -2480,55 +2480,239 @@ def linear_ener_model_args() -> Argument:
...
             float,
             optional=True,
             default=None,
-            doc=doc_only_pt_supported + doc_decay_rate,
+            doc=doc_decay_rate,
         ),
+        Argument("smooth", bool, optional=True, default=False, doc=doc_smooth),
     ]
-    return args
+    return _learning_rate_common_args(doc_stop_lr, extra_args=extra_args)
```
njzjz: `_learning_rate_common_args` should be put outside the `Variant`. `dargs` gives the freedom to do it, so these arguments will not belong to any `Variant`.

deepmd/tf/utils/learning_rate.py
```diff
@@ -1,102 +1,136 @@
...
+
+        Raises
+        ------
+        RuntimeError
+            If the schedule has not been built.
+        """
+        if self._base_lr is None:
+            raise RuntimeError("Learning rate schedule is not built yet.")
+        return float(np.asarray(self._base_lr.value(step)))
+
```
njzjz: It seems `float(np.asarray(` is not needed?

deepmd/tf/train/trainer.py
```diff
@@ -797,7 +809,7 @@ def _get_place_holders(self, data_dict) -> None:
                 prec = GLOBAL_ENER_FLOAT_PRECISION
             self.place_holders[kk] = tf.placeholder(prec, [None], name="t_" + kk)
             self.place_holders["find_" + kk] = tf.placeholder(
-                tf.float32, name="t_find_" + kk
+                GLOBAL_TF_FLOAT_PRECISION, name="t_find_" + kk
```
njzjz: What is the reason to conduct this change in this PR?

deepmd/tf/train/trainer.py
```diff
@@ -426,14 +431,21 @@ def train(self, train_data=None, valid_data=None) -> None:
...
+                lr0,
+                lr0,
+            )
+        else:
+            log.info(
+                "start training at lr %.2e (== %.2e), final lr will be %.2e",
+                run_sess(self.sess, self.learning_rate),
+                self.lr.value(cur_batch),
+                self.lr.value(stop_batch),
+            )
```
njzjz: Why does it need a `if...else` block?

deepmd/tf/train/trainer.py
```diff
@@ -242,8 +242,13 @@ def build(self, data=None, stop_batch=0, origin_type_map=None, suffix="") -> Non
...
         self.global_step = tf.train.get_or_create_global_step()
-        self.learning_rate = self.lr.build(self.global_step, self.stop_batch)
-        log.info("built lr")
+        if self.stop_batch == 0:
+            # Use constant start_lr when stop_batch is zero (no training)
+            self.learning_rate = tf.cast(self.lr.start_lr(), GLOBAL_TF_FLOAT_PRECISION)
+            log.info("built lr (constant start_lr for stop_batch=0)")
+        else:
+            self.learning_rate = self.lr.build(self.global_step, self.stop_batch)
+            log.info("built lr")
```
njzjz: Why does it need a if...else block?

deepmd/dpmodel/utils/learning_rate.py
```diff
@@ -29,135 +29,468 @@ def __new__(cls: type, *args: Any, **kwargs: Any) -> Any:
         return super().__new__(cls)
 
     def __init__(
-        self, start_lr: float, stop_lr: float, stop_steps: int, **kwargs: Any
+        self,
+        start_lr: float,
+        stop_lr: float | None = None,
+        stop_lr_ratio: float | None = None,
+        num_steps: int = 100000,
```
njzjz: Why do you give it a defualt value?

njzjz (changes requested):

deepmd/tf/utils/learning_rate.py
OutisLi: the lr calculation is pretty simple, do not need to be on GPU, just plain simple cos/exp

njzjz (approved):

iProzd (approved):

PR #5162, opened 2026-01-19
feat(plugin): enhance register method to support aliases for plugins

OutisLi:
## Release Notes

* **New Features**
  * Plugin registration now supports aliases, allowing plugins to be referenced by multiple names for improved flexibility and convenience.

### DIFF
--- a/deepmd/utils/plugin.py
+++ b/deepmd/utils/plugin.py
@@ -38,13 +38,17 @@ class Plugin:
         self.plugins.update(other.plugins)
         return self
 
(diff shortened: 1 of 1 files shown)


njzjz (commented):
Currently, objects can call `register` multiple times to have aliases.

https://github.com/deepmodeling/deepmd-kit/blob/501adb86c6189a072a44bccd8204d89556f570ab/deepmd/jax/descriptor/se_e2_a.py#L31-L32

Could you tell me what is necessary for adding a new argument?

PR #5168, opened 2026-01-20
fix(pt): Huber energy loss uses raw model energy

wanghan-iapcm:
Huber energy loss uses raw model energy instead of `energy_pred` that includes `atom_ener_coeff` weighting, so coefficients are silently dropped when use_huber=True

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed energy Huber loss to use the same precomputed energy values as other loss paths for consistent energy loss computation.
  * Improved energy aggregation handling to be more robust and correctly shaped during computation.

* **Tests**
  * Expanded loss tests to cover configurations with and without atom energy coefficients and updated test inputs to validate the revised behavior.

### DIFF
--- a/deepmd/dpmodel/loss/ener.py
+++ b/deepmd/dpmodel/loss/ener.py
@@ -131,8 +131,8 @@ class EnergyLoss(Loss):
             # E = - E(A) - E(B) + E(C) + E(D)
             # A, B, C, D could be put far away from each other
             atom_ener_coeff = label_dict["atom_ener_coeff"]
-            atom_ener_coeff = xp.reshape(atom_ener_coeff, xp.shape(atom_ener))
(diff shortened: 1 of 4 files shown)


iProzd (commented):
Huber energy in pd loss also needs to be fixed.

iProzd (changes requested):
`xp.shape(atom_ener)` should be `atom_ener.shape` in https://github.com/deepmodeling/deepmd-kit/blob/master/deepmd/dpmodel/loss/ener.py#L134 .

It seems that ut never reached here. @njzjz

njzjz:
> xp.shape(atom_ener) should be atom_ener.shape in

You are correct.

iProzd (approved):

PR #5181, opened 2026-01-29
fix(pt/pd): fix incompatibility between AutoBatchSize and eval hooks

### LINKED ISSUE #5180: [BUG] `dp eval-desc` give mismatch descriptors on certain System
### Bug summary

On certain system, `dp eval-desc` will generate mismatch number of descriptor frames against System (number of frames is larger, where descriptors have 241 frames and the dpdata.System only have 175 frames)

### DeePMD-kit Version

3.1.2 (modified as #5177)

### Backend and its version

Pytorch

Built from source

[descritor_mismatch_example.tar.gz](https://github.com/user-attachments/files/24929037/descritor_mismatch_example.tar.gz)

- ./data contain target dpdata.System
- ./desc contain mismatch descriptor

### Steps to Reproduce

`dp --pt eval-desc -s [data] -m DPA-3.1-3M.pt --head Hybrid_Perovskite`

_No response_

njzjz:
Fix #5180.

## Summary by CodeRabbit

* **New Features**
  * Optional automatic OOM-retry during evaluation that retries with adjusted batch size and a retry loop.
  * New public control to enable or disable OOM-retry mode for evaluation flows.
* **Bug Fixes**
  * Evaluation retry now guarantees cleanup: evaluation hooks and OOM-retry mode are reliably cleared after success or failure.
* **Tests**
  * Added unit tests verifying OOM-retry behavior, retry cleanup, and state/hook reset across attempts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/deepmodeling/deepmd-kit/pull/5181?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


wanghan-iapcm (commented):
The fix introduces three new code paths with zero test coverage. Please add unit tests for each — they exist solely because of this PR and will silently regress otherwise.                  
   
  ## 1. `RetrySignal` raise path in `AutoBatchSize.execute()`                                                                                                                                  
                                                                            
  **Target**: `deepmd/utils/batch_size.py:130-131`                                                                                                                                             
  
  Subclass `AutoBatchSize` with a fake executor that raises an OOM-sentinel on first call. Then:                                                                                               
  - With `set_oom_retry_mode(True)`: assert `RetrySignal` is raised, its `__cause__` is the original OOM, and `current_batch_size` was halved before the raise.
  - With `set_oom_retry_mode(False)`: assert it returns `(0, None)` instead — pins the flag actually gates behavior.                                                                           
                                                                                                                                                                                               
  ## 2. Recursive retry in `eval_descriptor` / `eval_fitting_last_layer` clears the hook between attempts                                                                                      
                                                                                                                                                                                               
  **Target**: `deepmd/{pt,pd}/infer/deep_eval.py`                                                                                                                                              
                                                                            
  This is the actual fix for #5180 (descriptor frame count doubling). Per backend:                                                                                                             
  - Inject an `AutoBatchSize` that raises `RetrySignal` on the first `execute`, succeeds on the second.
  - Call `eval_descriptor` with N frames.                                                                                                                                                      
  - Assert the returned array has exactly N frames (not 2N — the #5180 regression check).
  - Spy on `model.set_eval_descriptor_hook`; assert the call sequence is `[True, False, True, False]`.                                                                                         
  - Assert `auto_batch_size.oom_retry_mode == False` after return.          
                                                                                                                                                                                               
  Repeat for `eval_fitting_last_layer` with `set_eval_fitting_last_layer_hook`.                                                                                                                
                                      
  ## 3. `finally` clears state on non-`RetrySignal` exceptions                                                                                                                                 
                                                                            
  **Target**: same files as Test 2.                                                                                                                                                            
                                                                            
  The PR moves hook cleanup into `finally` — a real improvement, since arbitrary errors no longer leave the hook stuck `True`. Per backend, per method:                                        
  - Monkey-patch `self.eval` to raise a generic `RuntimeError`.
  - Call inside `pytest.raises(RuntimeError)`.                                                                                                                                                 
  - Assert `set_eval_*_hook` was called with `False` (sequence `[True, False]`).
  - Assert `auto_batch_size.oom_retry_mode == False`.                                                                                                                                          
                                                                                                                                                                                               
  ## Out of scope (do NOT add)        
                                                                                                                                                                                               
  - `oom_retry_mode == True` + no OOM — exercises no new code.                                                                                                                                 
  - `oom_retry_mode == False` + OOM returning `(0, None)` — pre-existing behavior.
                                                                                                                                                                                               
  ## Acceptance                                                             
                                                                                                                                                                                               
  - All new tests pass.                                                     
  - Mutation check: removing `raise RetrySignal` fails Test 1; removing the `finally` fails Tests 2 and 3.

wanghan-iapcm (commented):
The new tests address the three asks from my prior review, with one structural concern worth raising before merge.                                                                                                                                           
                                                                                                                                                                                                                                                               
  `DummyDeepEval.eval_descriptor` / `eval_fitting_last_layer` (lines 68-104 of `source/tests/common/test_oom_retry.py`) are hand-written copies of the production `try`/`except RetrySignal`/`finally` orchestration in `deepmd/pt/infer/deep_eval.py` and     
  `deepmd/pd/infer/deep_eval.py`. The tests call the dummy's methods, not production's. This is testing the test code, not the code that ships.                                                                                                                
                                                                                                                                                                                                                                                               
  Concrete evidence the gap matters: commit `b5f789ae` (the day after the tests were added) refactored the production code from recursion to iteration (`while True` + `retry` flag). The dummy still uses recursion. The tests passed throughout — they never 
  noticed the production-side change because they don't exercise it. If a future refactor drops the `finally` block from production, the same tests will keep passing.
                                                                                                                                                                                                                                                               
  The mutation-check guarantee from the prior review ("removing the `finally` fails Tests 2 and 3") does not currently hold for Tests 2 and 3. It holds for Test 1, which calls the real `AutoBatchSize.execute`.                                              
                                      
  Mocking is right for a UT — but mock the *dependencies* of the production method, not the production method itself. Roughly:                                                                                                                                 
                                                                           
  ```python                           
  class TestEvalDescriptorRetry(unittest.TestCase):
      def setUp(self):
          # Construct DeepEval without loading a real model.
          self.dp_eval = DeepPotPT.__new__(DeepPotPT)
          self.dp_eval.dp = MagicMock()
          self.dp_eval.dp.model = {"Default": MagicMock()}
          self.dp_eval.auto_batch_size = MagicMock()                                                                                                                                                                                                           
   
      def test_retry_clears_hook_between_attempts(self):                                                                                                                                                                                                       
          model = self.dp_eval.dp.model["Default"]                                                                                                                                                                                                             
          model.eval_descriptor.return_value = np.array([1, 2, 3])
          with patch.object(self.dp_eval, "eval",                                                                                                                                                                                                              
                            side_effect=[RetrySignal, None]):                                                                                                                                                                                                  
              result = self.dp_eval.eval_descriptor(
                  coords=..., cells=..., atom_types=...,                                                                                                                                                                                                       
              )                                                                                                                                                                                                                                                
          self.assertEqual(           
              model.set_eval_descriptor_hook.call_args_list,                                                                                                                                                                                                   
              [call(True), call(False), call(True), call(False)],          
          )                           
          np.testing.assert_array_equal(result, [1, 2, 3])

      def test_finally_clears_hook_on_runtime_error(self):
          with patch.object(self.dp_eval, "eval",
                            side_effect=RuntimeError("non-retry failure")):                                                                                                                                                                                    
              with self.assertRaisesRegex(RuntimeError, "non-retry failure"):
                  self.dp_eval.eval_descriptor(                                                                                                                                                                                                                
                      coords=..., cells=..., atom_types=...,               
                  )                   
          model = self.dp_eval.dp.model["Default"]
          self.assertEqual(                                                                                                                                                                                                                                    
              model.set_eval_descriptor_hook.call_args_list,
              [call(True), call(False)],                                                                                                                                                                                                                       
          )                                                                
  ```                                 

  The reusable `DummyModel` / `DummyAutoBatchSize` stubs (lines 24-49) are fine as collaborator fakes — keep them. Only `DummyDeepEval` should go; replace its usages with the real `DeepPotPT` / `DeepPotPD` constructed via `__new__` (or a tiny init helper)
   with the dependencies above patched in. Same shape, similar line count, but the assertions then pin the production code paths.
                                                                                                                                                                                                                                                               
  Test 1 (the `AutoBatchSize.execute` / `RetrySignal` raise path) is already correct — no changes needed there.

wanghan-iapcm (approved):

iProzd (approved):

PR #5194, opened 2026-02-06
feat: new backend pytorch exportable.

wanghan-iapcm:
- add new backend pytorch-exportable (pt-expt)
- implement the se_e2_a descriptor
- added UTs. 

## Release Notes

* **New Features**
  * Added PyTorch exportable backend enabling model export and deployment workflows.
  * Introduced PyTorch-compatible descriptor implementations with device-aware tensor operations.

* **Bug Fixes**
  * Improved error handling in array conversion operations with better device placement.
  * Added safeguards for PyTorch thread configuration to prevent runtime failures.

* **Refactoring**
  * Enhanced descriptor statistics computation with explicit device management for improved performance.

PR #5204, opened 2026-02-08
refact (pt_expt): provide infrastructure for converting dpmodel classes to PyTorch modules. 

wanghan-iapcm:
consider after the merge of #5194

automatically wrapping dpmodel classes (array_api_compat-based) as PyTorch modules. The key insight is to detect attributes by their **value type** rather than by hard-coded names. 

## Summary by CodeRabbit

* **New Features**
  * Registry-driven conversion for DP objects to PyTorch modules enabling automatic wrapper creation.
  * New PyTorch-friendly descriptor variants with stable forward outputs for se_e2_a and se_r.
  * PyTorch-wrapped exclude-mask utilities and a NetworkCollection of wrapped network types for proper module/state handling.
  * Device-aware tensor conversion and robust handling of numpy buffers and None-valued buffers for reliable serialization/movement.


deepmd/pt_expt/common.py
```diff
@@ -0,0 +1,278 @@
...
+this registry to find the appropriate converter.
+
+Examples of registered mappings:
+- AtomExcludeMaskDP → lambda v: AtomExcludeMask(v.ntypes, exclude_types=...)
+- NetworkCollectionDP → lambda v: NetworkCollection.deserialize(v.serialize())
+"""
+
+
+def register_dpmodel_mapping(
+    dpmodel_cls: type, converter: Callable[[Any], torch.nn.Module]
```
njzjz: Considering that any dpmodel class is a subclass of `NativeOP`, we could replace `Any` with `NativeOP` in this file.
```suggestion
    dpmodel_cls: type[NativeOP], converter: Callable[[NativeOP], torch.nn.Module]
```

deepmd/pt_expt/common.py
```diff
@@ -0,0 +1,278 @@
...
+    # clear an existing buffer to None
+    if value is None and "_buffers" in obj.__dict__ and name in obj._buffers:
+        obj._buffers[name] = None
+        return True, None
+
+    # dpmodel object → pt_expt module
+    if "_modules" in obj.__dict__:
+        converted = try_convert_module(value)
+        if converted is not None:
+            return False, converted
```
njzjz: Do we need to have a check here? For example, if an object is a `NativeOP` but not a `torch.nn.Module`, we think it can be converted. But if it is not in the map, we consider that something went wrong.

njzjz (approved):

PR #5221, opened 2026-02-13
fix(pt): remove compile in HybridMuon for FSDP2 compatibility use triton instead

OutisLi:
## Summary by CodeRabbit

* **Refactor**
  * Simplified orthogonalization routing to process matrix parameters per-entry and added a faster GPU path when available.

* **New Features**
  * Added optimizer option "flash_muon" (default: true) to enable an accelerated orthogonalization path with automatic fallback.

* **Tests**
  * Expanded tests to cover both accelerated and fallback paths and adjusted shape/validation cases.

* **Documentation**
  * Updated optimizer argument docs to describe flash_muon and clarified 1D/>=2D parameter handling.

### DIFF
--- a/deepmd/pt/optimizer/hybrid_muon.py
+++ b/deepmd/pt/optimizer/hybrid_muon.py
@@ -48,6 +48,8 @@ References
        https://arxiv.org/abs/2502.16982
 .. [3] Moonlight GitHub Repository.
        https://github.com/MoonshotAI/Moonlight
+.. [4] Flash-Muon: Triton-accelerated symmetric matmul for Newton-Schulz.
+       https://github.com/lintianyang/flash-muon (MIT License, Tianyang Lin)
 """
 
 from __future__ import (
@@ -71,6 +73,18 @@ if TYPE_CHECKING:
     )
 
 # ============================================================================
+# Triton availability detection
+# ============================================================================
+
+try:
+    import triton
+    import triton.language as tl
+
+    TRITON_AVAILABLE = True
+except ImportError:
+    TRITON_AVAILABLE = False
+
+# ============================================================================
 # Constants
 # ============================================================================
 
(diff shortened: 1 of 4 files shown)


deepmd/pt/optimizer/hybrid_muon.py
```diff
@@ -689,37 +611,16 @@ def step(
...
+                for entry, update_tensor in bucket_entries:
                     update_matrix = update_tensor.reshape(rows, cols)
                     if not update_matrix.is_contiguous():
                         update_matrix = update_matrix.contiguous()
 
-                    orth = _zeropower_via_newtonschulz5_2d(update_matrix)
+                    orth = _newton_schulz_orth(update_matrix)
                     orth.mul_(scale)
                     delta = orth.reshape(entry["param"].shape)
                     entry["param"].add_(delta, alpha=-lr)
```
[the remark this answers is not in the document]
OutisLi: After test, it is faster instead

wanghan-iapcm (approved):

deepmd/pt/optimizer/hybrid_muon.py
```diff
@@ -82,33 +106,138 @@
...
+                num_warps=n_warps,
+            )
+            for blk_m in [32, 64, 128]
+            for blk_k in [32, 64]
+            for n_stages in [3, 4, 5]
+            for n_warps in [4, 8]
+        ]
+
+    @triton.autotune(configs=_get_autotune_config(), key=["M", "K"])
+    @triton.jit
```
njzjz, line 139: Will it be expensive at the import time?

deepmd/pt/optimizer/hybrid_muon.py
```diff
@@ -82,33 +106,138 @@
...
+            tl.store(ct_ptrs, tl.permute(c, (1, 0)), mask=ct_mask)
+
+    def _matmul_transpose_assign(d_in: torch.Tensor, d_out: torch.Tensor) -> None:
+        """Compute d_out = d_in @ d_in.T using triton symmetric matmul kernel."""
+        d_in = d_in.contiguous()
+        M, K = d_in.shape
+        grid = lambda META: (  # noqa: E731
+            triton.cdiv(M, META["BLOCK_SIZE_M"]) * triton.cdiv(M, META["BLOCK_SIZE_M"]),
+        )
+        with torch.cuda.device(d_in.device.index):
```
njzjz, line 201: What is the situation for the CPU?

source/tests/pt/test_hybrid_muon.py
njzjz, line 1: It seems that triton is not installed in the CI.

deepmd/pt/optimizer/hybrid_muon.py
OutisLi: after test，just less than 5s

deepmd/pt/optimizer/hybrid_muon.py
OutisLi: cpu will not use triton, just raw realization. The same as the original flash-muon repo

njzjz (approved):

iProzd (approved):

PR #5266, opened 2026-02-25
chore(pt): mv the input stat update to model_change_out_bias

wanghan-iapcm:
cleanup the logic of change_out_bias: it only change the output bias, does not update the fitting input stat. 

## Summary by CodeRabbit

* **Refactor**
  * Adjusted how bias-adjustment triggers recomputation of fitting/input statistics during model training, streamlining the behavior and adding a conditional recompute path for certain model types.

* **Tests**
  * Added cross-checking tests to ensure the new and previous bias/statistic update paths produce consistent fitting statistics.

### DIFF
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -1755,6 +1755,13 @@ def model_change_out_bias(
     )
     new_bias = deepcopy(_model.get_out_bias())
 
+    from deepmd.pt.model.model.dp_model import (
+        DPModelCommon,
+    )
+
+    if isinstance(_model, DPModelCommon) and _bias_adjust_mode == "set-by-statistic":
+        _model.get_fitting_net().compute_input_stats(_sample_func)
+
     model_type_map = _model.get_type_map()
     log.info(
         f"Change output bias of {model_type_map!s} from {to_numpy_array(old_bias).reshape(-1)!s} to {to_numpy_array(new_bias).reshape(-1)!s}."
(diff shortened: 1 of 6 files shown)


deepmd/pt/train/training.py
```diff
@@ -1825,6 +1755,13 @@ def model_change_out_bias(
     )
     new_bias = deepcopy(_model.get_out_bias())
 
+    from deepmd.pt.model.model.dp_model import (
+        DPModelCommon,
+    )
+
+    if isinstance(_model, DPModelCommon) and _bias_adjust_mode == "set-by-statistic":
+        _model.get_fitting_net().compute_input_stats(_sample_func)
```
Chengqian-Zhang, line 1763: Maybe `_model.atomic_model.compute_fitting_input_stat(_sample_func)` is better? Because if we directly call `get_fitting_net().compute_input_stats(_sample_func)`, it will ignore `protection` parameter in function `compute_input_stats`. And I have specify the value of this parameter in function `atomic_model.compute_fitting_input_stat` (the value equals to keyword `data_stat_protect` specified by user).

Chengqian-Zhang (changes requested):
See below

deepmd/pt/train/training.py
wanghan-iapcm: 1. the protection in the dp atomic model is never serialized, thus it will be deserialized with a default value. This value is actually not not used anywhere else. 
2. it is not a good idea to direactly access the data member `atomic_model`.

Chengqian-Zhang (approved):

iProzd (approved):

PR #5269, opened 2026-02-26
feat(pt, dpmodel): use data stat for observed type

iProzd:
## Summary by CodeRabbit

* **New Features**
  * Models now record, persist, and expose observed element-type lists in model metadata; evaluation/display prefer that metadata when present.
  * Training and model APIs accept an optional preset observed-type override to control observed-type selection.
  * Utilities added to collect, save, and restore observed-type lists for caching and reuse.

* **Tests**
  * Added unit and integration tests covering collection, file I/O, training integration, metadata propagation, fallback behavior, and user presets.


njzjz (approved):

deepmd/dpmodel/model/base_model.py
```diff
@@ -142,6 +142,18 @@ def get_model_def_script(self) -> str:
...
+
+            Returns empty list if not available.
+            """
+            if self.model_def_script:
+                params = json.loads(self.model_def_script)
+                observed = params.get("info", {}).get("observed_type")
+                if observed is not None:
+                    return observed
+            return []
+
```
wanghan-iapcm: conflicting with #5260 , see https://github.com/wanghan-iapcm/deepmd-kit/blob/162c5a1bd9a1b9ad8e68af151383d1e00964262c/deepmd/dpmodel/model/make_model.py#L381-L398
plz update this pr after the merge of 5260. thanks!

deepmd/pt/model/atomic_model/dp_atomic_model.py
```diff
@@ -358,6 +364,17 @@ def wrapped_sampler() -> list[dict]:
...
+        if preset_observed_type is not None:
+            self._observed_type = preset_observed_type
+        else:
+            observed = _restore_observed_type_from_file(stat_file_path)
+            if observed is None:
+                sampled = wrapped_sampler()
+                observed = collect_observed_types(sampled, self.type_map)
+                _save_observed_type_to_file(stat_file_path, observed)
+            self._observed_type = observed
+
```
wanghan-iapcm: duplicated with linear atomic model

deepmd/pt/model/atomic_model/pairtab_atomic_model.py
```diff
@@ -231,6 +231,7 @@ def compute_or_load_stat(
         sampled_func: Callable[[], list[dict]] | list[dict],
         stat_file_path: DPPath | None = None,
         compute_or_load_out_stat: bool = True,
+        preset_observed_type: list[str] | None = None,
```
wanghan-iapcm, line 231: what should pairtab atomic model do the same thing as dp and linear atomic model?

deepmd/pt/train/training.py
```diff
@@ -394,7 +397,14 @@ def get_lr(lr_params: dict[str, Any]) -> BaseLR:
                 finetune_has_new_type=self.finetune_links["Default"].get_has_new_type()
                 if self.finetune_links is not None
                 else False,
+                preset_observed_type=model_params.get("info", {}).get("observed_type"),
             )
+            # Persist observed_type from stat into model_params and model_def_script
+            if not resuming and self.rank == 0:
+                observed = getattr(self.model.atomic_model, "_observed_type", None)
```
wanghan-iapcm: 1. dangerous to directly access data member
2. the data member is assumed to be private (starting with _ prefix)

source/tests/consistent/test_observed_type.py
```diff
@@ -0,0 +1,57 @@
...
+        fake = FakeModel()
+        fake.get_observed_type_list = BaseModel.get_observed_type_list.__get__(
+            fake, FakeModel
+        )
+        return fake
+
+    def test_with_observed_type_in_info(self) -> None:
+        script = json.dumps(
+            {
+                "info": {"observed_type": ["H", "O"]},
```
wanghan-iapcm: is it a good design to add "observed_type" in the "info" block?  the "info" is "Used only in multitask models" by the doc

wanghan-iapcm (changes requested):

source/tests/consistent/test_observed_type.py
iProzd: By design, "info" should be a metadata dict suitable for both single and multi-task model. Doc updated.

deepmd/pt/model/atomic_model/dp_atomic_model.py
iProzd: fixed

deepmd/pt/model/atomic_model/pairtab_atomic_model.py
iProzd: Improved. 

For stand alone pairtab atomic, it should do the same observed type calculation as other atomic models.

For linear atomic model, it computes observed type once, then propagates it to sub-models via preset_observed_type to avoid redundant computation.

source/tests/pt/test_observed_type.py
wanghan-iapcm, line 1: A similar UT should be provided to dpmodel backend.

wanghan-iapcm (changes requested):

wanghan-iapcm (approved):

PR #5287, opened 2026-03-05
feat(pt_expt): implementing energy hessian model

wanghan-iapcm:
### Summary

Add hessian (second derivative of energy w.r.t. coordinates) support to the pt_expt backend, mirroring the JAX backend's approach.

  - Compute hessian on extended coordinates in forward_common_atomic (via torch.autograd.functional.hessian), then let dpmodel's communicate_extended_output naturally map from nall×nall to nloc×nloc
  - make_hessian_model only overrides atomic_output_def() to set r_hessian=True — no forward_common override or r_hessian toggle hack needed
  - Hessian is enabled at runtime via EnergyModel.enable_hessian() and returned through the user-facing forward() interface

### Changed files

  - deepmd/pt_expt/model/make_model.py — add _cal_hessian_ext and _WrapperForwardEnergy for hessian computation in forward_common_atomic
  - deepmd/pt_expt/model/make_hessian_model.py — minimal wrapper: __init__, requires_hessian, atomic_output_def
  - deepmd/pt_expt/model/ener_model.py — enable_hessian(), hessian output in forward() and translated_output_def()
  - deepmd/pt_expt/model/__init__.py — export make_hessian_model
  - source/tests/pt_expt/model/test_ener_hessian_model.py — unit test: autograd hessian vs finite-difference, parametrized over nv={1,2}
  - source/tests/consistent/model/test_ener_hessian.py — cross-backend consistency test covering PT, pt_expt, and JAX
  

## Summary by CodeRabbit

* **New Features**
  * Adds Hessian (second-order derivative) support for energy models.
  * Provides a mechanism to enable Hessian calculations per model instance.
  * When enabled, Hessian tensors are included in model prediction outputs.

* **Tests**
  * Added comprehensive multi-backend validation of Hessian functionality.
  * Added unit tests comparing analytical Hessians to finite-difference references.


deepmd/dpmodel/model/make_hessian_model.py
njzjz, line 1: It looks to me that this file can be moved into `dpmodel`? It doesn't contain any PyTorch-specific modules.

wanghan-iapcm: fixed by d4dc9456

njzjz (approved):

PR #5294, opened 2026-03-05
feat(dp, pt): add force l2 norm loss & mae loss

iProzd:
## Summary by CodeRabbit

* **New Features**
  * Introduced loss_func ("mse" or "mae") to select MSE vs MAE for energy/force/virial/atom losses.
  * Added f_use_norm to enable vector‑norm MAE behavior when allowed.

* **Validation**
  * Enforced that f_use_norm is only valid when use_huber is enabled or loss_func="mae"; invalid combos are rejected.

* **Tests**
  * Extended loss tests and skipping logic to cover loss_func and f_use_norm combinations.

* **Documentation**
  * Updated docs to describe loss_func and resulting metric names (rmse_* vs mae_*).

* **Chores**
  * New options are persisted in serialized configurations.

* **Notes**
  * Some backends currently only support "mse" (MAE not yet available everywhere).


deepmd/pt/loss/ener.py
```diff
@@ -140,9 +144,14 @@ def __init__(
             raise RuntimeError(
                 "When generalized force loss is used, the dimension of generalized coordinates should be larger than 0"
             )
-        self.use_l1_all = use_l1_all
+        self.use_mae_loss = use_mae_loss
         self.inference = inference
         self.use_huber = use_huber
+        self.f_use_norm = f_use_norm
+        if self.f_use_norm and not (self.use_huber or self.use_mae_loss):
+            raise RuntimeError(
```
OutisLi, line 161: I think this is not necessary. When use_huber and use_mae_loss are both false, the value of f_use_norm does not change anything. As a result, f_use_norm can be true in any cases.

deepmd/dpmodel/loss/ener.py
```diff
@@ -50,8 +50,75 @@ def __init__(
...
+            Loss function type for energy, force, and virial terms.
+            Options: 'mse' (Mean Squared Error, L2 loss, default) or 'mae' (Mean Absolute Error, L1 loss).
+            MAE loss is less sensitive to outliers compared to MSE loss.
+            Future extensions may support additional loss types.
+        f_use_norm : bool
+            If true, use L2 norm of force vectors for loss calculation when loss_func='mae' or use_huber is True.
+            Instead of computing loss on force components, computes loss on ||F_pred - F_label||_2.
+        **kwargs
+            Other keyword arguments.
+        """
```
njzjz: Please move the doc to the class docstring. We follow numpydoc, which suggests this rule.

njzjz (changes requested):

deepmd/pt/loss/ener.py
iProzd: When using `mse` loss, `f_use_norm` = `True` is equivalent to `False`. 
See : mean_atoms(||ΔF_i||²) = mean_atoms(Δfx_i² + Δfy_i² + Δfz_i²) = 3 * mean_components(Δf²).
This will make users confused.

OutisLi: ok

OutisLi (approved):

njzjz (approved):

wanghan-iapcm (approved):

PR #5295, opened 2026-03-05
feat(dp, pt): add charge spin embedding

iProzd:
## Summary by CodeRabbit

* **New Features**
  * Optional charge/spin electronic embedding for DPA3 (toggleable).
  * Descriptor and model APIs accept optional per-frame parameters (fparam); when enabled, default per-frame params are forwarded into descriptor embeddings.

* **Tests**
  * Expanded test coverage for charge/spin embedding, fparam construction/propagation, backend paths, and serialization.

* **Documentation**
  * Added help/config entry for the charge/spin embedding option.

### DIFF
--- a/deepmd/dpmodel/atomic_model/dp_atomic_model.py
+++ b/deepmd/dpmodel/atomic_model/dp_atomic_model.py
@@ -73,6 +73,9 @@ class DPAtomicModel(BaseAtomicModel):
         if hasattr(self.fitting_net, "reinit_exclude"):
             self.fitting_net.reinit_exclude(self.atom_exclude_types)
         self.type_map = type_map
+        self.add_chg_spin_ebd: bool = getattr(
+            self.descriptor, "add_chg_spin_ebd", False
+        )
         super().init_out_stat()
 
     def fitting_output_def(self) -> FittingOutputDef:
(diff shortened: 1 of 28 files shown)


deepmd/dpmodel/atomic_model/dp_atomic_model.py
```diff
@@ -180,11 +183,38 @@ def forward_atomic(
         """
         nframes, nloc, nnei = nlist.shape
         atype = xp_take_first_n(extended_atype, 1, nloc)
+
+        # Handle default fparam if fitting net supports it
+        if (
```
wanghan-iapcm, line 188: the handling of the default fparam is actually not covered by ut. 
I suggest in the model's consistency we test : 
1. no fparam (before this pr)
2. explicitly provided fparam
3. fparam == None but fitting net has default fparam. 

In all the three cases the dpmodel should be consistent with the pt model .

wanghan-iapcm (changes requested):

deepmd/dpmodel/atomic_model/dp_atomic_model.py
iProzd: added in https://github.com/deepmodeling/deepmd-kit/pull/5295/commits/8979f8138966108e6d2aba273bba3b1e28b14e56

wanghan-iapcm (approved):

njzjz (approved):

PR #5317, opened 2026-03-16
feat(pt_expt): add linear energy model

wanghan-iapcm:
This is a **breaking change**: bump the model data versio of `LinearEnergyAtomicModel` from 2 to 3 due to the bug fixing. 

## Summary
  - Add `LinearEnergyModel` to pt_expt backend, enabling linear combination of multiple sub-models
  - Add `get_linear_model` factory in pt_expt for constructing from config dicts
  - Fix bugs in dpmodel/pt shared code:
    - `get_linear_model` (pt) not propagating `type_map` to sub-models
    - `LinearEnergyAtomicModel` (dpmodel) missing `weights` parameter, causing deserialization failure
    - `_compute_weight` calling `array_api_compat.array_namespace()` with Python list and using numpy dtype with torch

  ## Test plan
  - [x] Cross-backend consistency test (`source/tests/consistent/model/test_linear_ener.py`) — pt vs pt_expt, with parameterized exclude types
  - [x] Existing dpmodel/pt linear model tests still pass

## Summary by CodeRabbit

* **New Features**
  * Added LinearEnergyModel and a "linear_ener" fitting path to combine multiple sub-models.
  * Configurable weighting when combining sub-model energies: "mean", "sum", or a custom vector; weights are validated and stored.

* **Behavior**
  * Sub-model type mappings are propagated when omitted.
  * Model serialization now persists weight settings and advances the model version for compatibility.

* **Tests**
  * Added cross-backend and unit tests validating weighting behaviors, outputs, and selector updates.


deepmd/pt_expt/model/dp_linear_model.py
```diff
@@ -0,0 +1,193 @@
...
+            model_predict["virial"] = model_ret["energy_derv_c_redu"].squeeze(-2)
+            if do_atomic_virial and model_ret.get("energy_derv_c") is not None:
+                model_predict["extended_virial"] = model_ret["energy_derv_c"].squeeze(
+                    -2
+                )
+        if "mask" in model_ret:
+            model_predict["mask"] = model_ret["mask"]
+        return model_predict
+
+    def translated_output_def(self) -> dict[str, Any]:
```
njzjz, line 105: It seems that this method is not used anywhere.

wanghan-iapcm: yes but to be consistent with pt backend, we provide the api

njzjz (approved):

PR #5318, opened 2026-03-16
feat(pt_expt): add frozen model

wanghan-iapcm:
## Summary
  - Add `FrozenModel` to pt_expt backend for loading pre-frozen model files (`.pte`, `.pth`, `.dp`)
  - Create dpmodel-level `FrozenModel` (`NativeOP` + `BaseModel`) with all delegation methods, so pt_expt wraps it via `@torch_module` instead of duplicating code
  - pt_expt `FrozenModel` handles `.pte` natively via `serialize_from_file`, falls back to generic backend detection for other formats
  - Add pt_expt support to frozen model consistency test

  ## Test plan
  - [x] Cross-backend consistency test (`source/tests/consistent/model/test_frozen.py`) — pt_expt consistent_with_ref and self_consistent pass
  - [x] Existing pt/tf frozen model tests unaffected

## Summary by CodeRabbit

* **New Features**
  * Added support for loading and using frozen model files across workflows and exposed a FrozenModel in the Python API.
  * Broadened backend compatibility to include an additional experiment backend for frozen models.

* **Tests**
  * Added/updated tests to validate frozen-model loading and evaluation across supported backends, including the new experiment backend.

### DIFF
--- a/deepmd/pt_expt/model/frozen.py
+++ b/deepmd/pt_expt/model/frozen.py
@@ -1,16 +1,11 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 from typing import (
     Any,
-    NoReturn,
 )
 
-import torch
-
-from deepmd.dpmodel.output_def import (
-    FittingOutputDef,
-)
-from deepmd.utils.data_system import (
-    DeepmdDataSystem,
+from deepmd.dpmodel.model.frozen import FrozenModel as FrozenModelDP
+from deepmd.pt_expt.common import (
+    torch_module,
 )
 
 from .model import (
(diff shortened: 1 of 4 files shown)


deepmd/pt_expt/model/frozen.py
```diff
@@ -0,0 +1,46 @@
...
+        self.model.eval()
+
+    def serialize(self) -> dict:
+        """Serialize the model.
+
+        Unlike the pt backend (which must reconstruct from model_def_script
+        because its inner model is an opaque ScriptModule), pt_expt's inner
+        model is a real pt_expt model that can serialize directly.
+        """
+        return self.model.serialize()
```
njzjz: This method seems to be the same as that in the parent class, so it's not needed to write.

wanghan-iapcm: fixed by 4815fbd2

wanghan-iapcm: fixed by 4815fbd2

njzjz (approved):

PR #5322, opened 2026-03-17
feat(property): Support element type-wise bias in property fitting

Chengqian-Zhang:
### Summary
Introduced the `distinguish_types` option for property fitting to support more flexible atom contribution calculations.
- `distinguish_types=True`(Default) has consistent behavior with energy fitting(if the property is extensive), calculating individual atom contributions based on each specific atom type.
- `distinguish_types=False` treats atom contributions as element-agnostic. This is particularly useful for element extrapolation scenarios where the model needs to generalize across different chemical species.

### Key changes
- `deepmd/pt/model/task/property.py`: 
   - Add `distinguish_types` parameter.
   - Change the serialization version from 5 to 6.
- `deepmd/pt/utils/stat.py`:
   - Refactored `output_std`: Replaced from the constant "ones" padding with actual standard deviation values(required by function `apply_out_stat` in `deepmd/pt/model/atomic_model/property_atomic_model.py`)
   - Optimized the RMSE logging/printing logic for better clarity.
- `deepmd/utils/out_stat.py`:
   - Updated `compute_stats_from_redu` to include the `intensive` parameter.

## Summary by CodeRabbit

* **New Features**
  * Added a distinguish_types option for property fitting and an intensive mode for statistics computation.

* **Bug Fixes**
  * Output statistics can be applied per atom type when enabled; global standard deviations are now correctly broadcast to per-type shapes.
  * Normalization and RMSE reporting adjusted to handle intensive vs per-atom modes.

* **Tests**
  * Added/updated tests covering intensive statistics and updated expected statistic values.


source/tests/common/test_out_stat.py
```diff
@@ -142,6 +142,34 @@ def test_compute_stats_do_not_distinguish_types_extensive(self) -> None:
...
+                0.00001954897296228177,
+                0.000020281857747683162,
+                0.000010741237959989648,
+                0.000020258211828681347,
+            ]
+        )
+        np.testing.assert_allclose(
+            std,
+            reference_std,
+            rtol=1e-6,
```
wanghan-iapcm, line 170: the value is almost 1e-5, the rtol is too large for checking consistency.

wanghan-iapcm (changes requested):

source/tests/common/test_out_stat.py
wanghan-iapcm: I see it is rtol. this concern is resolved.

deepmd/dpmodel/fitting/property_fitting.py
```diff
@@ -131,7 +135,7 @@ def output_def(self) -> FittingOutputDef:
     @classmethod
     def deserialize(cls, data: dict) -> "PropertyFittingNet":
         data = data.copy()
-        check_version_compatibility(data.pop("@version"), 5, 1)
+        check_version_compatibility(data.pop("@version"), 6, 1)
```
wanghan-iapcm, line 138: Missing backward compatibility for version-5 checkpoints that lack `distinguish_types`. Old models always used `get_compute_stats_distinguish_types() -> False`, so the default should be `False`.

```suggestion
        check_version_compatibility(data.pop("@version"), 6, 1)
        data.setdefault("distinguish_types", False)
```

deepmd/pt/model/task/property.py
```diff
@@ -133,10 +137,14 @@ def get_intensive(self) -> bool:
...
 
+    def get_distinguish_types(self) -> bool:
+        """Get whether to distinguish atom types when computing output statistics."""
+        return self.distinguish_types
+
     @classmethod
     def deserialize(cls, data: dict) -> "PropertyFittingNet":
         data = data.copy()
-        check_version_compatibility(data.pop("@version", 1), 5, 1)
+        check_version_compatibility(data.pop("@version", 1), 6, 1)
```
wanghan-iapcm, line 147: Same issue here.

```suggestion
        check_version_compatibility(data.pop("@version", 1), 6, 1)
        data.setdefault("distinguish_types", False)
```

wanghan-iapcm (approved):

njzjz (approved):

iProzd (approved):

PR #5336, opened 2026-03-24
feat(pt): add full validation

OutisLi:
## Summary by CodeRabbit

* **New Features**
  * Full-training validation: configurable start/frequency/metric, per-system energy/force/virial MAE & RMSE, logging to val.log, and optional top‑K best‑checkpoint saving/rotation.
  * New evaluation utilities producing standardized metric outputs for energy/force/virial/spin.
  * Auto-batch-size gain a `silent` option to suppress informational logs.

* **Changes**
  * Read-only access to underlying data system from the dataset loader.
  * test report function now returns structured metric dictionaries.

* **Tests**
  * New unit and integration tests covering full-validation, config validation, and checkpoint rotation.


deepmd/pt/train/training.py
```diff
@@ -857,6 +861,57 @@ def single_model_finetune(
...
+            if validation_data is None:
+                raise ValueError(
+                    "validating.full_validation requires `training.validation_data` "
+                    "to be configured."
+                )
+            if self.zero_stage >= 2:
+                raise ValueError(
+                    "validating.full_validation only supports single-task energy "
+                    "training with training.zero_stage < 2."
+                )
```
wanghan-iapcm: make a method for all these value checks? so the streamline of the implementation is not disturbed too much.

wanghan-iapcm (changes requested):

deepmd/pt/train/validation.py
```diff
@@ -0,0 +1,759 @@
...
+        if self.is_distributed:
+            dist.broadcast_object_list(save_path, src=0)
+
+        if save_path[0] is not None:
+            try:
+                if not self.is_distributed or self.zero_stage == 0:
+                    if self.rank == 0:
+                        save_checkpoint(Path(save_path[0]), lr=lr, step=step_id)
+                else:
+                    save_checkpoint(Path(save_path[0]), lr=lr, step=step_id)
```
iProzd:
```suggestion
                if (self.is_distributed and self.zero_stage != 0) or self.rank == 0:
                    save_checkpoint(Path(save_path[0]), lr=lr, step=step_id)
```

iProzd (changes requested):

deepmd/pt/train/validation.py
```diff
@@ -0,0 +1,759 @@
...
+    natoms: int,
+    has_pbc: bool,
+) -> dict[str, tuple[float, float]]:
+    """Compute per-system full validation metrics."""
+    metrics: dict[str, tuple[float, float]] = {}
+    find_energy = bool(test_data.get("find_energy", 0.0))
+    find_force = bool(test_data.get("find_force", 0.0))
+    find_virial = bool(test_data.get("find_virial", 0.0))
+
+    if find_energy:
```
njzjz: This method is different from its name. It seems to only be used for energy, but the method name has no information with energy.

deepmd/pt/train/validation.py
njzjz, line 1: Also, it seems to me that most of codes is duplicated with dp test

njzjz (approved):
LGTM

iProzd (approved):

wanghan-iapcm (approved):

PR #5356, opened 2026-03-30
feat(pt/dpmodel): add use_default_pf

iProzd:
## Summary by CodeRabbit

* **New Features**
  * Added use_default_pf option to enable prefactor-force loss using a default atom preference of 1.0 so atom_pref.npy can be omitted.

* **Behavior Changes**
  * Prefactor-force loss will be computed when the option is enabled; label requirements expanded and serialization schema version bumped to record the new option. PyTorch supports the option; TensorFlow and Paddle explicitly reject it.

* **Documentation**
  * Training docs updated with guidance and an example config.

* **Tests**
  * Added unit and cross-backend tests for default-prefactor behavior and serialization/versioning.


anyangml (approved):
LGTM

njzjz (commented):
Could `use_default_pf` have a better name? @coderabbitai

wanghan-iapcm (changes requested):
consistency test is required to check the cross backend consistency for the support.

wanghan-iapcm (approved):

iProzd:
> I think the new documentation is currently placed in the wrong page.
> 
> `doc/model/train-se-a-mask.md` is a `se_a_mask` page whose supported backend is listed as TensorFlow only, but this PR explicitly raises `NotImplementedError` for `use_default_pf=True` in the TensorFlow and Paddle loss implementations. So the new example in this page describes a configuration that should not work for the documented backend.
> 
> Could you move this note/example to a generic energy-loss / `atom_pref` documentation page, or to a PyTorch/DPModel-supported training page, and keep the backend limitation visible there? The `argcheck.py` text already says this option is only effective for PyTorch/DPModel, so the standalone docs should be consistent with that.
> 
> — OpenClaw 2026.4.22 (00bd2cf) (model: gpt-5.5)

@njzjz-bot docs updated

PR #5393, opened 2026-04-11
perf(pt_expt): use inductor+dynamic for torch.compile training

wanghan-iapcm:
## Summary
- Replace `aot_eager` backend + manual nall padding with `inductor` backend + `dynamic=True` for training compilation
- Use `make_fx(tracing_mode="symbolic")` instead of `tracing_mode="real"` to capture shape-polymorphic ops
- Inductor options: `shape_padding=True`, `max_autotune=False`, `epilogue_fusion=False`, `triton.cudagraphs=False`, `max_fusion_size=8`
- Removes ~120 lines of manual padding/recompilation infrastructure (`_CompiledModel._recompile`, max_nall estimation from 20 sampled batches, etc.)
- Add `silut`/`custom_silu` activation support to pt_expt `_torch_activation` (needed for DPA3)
- Fix: duplicate trace sample to nframes=2 when nframes=1, so symbolic tracer creates a dynamic batch dimension (supports `batch_size: "auto"` with mixed-size systems)

### DPA1 (se_atten_compressible: rcut=6, sel=120, fitting=[240,240,240], float64)

| Mode | bs=1 | bs=4 |
|---|---|---|
| Uncompiled | 21.8 ms | 42.9 ms |
| Old compiled (aot_eager) | 18.1 ms (1.20x) | 38.3 ms (1.12x) |
| New compiled (inductor) | 9.8 ms (2.22x) | 20.4 ms (2.10x) |

### DPA2 (input_torch_small, float32, bs=1)

| Mode | time/step |
|---|---|
| Uncompiled | 63.1 ms |
| Compiled (inductor) | 22.1 ms (**2.85x**) |

### DPA3 (silut:10.0, static sel, float32, bs=1)

| Mode | time/step |
|---|---|
| Uncompiled | 164.5 ms |
| Compiled (inductor) | 45.7 ms (**3.60x**) |

### DPA1 (se_atten_compressible) — rmse_f_val

| step | Uncompiled | Compiled |
|---|---|---|
| 1 | 1.37 | 1.36 |
| 500 | 0.412 | 0.497 |
| 1000 | 0.291 | 0.316 |

### DPA2 — val loss / e_rmse / f_rmse

| step | Uncompiled | Compiled |
|---|---|---|
| 1 | 24.5 / 1.96e-01 / 0.776 | 27.2 / 1.98e-01 / 0.861 |
| 500 | 20.8 / 2.45e-02 / 0.659 | 17.2 / 1.87e-01 / 0.544 |
| 1000 | 7.32 / 3.63e-02 / 0.411 | 10.3 / 3.65e-02 / 0.576 |

### DPA3 (silut:10.0) — val loss / e_rmse / f_rmse

| step | Uncompiled | Compiled |
|---|---|---|
| 1 | 25.3 / 9.17e-02 / 0.800 | 22.4 / 9.45e-02 / 0.707 |
| 500 | 24.5 / 2.14e-02 / 0.776 | 24.0 / 1.66e-02 / 0.759 |
| 1000 | 13.1 / 1.13e-02 / 0.692 | 7.27 / 4.14e-03 / 0.384 |

All models converge comparably. Variation is within normal run-to-run noise from random seeds/batch ordering.

## Varying natoms + batch size

Compiled training with `batch_size: "auto"` across systems of different atom counts (192-atom + 6-atom) works correctly. Both `nframes` and `natoms` vary across steps. Tested locally (se_e2_a) and on V100 (DPA3 with 192+64 atoms).

## Known limitations
- DPA3 with `use_dynamic_sel: true` cannot be compiled (data-dependent `int()` in `get_graph_index`). Only static sel configs are supported.
- fparam/aparam with varying nframes is untested (the code handles it but no test exercises it).

## Test plan
- [x] All 13 training tests pass locally (including 2 new varying-natoms tests)
- [x] silut unit tests (make_fx, torch.export, gradient, branch coverage)
- [x] Cross-backend consistency tests (dpmodel, pt, pt_expt) for all activations including silut variants
- [x] Varying natoms compiled training (192-atom + 6-atom, batch_size: "auto", mocked system selection)
- [ ] CI passes

### DIFF
--- a/deepmd/pt_expt/train/training.py
+++ b/deepmd/pt_expt/train/training.py
@@ -233,8 +233,10 @@ def _trace_and_compile(
 
     # Override backend and dynamic — the inductor backend with
     # dynamic=True handles varying shapes automatically.
-    compile_opts.pop("dynamic", None)
-    compile_opts.pop("backend", None)
+    # Work on a copy to avoid mutating the caller-owned dict.
+    compile_opts = {
+        k: v for k, v in compile_opts.items() if k not in ("dynamic", "backend")
+    }
     if "options" not in compile_opts:
         compile_opts["options"] = {}
     opts = compile_opts["options"]
--- a/deepmd/pt_expt/train/training.py
+++ b/deepmd/pt_expt/train/training.py
@@ -153,14 +153,21 @@ def _trace_and_compile(
 ) -> torch.nn.Module:
     """Trace ``forward_lower`` with ``make_fx`` and compile with ``torch.compile``.
 
+    Uses symbolic tracing (``tracing_mode="symbolic"``) so the resulting
+    FX graph captures shape-polymorphic operations.  The graph is then
+    compiled with ``torch.compile(dynamic=True)`` and the inductor
+    backend, which automatically pads tensor shapes for efficient kernel
+    execution (``shape_padding=True``).
+
     Parameters
     ----------
     model : torch.nn.Module
-        The (uncompiled) model.  Temporarily set to eval mode for tracing.
+        The (uncompiled) model.
     ext_coord, ext_atype, nlist, mapping, fparam, aparam
-        Sample tensors (already padded to the desired max_nall).
+        Sample tensors used to drive the symbolic trace.
     compile_opts : dict
-        Options forwarded to ``torch.compile`` (excluding ``dynamic``).
+        Options forwarded to ``torch.compile``.  Keys ``dynamic`` and
+        ``backend`` are set internally and ignored if provided.
 
     Returns
     -------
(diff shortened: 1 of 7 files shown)


OutisLi (changes requested):
please add a test for common silu activation training perhaps

deepmd/pt_expt/train/training.py
OutisLi, line 197: should not detach the grad inside the function which will be traced, detach to a leaf node before tracing and used as compiled function

OutisLi (changes requested):

deepmd/pt_expt/train/training.py
wanghan-iapcm: addressed by #5397

wanghan-iapcm:
> please add a test for common silu activation training perhaps

solved by 7670e6e9

OutisLi (changes requested):
A few issues with the current implementation:

### 1. Indiscriminate detach removal is unsafe

`_remove_detach_nodes` removes every `aten.detach.default` node in the graph. Not all of them come from the autograd saved-tensor mechanism — user-explicit `.detach()` calls (e.g. gradient boundary management, virial computation) also appear as `aten.detach.default` and are semantically meaningful.

Although empirically removing all of them may not cause incorrect results in the current graph (because `make_fx` has already materialized the first-order backward, and the user-explicit detach only affects the already-frozen forward ops), this is fragile and does not preserve the original semantics. If the model structure changes in the future, blind removal could silently break user-intended gradient boundaries.

The two categories can be distinguished by graph topology alone, without hard-coding any op names. Autograd-inserted detach always forms a double-detach chain (`forward_op → detach_A → detach_B → backward_use`), while user-explicit detach is structurally isolated. Three rules suffice:

- **Chain inner**: input is another detach node.
- **Dead node**: no downstream users.
- **Chain head**: all users are detach nodes.

Anything else is user-explicit and should be preserved. Classification must happen on the original graph before any mutation (two-pass: collect, then remove), because removing chain-head nodes changes the input of chain-inner nodes, causing misclassification in a single pass.

Suggested implementation:

```python
def _strip_saved_tensor_detach(gm: torch.fx.GraphModule) -> None:
    _DETACH = torch.ops.aten.detach.default

    def _is_detach(n: torch.fx.Node) -> bool:
        return n.op == "call_function" and n.target == _DETACH

    # Pass 1: classify on the unmodified graph
    to_remove = []
    for node in gm.graph.nodes:
        if not _is_detach(node):
            continue
        inp = node.args[0]
        users = list(node.users.keys())
        is_chain_inner = _is_detach(inp)
        is_dead = len(users) == 0
        is_chain_head = len(users) > 0 and all(_is_detach(u) for u in users)
        if is_chain_inner or is_dead or is_chain_head:
            to_remove.append(node)

    # Pass 2: remove after classification is complete
    for node in to_remove:
        node.replace_all_uses_with(node.args[0])
        gm.graph.erase_node(node)

    gm.graph.lint()
    gm.recompile()
```

### 2. Must guard with training mode — eval-time stripping triggers `silu_backward` derivative error

The stripping should only be applied when tracing in training mode (`create_graph=True`). In eval mode (`create_graph=False`), the saved-tensor detach nodes are correct and harmless. Removing them in eval mode causes inductor to attempt higher-order differentiation of ops like `aten.silu_backward`, which do not have registered higher-order derivatives in PyTorch. This is a pre-existing PyTorch limitation, not something the PR can fix, so the correct approach is:

```python
traced = make_fx(compute_fn, ...)(*trace_args)

if self.training:
    _strip_saved_tensor_detach(traced)

compiled = torch.compile(traced, ...)
```

### 3. A unit test is needed, and it must use random parameter initialization

This bug is nearly invisible under default (zero or near-zero) initialization. At zero init, `tanh(0) = 0` and `sigmoid(0) = 0.5` produce saved tensors at special points where the second-order gradient contribution vanishes or is masked by numerical noise. In my testing, zero-init showed 0 parameter gradient mismatches; switching to random init immediately exposed ~80% relative error in 5 out of 54 parameters.

The test should:
- Build paired models (eager vs compiled) with **shared random weights**
- Compare force-loss (`(force**2).sum()`) backward gradients (second-order) across all parameters
- Use tight tolerance — the differences from this bug are O(1) relative, not floating-point noise

Minimal structure:

```python
def test_compile_second_order_grad_matches_eager(self):
    model_eager = build_model(use_compile=False)
    # Random init is critical — zero init hides the bug
    torch.manual_seed(42)
    with torch.no_grad():
        for p in model_eager.parameters():
            p.copy_(torch.randn_like(p) * 0.1)

    model_compiled = build_model(use_compile=True)
    model_compiled.load_state_dict(model_eager.state_dict())
    model_eager.train(); model_compiled.train()

    # Forward match
    out_e = model_eager(coord, atype, box=box)
    out_c = model_compiled(coord, atype, box=box)
    torch.testing.assert_close(out_e["force"], out_c["force"], ...)

    # Second-order: force-loss backward
    model_eager.zero_grad(); model_compiled.zero_grad()
    (out_e["force"] ** 2).sum().backward()
    (out_c["force"] ** 2).sum().backward()

    for (ne, pe), (_, pc) in zip(
        model_eager.named_parameters(),
        model_compiled.named_parameters(),
    ):
        ge = pe.grad if pe.grad is not None else torch.zeros_like(pe)
        gc = pc.grad if pc.grad is not None else torch.zeros_like(pc)
        torch.testing.assert_close(ge, gc, atol=1e-5, rtol=1e-5,
                                   msg=f"Grad mismatch: {ne}")
```

OutisLi:
Solution for `silu_backward derivative not implemented`:

```python
# Decompose silu_backward into primitive ops (sigmoid + mul + ...)
# so that inductor can compile the eval graph without requiring a
# higher-order derivative that PyTorch does not register for silu.
from torch._decomp import get_decompositions

decomp_table = get_decompositions([torch.ops.aten.silu_backward.default])

traced = make_fx(
    compute_fn,
    tracing_mode="symbolic",
    _allow_non_fake_inputs=True,
    decomposition_table=decomp_table,
)(
    trace_coord,
    trace_atype,
    trace_edge_index,
    trace_edge_vec,
    trace_edge_mask,
    trace_fp,
    trace_ap,
)
```

iProzd (approved):

wanghan-iapcm:
close because #5397 is merged.

PR #5410, opened 2026-04-22
fix(pt): typo in epoch training

OutisLi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed configuration lookup for the epoch count so training now reads and respects the configured number of epochs.

### DIFF
--- a/deepmd/pd/train/training.py
+++ b/deepmd/pd/train/training.py
@@ -141,7 +141,7 @@ class Trainer:
 
         # Iteration config
         self.num_steps = training_params.get("numb_steps")
-        self.num_epoch = training_params.get("num_epoch")
+        self.num_epoch = training_params.get("numb_epoch")
         self.num_epoch_dict = training_params.get("num_epoch_dict")
         self.acc_freq: int = training_params.get(
             "acc_freq", 1
--- a/deepmd/pt/train/training.py
+++ b/deepmd/pt/train/training.py
@@ -164,7 +164,7 @@ class Trainer:
 
         # Iteration config
         self.num_steps = training_params.get("numb_steps")
-        self.num_epoch = training_params.get("num_epoch")
+        self.num_epoch = training_params.get("numb_epoch")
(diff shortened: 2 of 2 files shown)


njzjz (commented):
I searched `num_epoch` and found many other places using `num_epoch`

https://github.com/search?q=repo%3Adeepmodeling%2Fdeepmd-kit%20num_epoch&type=code

OutisLi:
> I searched `num_epoch` and found many other places using `num_epoch`
> 
> https://github.com/search?q=repo%3Adeepmodeling%2Fdeepmd-kit%20num_epoch&type=code



> I searched `num_epoch` and found many other places using `num_epoch`
> 
> https://github.com/search?q=repo%3Adeepmodeling%2Fdeepmd-kit%20num_epoch&type=code

Thanks for the catch! I went through every num_epoch hit in that search.

The only other occurrence that had the same bug was in the Paddle backend, which I've now fixed in the latest commit:

deepmd/pd/train/training.py: training_params.get("num_epoch") → training_params.get("numb_epoch") (same root cause — the canonical key after normalize() is numb_epoch, while num_epoch is only an alias).
The remaining num_epoch mentions are intentional and not bugs:

deepmd/tf/entrypoints/train.py and change_bias.py: num_epoch is just a local variable name; the .get(...) call already uses the canonical "numb_epoch".
self.num_epoch, num_epoch_dict_config, test_sampler.py::num_epoch = 1.5, etc.: internal attribute/variable/parameter names, unrelated to the config key.
num_epoch_dict is itself the canonical name (not an alias of numb_epoch_dict), so all usages are correct.
num_epoch in docstrings and error messages in argcheck.py / TF entrypoints is user-facing text; since num_epoch is a documented alias, keeping it matches what most users actually write in their input files.
examples/water/se_e2_a/input_torch_num_epoch.json uses "num_epoch": 100, which is a legal alias and still works

njzjz (approved):
Thanks!

PR #5416, opened 2026-04-23
feat(pt_expt): make model.json optional in .pt2/.pte loading

OutisLi:
The pt_expt DeepEval's inference path runs through aoti_load_package / the exported module; `self._dpmodel` is only used to resolve metadata (rcut / type_map / atomic_output_def / dim_fparam / ...), which is already available in extra/metadata.json (the contract the C++ reader DeepPotPTExpt enforces). Drop the requirement that extra/model.json be present:

* _load_pt2 / _load_pte: model.json is optional; metadata.json is now the minimum contract.
* _init_from_metadata: reconstructs ModelOutputDef from the serialised fitting_output_defs and hoists sel / mixed_types to plain attributes.
* get_dim_fparam / get_dim_aparam / get_sel_type / model_type / get_use_spin: fall back to metadata when _dpmodel is None.
* eval_descriptor / eval_typeebd / eval_fitting_last_layer: raise a descriptive NotImplementedError in metadata-only mode (they inspect the dpmodel instance directly).

Also fixes two metadata-completeness gaps so metadata-only load is exact:
* _collect_metadata: add the `sel_type` field so get_sel_type works without a dpmodel round-trip (relevant for dipole / polar / wfc).
* _collect_metadata: force vdef.category to plain int for deterministic JSON serialisation across Python versions.

Archives produced by existing pt_expt serialisation still contain model.json and continue to use the dpmodel path unchanged. Regression covered by 77 existing tests in test_deep_eval.py + a dedicated new suite (test_deep_eval_metadata_only.py) that strips extra/model.json and asserts bitwise parity against the full archive.

## Summary by CodeRabbit

* **New Features**
  * Support loading archives in a metadata-only mode so inference can run without an embedded model file.

* **Behavior Changes**
  * Public accessors derive values from hoisted metadata when the model is absent; spin-related flags preserved.
  * Some introspection, descriptor and fitting-layer evaluations are disabled in metadata-only mode and raise NotImplementedError.
  * Archive metadata is now stored under model/extra/ and includes selection-type info for stable serialization.

* **Tests**
  * End-to-end and negative tests validate metadata-only loading, parity of inference outputs, and archive layout.

### DIFF
--- a/deepmd/pt_expt/infer/deep_eval.py
+++ b/deepmd/pt_expt/infer/deep_eval.py
@@ -148,6 +148,12 @@ class DeepEval(DeepEvalBackend):
 
         self.rcut = self._dpmodel.get_rcut()
         self.type_map = self._dpmodel.get_type_map()
+        # Hoist sel / mixed_types to plain attributes so the inference hot
+        # path (`_build_nlist_*`) does not need the dpmodel instance.  This
+        # matches what `_init_from_metadata` sets and keeps both code paths
+        # numerically identical.
+        self.sel = list(self._dpmodel.get_sel())
+        self.mixed_types = bool(self._dpmodel.mixed_types())
         if self._is_spin:
             self._model_output_def = ModelOutputDef(
                 FittingOutputDef(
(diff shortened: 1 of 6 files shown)


deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -266,9 +370,11 @@ def get_has_spin(self) -> bool:
 
     def get_use_spin(self) -> list[bool]:
         """Get the per-type spin usage of this model."""
-        if getattr(self, "_is_spin", False):
+        if not getattr(self, "_is_spin", False):
```
wanghan-iapcm: Here the `_is_spin` member should always present. no need to guard with `getattr`

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -159,47 +165,127 @@ def _init_from_model_json(self, model_json_str: str) -> None:
...
+        ``aoti_load_package`` / the exported module and uses plain
+        attributes (``self.rcut``, ``self.sel``, ``self.mixed_types``,
+        ``self._model_output_def``) for all metadata-level queries.
+        """
+        self._dpmodel = None
+        self._is_spin = bool(self.metadata.get("is_spin", False))
+        self.rcut = float(self.metadata["rcut"])
+        self.type_map = list(self.metadata["type_map"])
+        self.sel = [int(s) for s in self.metadata["sel"]]
+        self.mixed_types = bool(self.metadata["mixed_types"])
```
wanghan-iapcm: why some model attributes start with "_" while some are not?

wanghan-iapcm (commented):
**Test coverage gap: `.pt2` metadata-only branch is untested**                                                                                                                                                                                               
                                                                           
  The new `_init_from_metadata` fallback in `_load_pt2` (the `model_json_str == ""` branch) is not exercised end-to-end. `test_deep_eval_metadata_only.py` only does the zip-strip surgery on `.pte`.                                                          
                                                                                                                                                                                                                                                               
  Since `test_deep_eval.py::setUpClass` already builds a shared AOTI-compiled `.pt2` (paying the ~82 s compile cost once per class), adding a metadata-only `.pt2` case is essentially free: strip `model/extra/model.json` from the shared archive, reload via
   `DeepPot`, and assert numeric parity against the full archive — the same pattern already used for `.pte`. 


**Add a spin metadata-only parity test**                                                                                                                                                                                                                     
                                                                            
  The new `_init_from_metadata` reconstructs `ModelOutputDef` from the serialized `fitting_output_defs`, but `_init_from_model_json` builds the spin output def by hand (`energy`, `shape=[1]`, `magnetic=True`, etc.). Whether these two paths produce        
  equivalent defs for a spin model is not asserted anywhere — the existing `test_deep_eval_spin.py` only loads full archives.
                                                                                                                                                                                                                                                               
  Likewise, `get_use_spin`'s metadata fallback (`[bool(v) for v in self.metadata.get("use_spin", [])]`) only fires when a spin archive is loaded metadata-only, so it's currently dead from the test suite's perspective.                                      
                                                    
  Suggestion: reuse the `.pte` fixture in `test_deep_eval_spin.py`, strip `extra/model.json` (same pattern as `test_deep_eval_metadata_only.py::_strip_extra_model_json`), reload, and assert:                                                                 
                                                                            
  - bitwise numeric parity of `eval` (energy / force / virial / magnetic force) against the full archive,                                                                                                                                                      
  - `get_has_spin()`, `get_use_spin()`, `get_ntypes_spin()` agree,          
  - the reconstructed `_model_output_def` matches the hand-built one (e.g. compare `def_outp.get_data()` keys + per-variable `magnetic` / `r_differentiable` / `c_differentiable` flags).

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -995,13 +1088,44 @@ def get_model(self) -> torch.nn.Module:
...
+
+        Primary path: the :attr:`_is_spin` attribute set by the loaders
+        — this works for both ``model.json`` and metadata-only archives
+        (a spin ``.pt2`` carries ``is_spin=true`` in its metadata).
+
+        Legacy path: ``isinstance(_dpmodel, SpinModel)`` — retained for
+        tests that construct a non-spin archive and then swap
+        :attr:`_dpmodel` to a :class:`SpinModel` instance after load.
+        """
+        if bool(getattr(self, "_is_spin", False)):
```
wanghan-iapcm: use `self._is_spin` ?

wanghan-iapcm (approved):

PR #5431, opened 2026-05-05
feat(dpa3): decouple charge_spin from fparam

iProzd:
`add_chg_spin_ebd=True` previously hijacked `fparam` to smuggle the
  [charge, spin] scalars into DPA3, forcing users to set `numb_fparam=2`
  on the fitting net and blocking real frame parameters from coexisting
  with charge/spin. This PR plumbs `charge_spin: Tensor | None` as a
  first-class kwarg through every forward chain and adds an optional
  `default_chg_spin` fallback on the DPA3 descriptor.

  Backends covered: pt, dpmodel, pt_expt. The pd backend is left untouched.
  The C/C++/LAMMPS layer is unchanged.

  ## Forward chain

  `Calculator / deep_eval / dp test / lmdb_data / training.get_data`
    -> `wrapper.forward`
    -> `ener_model.forward / forward_lower`
    -> `make_model.forward_common / forward_common_lower`
    -> `base_atomic_model.forward_common_atomic`
    -> `dp_atomic_model.forward_atomic`     # default_chg_spin fallback here
    -> `descriptor.forward`                  # only DPA3 consumes it

  All other descriptors (se_e2_a, se_r, se_t, se_t_tebd, dpa1, dpa2,
  hybrid) only forward the kwarg through their signatures.

  ## New API surface

  On `BaseAtomicModel` and the wrapped model:

  - `has_chg_spin_ebd() -> bool`
  - `get_dim_chg_spin() -> int`               # 2 for DPA3, else 0
  - `has_default_chg_spin() -> bool`
  - `get_default_chg_spin() -> list[float] | Tensor | None`

  DPA3 descriptor gains a `default_chg_spin: list[float] | None = None`
  constructor arg (length 2, validated; round-trips through `serialize`).
  `descrpt_dpa3_args` exposes the matching `Argument` and the
  `add_chg_spin_ebd` doc no longer references fparam.

  ## Training data

  `charge_spin` is registered as a `DataRequirementItem(ndof=2,
  atomic=False, must=not has_default_cs, default=cs_default)`. The
  `get_data` path drops it (along with fparam) on frames where
  `find_charge_spin == 0`, so missing per-frame data falls back to
  `default_chg_spin` when one is configured.

  ## pt_expt specifics

  `forward_common_atomic`, `forward_common_lower_exportable`, the
  `make_fx`-traced inner `fn`, `_trace_and_compile`, and all wrapping
  energy/spin/dipole/dos/polar/property/dp_linear/dp_zbl model variants
  gained a `charge_spin` arg in lockstep so the export and inductor-
  compiled paths keep matching signatures. `deep_eval` no longer reuses
  `fparam` for charge/spin — it constructs `charge_spin_t` (with the
  metadata default-fallback) and passes it explicitly.

  ## Tests

  Three `cs_mode` cases are exercised everywhere it matters:
  `no_chg_spin`, `explicit_chg_spin`, `default_chg_spin`.

  - pt UT (`source/tests/pt/model/test_dpa3.py::test_consistency`)
    rewritten over the three modes; default mode also asserts that the
    default-fallback descriptor matches an explicit `[5,1]` peer.
  - pt_expt UT (`source/tests/pt_expt/descriptor/test_dpa3.py`) gains
    `test_consistency_chg_spin` covering explicit and default modes
    against dpmodel.
  - Universal tests: `DescriptorParamDPA3` learns `default_chg_spin`,
    parametrize gains `(None, [5.0, 1.0])`, and the
    `add_chg_spin_ebd` skip rule in `test_model.py` is replaced —
    the universal driver does not feed `charge_spin`, so chg_spin runs
    rely on the `default_chg_spin` fallback. 622 DPA3 model cases pass.
  - Consistent tests: `descriptor/common.py` threads `charge_spin`
    through every `eval_*` (pd ignores it). `test_dpa3.py` swaps
    `self.fparam` for `self.charge_spin`. `test_ener.py::
    TestEnerChgSpinEbdFparam` is reparametrized over the three modes
    and no longer touches `numb_fparam` / `default_fparam`.

  ## Smoke

  `examples/water/dpa3 dp --pt train input_torch_dynamic.json
  --skip-neighbor-stat` runs to batch 600 with monotonically decreasing
  loss.

  ## Test plan

  - [x] pytest source/tests/pt/model/test_dpa3.py -v
  - [x] pytest source/tests/pt_expt/descriptor/test_dpa3.py -v
  - [x] pytest source/tests/consistent/descriptor/test_dpa3.py -v
  - [x] pytest source/tests/consistent/model/test_ener.py::TestEnerChgSpinEbdFparam -v
  - [x] pytest source/tests/universal/dpmodel/model/test_model.py -k "DPA3 and 5"
  - [x] examples/water/dpa3 smoke training (600 batches, loss decreasing)

## Summary by CodeRabbit

* **New Features**
  * Optional per-frame charge-spin input supported end-to-end: data readers, batching, inference, training, export/tracing; prediction/training calls accept and forward it.
  * Models/descriptors expose capability-query and default-value helpers for charge-spin embeddings; exportable/traced APIs honor defaults.

* **Tests**
  * Tests updated/expanded to validate charge-spin embedding behavior and cross-backend consistency.

* **Chores**
  * Configuration normalization warns on legacy charge/spin packed into legacy parameters and documents migration.


deepmd/dpmodel/descriptor/dpa3.py
```diff
@@ -359,6 +359,7 @@ def __init__(
         use_loc_mapping: bool = True,
         type_map: list[str] | None = None,
         add_chg_spin_ebd: bool = False,
+        default_chg_spin: list[float] | None = None,
```
wanghan-iapcm, line 380: `default_chg_spin` could be given `[0.0, 0.0]`.

wanghan-iapcm (changes requested):

deepmd/pt_expt/model/dipole_model.py
wanghan-iapcm, line 1: do not introduce line-ending / trailing-whitespace re-formatting in this pr, which make the file diff very large

wanghan-iapcm (changes requested):
- this is a breaking change. when trying to use old model with `default_fparam` injecting to the fitting net, an error should be raised. 
- dipole_model.py, dos_model.py, dp_zbl_model.py, polar_model.py, property_model.py all show large diffs (+155/−142 etc.) where the substantive change is ~13 lines of charge_spin plumbing. The remainder is line-ending / trailing-whitespace re-formatting   that makes review and git blame hard.
- deepmd/pt/train/training.py:1909 and deepmd/pt_expt/train/training.py:141:  `has_chg_spin_ebd` is now defined on the base atomic model — so the False default can never be returned (it's always a method). The `callable()` check is dead-code-defensive. Simplify to a direct call. 
```python
has_chg_spin_ebd = getattr(_model, "has_chg_spin_ebd", False)                                                                                                                                                                                                
if callable(has_chg_spin_ebd):                                                                                                                                                                                                                               
     has_chg_spin_ebd = has_chg_spin_ebd()                
```

deepmd/dpmodel/descriptor/dpa3.py
```diff
@@ -414,6 +415,11 @@ def init_subclass_params(sub_data: dict | Any, sub_class: type) -> Any:
 
         self.use_econf_tebd = use_econf_tebd
         self.add_chg_spin_ebd = add_chg_spin_ebd
+        self.default_chg_spin = default_chg_spin
```
wanghan-iapcm: the validity of the length of `default_chg_spin` should be checked before assignment.

source/tests/pt/model/test_dpa3.py
```diff
@@ -158,6 +166,42 @@ def test_consistency(
                 atol=atol,
             )
 
+            # default_chg_spin should match explicit when value is the same.
```
wanghan-iapcm, line 178: I would recommend moving the consistency check to dpmodel backend, as pt will be deprecated in the future.

wanghan-iapcm (changes requested):

deepmd/dpmodel/descriptor/dpa3.py
iProzd: It's intentional. When the user sets `add_chg_spin_ebd=True` but does not set `default_chg_spin` , it stays None, and `charge_spin.npy` is then required explicitly. Missing data raises a clear error rather than silently falling back.

source/tests/pt/model/test_dpa3.py
iProzd: Removed the dpmodel-vs-pt comparison block from `source/tests/pt/model/test_dpa3.py` — the pt test now only covers pt-specific paths (serialization round-trip and default_chg_spin == explicit equivalence). 

Cross-backend numeric consistency for add_chg_spin_ebd is already covered by `source/tests/consistent/descriptor/test_dpa3.py`.

iProzd:
> * this is a breaking change. when trying to use old model with `default_fparam` injecting to the fitting net, an error should be raised.

Addressed with a strong warning rather than a hard error, in `deepmd/utils/argcheck.py::_check_dpa3_chg_spin_migration`. Reasoning:
 The legacy pattern was add_chg_spin_ebd=True + numb_fparam=2 (+ optionally default_fparam=[charge, spin]). After the decoupling, that exact same combination is now also a legitimate new-API config — a user may legitimately want DPA3's chg-spin embedding alongside two genuine frame parameters that just happen to have a default. From the JSON config alone we cannot distinguish "the numb_fparam=2 is leftover legacy charge/spin" from "the numb_fparam=2 is real new fparam," so any auto-raised error here would also reject valid new configs.

What we do instead:
- On every normalize() call, if descriptor.type=="dpa3" and add_chg_spin_ebd=True and fitting_net.numb_fparam==2, emit a warnings.warn(...) with explicit migration instructions (use charge_spin input or default_chg_spin).
 - Old saved models whose numb_fparam was set to 2 only because of legacy charge/spin will produce numerically wrong results if loaded as-is — but they will fail at the data-loading stage (charge_spin will be required) rather than silently mis-feeding fparam, so the user is forced to migrate.

source/tests/pt/model/test_dpa3.py
```diff
@@ -66,13 +65,20 @@ def test_consistency(
             [1, 2],  # n_multi_edge_message
             ["float64"],  # precision
             [False],  # use_econf_tebd
-            [False, True],  # add_chg_spin_ebd
+            ["no_chg_spin", "explicit_chg_spin", "default_chg_spin"],
```
wanghan-iapcm, line 87: you consistency test does not have such a complete coverage, therefore, other backends are not fully covered. at least you should cover the dpmodel backend by all these cases.

wanghan-iapcm (changes requested):

source/tests/pt/model/test_dpa3.py
iProzd: added new `default_chg_spin` test in https://github.com/deepmodeling/deepmd-kit/pull/5431/commits/03d21412ffd522c5118a91ac495eb81113a6c9aa

wanghan-iapcm (approved):

iProzd:
@copilot resolve the merge conflicts in this pull request

iProzd:
Update on the charge embedding range fix: kept the embedding size at 200 (not 201) to maintain backward compatibility. The valid charge range is now [-100, 99] with explicit validation, which maps to indices [0, 199]. This avoids resizing the embedding table while still preventing out-of-bounds access.

iProzd:
Thanks for the detailed review! I've addressed all 7 issues:

**1. JAX vmap shape issue** (`deepmd/jax/model/base_model.py`):
Fixed. Made `charge_spin` an explicit argument to `eval_output` and `eval_ce`, passing `charge_spin_[None, ...]` to `forward_common_atomic` to ensure correct per-frame slicing under vmap.

**2. Hybrid descriptor capability** (`deepmd/pt/model/descriptor/hybrid.py`):
Fixed. Changed from hard-coded `0` to `max(child.get_dim_chg_spin())` to properly aggregate child descriptor capabilities.

**3. Hybrid descriptor kwargs forwarding** (`deepmd/pt/model/descriptor/hybrid.py`):
Fixed. Now forwarding `comm_dict`, `fparam`, and `charge_spin` to child descriptors instead of only `charge_spin`. Also applied the same fix to `deepmd/dpmodel/descriptor/hybrid.py`.

**4. Type casting** (`deepmd/dpmodel/model/make_model.py`):
Fixed. Added `charge_spin` to `_input_type_cast` signature and cast it to model precision along with `coord`, `box`, `fparam`, and `aparam`.

**5. Charge embedding range** (`deepmd/pt/model/descriptor/dpa3.py`):
Fixed. Kept embedding size at 200 for backward compatibility. Valid charge range is now `[-100, 99]` with explicit validation, mapping to indices `[0, 199]`. Added range validation for both charge and spin before indexing.

**6. Shape validation** (`deepmd/pt_expt/infer/deep_eval.py`):
Fixed. Added explicit shape validation for `charge_spin` before reshape, raising clear `ValueError` if shape is not exactly `(nframes, 2)`.

**7. API documentation** (`deepmd/pt_expt/infer/deep_eval.py`):
Fixed. Added `charge_spin` parameter documentation to `eval` method docstring, including expected shape `(nframes, 2)` and `default_chg_spin` fallback behavior.

iProzd:
CI fix for JIT failures in hybrid descriptor tests:

The previous fix for the `get_default_chg_spin` aggregation in `hybrid.py` annotated the return type as `list[float] | None`, which caused TorchScript JIT failures when the caller (`dp_atomic_model.py`) tried to call `.unsqueeze(0)` on the returned value. The PT DPA3 descriptor returns `Optional[torch.Tensor]`, so the hybrid wrapper must use a matching tensor return type. Additionally, the `max(generator, default=0)` pattern used in `get_dim_chg_spin` was not JIT-friendly.

Changes:
- `deepmd/pt/model/descriptor/hybrid.py`: Changed `get_default_chg_spin` return type from `list[float] | None` to `Optional[torch.Tensor]` with `@torch.jit.export` decorator. Replaced `max(...)` generator with explicit for-loop for JIT compatibility.
- `deepmd/dpmodel/descriptor/hybrid.py`: Added matching `get_dim_chg_spin`, `has_default_chg_spin`, and `get_default_chg_spin` methods so the dpmodel hybrid correctly aggregates child capabilities (returns `list[float] | None` since dpmodel stores it as a list).

Verified locally that `test_jit`, `test_permutation`, `test_rot`, `test_get_dim_aparam`, `test_model_output_type`, and related tests on `DescriptorParamHybrid`/`DescriptorParamHybridMixed` now pass.

iProzd:
@njzjz-bot Updated. pt_expt now matches the existing PT reshape-compatible behavior and wraps invalid sizes in a clear ValueError instead of requiring the exact initial shape.

iProzd:
> Thanks for the updates — CI is now green and the previous PT_EXPT export/metadata/shape issues look fixed. I did one more pass and found one remaining issue in the hybrid descriptor aggregation.
> 
> `DescrptHybrid.get_dim_chg_spin()` currently returns the maximum child `get_dim_chg_spin()`, and `get_default_chg_spin()` returns the first child default. However, `forward()` passes the same frame-level `charge_spin` to every sub-descriptor, so the charge/spin input dimensionality is shared, not concatenated per child. With a hybrid containing multiple DPA3 descriptors that each have `add_chg_spin_ebd=True`, returning `max == 2` is the right runtime shape; summing would make export/training allocate `(nframes, 4)` and the DPA3 child would still only consume columns 0/1.
> 
> The remaining bug is `has_default_chg_spin()`: it returns true when _any_ child has a default, and `get_default_chg_spin()` returns that first default. That can silently hide a missing required `charge_spin` for another DPA3 child with `add_chg_spin_ebd=True` but no default. PT eager inference, PT_EXPT export sample inputs, and metadata-only inference may then all run with the first child’s default for every child instead of requiring explicit `charge_spin`.
> 
> I think hybrid default semantics should be stricter: if any child requires `charge_spin`, all charge/spin-consuming children should either have no default (so `charge_spin` remains required) or have the same shared default. If defaults differ or only some children provide one, raise during construction or make `has_default_chg_spin()` false so callers still require explicit `charge_spin`. A small hybrid DPA3 test would be useful for the mixed-default/missing-default case.
> 
> — OpenClaw 2026.5.12 (model: custom-chat-jinzhezeng-group/gpt-5.5)

Fixed. Hybrid descriptors now only expose a default charge_spin when all charge-spin-consuming children provide the same shared default; mixed or missing defaults now require explicit charge_spin. I also added DP/PT tests for the shared, missing, and mismatched default cases.

njzjz (approved):

Issue #5087, opened 2025-12-10
[BUG] _DeePMD-kit C API Error: DeePMD-kit Error: DeePMD-kit PyTorch backend error: The following operation failed in the TorchScript interpreter.


myyelishu:
### Bug summary

我首先在集群A上使用GPU训练出模型graph.pth，然后使用该模型在集群B的CPU节点运行LAMMPS，此时两个集群都是通过离线方式安装了GPU版本3.1.0的DeePMD-kit，在LAMMPS跑MD的过程中出错了。我的MD过程有两段：第一段是找到平衡态，第二段是施加剪切。我后续又进行了多次测试，总结错误一共两种：第一种是在第一段MD运行中途出错了，报错内容是“       477   14.671651     -1840.8705     -1840.4779      951095.46      1007.5899      957687.5       947820.98      947777.91      31.101126      2.3915695      13.546425      0              0              0            
       478   14.848564     -1840.4616     -1840.0643      956489.71      1006.907       963265.36      953626.93      952576.83      31.093733      2.3909153      13.544167      0              0              0            
ERROR on proc 34: DeePMD-kit C API Error: DeePMD-kit Error: DeePMD-kit PyTorch backend error: The following operation failed in the TorchScript interpreter.
Traceback of TorchScript, serialized code (most recent call last):
  File "code/__torch__/deepmd/pt/model/model/ener_model.py", line 66, in forward_lower
    comm_dict: Optional[Dict[str, Tensor]]=None) -> Dict[str, Tensor]:
    _6 = (self).need_sorted_nlist_for_lower()
    model_ret = (self).forward_common_lower(extended_coord, extended_atype, nlist, mapping, fparam, aparam, do_atomic_virial, comm_dict, _6, )
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
    _7 = (self).get_fitting_net()
    model_predict = annotate(Dict[str, Tensor], {})
  File "code/__torch__/deepmd/pt/model/model/ener_model.py", line 232, in forward_common_lower
    cc_ext, _40, fp, ap, input_prec, = _39
    atomic_model = self.atomic_model
    atomic_ret = (atomic_model).forward_common_atomic(cc_ext, extended_atype, nlist0, mapping, fp, ap, comm_dict, )
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
    _41 = (self).atomic_output_def()
    training = self.training
[...]

PR #5433, opened 2026-05-06
fix(pt): replace in-place nlist masking to enable CUDA graph capture

### LINKED ISSUE #5432: [Feature Request] `torch.tensor()` inside `repformers.forward()` prevents CUDA graph capture for DPA-2 models
### Summary

CUDA graph capture of `forward_lower` fails for DPA-2 models because `repformers.forward()` calls `torch.tensor(0, dtype=..., device='cuda')` on every forward pass. This dynamically allocates a new GPU tensor at runtime, which is forbidden during CUDA graph capture (`cudaErrorStreamCaptureUnsupported`). The result is that CUDA graph optimisation — which eliminates per-kernel-launch overhead and can give ~8–12× speedup on the GPU inference portion — is inaccessible for any DPA-2 model loaded via the LAMMPS C++ plugin.

### Detailed Description

**Exact error**

When attempting to capture `forward_lower` on a dedicated non-default CUDA stream from `DeepPotPT.cc`:

```
RuntimeError: CUDA error: operation not permitted when stream is capturing
Search for `cudaErrorStreamCaptureUnsupported'

TorchScript traceback:
  File "repformers.py", line 151, in forward
    _13 = torch.tensor(0, dtype=ops.prim.dtype(nlist0),
                          device=ops.prim.device(nlist0))
          ~~~~~~~~~~~~ <--- HERE
    _14 = annotate(List[Optional[Tensor]], [_12])
    _15 = torch.index_put_(nlist0, _14, _13)
```

---

**Root cause**

In `deepmd/pt/model/descriptor/repformers.py`:

```python
# Runs on every forward pass — allocates a new GPU tensor each call:
zero = torch.tensor(0, dtype=nlist.dtype, device=nlist.device)
nlist = torch.index_put_(nlist, [nlist == -1], zero)
```

`torch.tensor(scalar, device='cuda')` triggers a GPU memory allocation requiring CPU↔GPU synchronisation, which is explicitly disallowed during CUDA graph capture. Because this is baked into the frozen TorchScript IR of the `.pth` file, it cannot be patched from the C++ plugin side.

---

**Proposed fix**

Replace the dynamic allocation with a persistent module buffer registered at construction:

```python
# In Repformers.__init__:
self.register_buffer('_zero_idx', torch.zeros(1, dtype=torch.long))

# In forward, replace:
#   zero = torch.tensor(0, dtype=nlist.dtype, device=nlist.device)
zero = self._zero_idx.to(dtype=nlist.dtype)   # fixed address, no allocation
nlist = torch.index_put_(nlist, [nlist == -1], zero)
```

`register_buffer` allocates the tensor once at model load with a fixed GPU address. The `index_put_` call then becomes CUDA-graph-capturable.

The same pattern should be audited across all descriptor layers for any other `torch.tensor(scalar, device='cuda')` calls that run inside `forward`.

---

**Impact**

CUDA graphs are a standard PyTorch optimisation for repeated fixed-topology inference. Removing this incompatibility would allow:

- Capture of `forward_lower` in `DeepPotPT.cc` on a dedicated stream
- Replay with only a coordinate `copy_` per step instead of full kernel re-launch
- Estimated ~8–12× reduction in GPU inference time per MD step for message-passing models

Non-message-passing models (e.g. `se_e2_a`) do not have this issue and are already CUDA-graph-capturable.

---

**Additional note: `comm_dict` path**

Even after the above fix, it would be worth verifying that the full `forward_lower` call with `comm_dict` (used for ghost atom coordination on single-GPU LAMMPS runs) is fully CUDA-graph-capturable, as it may contain additional CPU↔GPU synchronisation points.

---

**References**

- PyTorch CUDA Graphs: https://pytorch.org/docs/stable/notes/cuda.html#cuda-graphs
- `cudaErrorStreamCaptureUnsupported`: https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html
- `torch.cuda.make_graphed_callables` requirements: all operations inside the callable must be graph-compatible

_No response_

wanghan-iapcm:
## Summary

- Replace `nlist[nlist == -1] = 0` in `DescrptBlockRepformers.forward` (DPA-2) and `DescrptBlockRepflows.forward` (DPA-3, two occurrences) with `nlist = torch.where(nlist == -1, 0, nlist)`.
- TorchScript was lowering the indexed-assignment into an `index_put_` whose value was `torch.tensor(0, device=...)` — a per-call scalar allocation that triggers a CPU↔GPU sync and is forbidden during CUDA stream capture (`cudaErrorStreamCaptureUnsupported`). This made `forward_lower` of DPA-2 / DPA-3 models non-capturable from the LAMMPS C++ plugin, blocking the ~8–12× CUDA-graph speedup reported in #5432.
- `aten::where` with a Python scalar embeds the literal in the kernel without allocating a device tensor, so the resulting frozen IR is graph-capturable. The pattern matches what `se_atten.py` and `se_t_tebd.py` already use.

## Notes

- `nlist` / `a_nlist` were already rebound to local copies via `torch.where(exclude_mask != 0, ...)` earlier in both functions, so dropping the in-place mutation has no observable effect on callers.
- Existing deployed `.pth` files won't benefit until re-exported (their TorchScript IR is frozen).
- The reporter also flagged that the `comm_dict` ghost-atom path in `forward_lower` may still contain other CPU↔GPU sync points needed for full LAMMPS capture; that audit is out of scope here.

Closes #5432.

## Test plan

- [x] `pytest source/tests/pt/model/test_dpa3.py` — passes (exercises `torch.jit.script` on the DPA-3 descriptor → covers patched `repflows.py` lines)
- [x] `pytest source/tests/pt/model/test_dpa2.py` — passes (numerical consistency of DPA-2 descriptor → covers patched `repformers.py` line)
- [x] Inline `torch.jit.script(DescrptDPA2(...))` succeeds (verifies new `torch.where(... == -1, 0, nlist)` form scripts cleanly)
- [ ] CI: full `test_jit.py` and consistency suites

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal tensor operations in descriptor modules for improved performance and consistency.

### DIFF
--- a/deepmd/pt/model/descriptor/repflows.py
+++ b/deepmd/pt/model/descriptor/repflows.py
@@ -496,8 +496,8 @@ class DescrptBlockRepflows(DescriptorBlock):
         a_sw = a_sw.masked_fill(~a_nlist_mask, 0.0)
         # set all padding positions to index of 0
         # if the a neighbor is real or not is indicated by nlist_mask
-        nlist[nlist == -1] = 0
-        a_nlist[a_nlist == -1] = 0
+        nlist = torch.where(nlist == -1, 0, nlist)
+        a_nlist = torch.where(a_nlist == -1, 0, a_nlist)
 
         # get node embedding
         # [nframes, nloc, tebd_dim]
--- a/deepmd/pt/model/descriptor/repformers.py
+++ b/deepmd/pt/model/descriptor/repformers.py
@@ -457,7 +457,7 @@ class DescrptBlockRepformers(DescriptorBlock):
 
         # set all padding positions to index of 0
         # if the a neighbor is real or not is indicated by nlist_mask
-        nlist[nlist == -1] = 0
+        nlist = torch.where(nlist == -1, 0, nlist)
         # nb x nall x ng1
         if comm_dict is None:
             assert mapping is not None


njzjz (approved):

ganfisher:
I wonder whether this bug is related to #5087, since I encountered the same problem with DPA3

d1l-netizen:
Thank you for the suggested changes! I applied the patches to repformers.py and reflow.py,        
re-froze the model with a fixed nsel for static tensor shapes, and ran a LAMMPS smoke test.       
                                                                                                
The CUDA graph model loads and runs correctly — libcudart.so.12 loads successfully and the run    
completes without errors.                                                                         
                                                                                                
Performance on a small test system (~700 atoms, L40S GPU):                                        
- Baseline (eager): ~10.7 steps/s                                                                 
- CUDA graph: ~10.4 steps/s (~0.97×, essentially no change)                                       
                                                                                                
No speedup at this system size, which is expected — at ~700 atoms the DPA-2 attention kernels are 
short enough that CUDA graph launch-overhead savings are negligible. I'll retest at a larger      
system size (~5k–10k atoms) where the crossover should occur.                                     
                                                                                                
One note for others: verify nsel matches your actual neighbour count at your system density       
before re-freezing.

PR #5434, opened 2026-05-07
fix(pt): base LambdaLR on configured start_lr

OutisLi:
- Fix PyTorch LR scheduler construction to use the configured `start_lr` as `LambdaLR`'s base LR.
- Reset optimizer param group `initial_lr` after loading restart checkpoints so stale checkpoint values do not scale resumed training.
- Avoid double-counting `start_step` by letting `last_epoch` handle scheduler resume position.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal learning-rate scheduler construction for more robust and maintainable training behavior.
* **Bug Fixes**
  * Added runtime validation to reject invalid starting learning-rate values, preventing training misconfiguration.
* **Tests**
  * Added a resume/restart learning-rate test and a unit test ensuring invalid start values are rejected; fixed a corrupted test.

### DIFF
--- a/deepmd/dpmodel/utils/learning_rate.py
+++ b/deepmd/dpmodel/utils/learning_rate.py
@@ -70,7 +70,11 @@ class BaseLR(ABC, PluginVariant, make_plugin_registry("lr")):
             The warmup learning rate starts from warmup_start_factor * start_lr.
             Default is 0.0.
         """
-        # === Step 1. Validate stop_lr and stop_lr_ratio (runtime check) ===
+        # === Step 1. Validate start_lr (runtime check) ===
+        if start_lr <= 0 or not np.isfinite(start_lr):
+            raise ValueError(f"start_lr ({start_lr}) must be positive and finite.")
+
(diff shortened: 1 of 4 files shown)


wanghan-iapcm (changes requested):
No regression test.  better to provide a test that:                                                                                                                                                                                                    
  - Constructs a Trainer, runs N steps, saves checkpoint.                 
  - Resumes from that checkpoint, runs M more steps.                                                                                                                                                                                                           
  - Asserts `scheduler.get_last_lr()[0]` at step N+k matches `lr_schedule.value(N+k)` to high tolerance.

wanghan-iapcm (approved):

iProzd (approved):
The double-offset bug was introduced in #5154 , which added `last_epoch=start_step-1` but kept `+start_step` in the lambda.

So UT for restart training was actually missing in #5154. Old version uses `last_epoch=-1` and `+start_step` in the lambda, which was correct.

njzjz (approved):

PR #5449, opened 2026-05-20
fix: try remove memory footprint

anyangml:
There are two issues hindering multitask DDP training in pt-expt compile mode:
- 1. memory footprint scales with N tasks; 
- 2. lazy compile cause NCCL timeout.

Solutions:
- 1. clean up memory footprint after compiling each task
- 2. compile all task before training start and sync.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced memory usage during model tracing/compilation by immediately releasing temporary trace and compiled artifacts and conditionally clearing GPU cache when CUDA is active.
  * Improved multi-task compilation stability by cleaning up per-task intermediate data right after each task.
  * Ensured logged and validation metrics are converted to Python scalars for accurate aggregation and reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/deepmodeling/deepmd-kit/pull/5449?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


wanghan-iapcm:
### Code review

Found 1 issue:

1. **Three new CUDA guards are inconsistent with each other and with the existing pt_expt convention.**

   The PR introduces three different CUDA-guard patterns:

   - `_trace_and_compile`: `model_uses_cuda and torch.cuda.is_available() and torch.cuda.is_initialized()`
   - `_compile_model`: `DEVICE.type == "cuda" and torch.cuda.is_initialized()`
   - `run()` display path: `torch.cuda.is_available() and torch.cuda.is_initialized()`

   The pre-existing pt_expt convention is simply `DEVICE.type == "cuda"`, used at [`pt_expt/train/training.py:660-662`](https://github.com/deepmodeling/deepmd-kit/blob/b5236de8f66a2b54e462b8f64734e7c0ffe68ec8/deepmd/pt_expt/train/training.py#L660-L662) (the only CUDA guard before this PR) and [`pt_expt/utils/serialization.py:883-885`](https://github.com/deepmodeling/deepmd-kit/blob/b5236de8f66a2b54e462b8f64734e7c0ffe68ec8/deepmd/pt_expt/utils/serialization.py#L883-L885). `torch.cuda.is_initialized()` is not used anywhere else in the codebase.

   - The `is_initialized()` half is dead in all three sites: if `DEVICE.type == "cuda"` (or any model parameter is on CUDA), the context is necessarily initialized.
   - The third site (display path) is also actively wrong on a CPU run on a CUDA-equipped host (`DEVICE=cpu` env var): `torch.cuda.is_available() and torch.cuda.is_initialized()` returns True if anything else in the process touched CUDA, causing a stray `empty_cache()` on an unrelated context. `DEVICE.type == "cuda"` correctly returns False.
   - The `_trace_and_compile` `model_uses_cuda = any(param.is_cuda for ...)` iterates every parameter and buffer on each call — same answer as `DEVICE.type == "cuda"` in pt_expt (every model is on `DEVICE`), at higher cost.

   Suggest collapsing all three sites to:
   ```python
   if DEVICE.type == "cuda":
       torch.cuda.empty_cache()
   ```

   `DEVICE` is already imported in `training.py`; for the module-level `_trace_and_compile` helper it can be imported from `deepmd.pt_expt.utils.env`.

   https://github.com/deepmodeling/deepmd-kit/blob/b5236de8f66a2b54e462b8f64734e7c0ffe68ec8/deepmd/pt_expt/train/training.py#L341-L348

wanghan-iapcm:
### Follow-up: lifeline workaround for `_traced_lower_ref`

The `object.__setattr__(compiled, "_traced_lower_ref", traced_lower)` workaround is correct and the comment explains the GC mechanism well. Two maintainability suggestions before this lands:

1. **Fail loud if PyTorch's `OptimizedModule` ever starts using the same attribute name.** The current code silently overwrites whatever may live there. Replace the bare setattr with:

   ```python
   if hasattr(compiled, "_traced_lower_ref"):
       raise RuntimeError(
           "torch.compile wrapper already exposes _traced_lower_ref; "
           "the FX-graph lifeline mechanism needs review against the "
           "current PyTorch version."
       )
   object.__setattr__(compiled, "_traced_lower_ref", traced_lower)
   ```

   `assert` would be wrong here — it's stripped under `python -O` / `PYTHONOPTIMIZE=1`, and this is a check at an external-library boundary (the shape of `torch.compile`'s return type), which is canonically a `RuntimeError`. The day PyTorch changes how it wraps modules, this turns a silent regression into a loud one.

2. **File the underlying bug upstream.** The root cause — `_remove_detach_nodes` leaving C++ view-metadata with raw pointers to FX-graph-owned `SymInt`s that get GC'd — is a real PyTorch issue worth a minimal reproducer. Without an upstream tracking issue, this workaround becomes permanent debt with no exit path. A short link in the in-code comment (`# upstream: pytorch/pytorch#NNNNN`) would let future maintainers check whether the workaround can be removed.

Neither is blocking; both are about making sure the workaround stays visible and removable rather than quietly accumulating.

https://github.com/deepmodeling/deepmd-kit/blob/b5236de8f66a2b54e462b8f64734e7c0ffe68ec8/deepmd/pt_expt/train/training.py#L335-L345

wanghan-iapcm (commented):
see my posted comments.

PR #5457, opened 2026-05-26
perf(pt_expt): share compiled forward_lower across multi-task shared-fitting tasks

anyangml:
make dataset embedding and energy bias as input not buffer for compile, this allows multitask training share compiled model thus resolve OOM and NCCL timeout issue. Since the empty_cache and del are removed, no GC complaints.

Regression Test
lcurve

## Summary by CodeRabbit

* **New Features**
  * Multi-task training groups models by structure and caches/reuses compiled computation graphs, with per-task buffer handling to support shared fitting nets.

* **Bug Fixes**
  * Checkpoint loading now skips extraneous per-task buffer entries so only original model parameters are restored.
  * Training aggregation coerces tensor-like loss/metric values to floats for accurate reporting.

* **Tests**
  * Added regression test ensuring compiled and eager outputs match per task for shared-fitting, different-descriptor setups.

[Review Change Stack](https://app.coderabbit.ai/change-stack/deepmodeling/deepmd-kit/pull/5457?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


anyangml:
> I think there is still a correctness issue in the compiled graph cache key for multitask models.
> 
> `_get_model_structure_key()` currently uses only the id of the first non-task-specific child under `fitting_net` as the structure key. That can incorrectly treat two task models as sharing the same compiled `forward_lower` graph when they share only the fitting net but have different/non-shared descriptors or other atomic-model state.
> 
> `forward_lower` is not just the fitting net path: it also includes descriptor/atomic-model computation. When the compiled graph traced for task/model 1 is reused for task/model 2, `_CompiledModel.forward()` only supplies the extra task buffers (`bias`, `case_embd`, `out_bias`, `out_std`, etc.). It does not pass descriptor or other atomic-model parameters as dynamic inputs. So if the descriptor is not actually shared, or has different parameters/outputs, task/model 2 may run the graph captured from task/model 1, producing incorrect predictions and gradients.
> 
> I would suggest making the cache key include the identity/structure of all shared components that participate in `forward_lower` — at least the descriptor/atomic-model path in addition to fitting — or, more conservatively, only reusing a compiled graph when both descriptor and fitting components are confirmed to be the same shared objects. Otherwise each task should compile separately.
> 
> A regression test would be helpful: construct two tasks that share `fitting_net` but use distinguishable non-shared descriptors, enable compile, and verify both tasks match the uncompiled results/gradients.
> 
> Minor follow-up: `_CompiledModel.forward()` currently swallows `AttributeError` while collecting task buffers and then passes an empty tuple, which will likely fail later with a less clear compiled-argument mismatch. It would be better to raise a direct `RuntimeError` with the missing task-buffer name/context.
> 
> Authored by OpenClaw (version: 2026.5.12 f066dd2, model: custom-chat-jinzhezeng-group/gpt-5.5)

Is that a valid use case?

wanghan-iapcm (commented):
### Code review

Found 2 issues:

1. Cache key under-specifies what is baked into the compiled graph. `_get_model_structure_key` returns `id()` of the fitting net's first non-task-specific child, but only `bias_atom_e`/`case_embd`/`out_bias`/`out_std` are promoted to FX placeholders — descriptor parameters/buffers (attention weights, type-embedding `davg`/`dstd`, exclude-mask, etc.) remain baked-in constants in the traced graph. Two tasks that share `fitting_net` but have different descriptors (or differ in `ntypes` / `dim_case_embd` / `sel` / `rcut`) produce the same structure key and silently reuse task 0's compiled graph, yielding wrong predictions and gradients. Same concern previously raised by njzjz-bot and Copilot; unresolved on the current head.

https://github.com/deepmodeling/deepmd-kit/blob/9ce8d3e135e7f9aa590a0c9d090a9cf301a49838/deepmd/pt_expt/train/training.py#L109-L130

https://github.com/deepmodeling/deepmd-kit/blob/9ce8d3e135e7f9aa590a0c9d090a9cf301a49838/deepmd/pt_expt/train/training.py#L1126-L1188

Suggested fix: include the descriptor (and any other non-fitting components participating in `forward_lower`) in the key, e.g. `(id(model.get_descriptor()), id(fitting_first_child))`, or build the key from `tuple(id(p) for _, p in model.named_parameters()) + tuple(id(b) for n, b in model.named_buffers() if n.rsplit(".", 1)[-1] not in _TASK_SPECIFIC_BUFFER_NAMES + _ATOMIC_MODEL_TASK_BUFFER_NAMES)`. A regression test where two tasks share fitting but have distinguishable descriptors and assert compiled outputs match eager would catch this.

2. The promoted-buffer set is incomplete. Only `bias_atom_e`, `case_embd`, `out_bias`, `out_std` are promoted to FX placeholders, but fitting nets carry other per-task statistics buffers that are silently baked in as task-0 constants — notably `fparam_avg`/`fparam_inv_std` and `aparam_avg`/`aparam_inv_std` (set per task by `make_stat_input` from each task's data distribution), and the descriptor's `EnvMat` `davg`/`dstd` if descriptor stats also vary per task. For multi-task configs that use `fparam`/`aparam` with shared fitting but task-local stats, all tasks would end up running with task 0's normalization, producing incorrect inputs to the fitting MLP.

https://github.com/deepmodeling/deepmd-kit/blob/9ce8d3e135e7f9aa590a0c9d090a9cf301a49838/deepmd/pt_expt/train/training.py#L72-L86

Suggested fix: either (a) expand `_TASK_SPECIFIC_BUFFER_NAMES` to include `fparam_avg`, `fparam_inv_std`, `aparam_avg`, `aparam_inv_std` and have `_get_task_buffers` enumerate them, or (b) auto-detect per-task buffers by diffing buffer identities across tasks in `_compile_model` rather than maintaining a hardcoded allow-list that will drift as descriptors and fittings evolve.

wanghan-iapcm (approved):

njzjz (approved):

PR #5459, opened 2026-05-26
test(universal): replace Cartesian product with curated matrices for descriptor test

iProzd:
## Summary by CodeRabbit

* **Tests**
  * Centralized descriptor parameter combinations into shared, reusable constants used across atomic/PT and model test suites to reduce repetition.
  * Added curated energy-model-specific parameter variants and a small helper to create fixed-argument variants for integration tests.
  * Removed an outdated descriptor helper and updated parameterized tests to consume the new centralized parameter collections.

[Review Change Stack](https://app.coderabbit.ai/change-stack/deepmodeling/deepmd-kit/pull/5459?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/source/tests/universal/dpmodel/descriptor/test_descriptor.py
+++ b/source/tests/universal/dpmodel/descriptor/test_descriptor.py
@@ -573,19 +573,6 @@ DescriptorParamDPA3List = parameterize_func(
 )
 
 
-def DescriptorParamDPA3DefaultChgSpin(ntypes, rcut, rcut_smth, sel, type_map, **kwargs):
-    return DescriptorParamDPA3(
-        ntypes,
-        rcut,
-        rcut_smth,
-        sel,
-        type_map,
-        **kwargs,
-        add_chg_spin_ebd=True,
-        default_chg_spin=[5.0, 1.0],
-    )
-
-
 # to get name for the default function
 DescriptorParamDPA3 = DescriptorParamDPA3List[0]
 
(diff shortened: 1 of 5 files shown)


njzjz (approved):

source/tests/universal/dpmodel/descriptor/test_descriptor.py
```diff
@@ -570,23 +573,257 @@ def DescriptorParamDPA3(
...
+
+DescriptorParamDPA1EnergyModelList = (
+    # Baseline coverage.
+    DescriptorParamDPA1,
+    # Single-option descriptor toggles.
+    _descriptor_param_variant(
+        DescriptorParamDPA1, "DescriptorParamDPA1_strip", tebd_input_mode="strip"
+    ),
+    _descriptor_param_variant(
+        DescriptorParamDPA1, "DescriptorParamDPA1_no_attention", attn_layer=0
```
wanghan-iapcm: **Coverage gap:** `_no_attention` duplicates the rebound baseline. At line 322 above, `DescriptorParamDPA1` is rebound to `DescriptorParamDPA1List[0]`, which has `attn_layer=0` (the first value in the OrderedDict tuple `(0, 2)`). This variant then explicitly sets `attn_layer=0` again, and no other variant in `DescriptorParamDPA1EnergyModelList` overrides it.

Net effect: after this PR, the DPA1 attention code path (`attn_layer=2`, `NeighborGatedAttention`) is no longer exercised at the model / atomic-model integration level — only at the descriptor level. Descriptor-internal regressions are still caught by the preserved Cartesian `DescriptorParamDPA1List`, but model-level glue bugs that surface only with attention enabled would slip through.

Non-blocking; trivial fix: change this variant to `attn_layer=2` (and rename to `_with_attention`) so the curated list covers both branches at the model level.

iProzd: fixed in https://github.com/deepmodeling/deepmd-kit/pull/5459/commits/7d80d36bebd146c538ad576ea00e2e681664217b

PR #5483, opened 2026-06-01
fix: dpa4 compile aliasing collision

anyangml:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved edge-list construction and tracing stability to avoid out-of-range/index issues during model tracing.

* **Performance**
  * Reused compiled callables across compatible multitask branches via shared caching for faster multi-task compilation and reduced duplication.

* **Changes**
  * Model export/tracing now consistently includes the charge_spin input slot to match runtime expectations.

* **Tests**
  * Updated multitask compile-cache test to expect callable reuse across compatible branches.

### DIFF
--- a/deepmd/pt/model/model/sezm_model.py
+++ b/deepmd/pt/model/model/sezm_model.py
@@ -374,6 +374,9 @@ import torch
 from einops import (
     rearrange,
 )
+from packaging.version import (
+    Version,
+)
 from torch.fx.experimental.proxy_tensor import (
     make_fx,
 )
@@ -2179,8 +2182,9 @@ class SeZMModel(DPModelCommon, SeZMModel_):
         mapping: torch.Tensor | None = None,
         fparam: torch.Tensor | None = None,
         aparam: torch.Tensor | None = None,
-        do_atomic_virial: bool = False,
         charge_spin: torch.Tensor | None = None,
+        *,
+        do_atomic_virial: bool = False,
     ) -> torch.nn.Module:
         """Trace ``forward_common_lower`` into an exportable FX ``GraphModule``.
 
--- a/deepmd/pt/model/model/sezm_model.py
+++ b/deepmd/pt/model/model/sezm_model.py
@@ -449,7 +449,7 @@ _dynamo_cfg.optimize_ddp = False
 # Maps (structure_key..., training, do_atomic_virial, has_coord_corr) to the
 # compiled callable.  Tasks whose descriptor AND fitting-net first child have
 # the same Python-object identity (after share_params) reuse a single compiled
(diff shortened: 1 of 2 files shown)


iProzd (approved):

deepmd/pt/model/model/sezm_model.py
```diff
@@ -1778,7 +2160,8 @@ def compile_dens(self) -> None:
                     "epilogue_fusion": False,
                     "triton.cudagraphs": False,
                     "shape_padding": True,
-                    "max_fusion_size": 64,
+                    "max_fusion_size": 8,
```
OutisLi: Should not modify the DeNS path

deepmd/pt/model/model/sezm_model.py
```diff
@@ -1960,10 +2353,9 @@ def build_edge_list_from_nlist(
         Build a compact edge list from DeePMD padded neighbor list.
 
         Edge vectors are computed via ``index_select`` on ``extended_coord``
-        so they remain differentiable w.r.t. the input coordinates.  Two
-        masked dummy edges are always appended to avoid data-dependent empty-edge
-        branches that ``make_fx`` cannot trace and singular edge-axis guards
-        in Inductor's batched matmul lowering.
+        so they remain differentiable w.r.t. the input coordinates.  One
```
OutisLi: Why change this to one dummy edge?

OutisLi (changes requested):

deepmd/pt/model/model/sezm_model.py
anyangml: I haven’t tested DeNS myself, so I’m not sure whether the fusion size causes any issues. I can revert it for now, but if the efficiency impact is negligible, I’d lean toward keeping it at 8 since that seems like the safer choice.

deepmd/pt/model/model/sezm_model.py
anyangml: Is this for single atom case?  If that's a real problem, we should probably restore this.

wanghan-iapcm (approved):

deepmd/pt/model/model/sezm_model.py
OutisLi: DeNS is just normal compile, so there should be bug-free, and the DeNS itself is experimental, I believe keep it as it is is better

deepmd/pt/model/model/sezm_model.py
OutisLi: > Is this for single atom case? If that's a real problem, we should probably restore this.

I ran into a bug before so I change this to 2, it solved the bug, but I forget what the bug is exactly

OutisLi (approved):

PR #5486, opened 2026-06-02
feat(pt/pt_expt): O(N) vesin neighbor list for Python/ASE inference

wanghan-iapcm:
## Motivation

The Python / ASE inference path builds the neighbor list with the native all-pairs **O(N²)** algorithm (`build_neighbor_list` materializes a dense `[nf, nloc, nall, 3]` displacement tensor). For large systems this dominates end-to-end throughput — exactly the front-end bottleneck described in the DPA4 manuscript §2.4 (the ASE throughput curves bend down at large N because of native O(N²) neighbor lists).

This PR adds an opt-in, O(N) neighbor-list backend based on the [`vesin`](https://github.com/Luthaf/vesin) cell list for the Python inference interfaces (`DeepPot`, the ASE `DP` calculator), for both the `pt` (`.pth`/`.pt`) and `pt_expt` (`.pte`/`.pt2`) backends. Model results are unchanged: the neighbor *search* is non-differentiable, so its result is fed to the model's own `forward_common_lower` + `communicate_extended_output`, leaving the autograd graph and outputs intact.

## API

New `nlist_backend` argument on `DeepPot` / `DeepEval` / the ASE `DP` calculator:

- `"auto"` (default): use the O(N) `vesin` cell list when applicable (`vesin.torch` installed, non-spin, non-hessian model) and silently fall back to the native builder otherwise. Keeps the fast path by default; never breaks spin/hessian.
- `"vesin"`: strict — raise `ValueError` if `vesin.torch` is missing, the model is spin/hessian, or (pt_expt) an explicit ASE `neighbor_list` is also given.
- `"native"`: force the native all-pairs builder.

## Design

- **torch-only, device-resident.** The builder uses `vesin.torch`, which runs the cell list on the device of the input tensors (CPU **or** CUDA). On GPU the whole extended-system build stays on the GPU (no host round-trip). Lives in `deepmd/pt_expt/utils/nlist.py`; both backends import it. `dpmodel` stays vesin-free (no torch code in the array-API backend).
- The builder converts `vesin`'s `(i, j, S)` edge list into deepmd's extended format (extended coords/types, `mapping`, candidate `nlist`) and lets `forward_common_lower.format_nlist` re-format (distance-sort, truncate to `sel`, type-split) — matching `forward_common` exactly.
- Works with `auto_batch_size`.
- `pyproject`: depends on `vesin[torch]` so the device-capable bindings ship by default.

## Verification

native vs vesin agree to floating-point round-off (energy bit-identical; forces/virials ~1e-16, the only difference being ghost-atom enumeration order):

| path | result |
|---|---|
| se_e2_a / se_atten energy, PBC & non-PBC, atomic, multiframe, auto_batch_size | E `0.0`, F/V `~1e-16` |
| message-passing DPA2/DPA3 (the manuscript's class) | E `0.0`, F/V `~5e-16` (spot-checked) |
| non-energy (dipole) | `0.0` (spot-checked) |
| fparam/aparam, charge_spin | pass-through, identical to native |

## Tests

- `source/tests/pt_expt/utils/test_nlist.py` — vesin.torch builder vs native (PBC/non-PBC/distinguish), input-device placement.
- `source/tests/pt/model/test_deeppot.py`, `source/tests/pt_expt/infer/test_deep_eval.py`, `test_deep_eval_spin.py`, `source/tests/pt/test_calculator.py` — every `nlist_backend` dispatch branch (invalid → ValueError, native, `auto` fallback and strict-`vesin` raise for vesin-missing / spin / hessian / ASE-conflict), native-vs-vesin consistency (PBC/non-PBC/atomic/multiframe/auto_batch_size), real-spin gate-off end-to-end, and the ASE calculator threading `nlist_backend` for both backends.

## Known limitations

- **No real-CUDA test**: the device-placement test runs on CPU in CI (no GPU runner). The CUDA execution path of `vesin.torch` is implemented but exercised only via the CPU branch here.
- **`vesin[torch]` is a compiled torch-ABI extension**: on a platform without a matching wheel, *install* could fail (runtime is guarded by `is_vesin_torch_available`, but pyproject is not). There is no numpy-vesin fallback by design — `auto` falls back to native, explicit `vesin` raises.
- **Not addressed** (separate follow-ups): the in-graph O(N²) builder used during training / `torch.export`; `extend_coord_with_ghosts` over-generating ~27×N ghosts; the (still dead) ASE `neighbor_list` arg on the `pt` backend; a committed throughput benchmark.
- `charge_spin` and `.pt2`-runner+vesin are covered by equivalence/`.pte` rather than dedicated committed tests; message-passing and dipole through vesin are spot-checked, not committed.

## Summary by CodeRabbit

* **New Features**
  * Added `nlist_backend` parameter to calculator and inference backends for selecting neighbor-list construction algorithms.
  * Supports "auto" (intelligent backend selection), "native", and "vesin" modes with validation and fallback handling.

* **Chores**
  * Added vesin[torch] as a runtime dependency.


wanghan-iapcm:
I will make deepmd-kit directly consume vesin style nlist.

PR #5494, opened 2026-06-04
feat(pt): add atomic charge population fitting and model

chrisahart:
We implement the DeepPopulation model, which can be used to fit the atomic charge population.

See the [preprint on arXiv](https://arxiv.org/abs/2606.01763) for details and an example of how the DeepPopulation model was used to study small polaron transport with DeepPolaron.

## The Fitting Network

The JSON of `population` type should be provided like

```json
"fitting_net": {
    "type": "population"
}
```

- `type` specifies which type of fitting net should be used. It should be `population`.
- The rest of the arguments have the same meaning as they do in `ener` mode.

## Loss

The loss function for DeepPopulation has the form:

$$
\begin{equation}
\begin{aligned}
\mathcal{L} =
& \lambda_{N_{\alpha}} \ell\!\left( N^{ML}_{\alpha},\, N^{DFT}_{\alpha} \right)  +
\\ &  \lambda_{N_{\beta}} \ell\!\left( N^{ML}_{\beta},\, N^{DFT}_{\beta} \right) +
\\ & \lambda_S \ell\!\left( S^{ML},\, S^{DFT} \right) +
\\ & \frac{\lambda_s}{N} \sum_{i=1}^{N} \ell\!\left( s_{i}^{ML},\, s_{i}^{DFT} \right) +
\\ & \frac{\lambda_\sigma}{N} \sum_{i=1}^{N} \ell\!\left( \sigma_{i}^{ML},\, \sigma_{i}^{DFT} \right)
\end{aligned}
\end{equation}
$$

This includes contributions from:

- The total number of $\alpha$ electrons $N_{\alpha}$,
- The total number of $\beta$ electrons $N_{\beta}$,
- The total spin $S$,
- The atomic spin moments $s_{i}$,
- The atomic populations $\sigma_{i}$.

The loss section should be provided like

```json
"loss": {
    "type": "population",
    "start_pref_pop_alpha_total": 1,
    "limit_pref_pop_alpha_total": 1,
    "start_pref_pop_beta_total": 1,
    "limit_pref_pop_beta_total": 1,
    "start_pref_spin_total": 1,
    "limit_pref_spin_total": 1,
    "start_pref_spin": 1,
    "limit_pref_spin": 1,
    "start_pref_pop": 1,
    "limit_pref_pop": 1,
    "loss_func": "mae",
    "metric": ["mae"]
}
```

## Training Data Preparation

The atomic charge population file should be named `atomic_population.npy`, with shape `[num_frames, num_atoms, 2]`, where `[num_frames, num_atoms, 0]` represents the alpha spin channel and `[num_frames, num_atoms, 1]` represents the beta spin channel.

## Summary by CodeRabbit

* **New Features**
  * End-to-end support for atomic population modeling: new model type, atomic-model variant, fitting network, inference wrapper, and runtime selection for population outputs.

* **Training & Loss**
  * New population training loss with multi-term prefactor scheduling, per-atom and total-population components, metrics, and training integration.

* **Documentation**
  * Added training guide describing data format, loss terms, example configuration, and workflow.

* **Examples**
  * Added example dataset and training configuration for population training.

* **Tests**
  * Added unit tests covering population loss behaviors and scaling.


chrisahart:
@coderabbitai review

chrisahart:
@coderabbitai resume

chrisahart:
@njzjz Can you please review this PR when you have a chance? This is my first time contributing to DeePMD-kit. Thank you!

chrisahart:
I believe all relevant requested changes have now been performed.

chrisahart:
Thank you @njzjz-bot. These should now be fixed.

chrisahart:
Thank you @njzjz-bot. I have made the requested changes.

chrisahart:
@njzjz Can you please review this PR when you have a chance? All tests are now successful, except for one which is unrelated to this pull request. Thank you!

wanghan-iapcm (changes requested):
**1. Training crashes on the first step — label key mismatch (must fix).**

`PopulationLoss.forward()` reads the label under the wrong key, so real training crashes on the first step. The read uses `label["atom_population"]`, but the loss's own `label_requirement` declares the field as `atomic_population` (and the example
  data file is `atomic_population.npy`, as documented). 

https://github.com/deepmodeling/deepmd-kit/blob/4164f947d5927fd99a3929649f021f5144b83c74/deepmd/pt/loss/population.py#L174-L176

https://github.com/deepmodeling/deepmd-kit/blob/4164f947d5927fd99a3929649f021f5144b83c74/deepmd/pt/loss/population.py#L282-L285

Since the data loader keys the label dict by the `DataRequirementItem` name, the label is stored under `atomic_population` and `forward()` hits `KeyError: 'atom_population'`. 

The unit tests don't   catch this because they bypass `label_requirement` and hand-build the label dict with the same typo'd key. 

Proposed fix: read `label["atomic_population"]`.

wanghan-iapcm:
Thank you very much for this nice contribution!

iProzd (changes requested):
Thanks for the contribution! Agree with @wanghan-iapcm . The `label["atomic_population"]` fix is correct — please also fix `test_loss_population.py` to use `atomic_population` so the test actually covers the real label path instead of masking the bug.
Others LGTM.

chrisahart:
@wanghan-iapcm @iProzd Thank you for your comments. However, can you please confirm whether you have observed the code crash? The confusion seems to be that deepmd-kit is doing some internal renaming from `atomic_` to `atom_`. Previously I had coded as you suggested using `atomic_population`, however this crashed and I realised that this is part of deepmd-kit itself such that the label is stored under `atom_population` at runtime. The test included does not mask the bug, rather it replicates the deepmd-kit internal renaming code.

https://github.com/deepmodeling/deepmd-kit/blob/b52c359ec6b6f6f34998e00f774c35d253cf09e5/deepmd/utils/data.py#L191

wanghan-iapcm:
> @wanghan-iapcm @iProzd Thank you for your comments. However, can you please confirm whether you have observed the code crash? The confusion seems to be that deepmd-kit is doing some internal renaming from `atomic_` to `atom_`. Previously I had coded as you suggested using `atomic_population`, however this crashed and I realised that this is part of deepmd-kit itself such that the label is stored under `atom_population` at runtime. The test included does not mask the bug, rather it replicates the deepmd-kit internal renaming code.
> 
> https://github.com/deepmodeling/deepmd-kit/blob/b52c359ec6b6f6f34998e00f774c35d253cf09e5/deepmd/utils/data.py#L191

@chrisahart You're right. thanks for the clear explanation, and apologies for the noise. 
The pr LGTM.

wanghan-iapcm (approved):

iProzd (approved):

njzjz (approved):

PR #5504, opened 2026-06-08
Chore: refactor pt-expt compile

anyangml:
## Summary by CodeRabbit

* **Refactor**
  * Moved compile/tracing helpers into a shared utilities module to reduce duplication.
  * Training trace path now uses shared helpers and selects a “safe prime” frame size, padding only necessary dimensions.

* **Bug Fixes**
  * Improved FX detach cleanup and graph rebuilding for more stable tracing and compilation (fewer symbolic-shape and re-trace failures).


iProzd (approved):

deepmd/pt/utils/compile_utils.py
wanghan-iapcm, line 1: mv this file to pt-expt backend? as pt will be deprecated in the future.

OutisLi (approved):

wanghan-iapcm (approved):

PR #5509, opened 2026-06-09
feat: add charge spin to cpp runtime

anyangml:
## Summary by CodeRabbit

* **New Features**
  * Compute APIs now accept an optional per-call charge_spin vector and expose dim_chg_spin; model defaults are used when omitted. Support is threaded end-to-end (PT/pt2, C API v27, mixed-type, per-frame, neighbor-list) and LAMMPS pair-style parsing accepts explicit charge_spin.

* **Tests**
  * Added generator and unit/integration tests verifying dim_chg_spin, default vs explicit charge_spin behavior, and inference across backends.


source/lmp/pair_deepmd.cpp
```diff
@@ -590,6 +598,7 @@ void PairDeepMD::settings(int narg, char** arg) {
     numb_types_spin = deep_pot_model_devi.numb_types_spin();
     dim_fparam = deep_pot_model_devi.dim_fparam();
     dim_aparam = deep_pot_model_devi.dim_aparam();
+    dim_chg_spin = deep_pot.dim_chg_spin();
```
wanghan-iapcm: ### Code review

The model-deviation (`numb_models > 1`) path silently drops `charge_spin`.

Here `dim_chg_spin` is set for the model-devi branch, and `settings()` accepts the `charge_spin` / `charge_spin_from_compute` keywords in this mode too. But the two `deep_pot_model_devi.compute(...)` calls in `PairDeepMD::compute()` (L318, L326) still pass only `(fparam, daparam)` — no `charge_spin` — and `DeepPotModelDevi` was not given `charge_spin`-aware `compute()` overloads.

Net effect: running model deviation against a `dim_chg_spin > 0` model parses the user's per-step `charge_spin` without warning, then drops it at compute time, so inference falls back to the model's stored `default_chg_spin`. Either thread `charge_spin` into the `deep_pot_model_devi.compute()` calls (and add the overloads), or reject the keyword in model-devi mode.

source/lmp/pair_deepmd.cpp
```diff
@@ -686,6 +696,28 @@ void PairDeepMD::settings(int narg, char** arg) {
...
+        if (iarg + 1 + ii >= narg || is_key(arg[iarg + 1 + ii])) {
+          char tmp[1024];
+          sprintf(tmp, "Illegal charge_spin, the dimension should be %d",
+                  dim_chg_spin);
+          error->all(FLERR, tmp);
+        }
+        charge_spin.push_back(atof(arg[iarg + 1 + ii]));
+      }
+      iarg += 1 + dim_chg_spin;
+    } else if (string(arg[iarg]) == string("charge_spin_from_compute")) {
```
iProzd: `charge_spin_from_compute` is unnecessary?

iProzd (changes requested):
The following should be completed for the main runtime paths before merge:

1. Please wire `charge_spin` through the C API / `deepmd.hpp` path.
   In common LAMMPS builds with `DP_USING_C_API`, `PairDeepMD` uses `deepmd::hpp::DeepPot` from `source/api_c/include/deepmd.hpp`. The new `charge_spin` argument is currently accepted there but discarded via `(void)charge_spin`, so `charge_spin` has no effect on this widely used path. Please add proper C API entry points in `source/api_c/include/c_api.h` / `source/api_c/src/c_api.cc` and have `deepmd.hpp` forward `charge_spin` through to the underlying `api_cc::DeepPot` implementation.

2. Please support the regular PyTorch `.pth` backend as well.
   `DeepPotPT` currently does not override the new `charge_spin`-aware `computew` overloads, so `.pth` models fall back to the default `DeepPotBackend` implementation and silently ignore `charge_spin`. If runtime `charge_spin` is a C++ runtime feature, it should also be implemented for the PT backend, not only `DeepPotPTExpt` / `.pt2`.

3. Please also thread `charge_spin` through model-deviation.
   LAMMPS model-deviation currently calls `deep_pot_model_devi.compute(..., fparam, aparam)` without `charge_spin`, so the main model path and deviation path can use different effective conditioning.

Without these, `charge_spin` support is incomplete and can silently no-op in common usage.

source/api_c/include/deepmd.hpp
```diff
@@ -57,7 +57,8 @@ inline void _DP_DeepPotCompute(DP_DeepPot* dp,
                                FPTYPE* force,
                                FPTYPE* virial,
                                FPTYPE* atomic_energy,
-                               FPTYPE* atomic_virial);
+                               FPTYPE* atomic_virial,
+                               const double* charge_spin);
```
iProzd: Should `charge_spin` be `const FPTYPE*` just after `aparam`? Since they should be the same type and also the input rather than output.

source/api_c/src/c_api.cc
```diff
@@ -251,7 +251,8 @@ inline void DP_DeepPotCompute_variant(DP_DeepPot* dp,
                                       VALUETYPE* force,
                                       VALUETYPE* virial,
                                       VALUETYPE* atomic_energy,
-                                      VALUETYPE* atomic_virial) {
+                                      VALUETYPE* atomic_virial,
+                                      const double* charge_spin = nullptr) {
```
iProzd: same as above

iProzd (changes requested):

iProzd (approved):

iProzd:
LGTM, @njzjz plz also review this.

source/api_c/include/c_api.h
```diff
@@ -675,6 +675,123 @@ extern void DP_DeepPotComputeNListf2(DP_DeepPot* dp,
...
+                                float* atomic_virial);
+
+/**
+ * @brief Evaluate the energy, force and virial by using a DP with the neighbor
+ *list. (double version, with charge_spin)
+ * @version 3
+ * @param[in] charge_spin The per-frame charge/spin input. The array can be of
+ *size nframes x dim_chg_spin. Pass NULL to use the model's stored
+ *default_chg_spin.
+ * @since API version 27
```
njzjz, line 799: Please give the doc for all arguments, not only the argument you added.

source/api_c/include/deepmd.hpp
```diff
@@ -1103,7 +1204,8 @@ class DeepPot : public DeepBaseModel {
       const std::vector<int>& atype,
       const std::vector<VALUETYPE>& box,
       const std::vector<VALUETYPE>& fparam = std::vector<VALUETYPE>(),
-      const std::vector<VALUETYPE>& aparam = std::vector<VALUETYPE>()) {
+      const std::vector<VALUETYPE>& aparam = std::vector<VALUETYPE>(),
+      const std::vector<VALUETYPE>& charge_spin = std::vector<VALUETYPE>()) {
```
njzjz, line 1212: Please add the doc for your newly added argument.

source/api_cc/include/DeepPot.h
```diff
@@ -259,7 +368,8 @@ class DeepPot : public DeepBaseModel {
                const std::vector<int>& atype,
                const std::vector<VALUETYPE>& box,
                const std::vector<VALUETYPE>& fparam = std::vector<VALUETYPE>(),
-               const std::vector<VALUETYPE>& aparam = std::vector<VALUETYPE>());
+               const std::vector<VALUETYPE>& aparam = std::vector<VALUETYPE>(),
+               const std::vector<double>& charge_spin = std::vector<double>());
```
njzjz, line 376: Please add the doc for your newly added argument.

source/lmp/pair_deepmd.cpp
```diff
@@ -686,6 +692,17 @@ void PairDeepMD::settings(int narg, char** arg) {
       do_compute_aparam = true;
       compute_aparam_id = arg[iarg + 1];
       iarg += 1 + 1;
+    } else if (string(arg[iarg]) == string("charge_spin")) {
```
njzjz, line 695: Please add the new argument to the user documentation.

njzjz (changes requested):

njzjz (approved):

wanghan-iapcm (approved):

PR #5519, opened 2026-06-12
feat(pt): avoid grad clip sync

OutisLi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection and handling of non-finite gradients (NaN/Inf) during training; checkpoint saves are now aborted if invalid gradients were observed since the last checkpoint.
  * Better diagnostics reporting which parameters or reduction steps caused non-finite gradient issues.

* **Chores**
  * Reworked gradient-clipping to use a consistent in-place L2 norm clipping and on-device norm reduction for improved numerical stability.

* **Tests**
  * Updated tests covering the new clipping behavior, empty/no-grad cases, stable-vs-unstable reductions, and the deferred non-finite detection guard.


deepmd/pt/train/training.py
```diff
@@ -1758,6 +1760,11 @@ def log_loss_valid(_task_key: str = "Default") -> dict:
                 )
                 or (display_step_id) == self.num_steps
             ) and (self.zero_stage > 0 or self.rank == 0 or dist.get_rank() == 0):
+                # Abort before writing if any gradient norm since the previous
+                # checkpoint was non-finite.
+                self.nonfinite_grad_guard.raise_if_nonfinite(
```
wanghan-iapcm, line 1762: DDP deadlock: under plain DDP (`zero_stage == 0`) this checkpoint block is gated by `self.zero_stage > 0 or self.rank == 0 or dist.get_rank() == 0` (line 1762), so only rank 0 enters and calls `raise_if_nonfinite`. If rank 0 raises on a non-finite gradient, the other ranks proceed to the next step and block on the next NCCL collective (the backward all-reduce), causing a hang / NCCL timeout. The previous per-step raise executed on all ranks, so divergence failed cleanly and collectively. Consider running the non-finite check/raise on all ranks, outside the rank-0 checkpoint gate.

source/tests/pt/test_train_utils.py
```diff
@@ -1,116 +1,100 @@
...
+        p = torch.nn.Parameter(torch.zeros(128, dtype=torch.float32, device="cpu"))
+        p.grad = torch.full((128,), 1.0e19, dtype=torch.float32, device="cpu")
+
+        total_norm = clip_grad_norm_([p], max_norm=2.0)  # stable=True (default)
+        clipped_norm = torch.linalg.vector_norm(p.grad.double())
+
         self.assertEqual(total_norm.dtype, torch.float64)
-        self.assertAlmostEqual(clipped_norm.item(), 3.0, places=5)
+        self.assertTrue(torch.isfinite(total_norm))
+        self.assertAlmostEqual(clipped_norm.item(), 2.0, places=3)
```
wanghan-iapcm: Tolerance loosened to `places=3` without justification. The stable path reduces in float64 (accurate to ~1e-7 here), and every other assertion in this file uses `places=5`; a 100x looser tolerance can mask a regression in the clipping math. Suggest `places=5`.

deepmd/pt/train/utils.py
```diff
@@ -24,137 +23,156 @@
...
+        ------
+        RuntimeError
+            If a non-finite gradient norm was recorded.
+        """
+        if self._nonfinite is None:
+            return
+        diverged = bool(self._nonfinite)
+        self._nonfinite = None
+        if diverged:
+            raise_nonfinite_gradient_norm(named_parameters())
```
wanghan-iapcm, line 133: Stale diagnostic: `named_parameters()` is consulted lazily here, at the moment the guard raises (checkpoint time), not at the step that diverged. By then the intervening backward passes have overwritten `p.grad`, so `raise_nonfinite_gradient_norm` scans the *current* gradients rather than the diverging ones. The reported list (and the docstring claim at line 115, "the offending parameters are reported") can therefore name a different set of parameters, or find them all finite and emit the misleading "overflow in the norm reduction" fallback. Consider capturing the offending parameter names on-device at the diverging step (e.g. record the per-parameter non-finite mask in `update`) instead of re-scanning at raise time, and softening the docstring to "current gradient state is reported."

source/tests/pt/test_train_utils.py
```diff
@@ -1,116 +1,100 @@
...
-            torch.stack([p0.grad.double().norm(), p1.grad.double().norm()])
-        )
-        self.assertTrue(torch.isfinite(total_norm))
+    def test_stable_norm_survives_float32_overflow(self) -> None:
+        # A single parameter whose gradient sum-of-squares overflows float32 is
+        # still clipped via the float64 reduction rather than aborting.
+        p = torch.nn.Parameter(torch.zeros(128, dtype=torch.float32, device="cpu"))
+        p.grad = torch.full((128,), 1.0e19, dtype=torch.float32, device="cpu")
+
+        total_norm = clip_grad_norm_([p], max_norm=2.0)  # stable=True (default)
```
wanghan-iapcm, line 40: Missing end-to-end non-finite coverage. This PR removed `test_stable_fallback_nan_individual_grad_raises` and `test_fsdp_path_nonfinite_raises`, and the new `TestNonFiniteGradGuard` tests inject pre-built non-finite tensors directly into the guard, bypassing `clip_grad_norm_`. No test feeds an actual NaN/Inf gradient through `clip_grad_norm_` and asserts that it returns a non-finite `total_norm` (which the guard then accumulates) — for either `stable=True` or `stable=False`. This is the new deferred-detection contract and it is exactly the untested path. Suggest adding something like:

```python
def test_clip_grad_norm_returns_nonfinite_for_nan_grad(self) -> None:
    for stable in (True, False):  # both branches of the stable flag
        p = torch.nn.Parameter(torch.zeros(4))
        p.grad = torch.full((4,), float("nan"))
        total_norm = clip_grad_norm_([p], max_norm=1.0, stable=stable)
        self.assertFalse(torch.isfinite(total_norm))
        guard = NonFiniteGradGuard()
        guard.update(total_norm)
        with self.assertRaises(RuntimeError):
            guard.raise_if_nonfinite(lambda: [("p", p)])
```

wanghan-iapcm (approved):

deepmd/pt/train/training.py
```diff
@@ -1412,12 +1414,12 @@ def step(_step_id: int, task_key: str = "Default") -> None:
...
                         ]
-                    total_norm = clip_grad_norm_with_stable_fallback(
+                    total_norm = clip_grad_norm_(
                         self.wrapper.parameters(),
                         self.gradient_max_norm,
-                        use_stable_fallback=self.zero_stage < 2,
-                        named_parameters=self.wrapper.named_parameters,
+                        stable=self.zero_stage < 2,
                     )
+                    self.nonfinite_grad_guard.update(total_norm)
```
njzjz, line 1422: Recording the non-finite norm here does not prevent the next `optimizer.step()` from running. When `clip_grad_norm_` sees `nan`/`inf`, it still calls `clip_grads_with_norm_`; with a non-finite total norm that can turn gradients into non-finite values and then update parameters/optimizer state before `raise_if_nonfinite()` runs at checkpoint time. Please either raise/skip the step immediately when the norm is non-finite, or make the clipping path leave gradients and optimizer state untouched until the guard aborts.

- codex

njzjz (changes requested):
Requesting changes for the non-finite gradient handling issue below.

Reviewed by codex.

deepmd/pt/train/training.py
OutisLi: That's the intention of this PR, which is to delay the raise to minimize the H2D sync

njzjz (approved):

PR #5524, opened 2026-06-13
fix(pt): sort nlist for compressed se_e2_a in forward_lower

### LINKED ISSUE #5092: the MD running is only working when using CPU
### Bug summary

Dear developers,

I found that my lammps running based on trained dp models will only work if I use CPU, while when I use GPU, the NVT temperature control crashes, and I'm quite confused why this happens, could you please give me some suggestions? I have attached all the related documents and hopefully they can be useful. Many thanks.

Best wishes,
Lingjun

### DeePMD-kit Version

3.1.1

### Backend and its version

PyTorch v2.8.0

Offline packages

### Steps to Reproduce

All input files, including the data file, the pth file and the in file were included as above in the zip files.

_No response_

wanghan-iapcm:
## Summary

Fixes wrong energy/forces (and unstable LAMMPS MD) for **compressed `se_e2_a` models evaluated through `forward_lower`** — the C++/LAMMPS inference path. Reported in discussion #5438.

## Root cause

The compressed `tabulate_fusion_se_a` op (`source/lib/src/tabulate.cc`, forward and grad kernels) has an `is_sorted`-gated early-termination:

```cpp
ago = em_x[ii * nnei + nnei - 1];           // last neighbor's em_x
if (ago == xx && ll[1]==0 && ll[2]==0 && ll[3]==0 && is_sorted) break;
```

It stops accumulating at the first neighbor whose env-mat **direction is zero**. Both `-1` padding *and* out-of-`rcut` neighbors (`sw==0`) have zero direction and the same `em_x == -davg/dstd` (`== ago`), so the op assumes all such neighbors are **trailing**. `is_sorted` defaults to `true` and the PT op never overrides it.

The C++/LAMMPS `forward_lower` neighbor list uses `rcut + skin` and is **not distance-sorted**. `_format_nlist` only filters out-of-`rcut` neighbors in its *sort* branch (`n_nnei > nnei`), which is skipped when the LAMMPS list is narrower than `sum(sel)` (pad-only branch). The zero-direction neighbors then land **before** real ones, so the op breaks early and silently drops real neighbors → wrong descriptor → wrong energy/forces → unstable MD.

Only the **compressed** path is affected:
- The uncompressed embedding-net path sums over neighbors and treats zero-direction fillers identically regardless of position, so it is order-invariant.
- Only `tabulate_fusion_se_a` (forward + grad) has this early-termination; `se_t`/`se_r` forward kernels do not.

It is **device-independent** (reproduces identically on CPU and GPU).

## Fix

The wiring already exists — the model calls `format_nlist(..., extra_nlist_sort=self.need_sorted_nlist_for_lower())` — but `DescrptBlockSeA.need_sorted_nlist_for_lower()` always returned `False`. Make it return `self.compress`:

```python
def need_sorted_nlist_for_lower(self) -> bool:
    return self.compress
```

When compression is enabled this forces the sort + `rcut`-filter branch (in-`rcut` neighbors first, all padding last), restoring the op's invariant. The standard (uncompressed) route is unchanged, so there is no added cost on the common path.

## Verification

- In LAMMPS (CPU and GPU) the compressed model now matches the uncompressed model to ~2.5e-14 (was ~0.5 eV off with scrambled forces).
- New regression test `source/tests/pt/model/test_compressed_se_a_forward_lower.py` runs compressed `forward_lower` with an unsorted, over-`rcut` neighbor list and compares energy + force to the uncompressed reference, parameterized over `type_one_side ∈ {True, False}`. It **fails without this fix and passes with it**; existing `test_compressed_descriptor_se_a.py` and `test_forward_lower.py` still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed compressed descriptor behavior so enabling compression preserves neighbor ordering invariants and no longer causes valid neighbors to be dropped; energies and forces remain correct with unsorted/padded neighbor lists.

* **Tests**
  * Added regression tests for multiple descriptor variants that validate compressed mode against uncompressed baselines using unsorted/over-cut neighbor lists, asserting energy and force fidelity.

### DIFF
--- a/deepmd/pt/model/descriptor/se_a.py
+++ b/deepmd/pt/model/descriptor/se_a.py
@@ -871,4 +871,13 @@ class DescrptBlockSeA(DescriptorBlock):
 
     def need_sorted_nlist_for_lower(self) -> bool:
         """Returns whether the descriptor block needs sorted nlist when using `forward_lower`."""
-        return False
+        # The compressed tabulate op (`tabulate_fusion_se_a`) uses an
+        # `is_sorted` early-termination that stops accumulating as soon as it
+        # meets a neighbor whose env-mat direction is zero (padding, or an
+        # out-of-rcut neighbor with sw==0). It therefore assumes such neighbors
+        # are trailing. The `forward_lower` neighbor list coming from C++/LAMMPS
+        # (rcut+skin, not pre-sorted) can interleave zero-direction neighbors
+        # before real ones, which would silently drop real neighbors. Forcing a
+        # sorted nlist (extra_nlist_sort) filters out-of-rcut neighbors and puts
+        # all padding last, restoring the op's invariant. See discussion #5438.
+        return self.compress
--- a/source/tests/pt/model/test_compressed_se_a_forward_lower.py
+++ b/source/tests/pt/model/test_compressed_se_a_forward_lower.py
@@ -1,0 +1,134 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+"""Regression test for discussion #5438.
+
+A compressed ``se_e2_a`` model produced wrong energy/forces when evaluated
+through ``forward_lower`` with a neighbor list that is *not* pre-sorted and
+contains out-of-``rcut`` (``sw == 0``) neighbors before the real ones -- exactly
+what the C++/LAMMPS inference path provides (its neighbor list uses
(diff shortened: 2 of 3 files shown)


njzjz (commented):
Does the same problem exist in `se_e2_r`?

njzjz (approved):

PR #5531, opened 2026-06-13
refactor(api_cc): load backends as plugins

njzjz:
## Summary
- split api_cc backend implementations into dynamically loaded backend plugins
- route DeepPot, DeepSpin, DeepTensor, and DipoleChargeModifier creation through a C ABI plugin layer
- keep libdeepmd_cc free of TensorFlow/PyTorch/Paddle runtime links; backend runtimes live in plugin libraries
- package backend plugin targets with C API packaging

## Validation
- ruff check .
- ruff format .
- git diff --check
- cmake -S source -B /tmp/deepmd-plugin-no-backend -DBUILD_CPP_IF=ON -DBUILD_PY_IF=ON -DBUILD_TESTING=OFF -DCMAKE_TESTING_ENABLED=OFF -DENABLE_TENSORFLOW=OFF -DENABLE_PYTORCH=OFF -DENABLE_JAX=OFF -DENABLE_PADDLE=OFF
- cmake --build /tmp/deepmd-plugin-no-backend --target deepmd_cc deepmd_c -j2
- cmake -S source -B /tmp/deepmd-plugin-tf -DBUILD_CPP_IF=ON -DBUILD_PY_IF=ON -DBUILD_TESTING=OFF -DCMAKE_TESTING_ENABLED=OFF -DENABLE_TENSORFLOW=ON -DENABLE_PYTORCH=OFF -DENABLE_PADDLE=OFF -DTENSORFLOW_ROOT=/home/jzzeng/miniconda3/lib/python3.13/site-packages/tensorflow
- cmake --build /tmp/deepmd-plugin-tf --target deepmd_backend_tf deepmd_backend_jax -j2
- ldd /tmp/deepmd-plugin-no-backend/api_cc/libdeepmd_cc.so
- nm -D /tmp/deepmd-plugin-tf/api_cc/libdeepmd_backend_tf.so

PyTorch plugin build was not validated locally because the installed torch package is CUDA-enabled and this environment does not provide CUDA toolkit files required by TorchConfig.cmake.

## Summary by CodeRabbit

* **New Features**
  * Added plugin-based runtime backend selection (TensorFlow, PyTorch, JAX, Paddle) with search paths configurable via `DP_BACKEND_PLUGIN_PATH`.
  * Backend op libraries are now loaded per selected backend at runtime.
  * Enabled TensorFlow pbtxt → pb conversion via the backend plugin mechanism.
  * Improved C API packaging/install to better include resolved runtime dependencies.
  * Updated C++ inference documentation and examples for plugin-based backend loading.
* **Breaking/Behavior Changes**
  * Backend op library loading now requires backend selection (backend-agnostic loading removed).
  * pbtxt → pb conversion no longer strictly depends on TensorFlow being built-in.


njzjz:
Addressed the review follow-ups in d6795db5b.

- Wrapped all backend plugin create calls with consistent exception-to-deepmd_exception handling, including plugin path and plugin error cleanup.
- Documented why the TensorFlow backend plugin compiles common.cc with BUILD_TENSORFLOW while libdeepmd_cc stays backend-neutral.
- Added C/C++ backend plugin documentation: plugin names, install locations, DP_BACKEND_PLUGIN_PATH search order, missing-plugin behavior, no-backend library usage, and the distinction from DP_PLUGIN_PATH. The C/C++ link examples no longer link TensorFlow directly.

Local validation: rebuilt deepmd_cc and deepmd_backend_tf, ran the C API DeepPot smoke test through the TensorFlow plugin, and ran ruff/pre-commit checks for the changed files.

doc/install/install-from-source.md
```diff
@@ -370,12 +370,29 @@ The installation requires CMake 3.25.2 or later for all platforms (CPU, CUDA, an
...
 ::::{tab-set}
 
+:::{tab-item} Backend-neutral C/C++ libraries
+
+To build only `libdeepmd_cc` and `libdeepmd_c` without backend plugins, leave all backend options disabled, or set them explicitly:
+
+```bash
+cmake -DBUILD_CPP_IF=ON -DBUILD_PY_IF=OFF \
+  -DENABLE_TENSORFLOW=OFF -DENABLE_PYTORCH=OFF \
+  -DENABLE_JAX=OFF -DENABLE_PADDLE=OFF \
```
wanghan-iapcm, line 387: This documented backend-neutral build command does not work as written. With all `ENABLE_*=OFF` and `BUILD_PY_IF=OFF` (and no `DEEPMD_C_ROOT`), the configure step aborts at `source/CMakeLists.txt:441` with `FATAL_ERROR "No backend is enabled."`. That guard is not relaxed in this PR, so the whole backend-neutral story documented here (and referenced in `doc/inference/cxx.md`) is unreachable. Either relax the guard (e.g. add an opt-in option, or drop `AND NOT BUILD_PY_IF`) in this same PR, or correct the documented command.

source/api_cc/src/BackendPlugin.cc
```diff
@@ -0,0 +1,466 @@
...
+            handle, candidate, deepmd::DEEPMD_DEEPPOT_PLUGIN_CREATE_SYMBOL));
+    plugin->delete_deeppot =
+        reinterpret_cast<deepmd::deepmd_delete_deeppot_backend_fn>(load_symbol(
+            handle, candidate, deepmd::DEEPMD_DEEPPOT_PLUGIN_DELETE_SYMBOL));
+    plugin->free_error = reinterpret_cast<deepmd::deepmd_free_backend_error_fn>(
+        load_symbol(handle, candidate,
+                    deepmd::DEEPMD_BACKEND_PLUGIN_FREE_ERROR_SYMBOL));
+    plugins[backend] = plugin;
+    return plugin;
+  }
```
wanghan-iapcm, line 285: The candidate loop only falls through to the next candidate when `open_library` returns null (L215-218). But if a candidate opens successfully yet is missing a required symbol, `load_symbol` (L223-231) throws, and the exception escapes the loop — so the remaining candidates (the dir next to `libdeepmd_cc`, then the bare-name loader search) are never tried. A stale or ABI-mismatched `libdeepmd_backend_*.so` sitting earlier in `DP_BACKEND_PLUGIN_PATH` will therefore hard-fail the load instead of falling through to a good plugin. The opened `handle` is also leaked on this path (`PluginHandle` has no `dlclose` in its destructor). Suggest wrapping the per-candidate symbol resolution in try/catch: record the error into `errors`, `dlclose` the handle, and `continue`.

source/api_cc/CMakeLists.txt
```diff
@@ -54,6 +54,92 @@ if(CMAKE_TESTING_ENABLED)
...
+  target_link_libraries(deepmd_backend_ptexpt PRIVATE "${TORCH_LIBRARIES}")
+  target_compile_definitions(deepmd_backend_ptexpt PRIVATE BUILD_PYTORCH)
+endif()
+
+if(ENABLE_JAX)
+  add_library(deepmd_backend_jax SHARED src/DeepPotJAX.cc
+                                        src/DeepPotJAXPlugin.cc)
+  deepmd_configure_backend_plugin(deepmd_backend_jax)
+  target_link_libraries(deepmd_backend_jax PRIVATE TensorFlow::tensorflow_c)
+  target_compile_definitions(deepmd_backend_jax PRIVATE BUILD_JAX)
```
wanghan-iapcm, line 129: Non-blocking: this gates the JAX plugin on `ENABLE_JAX` only. On master, `.savedmodel` loading was guarded `#if defined(BUILD_TENSORFLOW) || defined(BUILD_JAX)`, so an `ENABLE_TENSORFLOW=ON, ENABLE_JAX=OFF` build could load JAX2TF SavedModels (they run on the TF C API; `test_deeppot_jax.cc` exercises this). Since the JAX sources are absent from the `deepmd_backend_tf` target, a TF-only build now fails with "Unable to load jax backend plugin", and the `BUILD_TENSORFLOW ||` half of the guard kept in `DeepPotJAXPlugin.cc` is dead. Consider also building the JAX plugin under `ENABLE_TENSORFLOW`, or documenting the requirement change.

source/api_cc/src/BackendPlugin.cc
```diff
@@ -0,0 +1,466 @@
...
+
+  throw deepmd::deepmd_exception(
+      "Unable to load " + backend_name + " backend plugin (" + library_name +
+      "). Set DP_BACKEND_PLUGIN_PATH or install the plugin next to "
+      "libdeepmd_cc. Tried:" +
+      errors.str());
```
wanghan-iapcm, line 291: Non-blocking: this not-found path (and the env-var discovery / cross-ABI exception translation in this file and `BackendPluginFactory.h`) has no unit test — the existing C++ suite only exercises the happy path indirectly. A small install/runtime test for plugin discovery, especially this "plugin not found" case, would lock in the behavior.

iProzd (changes requested):
One remaining fallback edge case: `load_plugin()` now falls through when an opened candidate is missing the required DeepPot/free-error symbols, but operation-specific symbols are still resolved after the plugin is cached. For example, a stale `libdeepmd_backend_tf.so` earlier in `DP_BACKEND_PLUGIN_PATH` may export the DeepPot symbols but not `deepmd_create_deepspin_backend_v1` / `deepmd_create_deeptensor_backend_v1` / DCM / pbtxt-convert symbols. In that case the backend plugin is cached successfully, and a later DeepSpin/DeepTensor/DCM call fails at `load_typed_symbol()` without trying the remaining candidates.

Could we either validate the required symbol set for the requested operation before caching/returning the plugin, or make the operation-specific symbol resolution participate in the same candidate fallback/error aggregation path?

njzjz:
@iProzd Addressed in 1a36509a7.

The operation-specific paths now call `load_plugin_with_symbols(...)` with their required symbols before returning/caching a plugin:

- DeepSpin validates `deepmd_create_deepspin_backend_v1` / `deepmd_delete_deepspin_backend_v1`.
- DeepTensor validates `deepmd_create_deeptensor_backend_v1` / `deepmd_delete_deeptensor_backend_v1`.
- DipoleChargeModifier validates `deepmd_create_dipole_charge_modifier_backend_v1` / `deepmd_delete_dipole_charge_modifier_backend_v1`.
- pbtxt conversion validates `deepmd_convert_pbtxt_to_pb_v1`.

Those symbols are resolved inside the same candidate fallback loop. If an earlier plugin from `DP_BACKEND_PLUGIN_PATH` opens but is missing one of the requested operation symbols, the error is appended to the aggregated discovery message, the handle is closed, and discovery continues to the next candidate. The cache key is now `(backend, required-symbol-set)`, so a plugin cached from a prior DeepPot-only load cannot satisfy a later DeepSpin/DeepTensor/DCM/pbtxt request unless it also exports that operation-specific ABI.

I also checked `git diff --check origin/master...HEAD`, and the current head CI for 1a36509a7 is green.

iProzd (approved):

wanghan-iapcm (approved):

PR #5533, opened 2026-06-14
fix(jax): stabilize repflow dynamic selection export

njzjz:
## Summary
- add an internal fixed-capacity dynamic-selection layout for repflows so JAX/jax2tf export avoids runtime-sized edge/angle tensors
- skip unnecessary bincount in sum-only aggregate calls with a known owner count
- add regression coverage comparing compact and static dynamic selection outputs

## Validation
- ruff check .
- ruff format .
- pytest source/tests/universal/dpmodel/descriptor/test_descriptor.py::TestDPA3StaticDynamicSelDP::test_static_dynamic_sel_matches_packed_dynamic_sel -q
- dp convert-backend DPA-3.2-5M.pth DPA-3.2-5M.savedmodel

## Summary by CodeRabbit

* **New Features**
  * Added an export-friendly “static dynamic” execution mode for DPA3 repflow neighbor/edge/angle selection (fixed-capacity layout).
* **Bug Fixes**
  * Improved static-dynamic indexing and edge count handling to keep outputs consistent, including correct padded-slot behavior and layout-mode preservation on restore.
  * Updated export-time warning when static-capacity materialization is triggered.
* **Documentation**
  * Clarified DPA3 `repflow`/`use_dynamic_sel` angle-selection constraints for JAX export and memory scaling.
* **Chores**
  * Refined `aggregate` to compute bin counts only when required.
* **Tests**
  * Added tests verifying static-dynamic matches packed dynamic results and validating `aggregate` with explicit `num_owner`.

### DIFF
--- a/deepmd/jax/descriptor/repflows.py
+++ b/deepmd/jax/descriptor/repflows.py
@@ -16,7 +16,14 @@ from deepmd.jax.common import (
 
 @flax_module
 class DescrptBlockRepflows(DescrptBlockRepflowsDP):
-    pass
+    # JAX/jax2tf export cannot represent the compact dynamic layout where
+    # boolean indexing creates runtime-sized ``n_edge``/``n_angle`` arrays.
+    # Use the fixed-capacity dynamic layout instead:
+    #   edges  = nf * nloc * e_sel
+    #   angles = nf * nloc * a_sel * a_sel
+    # Invalid slots are still masked by switch weights, so DPA-3 outputs match
+    # the compact dynamic implementation.
+    _use_static_dynamic_sel = True
 
 
 @flax_module
@@ -26,3 +33,6 @@ class RepFlowLayer(RepFlowLayerDP):
         "e_residual",
         "a_residual",
     }
+    # Keep the layer-level graph operations in the same fixed-capacity layout
+    # selected by the owning descriptor block.
+    _use_static_dynamic_sel = True
--- a/deepmd/dpmodel/descriptor/repflows.py
+++ b/deepmd/dpmodel/descriptor/repflows.py
@@ -282,6 +282,7 @@ class DescrptBlockRepflows(NativeOP, DescriptorBlock):
                 "The JAX exportable dynamic-selection layout materializes "
                 "fixed angle capacity nf * nloc * a_sel * a_sel. Keep a_sel "
                 "modest for exportable DPA-3 models.",
+                stacklevel=2,
             )
         self.use_loc_mapping = use_loc_mapping
         self.sel_reduce_factor = sel_reduce_factor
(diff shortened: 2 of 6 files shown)


wanghan-iapcm (approved):

deepmd/jax/descriptor/repflows.py
```diff
@@ -16,7 +16,14 @@
...
 class DescrptBlockRepflows(DescrptBlockRepflowsDP):
-    pass
+    # JAX/jax2tf export cannot represent the compact dynamic layout where
+    # boolean indexing creates runtime-sized ``n_edge``/``n_angle`` arrays.
+    # Use the fixed-capacity dynamic layout instead:
+    #   edges  = nf * nloc * e_sel
+    #   angles = nf * nloc * a_sel * a_sel
+    # Invalid slots are still masked by switch weights, so DPA-3 outputs match
+    # the compact dynamic implementation.
+    _use_static_dynamic_sel = True
```
wanghan-iapcm, line 26: The static layout reintroduces — and for angles amplifies — the padded cost that dynamic selection exists to avoid, and that is silent here.

Compact dynamic-sel processes only the real edges (`int(sum(nlist_mask))`), so configs deliberately over-provision `sel`. The shipped examples make this explicit: `examples/water/dpa3/input_torch_dynamic.json` uses `e_sel=1200, a_sel=300` (10x the non-dynamic `input_torch.json`'s `120/30`). Under this static path those become fixed capacities:

- edges:  `nf * nloc * e_sel`       = `nf * nloc * 1200`   (10x the 120-baseline)
- angles: `nf * nloc * a_sel * a_sel` = `nf * nloc * 90000` (100x the 30-baseline, the `a_sel**2` term)

So a model trained with the recommended dynamic-sel example, exported via JAX, materializes an `nf x nloc x 300 x 300 x dim` angle tensor — the opposite of what the user picked dynamic selection for, with no diagnostic.

Suggestion: emit a one-time `log.warning` when this path is enabled with a large `a_sel` (the quadratic term dominates), and/or document in the export docs that JAX export materializes `nf * nloc * a_sel**2` angle slots, so `a_sel` should be kept modest for exportable models. A large `sel` under compact mode is correct/intended, so the warning should be scoped to the static path only — not to large `sel` in general.

njzjz: I added a `warning.warn` message.

iProzd (approved):
Two non-blocking suggestions:

1. The `a_sel**2` angle materialization is the main caveat. The `warnings.warn` helps, but it'd be good to also document in the export docs that JAX export materializes `nf * nloc * a_sel**2` angle slots, so `a_sel` should be kept modest for exportable models.
2. Patch coverage on `network.py` is low (~30%) — the new `aggregate` branches (the `num_owner` concat path and the `average=True` assert) aren't exercised. A small unit test targeting `aggregate` directly would prevent regressions.

LGTM otherwise.

PR #5537, opened 2026-06-15
fix(pt): enable second-order autograd for tabulate descriptors

### LINKED ISSUE #4994: PyTorch Backend: Missing autograd wrapper for se_t descriptor's second-order derivatives
### Summary

The se_t descriptor (TabulateFusionSeTOp) does not seem to support second-order derivative calculations (required for virial/stress), even though the underlying C++/CUDA kernel (deepmd::tabulate_fusion_se_t_grad_grad_gpu) and its wrapper function (TabulateFusionSeTGradGradForward) are present in the code.

### DeePMD-kit Version

devel latest

### Backend and its version

do not matter

### Python Version, CUDA Version, GCC Version, LAMMPS Version, etc

_No response_

### Details

1. Working Implementation (se_a): The se_a descriptor has a complete autograd chain. Its first-derivative calculation is wrapped in TabulateFusionSeAGradOp, whose backward method correctly calls TabulateFusionSeAGradGradForward. This makes the operation fully second-order differentiable.

2. Incomplete Implementation (se_t): For the se_t descriptor, the TabulateFusionSeTOp's backward method calls TabulateFusionSeTGradForward directly. There is no corresponding TabulateFusionSeTGradOp class to wrap this first-derivative calculation.

3. The Consequence: Because the autograd wrapper (GradOp) is missing, the TabulateFusionSeTGradGradForward function is defined but is never actually called by the PyTorch autograd engine. This effectively makes it dead code in the context of PyTorch's automatic differentiation and prevents training models with virial/stress labels when using the se_t descriptor.

njzjz:
## Summary
- wrap PyTorch tabulate descriptor first-derivative kernels in autograd Functions
- connect `se_a`, `se_atten`, `se_t`, `se_r`, and `se_t_tebd` backward paths to existing grad-grad kernels
- add second-order backward regression tests for all affected tabulate descriptor ops

Fixes #4994.

## Tests
- `pytest source/tests/pt/test_tabulate_fusion_se_a.py source/tests/pt/test_tabulate_fusion_se_atten.py source/tests/pt/test_tabulate_fusion_se_r.py source/tests/pt/test_tabulate_fusion_se_t.py source/tests/pt/test_tabulate_fusion_se_t_tebd.py -q`
- `ruff check .`
- `ruff format .`
- commit hook suite, including `clang-format`

## Summary by CodeRabbit

* **New Features**
  * Improved/extended second-order gradient support for tabulate fusion operations (embeddings and attention), improving correctness for higher-order differentiation workflows.

* **Tests**
  * Added second-order backward tests across multiple tabulate fusion ops.
  * Introduced shared finite-difference-based test utilities to validate autograd second-order gradients (including numerical tolerance checks).

### DIFF
--- a/source/tests/pt/test_tabulate_fusion_se_a.py
+++ b/source/tests/pt/test_tabulate_fusion_se_a.py
@@ -1488,6 +1488,34 @@ class TestTabulateFusionSeAOp(unittest.TestCase):
             rtol=self.prec,
         )
 
+    def test_second_order_backward(self) -> None:
+        forward_result = torch.ops.deepmd.tabulate_fusion_se_a(
+            self.table_tensor,
+            self.table_info_tensor,
+            self.em_x_tensor,
+            self.em_tensor,
+            self.last_layer_size,
+        )
+
+        descriptor_tensor = forward_result[0]
+        dy_tensor = torch.ones_like(descriptor_tensor, requires_grad=True)
+        dy_dem_x, dy_dem = torch.autograd.grad(
+            descriptor_tensor,
+            (self.em_x_tensor, self.em_tensor),
+            grad_outputs=dy_tensor,
+            create_graph=True,
+        )
+
+        dz_dy_tensor = torch.autograd.grad(
+            dy_dem_x.sum() + dy_dem.sum(),
+            dy_tensor,
+        )[0]
+
+        self.assertIsNotNone(dz_dy_tensor)
+        self.assertEqual(dz_dy_tensor.shape, descriptor_tensor.shape)
+        self.assertTrue(torch.isfinite(dz_dy_tensor).all())
+        self.assertGreater(dz_dy_tensor.abs().max().item(), 0.0)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/source/tests/pt/test_tabulate_fusion_se_a.py
+++ b/source/tests/pt/test_tabulate_fusion_se_a.py
@@ -13,6 +13,9 @@ from deepmd.pt.utils import (
 from ..consistent.common import (
     parameterized,
 )
+from .tabulate_test_utils import (
+    assert_second_order_backward_matches_finite_difference,
+)
 
 
 @parameterized((torch.float64, torch.float32))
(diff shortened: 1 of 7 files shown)


source/tests/pt/test_tabulate_fusion_se_a.py
```diff
@@ -1488,6 +1488,34 @@ def test_backward(self) -> None:
...
+
+        dz_dy_tensor = torch.autograd.grad(
+            dy_dem_x.sum() + dy_dem.sum(),
+            dy_tensor,
+        )[0]
+
+        self.assertIsNotNone(dz_dy_tensor)
+        self.assertEqual(dz_dy_tensor.shape, descriptor_tensor.shape)
+        self.assertTrue(torch.isfinite(dz_dy_tensor).all())
+        self.assertGreater(dz_dy_tensor.abs().max().item(), 0.0)
```
wanghan-iapcm: The five new `test_second_order_backward` tests only assert that the second-order gradient is non-None, shape-matches the descriptor, is all-finite, and has nonzero max-abs. They do not verify numerical correctness — no `torch.autograd.gradcheck`, finite-difference, or reference comparison.

This matters here because for se_atten/se_t/se_r/se_t_tebd the grad-grad kernels (`TabulateFusionSe*GradGradForward`) become reachable through PT autograd for the first time in this PR (previously dead code), so their values have never been validated end-to-end. A mis-wired but finite Hessian would pass all four assertions.

The adjacent first-order `test_backward` in each file already compares against a stored numerical reference (`assert_allclose(grad, expected, atol/rtol=self.prec)`). Suggest the second-order tests follow the same pattern — e.g. `torch.autograd.gradcheck` on the first-derivative op, or a finite-difference reference for `dz_dy` — so a wrong grad-grad value is actually caught rather than just confirming the path runs.

njzjz: Thanks, addressed in b0b422ddd.

I replaced the five second-order tests with a finite-difference check of `dz/dy`: the helper builds a non-uniform `dy`, projects the first derivatives with non-uniform weights, then compares the grad-grad result against central finite differences of that projected first-gradient value. This should catch finite but miswired grad-grad outputs instead of only confirming the path runs.

Validated with:
- `pytest source/tests/pt/test_tabulate_fusion_se_a.py source/tests/pt/test_tabulate_fusion_se_atten.py source/tests/pt/test_tabulate_fusion_se_t.py source/tests/pt/test_tabulate_fusion_se_r.py source/tests/pt/test_tabulate_fusion_se_t_tebd.py -k second_order_backward -q` -> `10 passed`
- same five files without `-k` -> `30 passed`
- `ruff check .`
- `ruff format .`

wanghan-iapcm (approved):

OutisLi (approved):

PR #5556, opened 2026-06-19
feat(pt/dpa4): add SO3 readout and small bug fix

OutisLi:
## Summary by CodeRabbit

* **New Features**
  * Added `so3_readout` configuration option for SeZM descriptor, enabling selection of computation methods (`"none"`, `"glu"`, `"mlp"`) for final descriptor channel extraction.

* **Improvements**
  * Optimized compiler version checks to run only when compilation is actually invoked, reducing unnecessary checks at initialization.
  * Enhanced neighbor list performance with device-aware method selection for both CPU and GPU execution.

### DIFF
--- a/deepmd/pt/model/descriptor/sezm.py
+++ b/deepmd/pt/model/descriptor/sezm.py
@@ -320,6 +320,16 @@ class DescrptSeZM(BaseDescriptor, nn.Module):
     message_node_so3
         If True, use the corresponding post-aggregation SO(3) Wigner-D grid-net
         branch. The message is the query and the node state is the context.
+    so3_readout
+        Read-out FFN mode for the final ``l=0`` descriptor. ``"none"`` applies a
+        degree-0 scalar FFN to the ``l=0`` slice only; ``l>0`` coefficients are
+        discarded before the read-out. ``"glu"`` and ``"mlp"`` apply a full
+        equivariant FFN whose degree equals the node degree of the last
+        interaction block, driven by the SO(3) Wigner-D grid, so ``l>0`` geometry
+        is folded into ``l=0`` before the scalar is extracted. The value selects
+        the quadratic grid product (``"glu"``) or the polynomial point-wise grid
+        MLP (``"mlp"``). The Wigner-D frame order follows ``kmax``. The residual
+        stays on the ``l=0`` channel.
     lebedev_quadrature
         Either one boolean applied to both S2 branches, or two booleans
         ``[so2_enabled, ffn_enabled]`` aligned with ``s2_activation``. If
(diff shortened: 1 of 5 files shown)


wanghan-iapcm:
### Code review

Found 1 blocking issue (CI is red because of it):

**`so3_readout` is added to the pt serialize format but not to the dpmodel `DescrptDPA4` counterpart, breaking DPA4 cross-backend consistency.**

The DPA4/SeZM descriptor is not pt-only — `deepmd/dpmodel/descriptor/dpa4.py` shares the identical serialize format (its `deserialize` consumes the pt `serialize()` output). This PR adds `so3_readout` to the pt side only:

https://github.com/deepmodeling/deepmd-kit/blob/1d244fd955596638bbd617fdda5004a67a899998/deepmd/pt/model/descriptor/sezm.py#L2079-L2081

but `deepmd/dpmodel/descriptor/dpa4.py` is unchanged (no `so3_readout` in its `__init__`, `serialize`, or `output_ffn`).

**Why this fails:** `source/tests/consistent/descriptor/test_dpa4.py::...::test_pt_consistent_with_ref` serializes the dpmodel descriptor, deserializes it into pt, re-serializes, and asserts the two config dicts are equal. After this PR the pt dict carries a `so3_readout` key the dpmodel dict lacks, so `np.testing.assert_equal(data1, data2)` fails — on the **default** (`so3_readout="none"`) config. That is why every `Test Python` shard is red, e.g. `TestDPA4_float64_1_1_1_False_True_gaussian::test_pt_consistent_with_ref`.

There is also a quieter consequence for the non-default modes: dpmodel `DescrptDPA4.__init__` has a `**kwargs` catch-all that silently swallows `so3_readout`, and its `output_ffn` is hard-wired to the `none` behavior. Since this PR makes `examples/water/dpa4/input.json` default to `"so3_readout": "mlp"`, a model trained from the example would silently produce a different descriptor under the dpmodel / jax / pt_expt backends (no error, wrong numerics).

This is the serialization-discipline rule: a new serialized field must be added to all backends that share the format, in the same change — otherwise the shared-format consistency test breaks (as it has here).

**Suggested fix:** mirror `so3_readout` in dpmodel `DescrptDPA4` — `__init__` param, `serialize()` emit, and the `output_ffn` construction + forward output-mixing block (a direct port of the pt diff) — and add `glu`/`mlp` cases to the parity/consistent test matrix so the non-default paths are covered cross-backend. (A separate dpmodel-only PR won't work: a one-sided serialize-format change breaks the same consistency test in the other direction, so the two sides must land together.)

wanghan-iapcm:
I may handle it after #5555 merges.

PR #5565, opened 2026-06-20
feat(data): add dpdata format conversion

### LINKED ISSUE #5237: Automatic Format Conversion in dp using dpdata
Enable `dp` to automatically convert public dataset formats (e.g., extxyz, ase) to DeepMD format by calling `dpdata`.

**Goals:**
- Streamline the user experience for using external datasets.
- Automatically detect and convert formats.

---
*Generated by OpenClaw*

njzjz:
## Summary

- Add automatic training/validation data conversion through `format` and `out_format` (`output_format` alias), with `dpdata>=1.1.0` as a runtime dependency.
- Default converted output to `deepmd/lmdb` and cache it beneath the working directory in `.deepmd_dpdata_cache`. Record source file membership, size, nanosecond mtime/ctime, and small-file content hashes in an atomic sidecar manifest. Restored inputs with older timestamps invalidate the cache; inputs changed during conversion fail without publishing a valid manifest.
- Support converted LMDB in TensorFlow, JAX, PyTorch, and PyTorch Exportable, with reproducible sampling and cleanup on training/setup failures. Preserve Python 3.10 compatibility.
- Make LMDB `prob_uniform` sample original systems with equal probability; this changes existing LMDB runs that previously sampled them proportionally to their frame counts. Share extended block probability calculation with the NPY loader.
- Serialize conversion with owned locks. Recover dead local writers, but never evict an identified remote writer solely because its heartbeat looks stale; report such locks for owner verification.

Closes #5237

## Configuration constraints

The `deepmd/lmdb` default requires exactly one resolved input path and an explicit `model/type_map`. LMDB does not support Paddle, data modifiers, or explicit `sys_probs`; unsupported settings fail clearly. Use `out_format: deepmd/hdf5` for those capabilities or multiple input paths. The training and validation option documentation states these constraints.

## Validation

All checks below ran on Python 3.10.12 with CPU backends:

- Built the Python package and TensorFlow/PyTorch native operators against TensorFlow 2.21.0 and PyTorch 2.13.0+cpu.
- `ruff check .`, `ruff format .`, and `git diff --check`.
- 178 focused tests passed across conversion/cache, shared task-map cleanup, common LMDB, PyTorch LMDB adapters, PyTorch Exportable routing, TensorFlow setup cleanup, and `TestDPTestEner::test_1frame` (plus 22 subtests).
- Two JAX setup-cleanup tests and four TF2 entrypoint tests passed, for 184 tests total.
- Real dpdata 1.1.0 conversion regression: replace energy 0 with energy 42, restore an older source mtime, and verify the LMDB reader sees energy 42.
- Four simultaneous fresh processes reused one converted cache; after restoring the input with an older mtime, four fresh processes all read the updated energy and left no conversion lock.
- `dp --version` and Python/TensorFlow imports succeeded.

Coding agent: Codex
Codex version: codex-cli 0.154.0
Model: gpt-6-astra
Reasoning effort: xhigh

## Summary by CodeRabbit

* **New Features**
  * Added automatic conversion from formats such as `.extxyz` to LMDB, HDF5, and other supported formats.
  * Added `format` and `out_format`/`output_format` options for training and validation datasets.
  * Improved LMDB loading with reproducible batch ordering, sampling controls, caching, and neighbor-statistics support.

* **Bug Fixes**
  * Improved handling of periodic and non-periodic data and unsupported backend configurations.
  * Improved cleanup when training or validation setup fails.

* **Tests**
  * Expanded coverage for conversion, caching, LMDB loading, validation, and failure handling.


njzjz:

- @wanghan-iapcm — 20 commits on changed files; 74 reviews on exact changed files (`deepmd/dpmodel/utils/lmdb_data.py`, `deepmd/pd/entrypoints/main.py`, `deepmd/pt/entrypoints/main.py`, `deepmd/pt_expt/entrypoints/main.py`, `deepmd/utils/argcheck.py`, `deepmd/utils/data_system.py`, `pyproject.toml`, `source/tests/pt_expt/test_lmdb_training.py`).
- @iProzd — 29 commits on changed files (`deepmd/dpmodel/utils/lmdb_data.py`, `deepmd/pt/entrypoints/main.py`, `deepmd/utils/argcheck.py`, `deepmd/utils/data_system.py`, `pyproject.toml`).

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/utils/data_system.py
```diff
@@ -815,14 +1120,214 @@ def prob_sys_size_ext(keywords: str, nsystems: int, nbatch: int) -> list[float]:
...
+        str(Path.cwd().resolve(strict=False)),
+        str(source.resolve(strict=False)),
+        fmt,
+        out_fmt,
+    )
+    if cache_key in _DPDATA_CONVERSION_CACHE:
+        return _DPDATA_CONVERSION_CACHE[cache_key]
+
+    output = _conversion_cache_path(source, fmt, out_fmt)
+    output.parent.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
njzjz: Fixed in b16b18067. Cached conversion entries are now revalidated with _is_conversion_current before reuse; stale entries are evicted and regenerated.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/utils/data_system.py
```diff
@@ -672,6 +716,267 @@ def _check_type_map_consistency(
...
+                "LMDB stores atom type indices and the training data adapter "
+                "must map them to element names."
+            )
+
+        self.lmdb_path = lmdb_path
+        self._reader = LmdbDataReader(
+            lmdb_path, type_map, batch_size, mixed_batch=False
+        )
+        self._type_map = list(type_map)
+        self.mixed_type = self._detect_mixed_type()
```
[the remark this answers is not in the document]
njzjz: Fixed in b16b18067. LmdbDataSystem now uses the reader mixed_type metadata instead of scanning every frame during initialization. The targeted conversion and LMDB tests pass.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/utils/data_system.py
```diff
@@ -30,6 +37,43 @@
 
 log = logging.getLogger(__name__)
 
+_DPDATA_CACHE_DIR = ".deepmd_dpdata_cache"
+_DPDATA_DEFAULT_OUT_FORMAT = "lmdb"
```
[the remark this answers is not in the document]
njzjz: Updated the PR description to document the actual per-working-directory .deepmd_dpdata_cache location and removed the developer-specific absolute path.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/dpmodel/utils/lmdb_data.py
```diff
@@ -936,8 +932,18 @@ def compute_block_targets(
...
+    # avoids an import cycle when the legacy adapter imports the LMDB reader.
+    block_weights = np.asarray([weight for _, _, weight in blocks], dtype=float)
+    assert np.all(block_weights >= 0), "the weight of a block should be no less than 0"
+    block_probs = block_weights / np.sum(block_weights)
+    sys_probs = np.zeros(nsystems, dtype=np.float64)
+    for block_idx, (stt, end, _weight) in enumerate(blocks):
+        block_frames = np.asarray(system_nframes[stt:end], dtype=float)
+        sys_probs[stt:end] = (
+            block_frames / np.sum(block_frames) * block_probs[block_idx]
+        )
```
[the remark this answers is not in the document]
njzjz: Resolved in 7ae5c2e32: block weights are now validated before use, including shape and value checks, with regression coverage.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/utils/data_system.py
```diff
@@ -815,14 +1110,220 @@ def prob_sys_size_ext(keywords: str, nsystems: int, nbatch: int) -> list[float]:
...
+            return True
+        time.sleep(1.0)
+    return False
+
+
+def _remove_path(path: Path) -> None:
+    if path.is_dir():
+        shutil.rmtree(path)
+    elif path.exists():
+        path.unlink()
```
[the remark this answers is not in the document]
njzjz: Resolved in 7ae5c2e32: cache cleanup now refuses to follow directory symlinks, preventing deletion outside the cache tree. The focused test suite passes (23 tests).

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/utils/data_system.py
```diff
@@ -1,12 +1,24 @@
...
 import warnings
 from functools import (
     cached_property,
 )
+from pathlib import (
+    Path,
+)
 from typing import (
     Any,
+    Self,
```
wanghan-iapcm, line 21: Blocking. `typing.Self` was added in Python 3.11 and `pyproject.toml` still declares `requires-python = ">=3.10"`, so this module cannot be imported on 3.10 at all. It is what the earlier bot review requested changes for, and every `Test Python (*, 3.10)` job plus the cp310 wheel build on this commit fail with `ImportError: cannot import name 'Self' from 'typing'`.

`typing_extensions` is already a transitive dependency, so `from typing_extensions import Self` is the smallest fix; dropping the annotation in favour of the class name works too.

deepmd/utils/data_system.py
```diff
@@ -885,15 +1826,40 @@ def process_systems(
...
+    conversion_inputs: list[str] = []
+    if fmt is not None:
+        for path in search_paths:
+            conversion_inputs.extend(_iter_conversion_inputs(path, patterns))
+        if (
+            _is_dpdata_lmdb_format(_canonical_dpdata_out_format(out_fmt))
+            and len(conversion_inputs) != 1
+        ):
+            raise ValueError(
+                "Automatic LMDB conversion requires exactly one resolved input "
```
wanghan-iapcm, line 1843: Under the defaults this makes the headline workflow fail for anything but a single input. `out_format` defaults to `deepmd/lmdb` for both training and validation data (`argcheck.py` 5431 and 5544), and `_iter_conversion_inputs` returns each `systems` entry as its own input, so `"systems": ["/A", "/B"], "format": "extxyz"` raises here, and so does a single directory with `rglob_patterns` matching more than one file. Nothing in `doc_out_format` tells the user the default carries this constraint.

The restriction itself is defensible for LMDB. What I would change is the default: either pick a format that keeps the multi-input contract, or make the doc state the single-input limitation and point at `deepmd/hdf5` explicitly, so the error is not the first place a user learns about it.

deepmd/utils/data_system.py
```diff
@@ -927,13 +1893,35 @@ def get_data(
...
+    lmdb_path = validate_lmdb_systems(systems, backend_name="legacy data loader")
+    if lmdb_path is not None:
+        validate_lmdb_sampling_options(jdata)
+        if type_map is None:
+            raise ValueError(
+                "LMDB training data requires model/type_map to be set. "
+                "Set model/type_map or choose training_data.out_format="
+                "'deepmd/hdf5' for automatic conversion."
+            )
+        return LmdbDataSystem(
```
wanghan-iapcm, line 1918: `modifier` is accepted by `get_data` and forwarded to `DeepmdDataSystem` on the NPY branch, but this branch discards it without a word. The TF entrypoint builds the modifier from `model/modifier` and passes it into every `get_data` call (`deepmd/tf/entrypoints/train.py` 264, 273, 312), so a TF run that configures `DipoleChargeModifier` and also sets `format` on its training data ends up training on unmodified data, silently, under the default `out_format`.

`sys_probs` already gets the fail-fast treatment through `validate_lmdb_sampling_options`; `modifier` should get the same, or be implemented. Same call site, lower stakes: `LmdbDataSystem.__init__` accepts `seed` and threads it into the batch sampler, but it is never passed here, while the pt path hands `DpLoaderSet` a per-rank seed and its sibling `LmdbDataset(...)` construction also omits it. Shuffle seeds configured in the training script therefore do not reach LMDB-backed datasets on either backend.

deepmd/utils/argcheck.py
```diff
@@ -5366,6 +5366,19 @@ def training_data_args() -> list[
...
+    doc_format = (
+        "The input data format passed to dpdata for automatic conversion. "
+        "If this key is not set, `systems` must already point to DeePMD data. "
+        "If this key is set to a non-DeePMD format, each selected input path is "
+        "loaded by dpdata and converted before training. Use dpdata format names "
+        "such as `extxyz`, `ase/structure`, `ase/traj`, or `auto`."
+    )
+    doc_out_format = (
+        "The output data format passed to dpdata for automatic conversion. "
+        "When `format` requests conversion from a non-DeePMD format, this key "
```
wanghan-iapcm, line 5378: This describes a conditional default, but the `Argument` below sets `default="deepmd/lmdb"` unconditionally, and the same wording is repeated for validation data at 5504. The practical consequence is on Paddle: a user who sets `format: extxyz` and leaves `out_format` alone gets `deepmd/lmdb` injected by argcheck normalisation and then hits `NotImplementedError: Paddle backend does not support LMDB data yet`, from a key they never wrote. Either make the wording match the code, or make the default genuinely conditional and let Paddle pick a format it can read.

wanghan-iapcm (changes requested):
The conversion pipeline itself looks solid to me: the cache key covers the resolved path, both formats, the schema and the dpdata version, so same-basename systems do not collide; dpdata 1.1.0's `deepmd/lmdb` writer stages to a temp directory and publishes with a single `os.replace`, so a killed conversion cannot leave a half-written database at the final path; and the multi-element `type_map` question that worried me turned out fine, because `MultiSystems` harmonises `atom_names` before `.to()` and the writer refuses to emit metadata without a global `type_map`, which the reader then remaps or rejects loudly. I checked each of those against the 1.1.0 sources rather than assuming.

What holds this up is one hard blocker and one design point that shows up in three places.

The blocker is the Python 3.10 import, inline below. It is the same thing the bot review already requested changes for, and every 3.10 job on this commit is red because of it.

The design point: `out_format` defaults to `deepmd/lmdb` unconditionally, but the LMDB path through `get_data` has a narrower contract than the NPY path it replaces. It accepts exactly one resolved input, drops `modifier` and the seed, and is refused by Paddle. Each of those is a reasonable limitation of an LMDB-backed dataset on its own; the problem is that a user who sets only `format` gets all of them silently, under the default, and the argcheck doc says the default is conditional when it is not. The three inline comments below are the concrete symptoms. Either the default should be a format that keeps the NPY contract, or each dropped capability needs to fail fast the way `sys_probs` already does through `validate_lmdb_sampling_options`.

Non-blocking, for the same pass or a follow-up:

- The lock-liveness code in `_lock_owner_is_alive` / `_recover_stale_conversion_lock` has several branches with no test (other-host owner, `PermissionError`, live owner, `owner_alive is None and lease_expired`), and `format: auto` plus `_looks_like_extxyz` are documented but never exercised, since every test passes `fmt="extxyz"`. The repository guideline asks for both sides of each boolean branch to be covered, and these are the branches that decide whether a live writer's lock gets deleted.
- Related: the 30 s stale lease is judged against a heartbeat that another node sees through NFS attribute caching, whose default window is longer than that. The repository already has a rank-0-plus-broadcast mechanism for exactly this shape of problem (`run_stat_on_chief`), and `torch.distributed` is initialised before conversion runs in the pt entrypoint, so it would be usable here and would also give waiters a propagated failure instead of a retry cascade.
- `deepmd/utils/data_system.py` now carries a third `LmdbDataSystem`-shaped implementation next to `deepmd/pt_expt/utils/lmdb_dataset.py` and pt's `LmdbDataset`, with the stat-group logic from #5944 copied into it. Worth converging on one before the copies drift.
- `compute_block_targets` in `lmdb_data.py` re-inlines `prob_sys_size_ext` (the function-local import used before already broke the cycle) and changes `prob_uniform` for existing LMDB runs from size-proportional to truly uniform. That is probably the right semantics, but it is a sampling change for current users and the PR description does not mention it.
- The PR body says `dpdata>=1.0.1`; `pyproject.toml` pins `>=1.1.0`, and the 1.1.0 writer's atomicity is what the docstring in `_write_dpdata_conversion` relies on, so 1.1.0 is the right one and the body should match.

deepmd/utils/data_system.py
```diff
@@ -844,14 +1313,483 @@ def prob_sys_size_ext(keywords: str, nsystems: int, nbatch: int) -> list[float]:
...
+    _DPDATA_SOURCE_MTIME_CACHE[cache_key] = (now, latest)
+    return latest
+
+
+def _is_conversion_current(
+    source: Path, output: Path, *, force_source_scan: bool = False
+) -> bool:
+    if not output.exists():
+        return False
+    return output.stat().st_mtime >= _source_mtime(
```
iProzd, line 1498: **Blocking: the cache can serve stale data silently.**

`output.st_mtime >= source.st_mtime` assumes a modified source carries a newer mtime. Every metadata-preserving copy or restore breaks that — `cp -p`, `rsync -a`, `tar -xp`, `git checkout` of a data file, a snapshot rollback — all give new content an *older* mtime than the cache that already exists.

Reproduced at this head:

1. convert `data.extxyz` carrying `energy=0.0`; cache is built;
2. replace it with `energy=42.0` and set its mtime one day back, as an archive restore would;
3. `process_systems(src, fmt="extxyz")` from a fresh process returns the cached LMDB, and `LmdbDataReader` serves `energy=0.0`.

Nothing is logged, so the run trains on the previous dataset. The directory branch of `_source_mtime` compares the same way, so a restored source directory behaves identically.

Fix: compare a recorded signature instead of mtime ordering — a sidecar manifest written at conversion time (per-file size and mtime, or a hash for small inputs) that must match exactly. Folding that signature into the `_conversion_cache_path` digest works too, since a changed source would then map to a different cache entry.

iProzd (commented):
Reviewed the full diff against `8cfd46e3` at head `b3a2715`. One new blocking issue; details inline.

**The conversion cache can silently serve stale data.** `_is_conversion_current` infers freshness from `output.st_mtime >= source.st_mtime`, which any mtime-preserving restore (`cp -p`, `rsync -a`, `tar -xp`, snapshot rollback) breaks. Reproduced on this head.

Already reported and still reproducing, so not re-raised: the `typing.Self` import (the single root cause of all 15 red checks), the single-input LMDB restriction, the dropped `modifier` and `seed` in `get_data`, the `doc_out_format` mismatch, and six open CodeQL/CodeRabbit threads.

Checked and found sound: the `dpdata>=1.1.0` pin and its transactional LMDB writer; the conversion lock under six concurrent processes; `LmdbTestDataNlocView` satisfying the JAX/TF2 validation contract; ordinary in-place source edits invalidating the cache correctly.

Not verified: multi-task, spin/Hessian, the C++/LAMMPS consumers, and the GPU runs listed in the description.

deepmd/pd/entrypoints/main.py
```diff
@@ -108,11 +110,39 @@ def prepare_trainer_input_single(
             validation_dataset_params["systems"] if validation_dataset_params else None
         )
         training_systems = training_dataset_params["systems"]
+        validate_backend_data_config(
```
wanghan-iapcm, line 113: This guard only runs inside `prepare_trainer_input_single`, which `train()` reaches via `get_trainer`. But the neighbor-stat block earlier in `train()` (line 336) calls the shared `get_data(config["training"]["training_data"], 0, type_map, None)` with no guard at all. With `format: extxyz` and the new default `out_format: deepmd/lmdb`, that call runs the whole dpdata to LMDB conversion, writes the cache entry, builds an `LmdbDataSystem` and passes it to `BaseModel.update_sel`, and only afterwards does `get_trainer` raise `NotImplementedError("Paddle backend does not support LMDB data yet")`.

That inverts the contract `validate_backend_data_config` documents for itself ("Reject unsupported converted output before dpdata performs any I/O"). On a large dataset the user pays the full conversion cost and disk footprint before being told the configuration is unsupported.

Calling `validate_backend_data_config` once before the neighbor-stat block would close it; putting the check inside the shared `get_data`/`process_systems` layer would also cover `deepmd/dpmodel/entrypoints/compress_common.py` and `deepmd/pt/entrypoints/compress.py`, which have the same unguarded shape.

deepmd/utils/data_system.py
```diff
@@ -701,6 +761,411 @@ def _check_type_map_consistency(
...
+            ):
+                array = array.reshape(nframes, -1)
+            out[key] = array
+
+        atype = np.asarray(batch["atype"], dtype=np.int32)
+        real_natoms_vec = np.asarray(
+            batch.get("real_natoms_vec", batch["natoms"]), dtype=np.int32
+        )
+        if real_natoms_vec.ndim == 1:
+            real_natoms_vec = np.tile(real_natoms_vec, (nframes, 1))
```
wanghan-iapcm, line 1046: `natoms_vec[2:]` is built here as a per-type maximum taken independently across the frames of the batch, so for a batch whose frames differ in composition `sum(natoms_vec[2:])` can exceed `natoms_vec[0]`.

The mixed-type convention from 5fe6e5756 (#2372) is the opposite: `DeepmdDataSystem._merge_batch_data` sets `natoms_vec = np.zeros(2 + ntypes)` and then `natoms_vec[0:3] = max_natoms`, leaving the per-type slots zero and keeping the per-frame truth in `real_natoms_vec`. That preserves `sum(natoms_vec[2:]) == natoms_vec[0]`, which TF consumers assume.

Most mixed-type TF paths read `real_natoms_vec`, so this is latent there, but `deepmd/tf/fit/ener.py` slices `atype_nall` with `tf.reduce_sum(natoms[2 : 2 + ntypes_atom])` on the spin path, and `get_data()` now returns `LmdbDataSystem` to the TF/TF2/JAX legacy trainers, so that slice can run past `nloc`. Suggest mirroring `_merge_batch_data` (`natoms_vec = np.zeros(2 + ntypes, dtype=np.int32); natoms_vec[0:3] = pad_nloc`), or an assertion plus a comment if the per-type values are deliberately relied upon somewhere.

wanghan-iapcm (commented):
All four threads from the earlier round are addressed at this head: `typing_extensions.Self` restores Python 3.10 (all twelve 3.10 shards and the cp310 wheel pass), the multi-input and `modifier` cases now fail fast with clear messages before any dpdata I/O, and the `out_format` docs describe the actual unconditional default. The new tests pass locally (conversion + train_data 44, lmdb_data + pt dataloader 190, pt_expt 47, whole dpmodel dir 1302), and the changes-requested verdict no longer stands.

Two inline points worth landing before merge, neither blocking. Smaller notes here:

- `validate_lmdb_systems` (https://github.com/deepmodeling/deepmd-kit/blob/2f3798826f858c5f6a5a772f0e6cdf91e496231d/deepmd/utils/data_system.py#L61) and `prob_sys_size_ext` (https://github.com/deepmodeling/deepmd-kit/blob/2f3798826f858c5f6a5a772f0e6cdf91e496231d/deepmd/utils/data_system.py#L795) are new public functions without the numpydoc `Parameters`/`Returns` sections the rest of the module uses.
- The cache signature walks the source with `os.walk(..., followlinks=False)` (https://github.com/deepmodeling/deepmd-kit/blob/2f3798826f858c5f6a5a772f0e6cdf91e496231d/deepmd/utils/data_system.py#L1502), so files under a symlinked subdirectory never enter the manifest. A dataset that links its frames in, then edits them, reuses the stale LMDB silently. Either follow links or record the link targets in the signature.
- On the multi-input thread: rejecting with a pointer to `deepmd/hdf5` is an acceptable resolution, but with `deepmd/lmdb` as the unconditional default it means the multi-directory aggregation that #5033 added now errors by default as soon as `format` is set. Worth a second thought on whether the default should be conditional on a single resolved input.

PR #5568, opened 2026-06-21
feat(pt/dpa4): add nvalchemi-toolkit warpper for DPA4

OutisLi:
## Release Notes

* **New Features**
  * Added `DPA4Wrapper`/`SeZMWrapper` integration for nvalchemi-toolkit inference, including support for frozen `.pt2` packages.
  * Added runnable water examples for single-point inference, NVE/NVT molecular dynamics, and geometry relaxation.

* **Documentation**
  * Added a third-party documentation page and an example README covering setup, runtime options, and stress conventions.
  * Updated documentation navigation to include the new nvalchemi guide.

* **Bug Fixes / Improvements**
  * Improved package import behavior with a clearer error and explicit install instructions when the optional toolkit is unavailable.

* **Tests**
  * Added comprehensive correctness tests comparing wrapper outputs to toolkit-native neighbor-list execution.


doc/third-party/nvalchemi.md
njzjz, line 1: This doc should be put under `doc/third-party`, which documents how third-party packages are integrated. The `inference` directory documents the programming language interface.

njzjz (changes requested):

deepmd/pt/nvalchemi/dpa4wrapper.py
```diff
@@ -0,0 +1,665 @@
...
+            model_ret = self.model.forward_lower(
+                model_inputs["coord"],
+                model_inputs["atype"],
+                model_inputs["edge_index"],
+                model_inputs["edge_vec"],
+                model_inputs["edge_scatter_index"],
+                model_inputs["edge_mask"],
+                do_atomic_virial=need_virial,
+                charge_spin=model_inputs["charge_spin"],
+            )
```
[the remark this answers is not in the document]
OutisLi: SeZM has a custom lower interface, and the wrapper already has passing unit tests + benchmarks through exactly these calls

deepmd/pt/nvalchemi/dpa4wrapper.py
```diff
@@ -0,0 +1,665 @@
...
+            model_inputs["coord"],
+            model_inputs["atype"],
+            model_inputs["edge_index"],
+            model_inputs["edge_vec"],
+            model_inputs["edge_scatter_index"],
+            model_inputs["edge_mask"],
+            None,
+            None,
+            model_inputs["charge_spin"],
+        )
```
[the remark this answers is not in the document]
OutisLi: SeZM has a custom lower interface, and the wrapper already has passing unit tests + benchmarks through exactly these calls

OutisLi:
Should be merged after PR #5571  is merged, this PR relies on that

deepmd/pt/nvalchemi/dpa4wrapper.py
```diff
@@ -0,0 +1,665 @@
...
+            model_inputs["atype"],
+            model_inputs["edge_index"],
+            model_inputs["edge_vec"],
+            model_inputs["edge_scatter_index"],
+            model_inputs["edge_mask"],
+            charge_spin=model_inputs["charge_spin"],
+            embedding_only=True,
+        )
+        n_node = data.num_nodes
+        node_embeddings = ret["descriptor"].reshape(n_node, self._descriptor_dim)
```
wanghan-iapcm, line 510: This call will crash at runtime. `forward_common_lower` has no `embedding_only` parameter (and no `**kwargs`), and its return dict does not contain a `"descriptor"` key. The current signature in `deepmd/pt/model/model/sezm_model.py` is `(coord, atype, edge_index, edge_vec, edge_scatter_index, edge_mask, fparam, aparam, comm_dict, extended_coord_corr, charge_spin, input_prec, use_compile)` and it returns `{energy, energy_redu, energy_derv_r, energy_derv_c, energy_derv_c_redu, mask}`. The token `embedding_only` does not exist anywhere in `deepmd/`. So `compute_embeddings` (and `embedding_shapes`/`_descriptor_dim`, which feed the same path) raise `TypeError: ... unexpected keyword argument 'embedding_only'` on first use, then `KeyError: 'descriptor'` if that were patched. The embeddings feature needs to be wired to whatever the model actually exposes for descriptor output.

source/tests/pt/model/test_sezm_nvalchemi.py
```diff
@@ -0,0 +1,391 @@
...
+        try:
+            super().tearDownClass()
+        finally:
+            ctx = getattr(cls, "_default_device_ctx", None)
+            if ctx is not None:
+                ctx.__exit__(None, None, None)
+                delattr(cls, "_default_device_ctx")
+
+
+@unittest.skipUnless(NVALCHEMI_AVAILABLE, "nvalchemi-toolkit is not installed")
```
wanghan-iapcm, line 103: The entire suite is gated on `NVALCHEMI_AVAILABLE`, but `nvalchemi-toolkit` is added only as a user-facing optional extra in `pyproject.toml` -- it is not in any test/CI dependency group (and is not installable from PyPI). So in upstream CI every test here silently skips and this ~665-line module gets zero executed coverage. Concrete evidence this hurts: the `compute_embeddings` runtime crash (`forward_common_lower(embedding_only=True)` / `ret["descriptor"]`) is exercised by `test_compute_embeddings_shapes` yet went uncaught, i.e. the suite was never actually run with nvalchemi installed. Please add nvalchemi-toolkit to the CI test dependencies (or otherwise ensure these tests run), so the parity tests provide real coverage.

deepmd/pt/nvalchemi/dpa4wrapper.py
```diff
@@ -0,0 +1,665 @@
...
+            if cell is None:
+                raise ValueError(
+                    "stress output requires a periodic `cell` for the volume."
+                )
+            atom_virial = model_output["extended_virial"].reshape(n_node, 9)
+            virial = torch.zeros(
+                n_graph, 9, dtype=atom_virial.dtype, device=atom_virial.device
+            ).index_add_(0, batch_idx, atom_virial)
+            volume = torch.det(cell.to(virial.dtype)).abs().view(n_graph, 1, 1)
+            output["stress"] = (virial.view(n_graph, 3, 3) / volume).to(out_dtype)
```
wanghan-iapcm, line 419: Warning -- please double-check the stress sign convention. This emits `stress = +virial / |det(cell)|` (unsymmetrized), whereas deepmd's own ASE calculator uses the opposite sign and symmetrizes: `stress = -0.5 * (v + v.T) / volume` (`deepmd/calculator.py:160-161`, "tensile stress is positive"), where `v` is the same virial quantity. If nvalchemi follows the standard ASE/tensile convention, this stress is sign-inverted (and missing symmetrization), which would drive NPT/NPH barostats the wrong way. The parity test only asserts `stress*volume == native_virial`, which is circular and cannot catch a wrong sign. Please confirm nvalchemi's expected stress convention and add a non-circular check (e.g. a finite-difference stress, or a known-pressure reference).

wanghan-iapcm, line 419: Following up on the sign: the current parity check (`stress * volume == native_virial`) is circular — it just re-asserts the wrapper's own `virial/volume` definition, so it cannot catch a wrong sign or the missing symmetrization. A non-circular UT would settle this for good. Two options:

**A. Finite-difference of energy w.r.t. strain (independent ground truth).** Stress is the energy derivative under cell strain, derivable from energies alone — fully independent of how the wrapper computes virial:

```python
# periodic system, model run in float64
def E(eps):                       # eps: symmetric 3x3 strain
    F = I + eps
    return model_energy(coord @ F.T, cell @ F.T)      # affine-deform atoms + cell

dE_deps[i, j] = (E(+d * E_ij) - E(-d * E_ij)) / (2 * d)   # central diff, d ~ 1e-4
stress_ref    = sign * dE_deps / det(cell)
torch.testing.assert_close(wrapper_stress, stress_ref)
```

Energy is a sign-unambiguous scalar, so `dE/deps` is the physically-correct stress; a flipped sign (or the missing symmetrization) fails the assert. Choosing the `sign` that matches nvalchemi's documented convention — and writing it explicitly — is exactly what resolves the open question, instead of asserting it in a comment.

**B. Compare against deepmd's own stress (simpler).** deepmd's ASE calculator already produces a Cauchy stress with a separately-tested convention: `stress = -0.5 * (v + v.T) / V` (`deepmd/calculator.py:160-161`). Assert the wrapper's stress equals that for the same structure. This is non-circular because the reference is deepmd's trusted, independently-validated convention — not the wrapper's `virial/volume` formula — so a sign flip (or asymmetry) relative to deepmd fails immediately.

Both need a periodic system with non-zero stress (the water example works), float64, and can be gated `@skipUnless(NVALCHEMI_AVAILABLE)` like the rest — now actually runnable in CI since `nvalchemi-toolkit` was added to the `test` deps. (A) is the gold standard (catches a bug even if deepmd and the wrapper shared one); (B) is a few lines and still catches a sign flip vs deepmd.

iProzd (approved):
I do not see a remaining blocking issue.

Two small non-blocking suggestions:
- Please consider guarding stress computation against non-finite or near-zero cell volumes before dividing by `det(cell)`.
- Since `.pt2` support is user-facing, a minimal `.pt2` wrapper smoke test would be helpful to pin the AOTI callable signature.

LGTM otherwise.

njzjz (approved):

wanghan-iapcm (approved):

PR #5571, opened 2026-06-22
feat(pt): add embedding sub cli for pt models

### LINKED ISSUE #5507: [BUG] `dp --pt eval-desc` OOMs for DPA4-Plus descriptors while `dp --pt test` inference completes
### Bug summary

`dp --pt eval-desc` runs out of GPU memory when extracting atomic descriptors with the DPA4-Plus pre-trained model on several larger systems, although the normal `dp --pt test` inference workflow with the same model/checkpoint completes successfully in the same environment.

The failure is reproducible on a single deepmd/npy/mixed system with 216 atoms and 59 frames. DeePMD auto batch size reduction reaches batch size 1, but the descriptor evaluation still fails with:

`deepmd.utils.errors.OutOfMemoryError: The callable still throws an out-of-memory (OOM) error even when batch size is 1!`

The CUDA traceback points to the SeZM/SO2 descriptor path:

`deepmd/pt/model/descriptor/sezm_nn/so2.py`, inside `torch.bmm(D_m_prime, x_src)`.

I would like to confirm whether this is expected behavior for DPA4-Plus `eval-desc`, or whether `eval-desc` should support further chunking/streaming to avoid materializing such a large intermediate descriptor computation.

### DeePMD-kit Version

DeePMD-kit v0.1.dev1+g27a18b604

### Backend and its version

PyTorch backend: torch 2.11.0+cu126

Built from source

Model:
  DPA4-Plus pretrained model, specifically `DPA4-Plus-OMat24-16M.pt`.

Minimal failing input:
  A single DPData system, `sampled_dpdata/216`, with 216 atoms and 59 frames.

  The DPData directory contains:

  ```text
  sampled_dpdata/216/type.raw
  sampled_dpdata/216/type_map.raw
  sampled_dpdata/216/set.000/box.npy
  sampled_dpdata/216/set.000/coord.npy
  sampled_dpdata/216/set.000/energy.npy
  sampled_dpdata/216/set.000/force.npy
  sampled_dpdata/216/set.000/real_atom_types.npy
  sampled_dpdata/216/set.000/spin.npy
  sampled_dpdata/216/set.000/virial.npy

  Failing command:

  dp --pt eval-desc \
    -s sampled_dpdata/216 \
    -m DPA4-Plus-OMat24-16M.pt \
    -o desc_216

  The original full-run command was:

  dp --pt eval-desc \
    -s sampled_dpdata \
    -m model.ckpt.pt \
    -o desc_train

  Successful inference comparison:

  dp --pt test \
    -s other_dpdata \
    -m model.ckpt.pt \
    -d results

  In the same workflow, four dp --pt test Slurm jobs completed successfully, while dp --pt eval-desc failed on the
  training descriptor job. The failing descriptor job first reached sampled_dpdata/216:

  [2026-06-05 03:10:23,331] DEEPMD INFO    # processing system : .../sampled_dpdata/216
  [2026-06-05 03:10:23,349] DEEPMD INFO    # evaluating descriptors for 59 frames

  Relevant traceback:

  File ".../site-packages/deepmd/pt/model/descriptor/sezm_nn/so2.py", line 1313, in forward
      x_local = torch.bmm(D_m_prime, x_src)  # (E, D_m, C_wide)

  torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 240.00 MiB.
  GPU 0 has a total capacity of 31.73 GiB of which 226.69 MiB is free.
  Including non-PyTorch memory, this process has 31.51 GiB memory in use.
  Of the allocated memory 30.94 GiB is allocated by PyTorch, and 162.59 MiB is reserved by PyTorch but unallocated.

  deepmd.utils.errors.OutOfMemoryError:
  The callable still throws an out-of-memory (OOM) error even when batch size is 1!

  Environment variables used in the Slurm job:

  export CUDA_HOME=/opt/devtools/nvidia/cuda-12.6.3
  export CUDAToolkit_ROOT=${CUDA_HOME}
  export TORCH_CUDA_ARCH_LIST=7.0
  export CMAKE_CUDA_ARCHITECTURES=70
  export DP_VARIANT=cuda
  export DP_ENABLE_PYTORCH=1
  export DP_ENABLE_TENSORFLOW=0
  export DP_COMPILE_INFER=0
  export TORCHDYNAMO_DISABLE=1
  export TORCHINDUCTOR_COMPILE_THREADS=1
  export DP_INTERFACE_PREC=high
  export OMP_NUM_THREADS=2

  Additional observation:
  A CPU fallback attempt for the same sampled_dpdata/216 system was also killed by Slurm due to memory usage, so the issue does not appear to be only CUDA memory fragmentation.

### Steps to Reproduce

1. Prepare a DeePMD-kit environment with the PyTorch backend and CUDA support.

2. Use the DPA4-Plus pretrained model:
     `DPA4-Plus-OMat24-16M.pt`.

3. Prepare the attached minimal DPData system:
     `sampled_dpdata/216`, containing 216 atoms and 59 frames.

4. Run descriptor extraction:

 ```bash
  dp --pt eval-desc \
    -s sampled_dpdata/216 \
    -m DPA4-Plus-OMat24-16M.pt \
    -o desc_216
```
5. Observe that dp --pt eval-desc fails with CUDA OOM inside the SeZM/SO2 descriptor block, even after auto batch size is reduced to 1.

6. For comparison, run normal inference with the same model/checkpoint:

  dp --pt test \
    -s <a compatible DPData test set> \
    -m DPA4-Plus-OMat24-16M.pt \
    -d results

Expected behavior:
  dp --pt eval-desc should either complete, expose a way to further chunk descriptor extraction, or provide a clearer
  documented limitation for DPA4-Plus descriptor extraction on larger systems.

Actual behavior:
  dp --pt eval-desc fails with OOM at batch size 1.

[issue_dpa4_plus_eval_desc_oom.tar.gz](https://github.com/user-attachments/files/28737322/issue_dpa4_plus_eval_desc_oom.tar.gz)

### PR DESCRIPTION
**Pull request by OutisLi**
Fixed #5507 

## Summary by CodeRabbit

* **New Features**
  * Added `dp embed` to export `descriptor`, `atomic_feature`, and `structural_feature` into a single HDF5 file with per-system groups.
  * Added `DeepPot.eval_embedding()` with selectable output precision (`fp32`, `fp64`, `native`).
  * Added embedding dtype CLI option and multi-task head selection (`--head` / `--model-branch`).
* **Bug Fixes**
  * Improved fail-fast validation for neighbor-list pairing when model sub-configurations don’t match.
* **Documentation**
  * Updated embedding docs, including Python API usage and embedding-only behavior.
* **Tests**
  * Added embedding-focused test coverage and removed obsolete OOM-retry tests.

### CODE REVIEW

File: deepmd/pt/infer/deep_eval.py [hunk 1]
```diff
@@ -1125,28 +1111,158 @@ def eval_fitting_last_layer(
...
+            if charge_spin is not None
+            else None
+        )
+        out = model.forward_embedding(
+            coord_input,
+            type_input,
+            box=box_input,
+            fparam=fparam_input,
+            aparam=aparam_input,
+            charge_spin=charge_spin_input,
```
wanghan-iapcm, line 1367: Back-compat break for models frozen before this PR. `forward_embedding` is a newly `@torch.jit.export`ed method; for a `.pth` file `self.dp` is the `torch.jit.load()`ed ScriptModule (line ~227), so a model frozen by an older deepmd version has no `forward_embedding` baked into its TorchScript. Calling `eval_embedding` -- and the now-rerouted `eval_descriptor`/`eval_fitting_last_layer`/`dp eval-desc` -- on such a model raises `AttributeError: 'ScriptModule' object has no attribute 'forward_embedding'`. The pt backend is expected to keep working with previously frozen artifacts, and no test loads a pre-existing frozen model through this path. Consider a graceful fallback (or a clear error) when the loaded module lacks `forward_embedding`.

deepmd/pt/infer/deep_eval.py
```diff
@@ -1125,28 +1111,158 @@ def eval_fitting_last_layer(
...
+        Raises
+        ------
+        NotImplementedError
+            If the loaded model does not support embedding extraction.
+        """
+        if self._has_spin:
+            raise NotImplementedError(
+                "eval_embedding is not supported for spin models in the "
+                "PyTorch backend."
+            )
```
wanghan-iapcm, line 1283: `eval_embedding` guards `_has_spin` but not the data modifier. `_eval_embedding` calls `model.forward_embedding(...)` directly (line ~1253), bypassing `ModelWrapper.forward`, which is where the data modifier is applied. The old `eval_descriptor`/`eval_fitting_last_layer` went through the modifier; rerouting them through `eval_embedding` introduces a **new silent** bypass -- a frozen `.pth` carrying a `data_modifier` (e.g. DPLR) would compute embeddings on unmodified inputs and return wrong results with no error. Note this same file already handles exactly this case for the strategy path (lines ~294-298: `elif self.modifier is not None: ... fall back ... so the modifier is still applied`). Add a corresponding `self.modifier is not None` guard/fallback here (mirroring the `_has_spin` pattern).

deepmd/entrypoints/embedding.py
```diff
@@ -0,0 +1,206 @@
...
+    while name in used_names:
+        name = f"{base}_{idx}"
+        idx += 1
+    used_names.add(name)
+    return name
+
+
+def embedding(
+    *,
+    model: str,
```
wanghan-iapcm, line 68: This entire new entrypoint is untested. `source/tests/pt/model/test_embedding.py` covers only the energy se_e2_a + SeZM-ener happy paths via the Python API; there is no test exercising `embedding()` here -- so the `datafile` vs `system` branches, `mixed_type` handling, box-None (`not data.pbc`), fparam/aparam presence, the `_unique_group_name` collision branch, and the `RuntimeError("Did not find valid system")` path all go uncovered. More broadly, several new branches elsewhere are untested too: dipole/polarizability `return_atomic_feature=True` (which the docs claim are supported), and the `EnergyFittingNetDirect` / `_has_spin` / linear-no-embedding `NotImplementedError` guards. Per the repo's test-coverage expectation, please add tests that exercise these reachable paths (at minimum a CLI/entrypoint test and a dipole/polar embedding test plus the negative-path raises).

deepmd/pt/infer/deep_eval.py
```diff
@@ -1125,28 +1111,158 @@ def eval_fitting_last_layer(
...
+            aparam=aparam_input,
+            charge_spin=charge_spin_input,
+        )
+        # Single output-precision boundary: the model produces embeddings in its
+        # native precision, and they are emitted as float32 (ample for the
+        # downstream analysis these embeddings are meant for).
+        return (
+            out["descriptor"].detach().to(torch.float32).cpu().numpy(),
+            out["atomic_feature"].detach().to(torch.float32).cpu().numpy(),
+            out["structural_feature"].detach().to(torch.float32).cpu().numpy(),
```
wanghan-iapcm: Non-blocking: this float32 cast is a behavioral change for the pre-existing `eval_descriptor`/`eval_fitting_last_layer` APIs (and `dp eval-desc` `.npy` output), which previously returned the model's native precision (float64 for a double model). Both now route through `eval_embedding` and silently downcast to float32 -- the PR loosens `source/tests/infer/test_models.py` from 7/12 to 5 places to match. Also note the Paddle backend still returns native precision, so the same public API now differs by backend. Suggest keeping the legacy `eval_descriptor`/`eval_fitting_last_layer` at native precision and downcasting only in the new `dp embed`/`eval_embedding` output (or expose a dtype option).

deepmd/pt/infer/deep_eval.py
OutisLi: I think a modifier guard is not needed here. In the PyTorch backend the data modifier is not an input preprocessor; ModelWrapper applies it after the model forward by adding modifier predictions to physical outputs such as energy/force/virial. The descriptor and fitting-last-layer feature are internal tensors of the neural model and the modifier has no corresponding embedding representation.

The previous eval_descriptor/eval_fitting_last_layer implementation also captured these tensors inside the model forward via hooks before ModelWrapper added modifier outputs, so modifier presence did not affect the returned descriptor/features. The new eval_embedding path intentionally returns the same neural-model embeddings.

iProzd (changes requested):
Thanks, this is a useful embedding extraction feature and the backend/docs compatibility issues look much better now.

I still think one back-compat issue should be fixed before merge. `eval_descriptor()` and `eval_fitting_last_layer()` are existing public APIs, but they now route through `eval_embedding()`, which requires the frozen TorchScript model to expose the newly exported `forward_embedding()` method. Models frozen before this PR cannot have that method, even though the old hook-based descriptor / fitting-last-layer path used to work for them.

The explicit `NotImplementedError` is fine for the new `eval_embedding()` API, but it is a regression for the existing descriptor APIs and for `dp eval-desc`. Could you keep a fallback to the old hook-based path when `forward_embedding` is unavailable, and add a regression test for a legacy-frozen / legacy-hook model path?

After that, LGTM.

wanghan-iapcm (approved):

OutisLi:
Closes  #5507

njzjz (approved):

iProzd (approved):

PR #5581, opened 2026-06-24
feat(dpmodel): NeighborGraph foundation — contract, segment toolkit, numpy builder, edge force/virial

wanghan-iapcm:
## Summary

First, foundational slice of the unified **`NeighborGraph`** neighbor-list design (design discussion: wanghan-iapcm/deepmd-kit#4). This PR adds the backend-agnostic `deepmd/dpmodel` core only — **pure numpy / array-API, no torch/jax/model dependencies** — so it is self-contained and not yet wired into any model.

## What this adds (`deepmd/dpmodel/utils/`)

- **`neighbor_graph.py`** — `NeighborGraph` dataclass (the edge-graph contract: `n_node`, `edge_index (2,E)` SoA `[src,dst]`, `edge_vec (E,3) = r_src − r_dst`, `edge_mask`, optional `n_local`/`angle_index`/`angle_mask`) and `GraphLayout` (edge/angle/node/frame capacities + `min_edges`). Node validity is **derived** (`arange(N) < sum(n_node)`), not a stored field; `node_validity_mask` and `pad_and_guard_edges` (compact-prefix layout, capacity/guard, overflow) helpers.
- **`segment.py`** — mask-aware, backend-dispatched `segment_sum` / `segment_mean` (on `xp_add_at`).
- **`numpy_neighbor_list.py`** — `NumpyNeighborList`, an all-pairs reference builder (coords → `NeighborGraph`, periodic + non-periodic, frame-offset).
- **`edge_derivatives.py`** — `edge_force_virial`: assembles per-node force, atom virial (attributed **full-to-`src`**, the canonical TF==PT convention), and global virial from a per-edge gradient `g_e`.

## Tests

`source/tests/common/dpmodel/`: 22 tests, all passing — dataclasses, `node_validity_mask`, segment ops, `pad_and_guard_edges`, builder correctness (brute-force-verified, periodic + non-periodic + multi-frame + capacity padding), and force/virial assembly (force formula, global virial, full-to-`src` atom virial summing to global, masked-edge no-op).

## Scope / deferred

Numpy-only foundation. Deferred to follow-ups: torch/jax backend dispatch + O(N) vesin/nvalchemiops builders; the `forward_common_lower_graph` model seam + `edge_energy_deriv` autograd; descriptor ports; attention `segment_softmax` + the 3-body angle list; jax static-`E_max` capacity/bin-packing; C/C++ ABI + LAMMPS + multi-rank.

## Known limitations

- `NumpyNeighborList` is an O(N²·shells) reference builder with an orthorhombic-shell heuristic — reference/test use; production O(N) builders come later.
- `edge_force_virial` consumes a pre-computed `g_e`; the autograd that produces it is a follow-up.
- `segment_max`/`segment_softmax` not yet implemented (only needed by attention descriptors later).
- Rectangular multi-frame only; ragged bin-packing is a later (jax) step.

## Summary by CodeRabbit

* **New Features**
  * Added neighbor-graph utilities for building graphs from inputs, including padding/guard edges and node validity masking.
  * Added segment reduction helpers (sum/mean) with safe handling for empty segments.
  * Added edge-based force and virial assembly from per-edge gradients and geometry.
  * Expanded top-level exports to make these utilities available directly.
* **Bug Fixes**
  * Ensured padded/guard edges are fully excluded (contribute zero to forces and virials).
* **Tests**
  * Added extensive unit tests covering graph construction, edge/virial behavior (including masking and multi-frame cases), and segment reductions.


deepmd/dpmodel/utils/neighbor_graph/builder.py
```diff
@@ -0,0 +1,137 @@
...
+    mixed_types: bool = True,
+    layout: GraphLayout | None = None,
+) -> NeighborGraph:
+    """Build a NeighborGraph by reusing the tested dense nlist (dpmodel default).
+
+    Calls ``extend_input_and_build_neighbor_list`` (general-cell, tested) then
+    :func:`neighbor_graph_from_extended`. With ``sel`` large enough that no real
+    neighbor is truncated, the result is exactly the in-``rcut`` environment (the
+    ``sel``-as-normalization regime; see memory/spec_unified_edge_nlist.md).
+    """
```
iProzd, line 173: I think the `sel` semantics still need to be made explicit in this foundational API.

`build_neighbor_graph()` currently reuses `extend_input_and_build_neighbor_list()`, so the resulting graph inherits the legacy `sel` truncation behavior. This is fine if the intent of this builder is “legacy-compatible graph adapter”. However, the broader `NeighborGraph` design describes the graph as carrying the complete in-`rcut` environment.

Those are two different contracts:

- a graph-native contract: all real neighbors within `rcut`;
- a legacy-compatible contract: neighbors after descriptor-specific `sel` selection/truncation.

Right now `build_neighbor_graph()` only documents “with `sel` large enough”, but it does not enforce that, detect truncation, or encode which contract the returned graph satisfies. That can become a correctness footgun for later descriptor ports, because a downstream graph consumer may assume it sees all neighbors while the builder has silently truncated them.

Could we make this explicit in the API/docstring, or split the naming into something like `neighbor_graph_from_legacy_nlist` vs a future all-neighbor graph builder? If this function is meant to preserve legacy `sel`, I think the contract should say so directly.

deepmd/dpmodel/utils/neighbor_graph/derivatives.py
```diff
@@ -0,0 +1,47 @@
...
+    g = g_e * xp.astype(edge_mask[:, None], g_e.dtype)
+    src = edge_index[0]
+    dst = edge_index[1]
+    # force
+    force = segment_sum(g, dst, n_node_total) - segment_sum(g, src, n_node_total)
+    # per-edge virial w_e[k, j] = -g_e[k] * edge_vec[j]  (broadcast, no einsum)
+    w_edge = -(g[:, :, None] * edge_vec[:, None, :])  # (E, 3, 3)
+    # atom virial: full-to-src
+    atom_virial = segment_sum(w_edge, src, n_node_total)  # (N, 3, 3)
+    global_virial = xp.sum(w_edge, axis=0)  # (3, 3)
```
iProzd: `edge_force_virial()` currently collapses virials from all frames into a single `(3, 3)` tensor. Also see chatgpt comments above.

iProzd (changes requested):

source/tests/common/dpmodel/test_numpy_neighbor_list.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def test_static_capacity_padding(self) -> None:
+        ng = NumpyNeighborList().build(
+            self.coord, box=None, rcut=self.rcut, layout=GraphLayout(edge_capacity=64)
+        )
+        self.assertEqual(ng.edge_index.shape[1], 64)
+        self.assertEqual(ng.edge_vec.shape[0], 64)
+        self.assertEqual(int(ng.edge_mask.sum()), int(ng.edge_mask[:].sum()))
+        # masked-out tail contributes no real edges
+        self.assertTrue(np.all(ng.edge_vec[~ng.edge_mask] == 0.0))
```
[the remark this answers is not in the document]
wanghan-iapcm: Thanks. The tautological assertion was already removed in the current revision, but I took the underlying suggestion: `test_static_capacity_padding` now asserts the exact real-edge count (`edge_mask.sum() == n_real` from the brute-force oracle) and the compact-prefix padding (`edge_mask[:n_real]` all True, `[n_real:]` all False). Done in 2d9fd9856.

deepmd/dpmodel/utils/neighbor_graph/builder.py
```diff
@@ -0,0 +1,217 @@
...
+    from deepmd.dpmodel.utils.nlist import (
+        extend_input_and_build_neighbor_list,
+    )
+
+    extended_coord, _extended_atype, mapping, nlist = (
+        extend_input_and_build_neighbor_list(
+            coord, atype, rcut, sel, mixed_types=mixed_types, box=box
+        )
```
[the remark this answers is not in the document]
wanghan-iapcm: Thanks. This is a type-contract cleanup rather than a runtime bug: `extend_input_and_build_neighbor_list` only forwards `sel` to `build_neighbor_list`, which already accepts `int`, so an integer `sel` ran correctly (verified it reproduces the list-`sel` graph). I normalized it to list form anyway so the annotated `list[int]` boundary is honored, and added `test_int_sel_matches_list_sel`. Done in 2d9fd9856.

deepmd/dpmodel/utils/neighbor_graph/builder.py
wanghan-iapcm: Agreed — we've made the contract explicit rather than enforcing one behavior. As of the latest push, the docstrings state that `neighbor_graph_from_extended` / `build_neighbor_graph` are the **legacy-compatible** builders: the returned graph INHERITS the dense nlist's `sel` truncation, and it coincides with the complete in-`rcut` environment only when `sel` is non-binding.

The carry-all, graph-native (`sel`-as-normalization) contract will be a **separate** builder — `from_ijs`, fed by ASE/vesin — so the two contracts live in distinct functions instead of one ambiguous `build_neighbor_graph`. This matches your "split the naming" suggestion. Thanks for catching it.

deepmd/dpmodel/utils/neighbor_graph/derivatives.py
wanghan-iapcm: Fixed. `edge_force_virial` now returns a **per-frame** virial `(nf, 3, 3)` instead of the collapsed `(3, 3)`: it takes `n_node`, assigns each edge to the frame of its `dst` via `searchsorted(cumulative_sum(n_node))`, and `segment_sum`s per frame.

Added regression coverage that the single-frame-only test was missing: multi-frame with ragged node/edge counts, padded edge & node axes, an empty middle frame (`n_node=[3,0,5]`), and the all-empty (zero real edges) case.

deepmd/dpmodel/utils/neighbor_graph/derivatives.py
```diff
@@ -0,0 +1,46 @@
...
```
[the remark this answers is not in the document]
wanghan-iapcm: Fixed — virial is now reduced **per frame** (`nf, 3, 3`) using `n_node` / frame offsets (`searchsorted(cumulative_sum(n_node))` to assign each edge to its `dst`'s frame), so batched systems no longer have their virials summed together. Regression tests cover ragged multi-frame, padding, empty frames, and the all-empty case.

deepmd/dpmodel/utils/numpy_neighbor_list.py
```diff
@@ -0,0 +1,98 @@
...
+        self,
+        coord: np.ndarray,
+        box: np.ndarray | None,
+        rcut: float,
+        layout: GraphLayout | None = None,
+    ) -> NeighborGraph:
+        if layout is None:
+            layout = GraphLayout()
+        coord = np.asarray(coord, dtype=np.float64)
+        nf, nloc = coord.shape[0], coord.shape[1]
```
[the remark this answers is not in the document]
wanghan-iapcm: This file was removed: the from-scratch all-pairs `NumpyNeighborList` was dropped in favor of reusing deepmd's tested `extend_input_and_build_neighbor_list` (via `neighbor_graph_from_extended`). The independent all-pairs check now lives only as a test oracle (`brute_force_neighbor_sets` in `test_neighbor_graph_builder.py`), so this comment no longer applies.

iProzd (approved):
LGTM now.

deepmd/dpmodel/utils/neighbor_graph/__init__.py
```diff
@@ -0,0 +1,40 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+"""NeighborGraph: backend-agnostic edge-graph neighbor-list subsystem.
+
+The unified edge/graph neighbor-list contract and its supporting machinery:
+``graph`` (the ``NeighborGraph``/``GraphLayout`` contract + derived node-validity
++ edge padding), ``builder`` (the carry-all ``build_neighbor_graph`` dispatcher +
+the ``from_dense_quartet`` legacy converter), ``segment`` (mask-aware
+segment-reduction toolkit), and ``derivatives`` (edge force/virial assembly).
+See memory/spec_unified_edge_nlist.md.
```
OutisLi, line 9: These docstrings reference `memory/spec_unified_edge_nlist.md`, but neither that file nor a `memory/` directory exists in this PR or the repo. The same dangling reference recurs in `graph.py`, `segment.py`, `builder.py`, and `derivatives.py` (along with the `decision #5/#17` callouts). For anyone else reading the code this points at a spec they can't find.

Could we vendor the spec under `doc/`, or swap these for the public design discussion (wanghan-iapcm#4)? Not blocking, but worth resolving before merge so the references stay valid.

deepmd/dpmodel/utils/neighbor_graph/derivatives.py
```diff
@@ -0,0 +1,87 @@
...
+        (N, 3) per-node force.
+    atom_virial
+        (N, 3, 3) per-node virial, full-to-``src`` attribution.
+    virial
+        (nf, 3, 3) PER-FRAME virial. A multi-frame graph keeps each frame's
+        virial separate (DeePMD virials are per frame); edges are assigned to a
+        frame via the frame of their ``dst`` node.
+    """
+    xp = array_api_compat.array_namespace(g_e)
+    n_real = int(xp.sum(n_node))  # real node count
```
OutisLi, line 65: `n_real = int(xp.sum(n_node))` runs unconditionally, so even when a static `node_capacity` is supplied (the intended jax path) this still calls `int()` on a traced array and would raise `ConcretizationTypeError` under `jax.jit`. Since the docstring already advertises the jax static-`N_max` use, maybe short-circuit so the static path never takes `int()` of a traced sum:

```python
n_out = int(node_capacity) if node_capacity is not None else int(xp.sum(n_node))
```

This also mirrors how `node_validity_mask` deliberately avoids `int(sum)`. numpy-only is unaffected today, so purely forward-looking.

deepmd/dpmodel/utils/neighbor_graph/derivatives.py
```diff
@@ -0,0 +1,87 @@
...
+  per-edge virial w_e = -g_e (x) edge_vec_e
+  atom virial attributed FULL-TO-src (canonical TF==pt-legacy convention)
+  per-frame virial = sum over the edges of that frame of w_e  (DeePMD virials
+  are per frame; a multi-frame NeighborGraph must NOT collapse frames)
+Padding/guard edges (edge_mask == 0) are zeroed before any scatter.
+"""
+
+import array_api_compat
+
+from deepmd.dpmodel.array_api import (
```
OutisLi, line 18: Minor consistency nit: `Array` is imported at runtime here, while `graph.py` and `builder.py` keep it under a `TYPE_CHECKING` guard. As `Array` is a pure type alias, moving this under `TYPE_CHECKING` (with `from __future__ import annotations`) would keep the subpackage uniform. Non-blocking.

OutisLi (commented):
Thanks @wanghan-iapcm — clean, self-contained foundation, and the test coverage is excellent (force sign, the full-to-`src` ↔ per-frame virial closure, empty/ragged frames, padding/guard edges, and the cross-PBC-image case). I re-derived the force/virial assembly independently and it checks out. A few non-blocking points inline. The only one I'd suggest addressing before merge is the dangling `memory/spec_unified_edge_nlist.md` references, since they'll be confusing for other maintainers.

PR #5582, opened 2026-06-24
feat(core): batch prod env mat over frames

### LINKED ISSUE #2618: [Feature Request] Compute `prod_env_mat` OP in parallel in the dimension of the frame

### LINKED ISSUE #5584: Clarify padded A env-mat normalization semantics across GPU, CPU, and dpmodel
While reviewing #5582, I noticed a subtle mismatch in how padded neighbor slots are normalized in the A environment matrix. This issue is for documenting and confirming the intended semantics, rather than claiming a confirmed downstream `se_atten` regression.

## Observation

For an A descriptor neighbor slot with `nlist[ii] < 0`, the raw environment-matrix value is conceptually a zero descriptor. If normalization is applied uniformly, the normalized value should be

```text
(0 - avg[type][slot_component]) / std[type][slot_component]
```

for all 4 A components of that padded slot.

In the older TF GPU kernel, real neighbor slots normalized all 4 components, but the padded-slot branch only touched the first component:

https://github.com/deepmodeling/deepmd-kit/blob/1a8586b8925ac6e2e9953b0f61dcfd8b96c748b5/source/lib/src/gpu/prod_env_mat.cu#L513-L520

```cpp
for (int ii = 0; ii < 4; ii++) {
  row_descript[idx_value + ii] =
      (dd[ii] - avg[type[i_idx_nall] * ndescrpt + idx_value + ii]) /
      std[type[i_idx_nall] * ndescrpt + idx_value + ii];
}
} else {
  row_descript[idx_value] -= avg[type[i_idx_nall] * ndescrpt + idx_value] /
                           std[type[i_idx_nall] * ndescrpt + idx_value];
}
```

So for padded slots, components 1..3 were not normalized or assigned by this branch. This is observable in low-level env-mat output if `avg` is nonzero.

In #5582, I changed the GPU branch to apply the same normalized-zero rule to all 4 components:

https://github.com/deepmodeling/deepmd-kit/blob/048376baa460d0082c3fbfef1a31c244d3f1a727/source/lib/src/gpu/prod_env_mat.cu#L513-L523

```cpp
} else {
  for (int ii = 0; ii < 4; ii++) {
    row_descript[idx_value + ii] -=
        avg[type[i_idx_nall] * ndescrpt + idx_value + ii] /
        std[type[i_idx_nall] * ndescrpt + idx_value + ii];
  }
}
```

## Comparison with `dpmodel`

`dpmodel` appears to follow the uniform-normalization semantics.

First, `_make_env_mat` masks absent neighbors, so the raw env-mat for padded slots is zero:

https://github.com/deepmodeling/deepmd-kit/blob/b52c359ec6b6f6f34998e00f774c35d253cf09e5/deepmd/dpmodel/utils/env_mat.py#L88-L104

```python
weight = weight * xp.astype(xp.expand_dims(mask, axis=-1), weight.dtype)
if radial_only:
    env_mat = t0 * weight
else:
    env_mat = xp.concat([t0, t1], axis=-1) * weight
```

Then `EnvMat.call` subtracts `davg` and divides by `dstd` for the whole `em` tensor:

https://github.com/deepmodeling/deepmd-kit/blob/b52c359ec6b6f6f34998e00f774c35d253cf09e5/deepmd/dpmodel/utils/env_mat.py#L158-L165

```python
if davg is not None:
    em -= xp.reshape(xp.take(davg, xp.reshape(atype, (-1,)), axis=0), em.shape)
if dstd is not None:
    em /= xp.reshape(xp.take(dstd, xp.reshape(atype, (-1,)), axis=0), em.shape)
```

Therefore, for padded A slots, `dpmodel` produces `-davg/dstd` for all 4 components, matching the #5582 GPU change and not the older TF GPU padded-slot branch.

## Downstream masking in `se_atten` / DPA1

This mismatch may not affect the usual final descriptor in the standard attention path, because padded neighbors are also masked downstream.

For TF `se_atten`, the op returns `nmask`, which is converted to `negative_mask`:

https://github.com/deepmodeling/deepmd-kit/blob/b52c359ec6b6f6f34998e00f774c35d253cf09e5/deepmd/tf/descriptor/se_atten.py#L639-L644

```python
self.nmask = tf.cast(
    tf.reshape(self.nmask, [-1, 1, self.sel_all_a[0]]),
    self.filter_precision,
)
self.negative_mask = -(2 << 32) * (1.0 - self.nmask)
```

The attention weights are masked for absent neighbors:

https://github.com/deepmodeling/deepmd-kit/blob/b52c359ec6b6f6f34998e00f774c35d253cf09e5/deepmd/tf/descriptor/se_atten.py#L968-L978

```python
attn *= self.nmask
attn += self.negative_mask
attn = tf.nn.softmax(attn, axis=-1)
...
attn *= tf.reshape(self.nmask, [-1, attn.shape[-1], 1])
```

The final contraction then uses the attention result:

https://github.com/deepmodeling/deepmd-kit/blob/b52c359ec6b6f6f34998e00f774c35d253cf09e5/deepmd/tf/descriptor/se_atten.py#L1309-L1313

```python
return tf.matmul(
    tf.reshape(inputs_i, [natom, shape_i[1] // 4, 4]),
    xyz_scatter_att,
    transpose_a=True,
)
```

For `dpmodel` DPA1, the env mat is normalized first:

https://github.com/deepmodeling/deepmd-kit/blob/b52c359ec6b6f6f34998e00f774c35d253cf09e5/deepmd/dpmodel/descriptor/dpa1.py#L1096-L1102

and then `nlist_mask = nlist != -1` is passed into attention:

https://github.com/deepmodeling/deepmd-kit/blob/b52c359ec6b6f6f34998e00f774c35d253cf09e5/deepmd/dpmodel/descriptor/dpa1.py#L1117-L1122

https://github.com/deepmodeling/deepmd-kit/blob/b52c359ec6b6f6f34998e00f774c35d253cf09e5/deepmd/dpmodel/descriptor/dpa1.py#L1220-L1222

and absent neighbors are masked in attention:

https://github.com/deepmodeling/deepmd-kit/blob/b52c359ec6b6f6f34998e00f774c35d253cf09e5/deepmd/dpmodel/descriptor/dpa1.py#L1663-L1670

This suggests that, for the usual attention path, padded env-mat values should not contribute physically even when they differ. Also, if `davg` is zero, this difference is numerically invisible.

## Questions

1. Should env-mat operators define padded A slots as normalized zero descriptors, i.e. `-avg/std` for every component, even if most consumers also carry an explicit mask?
2. Or should padded slots be forced to zero after normalization, making consumers less sensitive to whether they mask correctly?
3. Are there non-attention paths, `attn_layer == 0` paths, compression/tabulation paths, or standalone op users where padded env-mat values can affect results when `davg` is nonzero?
4. Should we add explicit tests asserting CPU/GPU/`dpmodel` consistency for padded slots with nonzero `avg`?

My current reading is that the #5582 GPU change makes the low-level TF GPU env-mat behavior consistent with `dpmodel` and with uniform normalization. I would like to confirm that this is the intended public semantics before relying on it more broadly.

njzjz:
## Summary
- batch CPU and GPU prod env mat over frames for non-LAMMPS neighbor lists
- support no-PBC self-built nlists, manual PBC coord copy, and mesh-provided nlists
- keep LAMMPS external nlists on the existing per-sample path
- add CPU/GPU multiple-frame neighbor-list and env-mat coverage

Fixes #2618

## Tests
- cmake --build source/build-codex-cuda --target runUnitTests_lib deepmd_op -j2
- source/build-codex-cuda/lib/tests/runUnitTests_lib --gtest_filter=TestNeighborList.cpu:TestNeighborList.cpu_multiple_frames:TestNeighborList.cpu_lessmem:TestEnvMatA.prod_cpu:TestEnvMatA.prod_cpu_equal_cpu:TestEnvMatA.prod_cpu_multiple_frames:TestEnvMatR.prod_cpu:TestEnvMatR.prod_cpu_equal_cpu:TestEnvMatR.prod_cpu_multiple_frames:TestEnvMatAMix.prod_cpu:TestEnvMatAMix.prod_cpu_equal_cpu
- srun --gres=gpu:1 source/build-codex-cuda/lib/tests/runUnitTests_lib --gtest_filter=TestEnvMatA.prod_gpu_multiple_frames:TestNeighborList.gpu_multiple_frames:TestEnvMatA.prod_gpu:TestEnvMatA.prod_gpu_equal_cpu:TestEnvMatR.prod_gpu:TestEnvMatR.prod_gpu_equal_cpu:TestFormatNlist.gpu:TestFormatNlistShortSel.gpu
- pytest source/tests/tf/test_prod_env_mat.py::TestProdEnvMat::test_pbc_self_built_nlist source/tests/tf/test_prod_env_mat.py::TestProdEnvMat::test_nopbc_self_built_nlist source/tests/tf/test_prod_env_mat.py::TestProdEnvMat::test_pbc_self_built_nlist_deriv source/tests/tf/test_prod_env_mat.py::TestProdEnvMat::test_nopbc_self_built_nlist_deriv -q

## Summary by CodeRabbit

* **New Features**
  * Neighbor-list, neighbor formatting, and environment-matrix operations now support multi-frame processing in a single call via an `nframes` argument (CPU and GPU).
  * Neighbor-list construction can optionally filter candidates using a per-atom type array.
* **Bug Fixes**
  * Invalid/filtered center or neighbor types are handled consistently (including zeroing `em`/`em_deriv` where applicable).
  * Improved GPU robustness with safer temporary-memory management and clearer error handling.
* **Tests**
  * Added/extended CPU and GPU tests to validate multi-frame neighbor-list and environment-matrix behavior, including negative-type cases.


source/lib/tests/test_env_mat_a.cc
```diff
@@ -671,6 +764,132 @@ TEST_F(TestEnvMatA, prod_gpu) {
...
+  deepmd::delete_device_memory(memory_dev);
+  deepmd::free_nlist_gpu_device(gpu_inlist);
+
+  for (int ff = 0; ff < nframes; ++ff) {
+    for (int ii = 0; ii < nloc; ++ii) {
+      for (int jj = 0; jj < nnei; ++jj) {
+        for (int dd = 0; dd < 4; ++dd) {
+          const int_64 idx =
+              (static_cast<int_64>(ff) * nloc + ii) * nnei * 4 + jj * 4 + dd;
+          EXPECT_LT(fabs(em[idx] - expected_multi[idx]), 1e-5);
```
wanghan-iapcm: Two items on this batched-frames change: a test tolerance and a doc gap.

**1. GPU multi-frame test tolerance is too loose (`1e-5`).** `prod_gpu_multiple_frames` compares the batched GPU `em` against a per-frame fp64 CPU reference (`prod_env_mat_a_cpu`) at `1e-5`. But `prod_gpu_equal_cpu` in this same file already compares GPU-vs-CPU `em`/`em_deriv`/`rij` at `1e-10` and passes, and CPU `prod_cpu_multiple_frames` uses `1e-10`. Both sides run the same fp64 algorithm, so batching over frames gives no numerical reason to relax 5 orders of magnitude — the `1e-5` looks inherited from the golden-literal `prod_gpu` test, whose hardcoded expected values only carry ~6 digits. A loose bound can hide a per-frame offset/stride bug in the batched path, which is the exact regression this test exists to catch. Suggest `1e-10` here and in `test_env_mat_a_mix.cc::prod_gpu_multiple_frames`.

**2. New `build_nlist_*` params are undocumented.** `build_nlist_cpu`/`build_nlist_gpu` gained `const int& nframes = 1` and `const int* type = nullptr`, but the `// inputs` block above each declaration is unchanged — it still lists a nonexistent `region` and omits both new params. `type` is load-bearing and non-obvious: it is a per-atom type array indexed `type[frame_idx * nall + atom]` (stride `nall`, so ghost atoms need entries too), and any pair whose center or neighbor has `type < 0` is excluded from the built list; `nullptr` disables the filter (legacy behavior). Worth documenting the indexing convention and the `<0` semantics so a caller doesn't pass an `nloc`-length array.

https://github.com/deepmodeling/deepmd-kit/blob/048376baa460d0082c3fbfef1a31c244d3f1a727/source/lib/include/neighbor_list.h#L144-L164

source/lib/tests/test_env_mat_r.cc
```diff
@@ -358,6 +393,89 @@ TEST_F(TestEnvMatR, prod_cpu_equal_cpu) {
   }
 }
 
+TEST_F(TestEnvMatR, prod_cpu_multiple_frames) {
```
wanghan-iapcm, line 397: Non-blocking, test coverage for the batched GPU path:

- **No GPU multi-frame test for the R descriptor.** `compute_env_mat_r` received the same `nframes` frame-indexing rewrite as the A kernel, but `test_env_mat_r.cc` adds only CPU multi-frame coverage (this test) — `test_env_mat_a.cc` and `test_env_mat_a_mix.cc` each add a `prod_gpu_multiple_frames`, while R does not. The batched GPU R path is currently unexercised.
- **The A / A-mix GPU multi-frame tests assert only on `em`.** They copy back only `em` ([test_env_mat_a.cc#L867](https://github.com/deepmodeling/deepmd-kit/blob/048376baa460d0082c3fbfef1a31c244d3f1a727/source/lib/tests/test_env_mat_a.cc#L866-L869)) and never validate `em_deriv` or `rij` under batching, yet those use distinct frame-offset arithmetic (`ff*nloc*ndescrpt*3`, `bid*nnei*3`) from `em`. The CPU multi-frame tests check all three at `1e-10`; the GPU ones should too.

njzjz: Addressed in 0c4e30d69.

- Added `TestEnvMatR.prod_gpu_multiple_frames` for the batched GPU R path.
- Extended the A and A-mix GPU multi-frame tests to copy back and validate `em_deriv`, `rij`, and `nlist` against per-frame CPU results, with `1e-10` checks for floating outputs.
- Also aligned the A-mix CPU multi-frame test to validate the same outputs.

Validation: `ruff check .`, `ruff format .`, TensorFlow-only C++ build of `runUnitTests_lib`, and `TestEnvMatA.*:TestEnvMatAMix.*:TestEnvMatR.*` all passed locally. GPU tests were not runnable locally because this machine has no CUDA toolkit.

njzjz: Supplementing the previous reply with the point-by-point mapping to the original review:

1. R descriptor GPU multi-frame coverage is now added in `source/lib/tests/test_env_mat_r.cc` as `TestEnvMatR.prod_gpu_multiple_frames`. It builds a 2-frame input, perturbs the second frame, computes per-frame CPU references, runs `prod_env_mat_r_gpu(..., nframes, ...)`, and compares all batched outputs.

2. A / A-mix GPU multi-frame coverage now validates more than `em`:
   - `source/lib/tests/test_env_mat_a.cc::TestEnvMatA.prod_gpu_multiple_frames` now copies back `em_deriv`, `rij`, and `nlist` in addition to `em`; it checks `em`, `em_deriv`, and `rij` against per-frame CPU references at `1e-10`, and checks `nlist` exactly.
   - `source/lib/tests/test_env_mat_a_mix.cc::TestEnvMatAMix.prod_gpu_multiple_frames` has the same coverage for `em`, `em_deriv`, `rij`, and `nlist`.

3. The expected buffers are filled using the same frame offsets that the GPU path needs to get right: `ff * nloc * ndescrpt` for `em`, `ff * nloc * ndescrpt * 3` for `em_deriv`, and `ff * nloc * nnei * 3` for `rij`. This specifically covers the offset arithmetic called out in the review.

4. I also extended `TestEnvMatAMix.prod_cpu_multiple_frames` to compare `em_deriv`, `rij`, and `nlist`, so CPU and GPU multi-frame tests now check the same output set.

Validation done locally: `ruff check .`, `ruff format .`, TensorFlow-only C++ build of `runUnitTests_lib`, and `./runUnitTests_lib --gtest_filter='TestEnvMatA.*:TestEnvMatAMix.*:TestEnvMatR.*'` passed. I could not execute the GPU-only tests locally because this machine does not have the CUDA toolkit; they are covered by the added GPU test code and should run in the GPU CI environment.

source/lib/include/neighbor_list.h
```diff
@@ -159,7 +159,9 @@ int build_nlist_cpu(InputNlist& nlist,
                     const int& nloc,
                     const int& nall,
                     const int& mem_size,
-                    const float& rcut);
+                    const float& rcut,
+                    const int& nframes = 1,
+                    const int* type = nullptr);
```
wanghan-iapcm, line 175: Follow-up on my earlier review (the doc-gap half of the tolerance/doc comment — the test-tolerance and R-GPU-test parts are now addressed, thanks).

The `// inputs` block just above this declaration (and the matching one above `build_nlist_gpu`) is still stale:

```
// inputs
//	c_cpy, nloc, nall, mem_size, rcut, region
```

It lists `region`, which is not a parameter, and omits the two new params added here: `nframes` and `type`. Worth documenting, especially `type`, which is load-bearing and non-obvious:

- it is a per-atom type array indexed `type[frame_idx * nall + atom]` (stride `nall`, so ghost atoms need entries too), and
- any pair whose center or neighbor has `type < 0` is excluded from the built list; `nullptr`/`NULL` disables the filter (legacy behavior).

Documenting the indexing convention and the `< 0` semantics would stop a caller from passing an `nloc`-length array.

njzjz: Addressed in d6ea9eb5d9639ae971c2e07d3eabbe031879973c (`docs(lib): document batched neighbor list inputs`).

The comments above both `build_nlist_cpu` and `build_nlist_gpu` now:

- remove the stale `region` entry,
- list the new `nframes` and `type` parameters,
- document that `c_cpy` is frame-major with `nall` atoms per frame, including ghosts,
- document that `type` uses `type[frame_idx * nall + atom]`,
- document that `nullptr`/`NULL` disables filtering, and
- document that pairs with center or neighbor `type < 0` are excluded.

I also documented the GPU `nlist_data` workspace size and corrected the return-condition notes for the CPU/GPU helpers.

iProzd (approved):

wanghan-iapcm (approved):

PR #5583, opened 2026-06-25
feat(pt_expt): dpa1(attn_layer=0) graph-native NeighborGraph forward

wanghan-iapcm:
## Summary

Adds the graph-native forward path for `dpa1(attn_layer=0)` (the factorizable, mixed-types case), built on the `NeighborGraph` foundation from #5581. Geometry enters the descriptor only through per-edge `edge_vec`; the neighbor-axis reduction becomes a `segment_sum` over edge centers. For `pt_expt` this becomes the **default** forward (force/virial via a single autograd backward through `edge_vec`).

## What it adds

- **dpmodel**: `edge_env_mat` (per-edge env-mat 4-vector), `DescrptBlockSeAtten._call_graph` + `DescrptDPA1.call_graph`, model `call_lower_graph` (energy), `neighbor_graph_from_ijs` + an optional **ASE** O(N) carry-all builder.
- **pt_expt**: `edge_energy_deriv` (autograd `grad(E, edge_vec)` → `edge_force_virial`) + `forward_common_lower_graph` (energy + force + virial + atom_virial).
- The dense `DescrptDPA1.call` becomes a thin adapter (`from_dense_quartet → call_graph`) preserving the 5-tuple ABI; a **shape-static** converter keeps it `jax.jit` / `torch.export`-traceable.

## Default behavior

- **pt_expt** defaults graph-eligible `dpa1(attn_layer=0, concat tebd, no exclude_types)` models to the carry-all graph (it has the autograd force/virial path).
- **dpmodel/jax** keep the dense default (they compute force/virial analytically; the graph lower is energy-only), and **agree with pt_expt at non-binding `sel`**.
- Ineligible configs (attention, strip tebd, `exclude_types`, linear/ZBL) fall back to the dense path unchanged. `neighbor_graph_method="legacy"` forces dense; `"dense"`/`"ase"` force the graph.

## Parity (graph vs legacy dense lower, fp64 CPU)

| | energy | force | virial | atom_virial |
|---|---|---|---|---|
| max abs diff | 0 | ~1e-19 | ~1e-18 | **~1e-18** |

atom_virial matches the canonical TF==pt-legacy full-to-src convention. dpa1 descriptor + model consistency suites green across dp/jax/pt_expt.

## Known limitations

- Default-flip is **pt_expt-only**; full carry-all default for dp/jax needs analytical/jax graph force (follow-up).
- `make_fx` (forward + grad) traces; **full `.pt2` AOTI export is a follow-up** (PR-B). The carry-all builders (`build_neighbor_graph`/`from_ijs`) still use `nonzero` (eager-only); their static variants land with the export PR.
- Single-rank only; CUDA unvalidated (CPU box); ASE is opt-in O(N) (vesin O(N) is a follow-up); no jax graph force / dpa2-3 message-passing yet.

Also folds in three follow-up fixes to the #5581 foundation from @OutisLi's review (dangling spec refs → design discussion, `edge_force_virial` jax int-sum short-circuit, `Array` typing).

## Summary by CodeRabbit

* **New Features**
  * Added graph-native “lowering” for DPA1 when compatible, including graph-native descriptor/forward execution and graph-native descriptor→model output conversion.
  * Introduced opt-in `neighbor_graph_method` routing for energy/force/virial, with carry-all neighbor graphs and graph-output fitting/post-processing.
  * Added new neighbor-graph utilities (including ASE-based carry-all building, `(i,j,S)` conversion, and per-edge environment-matrix computation), exported as part of the public API.
* **Bug Fixes**
  * Improved stability for masked/padded edges, virtual atom handling, and parameter protection consistency; refined traced virial assembly when node-capacity is used.
* **Tests**
  * Expanded parity/regression suites for graph lowering, energy/force/virial, conversion correctness, ragged graphs, and FX tracing.


deepmd/dpmodel/model/make_model.py
```diff
@@ -285,6 +294,17 @@ def call_common(
...
+                Opt-in CARRY-ALL graph energy forward (Option B). ``None``
+                (default) keeps the existing dense nlist path UNCHANGED. When
+                set to ``"dense"`` (in-tree all-pairs search) or ``"ase"``
+                (O(N) ASE cell list), the model builds a carry-all
+                :class:`NeighborGraph` and routes the ENERGY forward through
+                :meth:`call_lower_graph`. Requires a ``mixed_types`` descriptor
+                with a graph lower (dpa1 ``attn_layer == 0``). At non-binding
+                ``sel`` this matches the dense path exactly; at binding ``sel``
+                the carry-all graph keeps neighbors the dense path truncates, so
+                energy intentionally differs.
```
iProzd, line 320: `neighbor_graph_method="dense"` is a bit ambiguous, since it means the graph all-pairs carry-all builder rather than the legacy dense-nlist lower; could we rename it to something like `"graph_dense"` / `"all_pairs"` or document this explicitly?

iProzd, line 320: `legacy` option is missed. Could options be listed one per line?

deepmd/dpmodel/utils/neighbor_graph/builder.py
```diff
@@ -85,50 +86,112 @@ def from_dense_quartet(
...
         edge-axis length policy; ``None`` => dynamic (torch) with ``min_edges`` guards.
+    compact
+        If True (default), COMPACT real edges with ``nonzero`` and pad/guard via
+        :func:`pad_and_guard_edges` -- the data-dependent output shape breaks
+        jax.jit / torch.export. If False, emit a SHAPE-STATIC graph: every nlist
+        slot becomes an edge (``E = nf * nloc * nsel``, a static shape), invalid
+        slots (``nlist == -1``) get ``edge_mask=False``, zero ``edge_vec`` and a
+        ``src`` pointing at the center (in-range, masked) -- so no ``nonzero`` is
+        used and the converter is jit/export-traceable. The masked edges contribute
+        zero in a downstream ``segment_sum``, so the descriptor output is unchanged.
```
iProzd, line 100: It would help to document that `from_dense_quartet(...)` is an internal legacy-quartet-to-graph adapter that preserves legacy `sel` truncation, while `"dense"` / `"ase"` build carry-all graphs from raw coordinates.

deepmd/dpmodel/model/make_model.py
```diff
@@ -297,23 +317,157 @@ def call_common(
...
+                    model_output_def=self.model_output_def(),
+                    coord=cc,
+                    atype=atype,
+                    box=bb,
+                    fparam=fp,
+                    aparam=ap,
+                    do_atomic_virial=do_atomic_virial,
+                    coord_corr_for_virial=coord_corr_for_virial,
+                    charge_spin=cs,
+                    neighbor_list=neighbor_list,
```
iProzd, line 377: If a user passes an explicit `neighbor_list`, the pt_expt graph default should not silently ignore it; please either keep the legacy route or raise on conflicting `neighbor_list` + graph settings.

deepmd/dpmodel/utils/neighbor_graph/graph.py
```diff
@@ -102,11 +123,48 @@ def pad_and_guard_edges(
...
+
+    Returns
+    -------
+    frame_id
+        Frame index of each flat node, compact-prefix frame-major.
+        Shape ``(N,)`` int64, where ``N = sum(n_node)``.
+    """
+    xp = array_api_compat.array_namespace(n_node)
+    dev = array_api_compat.device(n_node)
+    n_total = int(xp.sum(n_node))
```
iProzd: `frame_id_from_n_node()` still does `int(xp.sum(n_node))`, which is not trace-friendly for JAX/export; could it accept a static `n_total` / capacity instead?

deepmd/dpmodel/atomic_model/dp_atomic_model.py
```diff
@@ -248,6 +254,56 @@ def forward_atomic(
         )
         return ret
 
+    def forward_atomic_graph(
+        self,
+        graph: "NeighborGraph",
+        atype: Array,
+        fparam: Array | None = None,
+        aparam: Array | None = None,
+    ) -> dict[str, Array]:
```
iProzd, line 264: Even though this DPA1 graph path does not use `charge_spin` yet, it may be better to include it in the `forward_atomic_graph()` / graph-lower interface from the beginning so the API does not need to change later for charge/spin-conditioned descriptors.

iProzd (changes requested):

deepmd/dpmodel/atomic_model/dp_atomic_model.py
OutisLi: +1

deepmd/dpmodel/model/make_model.py
wanghan-iapcm: Documented explicitly rather than renamed (the option strings are a locked design term, and renaming would churn the API/tests). The `neighbor_graph_method` docstring now lists each option one-per-line and states that `"dense"`/`"ase"` are the carry-all **graph** builders (in-tree all-pairs / ASE cell list, from raw coordinates) — NOT the dense-nlist lower. (5271edb0)

wanghan-iapcm: Done — options are now listed one per line, including `"legacy"`: `None` (default), `"legacy"` (force dense), `"dense"` (all-pairs carry-all graph), `"ase"` (O(N) ASE carry-all graph). (5271edb0)

deepmd/dpmodel/utils/neighbor_graph/builder.py
wanghan-iapcm: Done — `from_dense_quartet` now documents that it is the legacy-quartet→graph CONVERTER that performs NO search and INHERITS the `sel` truncation already baked into `nlist`, and explicitly contrasts it with `build_neighbor_graph` (the `"dense"` all-pairs route) and `build_neighbor_graph_ase` (the `"ase"` route), which search from RAW coordinates and apply NO `sel` truncation. (5271edb0)

deepmd/dpmodel/model/make_model.py
wanghan-iapcm: Fixed — `call_common` no longer lets the graph default silently ignore `neighbor_list`. If an explicit `neighbor_list` is combined with an explicit `neighbor_graph_method`, it raises `ValueError`; otherwise the supplied `neighbor_list` is honored by taking the dense route (`graph_method = None`), so the pt_expt default-flip can't drop it. Tests added: conflict-raise + dense-route fallback. (5271edb0)

deepmd/dpmodel/utils/neighbor_graph/graph.py
wanghan-iapcm: Fixed — `frame_id_from_n_node` now takes an optional static `n_total` (mirroring `node_validity_mask`). When supplied it avoids `int(xp.sum(n_node))`, keeping the function jit/export-traceable; padding nodes `[sum(n_node), n_total)` are clamped to the last frame so a padded node axis stays in range for `segment_sum`. The default (`None`) keeps the eager `int(sum)` behavior. Test added (exact + padded). (5271edb0)

deepmd/dpmodel/atomic_model/dp_atomic_model.py
wanghan-iapcm: Threaded `charge_spin` through the graph interface (`forward_atomic_graph` → `forward_common_atomic_graph` → `call_common_lower_graph` / `forward_common_lower_graph`) as accept-for-ABI-stability, matching the existing `comm_dict`/`n_local` pattern, so the signatures won't need to change for charge/spin-conditioned descriptors. (5271edb0)

To be precise on *consumption*: dpa1 does NOT consume `charge_spin` (`get_dim_chg_spin() == 0`, and the dense atomic model already passes `None` to the dpa1 descriptor since `add_chg_spin_ebd` is False). Actual charge/spin embedding is a dpa3/dpa4 feature; wiring its consumption into the graph lower (with parity vs the dense embedding) belongs to PR-G. For now I added an invariance test pinning that the dpa1 graph-lower output is unchanged by `charge_spin` (dpmodel energy/mask; pt_expt energy/force/virial/atom-virial) — combined with the existing non-binding-`sel` `graph == dense` parity this gives `graph(charge_spin) == graph(None) == dense`. (83583a31)

OutisLi (approved):

iProzd (approved):

PR #5590, opened 2026-06-25
feat(jax): add JAX-MD interface

njzjz:
## Summary
- add a JAX-MD adapter for DeePMD JAX models, including energy, force, neighbor-list, and `as_jax_md` helpers
- add focused JAX tests for energy/force evaluation, dense neighbor conversion, and real JAX-MD neighbor lists
- add a water JAX-MD smoke example using dpdata to read the LAMMPS data file
- document the JAX-MD interface in the third-party integration docs

## Tests
- `ruff check .`
- `PYTHONPATH=. pytest source/tests/jax/test_jax_md.py -q`
- JAX-MD water smoke run on GPU

## Summary by CodeRabbit

* **New Features**
  * Added JAX-MD support for DeePMD JAX models, including energy, force, neighbor-list, and model-loading utilities.
  * Added a new water simulation example that runs a short JAX-MD NVE smoke test from the command line.

* **Documentation**
  * Added setup and usage guidance for the JAX-MD integration.
  * Included a new third-party package entry for JAX-MD in the docs.


OutisLi (commented):
Perhaps `jax_md.py` should not be placed at the root folder of jax backend

njzjz:
> Perhaps `jax_md.py` should not be placed at the root folder of jax backend

@OutisLi which folder do you suggest?

deepmd/jax/jax_md.py
```diff
@@ -0,0 +1,423 @@
...
+        axis=0,
+    )[None, ...]
+    return extended_coord, extended_atype, lower_nlist[None, ...], mapping
+
+
+def _extract_energy(ret: Any) -> Array:
+    """Extract a scalar total energy from a DeePMD model return value."""
+    if isinstance(ret, tuple):
+        ret = ret[0]
+    for key in ("energy", "energy_redu"):
```
wanghan-iapcm: For `.hlo` models this returns the first atom's atomic energy, not the total energy.

The key order `("energy", "energy_redu")` only works for `.jax` checkpoints. `EnergyModel.call`/`call_lower` remap the output so `"energy"` is the reduced total and there is no `"energy_redu"` key:

https://github.com/deepmodeling/deepmd-kit/blob/158504866cf9ef9ddb44db036ecb8800d4b12e55/deepmd/dpmodel/model/ener_model.py#L159-L161

But `.hlo` models (an advertised input — `load_model` accepts `.hlo` and the docs list it) route through `model_call_from_call_lower` → `communicate_extended_output`, which emits **both** `"energy"` (per-atom, `nf x nloc x 1`) **and** `"energy_redu"` (total, `nf x 1`):

https://github.com/deepmodeling/deepmd-kit/blob/158504866cf9ef9ddb44db036ecb8800d4b12e55/deepmd/dpmodel/model/transform_output.py#L105-L111

Since the loop checks `"energy"` first, `jnp.ravel(ret["energy"])[0]` returns atom 0's atomic energy. Both the potential and the `-jax.grad` forces are then silently wrong for `.hlo` models (affects both the `neighbor=None` path via `HLO.call` and the neighbor path via `HLO.call_lower`).

Suggest preferring the reduced key:

```python
for key in ("energy_redu", "energy"):
```

This returns the total for `.hlo` (where `"energy_redu"` exists) and still falls through to `"energy"`=total for the `.jax`/mock case (where `"energy_redu"` is absent).

The current tests don't catch this: `HarmonicModel`/`EdgeModel` set `"energy"` to the total with no `"energy_redu"` key, so the dual-key `.hlo` case is never exercised.

OutisLi:
> > Perhaps `jax_md.py` should not be placed at the root folder of jax backend
> 
> @OutisLi which folder do you suggest?

Perhaps a third party folder or inference folder

deepmd/jax/jax_md.py
njzjz: Thanks, I re-checked the current PR state. `.hlo` is now rejected by `load_model()` with `NotImplementedError`, so the HLO dual-key path is not supported by this adapter.

For the supported `.jax` path, `EnergyModel.call()` and `EnergyModel.call_lower()` both translate the internal lower output so public `ret["energy"]` is the total energy (`energy_redu` is an internal/lower key). I updated `_extract_energy()` to only accept the public `energy` key instead of falling back between `energy` and `energy_redu`.

Fixed in c274f1dd4.

OutisLi (approved):

wanghan-iapcm (approved):

iProzd (changes requested):
I do not see a blocking issue.
One small documentation fix: `load_model()` now explicitly rejects `.hlo` with `NotImplementedError`, but `doc/third-party/jaxmd.md` still says `.hlo` is accepted. Please update that section to say that only `.jax` checkpoints (or an already constructed model object) are currently supported by the JAX-MD adapter, and that `.hlo` is not supported because the adapter needs JAX-differentiable energy functions.
LGTM after that doc tweak.

iProzd (approved):

PR #5592, opened 2026-06-26
feat(dpmodel): add descriptor compression

njzjz:
## Summary
- add dpmodel compression entrypoints and wire dp --dp/--jax compress
- persist and restore compressed dpmodel/JAX descriptor state, including HLO export metadata
- implement dpmodel compression for se_e2_a, se_e2_r, and se_atten/dpa1 (with se_atten_v2 state sync)
- add common dpmodel and JAX compression coverage

## Tests
- python -m pytest source/tests/common/dpmodel/test_model_compression.py -q
- python -m pytest source/tests/jax/test_model_compression.py -q
- python -m pytest source/tests/common/test_argument_parser.py -q
- python -m pytest source/tests/pt_expt/descriptor/test_se_r.py source/tests/pt_expt/descriptor/test_dpa1.py source/tests/pt_expt/descriptor/test_se_atten_v2.py -q
- python -m pytest source/tests/pt_expt/model/test_model_compression.py -q
- ruff check .
- ruff format --check .

## Summary by CodeRabbit

* **New Features**
  * Added model compression support for DPModel and JAX workflows.
  * Introduced a new `compress` command in the CLI for creating compressed model files.
  * Expanded support for compressed descriptors, including saving and restoring compressed models.

* **Documentation**
  * Updated CLI help text with new examples and supported model file formats.

* **Tests**
  * Added end-to-end and descriptor-level coverage for compression behavior and model reloads.


deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -1168,7 +1437,14 @@ def call(
...
+                if self.tebd_compress:
+                    tt_full = xp.asarray(
+                        self.type_embd_data[...],
+                        dtype=rr.dtype,
+                        device=array_api_compat.device(rr),
+                    )
+                else:
+                    tt_full = self.cal_g_strip(type_embedding, 0)
                 # (nf x nl x nnei) x ng
                 gg_t = xp_take_along_axis(tt_full, nei_type_index, axis=0)
```
wanghan-iapcm, line 1570: Non-blocking: the `type_one_side=True` path for DPA1 compression looks unexercised. This branch (and the matching one in `type_embedding_compression`) builds `tt_full` differently and gathers with `xp_take_along_axis(tt_full, nei_type_index, axis=0)`, i.e. a different type-embedding indexing than the `type_one_side=False` (default) case. Every DPA1 compression test in `source/tests/common/dpmodel/test_model_compression.py` and `source/tests/jax/test_model_compression.py` uses the default `type_one_side=False`, so this gather is never run. `se_e2_a` already parametrizes both `type_one_side` values — worth adding a `type_one_side=True` compressed-vs-uncompressed case for DPA1 too, since the indexing differs and an off-by-one there would silently produce wrong type embeddings.

njzjz: Addressed in b8a4c977a. I parameterized the DPA1 compression tests over `type_one_side=True` and `False` in both the common dpmodel and JAX suites, so the one-sided compressed type-embedding gather path is now compared against the uncompressed output and the serialized/reloaded compressed descriptor.

Validation:
- `python -m pytest source/tests/common/dpmodel/test_model_compression.py::TestDPModelCompression::test_se_atten_enable_compression -v`
- `python -m pytest source/tests/jax/test_model_compression.py::TestJAXModelCompression::test_se_atten_enable_compression -v`
- `ruff format .`
- `ruff check .`

deepmd/dpmodel/descriptor/se_e2_a.py
```diff
@@ -388,6 +391,182 @@ def cal_g(
...
+        self.compress_info = compress_info
+
+    def _tabulate_fusion_se_a(
+        self,
+        table: Array,
+        table_info: Array,
+        em_x: Array,
+        em: Array,
+        last_layer_size: int,
+    ) -> Array:
```
wanghan-iapcm, line 475: Non-blocking refactor suggestion: the three `_tabulate_fusion_se_{a,r,atten}` methods (here, `se_r._tabulate_fusion_se_r`, and `dpa1._tabulate_fusion_se_atten`) share ~70 lines that are byte-for-byte identical — the entire locator + interpolation + C1 extrapolation core: the nested `xp.where` `table_idx`, `dx`, `extrapolate_delta`, the `xp.take` of `coeff`, the Horner `values`/`values_grad`, and the final `values = values + values_grad * extrapolate_delta`.

The only real difference is the reduction tail:
- se_r: returns the per-pair `values`
- se_a: `xp.sum(em[:,:,:,None] * values[:,:,None,:], axis=1)`
- se_atten: `values = values * two_embed + values` then the same `xp.sum(...)`

Consider extracting the shared core into one helper that returns the per-pair `values` of shape `(nloc, nnei, last_layer_size)`, e.g. a module-level `tabulate_fusion(table, table_info, em_x, last_layer_size)` in `deepmd/dpmodel/utils/`, with each descriptor keeping only its reduction:

```python
# se_r:      return tabulate_fusion(...)
# se_a:      v = tabulate_fusion(...);                    return xp.sum(em[..., None] * v[:, :, None, :], axis=1)
# se_atten:  v = tabulate_fusion(...); v = v * two_embed + v;  return xp.sum(...)
```

This is the most error-prone part (the new C1 extrapolation that must track the C++ side in #5601), so keeping it single-source avoids three-way drift on any future correction. Unlike the C++ kernels (kept separate for fused reduction), the array-API path already computes `values` and reduces separately, so there's no performance reason to duplicate. A free function is cleaner than a base method since the three live in unrelated class hierarchies.

deepmd/dpmodel/descriptor/se_e2_a.py
```diff
@@ -513,6 +705,88 @@ def call(
...
+            sort_idx = xp.argsort(atype_loc)
+            unsort_idx = xp.argsort(sort_idx)
+            rr_s = xp.take(rr, sort_idx, axis=0)
+            mask_s = xp.take(exclude_mask, sort_idx, axis=0)
+            dev = array_api_compat.device(rr)
+            gr_s = xp.zeros([nfnl, 4, ng], dtype=rr.dtype, device=dev)
+            type_ends = []
+            offset = 0
+            for ti in range(self.ntypes):
+                offset += int(xp.sum(xp.astype(atype_loc == ti, xp.int32)))
```
wanghan-iapcm, line 683: Non-blocking (jax export): this `type_one_side=False` branch isn't jit/HLO-traceable. `int(xp.sum(...))` concretizes a traced value and the following `rr_s[s:e]` is a data-dependent slice, so under `jax.jit`/`jax_export.export` (the `.hlo` write path in `deepmd/jax/utils/serialization.py`) it raises `ConcretizationTypeError`. Since jax inherits this code (`DescrptSeA(DescrptSeADP)`), a compressed se_e2_a/dpa1 with the default `type_one_side=False` can't be exported to `.hlo`. This is the same de-facto limitation pt_expt has (its `type_one_side=False` compressed path uses boolean-mask indexing `rr[ti_mask, ...]`, equally non-exportable).

Two suggestions, consistent with pt_expt:

1. **Fail fast on `type_one_side=False` compressed jax export** with a clear message, rather than letting it die with a cryptic XLA `ConcretizationTypeError` deep in tracing. The eager dpmodel/numpy path can keep working; only the jit/HLO path needs to reject it.

2. **Extend HLO-export test coverage.** Today only `test_jax_compress_entrypoint_can_write_hlo` actually traces to StableHLO, and only for se_e2_a with `type_one_side=True`. `test_jax_compress_entrypoint` exercises dpa1 too but writes `.jax` (no tracing), so compressed **dpa1 and se_r HLO export are unverified**. Worth adding `.hlo` write cases for dpa1 + se_r, plus a negative test asserting `type_one_side=False` raises the clear error from (1).

The `type_one_side=True` path is fine — it traces (static loop over `ntypes`, unrolled into the graph) and is covered for se_e2_a.

deepmd/dpmodel/descriptor/se_e2_a.py
njzjz: Addressed in a6817154. I extracted the shared array-API spline lookup/interpolation/Horner/C1 extrapolation core into `deepmd/dpmodel/utils/tabulate.py::tabulate_fusion`, and kept only the descriptor-specific reductions in `se_r`, `se_e2_a`, and `dpa1`.

Validation:
- `venv/bin/python -m pytest source/tests/common/dpmodel/test_model_compression.py -v`
- `venv/bin/python -m pytest source/tests/jax/test_model_compression.py -v`
- `ruff format .`
- `ruff check .`

deepmd/dpmodel/descriptor/se_e2_a.py
njzjz: Addressed in a6817154. The `.hlo` serialization path now fails fast for compressed `se_e2_a`/`dpa1` descriptors with `type_one_side=False`, with a clear message explaining that the compressed path uses data-dependent type slices and suggesting `type_one_side=True` or `.jax` output.

I also extended HLO coverage so compressed `se_e2_a(type_one_side=True)`, `se_r`, and `dpa1(type_one_side=True)` are exported, and added negative coverage for `se_e2_a`/`dpa1` with `type_one_side=False`.

Validation:
- `venv/bin/python -m pytest source/tests/jax/test_model_compression.py -v`
- `ruff check .`

wanghan-iapcm (approved):

PR #5599, opened 2026-06-28
refactor(dpmodel/dpa4)

OutisLi:
This is first PR of porting the whole dpa4 descriptor

## Summary by CodeRabbit

* **New Features**
  * Added Cartesian tensor-product mixing options for SeZM/DPA4 descriptors (`edge_cartesian`, `node_cartesian`) with `mixing_layers` control.
  * Added LoRA fine-tuning support (adapter injection, merging, and LoRA-free checkpoint export).
  * Added opt-in performance acceleration (Triton when available) and eval-time activation checkpointing for the interaction block.

* **Bug Fixes**
  * Improved destination-wise edge aggregation/softmax and sparse-edge masking; more robust serialization/deserialization round-trips and dtype/precision restoration.

* **API Changes**
  * Renamed SeZM SO(2) parameter: `so2_layers` → `mixing_layers` (kept as alias).


deepmd/dpmodel/descriptor/dpa4.py
```diff
@@ -1232,8 +2327,10 @@ def serialize(self) -> dict[str, Any]:
                 "radial_mlp": self.radial_mlp,
                 "use_env_seed": self.use_env_seed,
                 "random_gamma": self.random_gamma,
+                "edge_cartesian": self.edge_cartesian,
+                "node_cartesian": self.node_cartesian,
                 "so2_norm": self.so2_norm,
-                "so2_layers": self.so2_layers,
+                "mixing_layers": self.mixing_layers,
```
wanghan-iapcm, line 2333: Non-blocking (serialization discipline): `serialize()` gains `edge_cartesian`/`node_cartesian` here, and the serialized key `so2_layers` was renamed to `mixing_layers`, but `@version` / `LATEST_VERSION` wasn't bumped (still `1.1` at line 434, and `deserialize` still does `check_version_compatibility(version, cls.LATEST_VERSION, 1)`).

In practice this is harmless today — DPA4 is unreleased, `deserialize` does `cls(**config)`, so old configs fill the new fields via constructor defaults + the `so2_layers` alias, and new configs round-trip. But the repo's convention is to bump `@version` and add the matching `deserialize` back-compat in the same change whenever serialized fields are added or renamed. Worth bumping now while it's free, so the first released version has a clean version boundary for these fields.

deepmd/dpmodel/descriptor/dpa4.py
```diff
@@ -1048,14 +2058,38 @@ def dim_emb(self) -> int:
...
+
+        resume
+            Unused for SeZM; kept for interface compatibility.
+
+        Raises
+        ------
+        NotImplementedError
+            If ``shared_level`` is not ``0`` or ``1``.
+        """
+        raise NotImplementedError("share_params is not yet implemented for DescrptDPA4")
```
wanghan-iapcm, line 2092: Non-blocking (docs): this `share_params` docstring (copied from the pt `sezm.py`) documents working `shared_level == 0` / `shared_level == 1` behavior and a conditional `Raises NotImplementedError — If shared_level is not 0 or 1`, but the body unconditionally raises `NotImplementedError` for every input. So the documented level-0/level-1 semantics don't actually exist, and the `Raises` clause is inaccurate (it raises for all inputs, not just invalid levels).

Since pt_expt inherits this dpmodel method, DPA4 multitask parameter-sharing is effectively unimplemented across dpmodel/jax/pt_expt. The fail-fast itself is fine — but the docstring should be trimmed to state plainly that it's not implemented (a known limitation), rather than describe behavior that isn't there.

deepmd/dpmodel/descriptor/dpa4.py
OutisLi: no need

deepmd/dpmodel/descriptor/dpa4.py
OutisLi: will handle in the next pr

wanghan-iapcm (approved):

PR #5601, opened 2026-06-28
fix(tabulate): add c1 extrapolation outside tables

### LINKED ISSUE #5593: Tabulate grad kernels ignore clamp derivative outside table data range
## Summary

The tabulation code clamps lookup coordinates outside the tabulated data range, but the gradient implementations still differentiate the endpoint polynomial as if the lookup coordinate depended on the original input. This can produce non-zero `grad` / `grad_grad` contributions from the table interpolation outside the data range.

This appears to affect both CPU and GPU implementations.

## Why this looks inconsistent

For `se_a` / `se_r`, the locator clamps out-of-range inputs by selecting the first/last table segment and resetting the local coordinate to `0`:

```cpp
if (xx < lower) {
  table_idx = 0;
  xx = (FPTYPE)0.;
...
} else {
  table_idx = int((upper - lower) / stride0) +
              (int)((max - upper) / stride1) - 1;
  xx = (FPTYPE)0.;
}
```

CPU: [`source/lib/src/tabulate.cc#L20-L40`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/tabulate.cc#L20-L40)
GPU: [`source/lib/src/gpu/tabulate.cu#L51-L72`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/gpu/tabulate.cu#L51-L72)

`se_t` and `se_t_tebd` do the same kind of clamp for the lower/upper outside regions:

CPU: [`tabulate.cc#L45-L73`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/tabulate.cc#L45-L73)
GPU: [`tabulate.cu#L76-L104`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/gpu/tabulate.cu#L76-L104)

Once `xx` is set to `0` in these branches, the table lookup value is locally constant with respect to the original out-of-range input coordinate, i.e. the chain-rule factor `d(clamped_xx) / d(original_xx)` should be `0` outside the tabulated range.

However, the gradient code computes the polynomial derivative at this clamped local coordinate and uses it directly.

### CPU `se_a` grad

After `locate_xx(...)` may clamp `xx` to `0`, `g` is still used as the derivative with respect to the original input:

```cpp
locate_xx(lower, upper, _max, stride0, stride1, xx, table_idx);
...
FPTYPE g =
    (a1 + (2 * a2 + (3 * a3 + (4 * a4 + 5 * a5 * xx) * xx) * xx) * xx);
...
grad += g * dotllrr;
...
dy_dem_x[ii * nnei + jj] = grad;
```

[`tabulate.cc#L202-L253`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/tabulate.cc#L202-L253)

The second derivative path has the same pattern via `var_grad * dz_xx`:

[`tabulate.cc#L305-L375`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/tabulate.cc#L305-L375)

### GPU `se_a` grad

The CUDA/HIP kernel follows the same pattern:

```cpp
locate_xx_se_a(xx, table_idx, lower, upper, max, stride0, stride1);
...
Csub +=
    (nnei - breakpoint) *
    (var[1] + ((FPTYPE)2. * var[2] +
               ((FPTYPE)3. * var[3] +
                ((FPTYPE)4. * var[4] + (FPTYPE)5. * var[5] * xx) * xx) * xx) * xx) *
    (enable_se_atten ? res * t + res : res);
...
dy_dem_x[block_idx * nnei + ii] = Csub;
```

[`tabulate.cu#L337-L388`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/gpu/tabulate.cu#L337-L388)

The GPU `grad_grad` path similarly uses `res_grad * dz_xx` after the locator may have clamped `xx`:

[`tabulate.cu#L439-L492`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/gpu/tabulate.cu#L439-L492)

### Other affected paths

The same issue seems to appear in the other tabulate variants:

- CPU `se_t_grad_cpu`: [`tabulate.cc#L430-L482`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/tabulate.cc#L430-L482)
- CPU `se_t_grad_grad_cpu`: [`tabulate.cc#L489-L537`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/tabulate.cc#L489-L537)
- CPU `se_t_tebd_grad_cpu`: [`tabulate.cc#L593-L637`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/tabulate.cc#L593-L637)
- CPU `se_t_tebd_grad_grad_cpu`: [`tabulate.cc#L644-L687`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/tabulate.cc#L644-L687)
- CPU `se_r_grad_cpu`: [`tabulate.cc#L731-L762`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/tabulate.cc#L731-L762)
- CPU `se_r_grad_grad_cpu`: [`tabulate.cc#L775-L811`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/tabulate.cc#L775-L811)
- GPU `se_t_grad`: [`tabulate.cu#L575-L606`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/gpu/tabulate.cu#L575-L606)
- GPU `se_t_grad_grad`: [`tabulate.cu#L635-L658`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/gpu/tabulate.cu#L635-L658)
- GPU `se_t_tebd_grad`: [`tabulate.cu#L750-L783`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/gpu/tabulate.cu#L750-L783)
- GPU `se_t_tebd_grad_grad`: [`tabulate.cu#L820-L840`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/gpu/tabulate.cu#L820-L840)
- GPU `se_r_grad`: [`tabulate.cu#L905-L927`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/gpu/tabulate.cu#L905-L927)
- GPU `se_r_grad_grad`: [`tabulate.cu#L955-L970`](https://github.com/deepmodeling/deepmd-kit/blob/330fa7558efb4325320f15a2f8504b9d6c6508db/source/lib/src/gpu/tabulate.cu#L955-L970)

For `se_t`-style code, there may be an explicit original-coordinate factor such as `tmp * res(clamped_xx)`. The explicit `tmp` derivative should still be kept if that is the intended model. The suspicious part is specifically the interpolation derivative `dres/dxx`, which should be multiplied by the derivative of the locator/clamp mapping.

## Expected behavior

If out-of-range inputs are intentionally clamped by the locator, then the derivative contribution from the table interpolation should be masked outside the data range:

```cpp
orig_xx < lower_or_min || orig_xx >= max  =>  d(clamped_xx) / d(orig_xx) = 0
inside range                              =>  d(clamped_xx) / d(orig_xx) = 1
```

So terms like `g`, `var_grad`, `res_grad`, and `dres_dxx` should be multiplied by this factor when they are used as derivatives with respect to the original input coordinate.

## Suggested fix

One possible fix is to make the locator return an additional chain-rule factor, for example:

```cpp
FPTYPE dxx_dorig = 1;
if (xx < lower) {
  table_idx = 0;
  xx = 0;
  dxx_dorig = 0;
} else if (...) {
  ...
} else {
  table_idx = last_table_idx;
  xx = 0;
  dxx_dorig = 0;
}
```

Then multiply only the table-interpolation derivative by `dxx_dorig`:

```cpp
dy_dem_x = ... * polynomial_derivative(xx) * dxx_dorig;
```

For `se_t` variants, if the forward expression contains both an explicit original-coordinate factor and a table lookup factor, only the table lookup derivative should be masked; the explicit-coordinate derivative should remain unchanged.

A small regression test could set `em_x` below `lower` and above `max` with non-zero first/last polynomial slopes and verify that the interpolation-related gradient is zero outside the table range, while CPU/GPU remain consistent.

---

Authored by OpenClaw (model: custom-chat-jinzhezeng-group/gpt-5.5)

njzjz:
Fixes #5593.

This implements the C1 linear extrapolation suggested in https://github.com/deepmodeling/deepmd-kit/issues/5593#issuecomment-4826255625:

- keep the original out-of-range coordinate and evaluate the endpoint value/slope at the proper boundary offset
- linearly extend table values outside the covered range while keeping gradients at the constant boundary slope
- apply the same behavior to se_a, se_r, se_t, and se_t_tebd on CPU and GPU paths
- add focused lib tests for lower/high tails, boundary continuity, finite-difference force consistency, and zero tail curvature

Validation:

- ruff check .
- ruff format .
- cmake --build source/build --target runUnitTests_lib -j2
- source/build/lib/tests/runUnitTests_lib --gtest_filter='TestTabulateSeA.tabulate_fusion_se_a_cpu:TestTabulateSeA.tabulate_fusion_se_a_grad_cpu:TestTabulateSeR.tabulate_fusion_se_r_cpu:TestTabulateSeR.tabulate_fusion_se_r_grad_cpu:TestTabulateSeT.tabulate_fusion_se_t_cpu:TestTabulateSeT.tabulate_fusion_se_t_grad_cpu:TestTabulateSeTTebd.tabulate_fusion_se_t_tebd_cpu:TestTabulateSeTTebd.tabulate_fusion_se_t_tebd_grad_cpu:TabulateExtrapolate.*'

Note: I used the existing source/build directory reconfigured as TensorFlow-only locally because the previous cache had ENABLE_PYTORCH=ON with a CUDA PyTorch install but no CUDA toolkit available.

## Summary by CodeRabbit

* **New Features**
  * Tabulated SE polynomial evaluations now use an extrapolation-aware form for out-of-range inputs, improving smoothness and continuity at both lower and upper table boundaries across SE-a, SE-t, SE-t-tebd, and SE-r.

* **Bug Fixes**
  * Corrected boundary/tail handling so value and gradient computations remain consistent near min/max, including off-grid max behavior; gradients now incorporate extrapolated results where needed.

* **Tests**
  * Added CPU and GPU test coverage validating value/gradient/grad-grad consistency via polynomial reference comparisons and finite-difference checks for all supported SE variants.

### DIFF
--- a/source/lib/tests/test_tabulate_extrapolate.cc
+++ b/source/lib/tests/test_tabulate_extrapolate.cc
@@ -549,6 +549,226 @@ double se_t_tebd_grad_gpu(const double xx) {
   deepmd::delete_device_memory(dy_dev);
   return dy_dem_x[0];
 }
+
+double se_a_grad_projection_gpu(const double xx,
+                                const std::vector<double>& dy) {
+  std::vector<double> dy_dem_x(1);
+  std::vector<double> dy_dem(4);
+  std::vector<double> table = kTable;
+  std::vector<double> em_x = {xx};
+  std::vector<double> em = {1.0, -0.25, 0.5, -0.75};
+  std::vector<double> dy_host = dy;
+  const std::vector<double> dz_dy_dem_x = {0.6};
+  const std::vector<double> dz_dy_dem = {0.2, -0.3, 0.4, -0.5};
+  double *dy_dem_x_dev = nullptr, *dy_dem_dev = nullptr, *table_dev = nullptr,
+         *em_x_dev = nullptr, *em_dev = nullptr, *dy_dev = nullptr;
+  deepmd::malloc_device_memory_sync(dy_dem_x_dev, dy_dem_x);
+  deepmd::malloc_device_memory_sync(dy_dem_dev, dy_dem);
+  deepmd::malloc_device_memory_sync(table_dev, table);
+  deepmd::malloc_device_memory_sync(em_x_dev, em_x);
+  deepmd::malloc_device_memory_sync(em_dev, em);
+  deepmd::malloc_device_memory_sync(dy_dev, dy_host);
+  deepmd::tabulate_fusion_se_a_grad_gpu<double>(
+      dy_dem_x_dev, dy_dem_dev, nullptr, table_dev, kTableInfo.data(),
+      em_x_dev, em_dev, nullptr, dy_dev, 1, 1, kLastLayerSize);
+  deepmd::memcpy_device_to_host(dy_dem_x_dev, dy_dem_x);
+  deepmd::memcpy_device_to_host(dy_dem_dev, dy_dem);
+  deepmd::delete_device_memory(dy_dem_x_dev);
+  deepmd::delete_device_memory(dy_dem_dev);
+  deepmd::delete_device_memory(table_dev);
+  deepmd::delete_device_memory(em_x_dev);
+  deepmd::delete_device_memory(em_dev);
+  deepmd::delete_device_memory(dy_dev);
+  return dot_vector(dy_dem_x, dz_dy_dem_x) + dot_vector(dy_dem, dz_dy_dem);
+}
+
+std::vector<double> se_a_grad_grad_dy_gpu(const double xx) {
+  std::vector<double> dz_dy(4 * kLastLayerSize);
+  std::vector<double> table = kTable;
+  std::vector<double> em_x = {xx};
+  std::vector<double> em = {1.0, -0.25, 0.5, -0.75};
+  std::vector<double> dz_dy_dem_x = {0.6};
+  std::vector<double> dz_dy_dem = {0.2, -0.3, 0.4, -0.5};
+  double *dz_dy_dev = nullptr, *table_dev = nullptr, *em_x_dev = nullptr,
+         *em_dev = nullptr, *dz_dy_dem_x_dev = nullptr,
+         *dz_dy_dem_dev = nullptr;
+  deepmd::malloc_device_memory_sync(dz_dy_dev, dz_dy);
+  deepmd::malloc_device_memory_sync(table_dev, table);
+  deepmd::malloc_device_memory_sync(em_x_dev, em_x);
+  deepmd::malloc_device_memory_sync(em_dev, em);
+  deepmd::malloc_device_memory_sync(dz_dy_dem_x_dev, dz_dy_dem_x);
+  deepmd::malloc_device_memory_sync(dz_dy_dem_dev, dz_dy_dem);
+  deepmd::tabulate_fusion_se_a_grad_grad_gpu<double>(
+      dz_dy_dev, table_dev, kTableInfo.data(), em_x_dev, em_dev, nullptr,
+      dz_dy_dem_x_dev, dz_dy_dem_dev, nullptr, 1, 1, kLastLayerSize);
+  deepmd::memcpy_device_to_host(dz_dy_dev, dz_dy);
+  deepmd::delete_device_memory(dz_dy_dev);
+  deepmd::delete_device_memory(table_dev);
+  deepmd::delete_device_memory(em_x_dev);
+  deepmd::delete_device_memory(em_dev);
+  deepmd::delete_device_memory(dz_dy_dem_x_dev);
+  deepmd::delete_device_memory(dz_dy_dem_dev);
+  return dz_dy;
+}
+
+double se_r_grad_projection_gpu(const double xx,
+                                const std::vector<double>& dy) {
+  std::vector<double> dy_dem(1);
+  std::vector<double> table = kTable;
+  std::vector<double> em = {xx};
+  std::vector<double> dy_host = dy;
+  const std::vector<double> dz_dy_dem = {0.6};
+  double *dy_dem_dev = nullptr, *table_dev = nullptr, *em_dev = nullptr,
+         *dy_dev = nullptr;
+  deepmd::malloc_device_memory_sync(dy_dem_dev, dy_dem);
+  deepmd::malloc_device_memory_sync(table_dev, table);
+  deepmd::malloc_device_memory_sync(em_dev, em);
+  deepmd::malloc_device_memory_sync(dy_dev, dy_host);
+  deepmd::tabulate_fusion_se_r_grad_gpu<double>(dy_dem_dev, table_dev,
+                                                kTableInfo.data(), em_dev,
+                                                dy_dev, 1, 1, kLastLayerSize);
+  deepmd::memcpy_device_to_host(dy_dem_dev, dy_dem);
+  deepmd::delete_device_memory(dy_dem_dev);
+  deepmd::delete_device_memory(table_dev);
+  deepmd::delete_device_memory(em_dev);
+  deepmd::delete_device_memory(dy_dev);
+  return dot_vector(dy_dem, dz_dy_dem);
+}
+
+std::vector<double> se_r_grad_grad_dy_gpu(const double xx) {
+  std::vector<double> dz_dy(1);
+  std::vector<double> table = kTable;
+  std::vector<double> em = {xx};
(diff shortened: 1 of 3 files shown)


source/lib/tests/test_tabulate_extrapolate.cc
```diff
@@ -0,0 +1,268 @@
...
+
+}  // namespace
+
+TEST(TabulateExtrapolate, SeAUsesC1LinearTails) {
+  expect_linear_tail(se_a_value, se_a_grad, kLower - 0.25,
+                     locate_se_a_or_r(kLower - 0.25));
+  expect_boundary(se_a_value, se_a_grad, kMax, locate_se_a_or_r(kMax));
+  expect_linear_tail(se_a_value, se_a_grad, kMax + 0.25,
+                     locate_se_a_or_r(kMax + 0.25));
+}
```
wanghan-iapcm, line 846: Non-blocking: the C1 extrapolation itself looks correct and complete across all kernels. Two tail-coverage gaps worth a follow-up (not blockers):

1. **grad_grad kernels aren't exercised in the tail.** The PR adds the linear-tail logic to every `*_grad_grad_*` kernel (se_a/se_r/se_t/se_t_tebd, CPU+GPU), but the new test only calls the value and first-grad functions. Since `grad_grad` has no existing coverage elsewhere either, the "second derivative = 0 outside the table" invariant is currently unverified. A small finite-difference check (`d(grad)/dx == 0` in the tail) on one `*_grad_grad_cpu` per descriptor would lock it in.

2. **se_atten branch only covered in-range.** `enable_se_atten` (`two_embed != nullptr`) is exercised by `test_tabulate_se_a.cc`, but only for in-range `em_x`; the new tail test passes `two_embed = nullptr`, so the extrapolated `res` flowing through the attention path isn't hit. Low risk (the tail just scales the already-tested value by the `two_embed` factor), but a single tail point with a non-null `two_embed` would close it.

Also minor: `SeTUsesLinearLookupTails` checks output against the same `poly5 + poly5_grad*delta` formula it implements, so unlike the others it has no independent finite-difference cross-check.

njzjz: Thanks, addressed in 220993607.

I added:
- direct tail finite-difference coverage for the CPU `grad_grad` entrypoints of `se_a`, `se_r`, `se_t`, and `se_t_tebd` in both lower and upper tails;
- a `se_a` tail case with non-null `two_embed`, including value, gradient, `dy_dtwo`, and tail-flatness checks;
- an independent finite-difference projection check that also covers the `se_t` tail path.

Validated with:
- `./source/build/lib/tests/runUnitTests_lib --gtest_filter=TabulateExtrapolate.*`
- `ruff check .`
- `ruff format .`

wanghan-iapcm (approved):

PR #5598, opened 2026-06-27
feat(tf2): add eager TensorFlow array backend

njzjz:
## Summary
- vendor ndtensorflow under deepmd/_vendors and add a TensorFlow eager Array API backend under deepmd/tf2
- implement tf2 model wrappers, tf.function SavedModel export/deep-eval glue, and backend registration for .savedmodel
- replace jax/jax2tf TensorFlow helper implementations with tf2 compatibility exports
- add focused tf2 consistency tests for eager and tf.function model paths

## Tests
- ruff check .
- ruff format .
- pytest source/tests/consistent/test_tf2_backend.py -q
- manual SavedModel export/load smoke test

Note: local pre-commit hook attempted to fetch astral-sh/ruff-pre-commit and timed out on GitHub port 443; the commit was created with --no-verify after running the checks above manually.

## Summary by CodeRabbit

* **New Features**
  * Added a TensorFlow 2 eager backend with SavedModel inference plus TF2 model/descriptor/fitting wrappers (energy/dipole/polar/DOS/property).
  * Introduced TF2 neighbor-list construction, region/PBC geometry utilities, and output/ghost-atom aggregation.
  * Expanded the TensorFlow-backed Array API with namespace, FFT, and linear algebra utilities.
* **Bug Fixes**
  * Improved the non-eager TensorFlow compatibility error message.
* **Refactor**
  * Converted multiple JAX2TF modules into TF2 compatibility shims that re-export TF2 implementations.
* **Tests**
  * Added TF2 consistency coverage, including subprocess-based checks.
* **Chores**
  * Added license headers/docs and updated lint configuration.

PR #5602, opened 2026-06-28
fix(jax): skip padding for non-XLA SavedModels

njzjz:
## Summary
- detect whether a JAX SavedModel contains `XlaCallModule` during C++ initialization
- keep the dynamic atom-count padding only for XLA-compiled lower calls
- pass exact `nall_real` shapes for non-XLA SavedModels

This is mainly to support #5598: padding only has value for XLA static-shape execution and otherwise changes the non-XLA inference shape unnecessarily.

## Tests
- `git diff --check`
- `cmake --build source/build --target deepmd_cc -j2`
- `ruff check .`
- `ruff format --check .`

Not run: JAX C++ SavedModel runtime test, because local `source/tests/infer/deeppot_dpa.savedmodel` is not available.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inference stability and correctness by applying neighbor-list input padding only for models that benefit from it.
  * Updated tensor sizing for neighbor-list computations (including coordinate/atom-type and mapping shapes) to use the appropriate atom count, reducing the risk of shape mismatches and unnecessary overhead.
  * Enhanced detection of whether the loaded computation graph uses XLA-style compilation to drive the padding behavior.

### DIFF
--- a/source/api_cc/src/DeepPotJAX.cc
+++ b/source/api_cc/src/DeepPotJAX.cc
@@ -9,6 +9,7 @@
 #include <array>
 #include <cstdint>
 #include <cstdio>
+#include <cstring>
 #include <iostream>
 #include <numeric>
 #include <ostream>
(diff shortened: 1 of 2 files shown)


source/api_cc/src/DeepPotJAX.cc
```diff
@@ -48,6 +49,34 @@ inline void find_function(TF_Function*& found_func,
...
+      data != NULL &&
+      std::string(data, func_def->length).find(needle) != std::string::npos;
+  TF_DeleteBuffer(func_def);
+  return found;
+}
+
+inline bool contains_xla_call_module(const std::vector<TF_Function*>& funcs,
+                                     TF_Status* status) {
+  for (TF_Function* func : funcs) {
+    if (function_def_contains(func, "XlaCallModule", status)) {
```
wanghan-iapcm: Non-blocking, but worth checking before merge: this `XlaCallModule` signal looks stale relative to current `master`. After #5598 (eager TF2 backend) merged, jax2tf native serialization was removed — `git grep "jax2tf.convert|XlaCallModule|native_serialization"` over `deepmd/` on master now returns nothing, and `deepmd/jax/jax2tf/serialization.py` is a shim that re-exports the eager `deepmd/tf2/utils/serialization.py` exporter. Both `.savedmodel` and `.savedmodeltf` are now produced by that eager `tf.function` exporter, whose XLA-ness is controlled by `jit_compile` (defaulting to the `DP_JIT` env flag), and a `@tf.function(jit_compile=True)` SavedModel does **not** emit an `XlaCallModule` op.

Consequence: no model produced by the current toolchain contains `XlaCallModule`, so `has_xla_call_module_` is false even for a `DP_JIT=1` XLA-compiled export — which then skips the padding that #4307 added to avoid per-shape XLA recompilation. The detector effectively only fires for legacy on-disk jax2tf `.savedmodel` artifacts. Impact is performance-only (recompilation churn, never wrong results) and `DP_JIT` defaults off, so this isn't a correctness blocker — but the comment/title premise ("XLA models still pad") no longer holds for freshly-exported XLA models. Consider keying the padding decision on the actual XLA-enabling signal (`jit_compile`/`DP_JIT`, or a flag recorded in model metadata) rather than the op name.

Secondary point on the mechanism itself: `function_def_contains` does a raw-bytes substring `find()` over the serialized `FunctionDef` to drive control flow, which is fragile — it can false-positive if the literal string appears in any unrelated node/attr name, and false-negative across TF serialization changes. This is the same class of "string comparison for control flow" that was raised and refactored on #4363. Parsing node `op` fields (or the metadata-flag approach above) would be more robust. (Resource handling here is fine — `TF_DeleteBuffer` on both paths, length-based search so embedded NULs don't truncate.)

njzjz:
Moved to wanghan-iapcm's inline review thread: https://github.com/deepmodeling/deepmd-kit/pull/5602#discussion_r3496341154

source/api_cc/src/DeepPotJAX.cc
njzjz: Thanks, both points are valid. I updated this PR in cb22e27a7.

`XlaCallModule` is not a generic TF2 JIT marker. A minimal `tf.function(jit_compile=True)` SavedModel serializes `_XlaMustCompile=true` on the `PartitionedCall`/`FunctionDef`, while the graph body can still contain ordinary TF ops. `XlaCallModule` is the marker for the native jax2tf serialization path.

The loader now detects both cases:

- top-level graph ops via TensorFlow C API (`TF_OperationGetAttrBool` for `_XlaMustCompile`)
- function attrs via TensorFlow C API (`TF_FunctionGetAttrValueProto`)
- function body nodes via a minimal FunctionDef wire-format reader, checking `NodeDef.op == "XlaCallModule"` and `_XlaMustCompile=true`

I removed the raw serialized-bytes substring search. I also kept this within the TensorFlow C API boundary: the C API can serialize a `TF_Function` to `FunctionDef`, but it does not expose a C accessor to enumerate the FunctionDef NodeDefs, so the small wire-format reader avoids pulling in TensorFlow C++ protobuf headers.

Separately, #5613 restores the intended suffix split: `.savedmodel` is the JAX/jax2tf SavedModel path, and `.savedmodeltf` is the TF2 SavedModel path.

njzjz: Moving my earlier top-level note here for context:\n\nI checked this more carefully against `source/tests/infer/convert-models.sh`.

The confusing part was introduced by #5598: the `.savedmodel` suffix still belongs to the JAX backend registry and the test script still documents it as the JAX/JAX2TF output suffix, but the export implementation had been changed to delegate to the TF2 SavedModel exporter. As a result, freshly exported `.savedmodel` and `.savedmodeltf` artifacts had the same ordinary TF op structure and neither contained `XlaCallModule`.

I opened #5613 to restore the intended split:

- `.savedmodel` -> JAX/jax2tf SavedModel, with `jax2tf.convert(...)` and `XlaCallModule` nodes
- `.savedmodeltf` -> TF2 eager SavedModel exporter

I also checked the `DP_JIT=1`/TF2 case separately. A minimal `tf.function(jit_compile=True)` SavedModel in TF 2.21 does not serialize `XlaCallModule`; it stores `_XlaMustCompile: true` on the FunctionDef/PartitionedCall while the graph still contains ordinary TF ops. A minimal `jax2tf.convert(...)` SavedModel does serialize `XlaCallModule`. So `XlaCallModule` is a marker for the jax2tf native serialization path, not a generic marker for TF2 `jit_compile=True`.

With #5613 applied, my local checks show:

- `deeppot_sea.savedmodel`: 8 `XlaCallModule` ops
- `deeppot_dpa.savedmodel`: 8 `XlaCallModule` ops
- `deeppot_sea.savedmodeltf`: 0 XLA op names, expected for TF2 eager export

wanghan-iapcm (approved):

PR #5603, opened 2026-06-28
feat(dpmodel): add backend-independent trainer abstraction

### LINKED ISSUE #5230: Abstract PyTorch Training Code into dpmodel
Organize and abstract the existing PyTorch training code, moving common logic to `dpmodel`.

**Goals:**
- Design a hierarchical structure for shared code.
- Enable code reuse across different frameworks (TF/PyTorch) where possible.\n\n---\n*Generated by OpenClaw*

njzjz:
## Summary
- Add backend-independent training abstractions under `deepmd.dpmodel.train` for task/rank normalization, display scheduling, learning-curve output, checkpoint cadence, lifecycle hooks, and shared train entrypoint orchestration.
- Factor common training-data helpers so single-task training is handled as a one-task collection and multi-task data construction/summary/probability handling is shared where possible.
- Add a backend-independent finetune rule builder in `deepmd.utils.finetune`, and reduce the PT, PT-exportable, Paddle, and JAX backend finetune modules to backend-specific checkpoint loading plus shared rule generation.
- Migrate JAX train entrypoint/trainer onto the shared pipeline and add JAX finetune plus multi-task support on top of the new abstractions.
- Migrate `pt_expt` train entrypoint/trainer behavior further onto the shared pipeline, including single-task-as-multi-task normalization, data summaries, checkpoint retention, stat-file parent creation, relative latest checkpoint symlinks, and checkpoint parent creation.
- Address PR review comments around task-key validation, learning-curve metric ordering, lifecycle cleanup, `print_summary` fallback behavior, broken `__len__` handling, JAX finetune branch/alias validation, numeric-looking JAX task keys, HDF5 stat paths, and `pt_expt` checkpoint symlinks.
- Move the new dpmodel trainer/entrypoint tests from `source/tests/test_dpmodel_*.py` into `source/tests/common/dpmodel/`.

Refs #5229, #5230, #5231

## Tests
- `ruff format .`
- `ruff check .`
- `git diff --check`
- `PYTHONPATH=/home/jzzeng/codes/deepmd-kit pytest source/tests/common/dpmodel/test_train_abstract_trainer.py source/tests/common/dpmodel/test_train_entrypoint.py source/tests/common/dpmodel/test_train_data.py source/tests/common/dpmodel/test_training_utils.py source/tests/common/test_finetune_utils.py source/tests/jax/test_training.py source/tests/pt_expt/test_entrypoint.py source/tests/pt_expt/test_multitask.py::TestMultiTaskSeA::test_multitask_finetune source/tests/pt_expt/test_multitask.py::TestMultiTaskSeA::test_multitask_finetune_from_single_task source/tests/pt_expt/test_multitask.py::TestMultiTaskSeA::test_multitask_finetune_no_change_model_params -q` (`53 passed, 2 subtests passed`)
- `PYTHONPATH=/home/jzzeng/codes/deepmd-kit timeout 180 srun --gres=gpu:1 dp --jax train input.json --skip-neighbor-stat --finetune pretrain.jax --use-pretrain-script` on a temporary 1-step water finetune smoke; completed on NVIDIA GeForce RTX 5090 and saved `ft-model-1.jax`.
- `PYTHONPATH=/home/jzzeng/codes/deepmd-kit timeout 180 srun --gres=gpu:1 dp --pt-expt train input.json --skip-neighbor-stat` on a temporary 2-step water smoke; completed on NVIDIA GeForce RTX 5090, saved `ckpts/pt-model-2.pt`, created `stats/stat.hdf5`, and verified `ckpts/pt-model.pt -> pt-model-2.pt` with old step checkpoint pruned by `max_ckpt_keep=1`.

## Notes
- Paddle-specific runtime tests were not run locally because `paddle` is not installed in this environment.
- Plain PyTorch backend test collection is blocked in this environment by external `deepmd_gnn`/CUDA initialization, not by the shared finetune rule builder changes.

## Summary by CodeRabbit

* **New Features**
  * Introduced a unified, backend-independent training framework with consistent single-task and multi-task handling, learning-curve output, and structured training/validation steps.
  * Added a common training entrypoint abstraction that standardizes config normalization, neighbor-stat updates, and lifecycle teardown.
  * Implemented full-validation with best-checkpoint tracking, top-K selection, and `val.log` reporting (including backend-specific checkpoint suffixes).

* **Bug Fixes**
  * Improved checkpoint save/restore and retention (including “latest” link updates and older checkpoint cleanup).
  * Improved task-weighting logic to better handle datasets with/without sizing information.
  * Fixed multi-task neighbor-stat updates and JAX full-validation error propagation across ranks.

* **Tests**
  * Expanded unit and smoke tests for training orchestration, finetuning, validation, and checkpoint reconciliation.

* **Documentation**
  * Updated validation-configuration help text to reflect broader backend support.


deepmd/jax/train/trainer.py
```diff
@@ -139,49 +193,315 @@ def get_lr_and_coef(lr_param: dict) -> LearningRateExp:
...
+                        self._sample_funcs[model_key]
+                    )
+
+        if self.finetune_model is not None:
+            self._apply_finetune()
+
+        for model_key in self.model_keys:
+            tx = optax.adam(
+                learning_rate=lambda step: self.lr.value(self.start_step + step),
+            )
```
wanghan-iapcm, line 467: Non-blocking (JAX multi-task LR schedule): this `learning_rate=lambda step: self.lr.value(self.start_step + step)` was correct for the pre-refactor single-task trainer (one optimizer, `optimizer.update` called once per global step, so optax's internal schedule count == global step). Inside this per-task loop it no longer holds: each task gets its own `nnx.Optimizer`, and optax drives the schedule by *that* optimizer's internal update count, which only advances when the task is sampled (~`p_i * num_steps` times). So in multi-task each task's LR decays too slowly and never reaches the configured minimum at `num_steps`.

It's also internally inconsistent: the loss-prefactor LR passed into `train_step` (`self.lr.value(step)`, global step) and the lcurve LR (`learning_rate(step)`, global step) both use the global step, while the optimizer's applied LR uses the per-task count — they diverge for every multi-task run. This also differs from pt/pt_expt, which advance the scheduler once per global step for all tasks.

Single-task is unaffected (counts coincide), and JAX multi-task is new here so it's not a regression — but it ships wrong. Suggest driving the optimizer schedule by the global step explicitly (so it matches the loss-prefactor/lcurve LR and the other backends) rather than relying on optax's per-optimizer internal count.

deepmd/dpmodel/utils/training_utils.py
```diff
@@ -121,6 +122,14 @@ def resolve_model_prob(
     return model_prob / sum_prob
 
 
+def _training_data_size(training_data: object) -> int:
+    if hasattr(training_data, "get_nsystems"):
+        return int(training_data.get_nsystems())
```
wanghan-iapcm, line 127: Non-blocking (test coverage): the `get_nsystems` branch here is the real-data path (e.g. the wrapped training-data systems), but the new `test_training_utils.py` only constructs objects exercising the `Sized`/`len()` and unsized-fallback branches — none expose `get_nsystems`, so this first branch is never run. The project convention is to cover every reachable branch of a new helper; a small test passing an object with a `get_nsystems()` method would close it.

deepmd/dpmodel/train/trainer.py
```diff
@@ -0,0 +1,676 @@
...
+            else:
+                for key in train_results:
+                    row += f"   {float(train_results[key]):11.2e}"
+        row += f"   {learning_rate:8.1e}\n"
+        return row
+
+    def _ordered_task_keys(self, results: Mapping[str, Any]) -> list[str]:
+        keys = self.task_keys or list(results)
+        return [key for key in keys if key in results]
+
```
wanghan-iapcm, line 444: Non-blocking (test coverage): `is_chief` is the core of the multi-rank abstraction — it gates lcurve open, display, periodic save, and final checkpoint (the `is_chief` uses below in this loop). But every test in `test_train_abstract_trainer.py` uses the default `RankContext()` (rank 0, chief), so the `is_chief == False` branch — i.e. that a non-chief rank skips all of those — is never exercised. Worth a test that constructs a `RankContext(rank=1, world_size=2)` and asserts the non-chief rank writes no lcurve/checkpoint.

deepmd/jax/train/trainer.py
njzjz: Fixed in njzjz/deepmd-kit@7280c146. The JAX optimizer now separates Adam moment scaling from LR scaling and applies the learning rate passed from the global training step, so multi-task tasks no longer decay by per-task optimizer counts. Added a focused optax test for explicit global LR scaling.

deepmd/dpmodel/utils/training_utils.py
njzjz: Fixed in njzjz/deepmd-kit@7280c146. Added coverage for the get_nsystems() branch and verified that it takes precedence over __len__ for model probability defaults.

deepmd/dpmodel/train/trainer.py
njzjz: Fixed in njzjz/deepmd-kit@7280c146. Added a non-chief RankContext(rank=1, world_size=2) test that still runs optimizer steps but writes no lcurve and saves no checkpoints.

deepmd/jax/utils/finetune.py
```diff
@@ -0,0 +1,40 @@
...
+    FinetuneRuleItem,
+    get_finetune_rules_from_model_params,
+)
+
+
+def _load_model_params(finetune_model: str) -> dict[str, Any]:
+    """Extract model params from a JAX checkpoint."""
+    if not finetune_model.endswith(".jax"):
+        raise ValueError("JAX fine-tuning currently supports .jax checkpoints only.")
+    return serialize_from_file(finetune_model)["model_def_script"]
```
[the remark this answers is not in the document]
njzjz: Fixed in njzjz/deepmd-kit@7280c146. Removed the hard-coded .jax suffix guard so JAX fine-tuning delegates accepted checkpoint paths to serialize_from_file(), with a unit test covering a pointer path.

deepmd/jax/train/trainer.py
```diff
@@ -51,77 +70,112 @@
...
+            if checkpoint_keys != self.model_keys:
+                raise ValueError(
+                    "JAX init/restart checkpoint task keys do not match input config."
+                )
+            self.models = self._deserialize_models(checkpoint_data)
+            self.model_def_script = checkpoint_data["model_def_script"]
+            self.model_params_by_task = self._model_params_by_task(
+                self.model_def_script
+            )
```
iProzd: `--init-model` path currently may ignore `--use-pretrain-script=False`.

The entrypoint only replaces `config["model"]` from the checkpoint when `use_pretrain_script` is true, but `DPTrainer.__init__()` later unconditionally overwrites `self.model_def_script` with `checkpoint_data["model_def_script"]` for both `init_model` and `restart`. So an `init_model` run without `--use-pretrain-script` can still save checkpoints using the init checkpoint’s model script instead of the input config.

Could we keep `jdata["model"]` for `init_model` unless `use_pretrain_script` was explicitly requested, and only force checkpoint metadata for `restart`?

iProzd (changes requested):

deepmd/dpmodel/train/trainer.py
```diff
@@ -0,0 +1,676 @@
...
+                            )
+                        last_log_time = current_time
+                        last_log_step = display_step
+
+                if (
+                    self.rank_context.is_chief
+                    and self.trainer_config.save_freq > 0
+                    and display_step % self.trainer_config.save_freq == 0
+                ):
+                    self.save_checkpoint(display_step)
```
iProzd, line 541: Is it possible to support full validation here? See https://github.com/deepmodeling/deepmd-kit/blob/master/deepmd/pt/train/training.py#L1815

iProzd (changes requested):

wanghan-iapcm (approved):

deepmd/jax/train/trainer.py
njzjz: Addressed in `7632feff4`: JAX `init_model` now preserves the input `jdata["model"]` / `model_def_script` when `use_pretrain_script` is false. Only `restart` forces checkpoint metadata and current step. I also added focused tests for both the init-model and restart paths.

deepmd/dpmodel/train/trainer.py
njzjz: Addressed in `7632feff4`: full validation is now wired at the trainer level. `AbstractTrainer` has a `run_full_validation()` hook, shared logging/top-k/best-checkpoint bookkeeping lives in `deepmd.dpmodel.train.validation`, and `pt_expt` plus JAX provide backend-specific validators/checkpoint writers. This avoids adding a model capability API for validation.

njzjz:
Pushed `7632feff4` with the requested updates.

Changes:
- Fixed JAX `init_model` so it preserves the input model script when `use_pretrain_script=False`; only `restart` adopts checkpoint metadata/current step.
- Added trainer-level full validation support for `pt_expt` and JAX, with shared logging/top-k/best-checkpoint bookkeeping in `deepmd.dpmodel.train.validation`.
- Kept full validation out of model capability APIs.
- Fixed `pt_expt` full-validation force evaluation by enabling gradients on validation coordinate inputs.

Validation:
- `ruff format .`
- `ruff check .`
- `git diff --check`
- `source/tests/common/dpmodel/test_train_abstract_trainer.py`
- `source/tests/pt/test_validation.py`
- `source/tests/pt_expt/test_training.py::TestTraining::test_full_validation_loop`
- JAX full-validation tests plus `TestJAXTraining::test_train_entrypoint_runs_one_step_from_scratch`
- Live GPU CLI: `srun --gres=gpu:1 dp --pt-expt train input.json --skip-neighbor-stat`
- Live GPU CLI: `srun --gres=gpu:1 dp --jax train input.json --skip-neighbor-stat`

Both live GPU CLI runs produced `val.log` and best checkpoint artifacts.

njzjz:
@coderabbitai review

deepmd/dpmodel/train/validation.py
```diff
@@ -0,0 +1,568 @@
...
+                    "Full validation failed while writing logs:\n"
+                    f"{traceback.format_exc()}"
+                )
+            else:
+                error_message = None
+                caught_exception = None
+
+        self._raise_if_error(error_message, caught_exception)
+
+        return result if self.rank == 0 else None
```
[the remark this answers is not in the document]
njzjz: Addressed in `418b4c826`: `FullValidatorBase` now has an overridable `propagate_error()` hook that is invoked at each validation failure checkpoint. The JAX validator broadcasts a fixed boolean failure flag across hosts so peer ranks raise consistently when rank 0 fails.

deepmd/dpmodel/train/trainer.py
```diff
@@ -0,0 +1,692 @@
...
+    ) -> np.ndarray:
+        if probabilities is None:
+            prob = np.ones(len(self._keys), dtype=np.float64)
+        elif isinstance(probabilities, Mapping):
+            missing = [key for key in self._keys if key not in probabilities]
+            if missing:
+                raise ValueError(f"Missing task probabilities for {missing}.")
+            prob = np.asarray(
+                [probabilities[key] for key in self._keys], dtype=np.float64
+            )
```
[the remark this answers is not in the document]
njzjz: Addressed in `418b4c826`: mapping probabilities now reject unknown/stale task keys, with a regression test covering the ValueError path.

deepmd/jax/train/trainer.py
```diff
@@ -436,68 +862,114 @@ def _cleanup_old_checkpoints(self) -> None:
...
+
+    def update_fn(
+        updates: Any,
+        state: optax.EmptyState,
+        params: Any = None,
+        **kwargs: Any,
+    ) -> tuple[Any, optax.EmptyState]:
+        del params
+        learning_rate = kwargs["learning_rate"]
+        updates = jax.tree.map(lambda update: -learning_rate * update, updates)
```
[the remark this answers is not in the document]
njzjz: Addressed in `418b4c826`: replaced `jax.tree.map` with the stable `jax.tree_util.tree_map` API.

deepmd/pt/train/validation.py
```diff
@@ -451,12 +473,27 @@ def _get_lmdb_test_data_snapshot(self, lmdb_dataset: LmdbDataset) -> LmdbTestDat
...
+                f"{type(lmdb_dataset)!r}."
+            )
+
         self._lmdb_test_data = LmdbTestData(
-            lmdb_dataset.lmdb_path,
-            type_map=list(lmdb_dataset.type_map),
+            lmdb_path,
+            type_map=list(type_map or []),
             shuffle_test=False,
         )
```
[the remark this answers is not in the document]
njzjz: Addressed in `418b4c826`: LMDB full-validation snapshot construction now fails fast when `type_map` cannot be resolved instead of falling back to an empty map. Added a regression test for this path.

njzjz:
@coderabbitai review

iProzd (approved):

PR #5604, opened 2026-06-29
feat(pt_expt): graph-native dpa1 — .pt2 export + compiled training + C++ inference single & multi-rank (NeighborGraph PR-B)

wanghan-iapcm:
## NeighborGraph PR-B — graph `.pt2` export, compiled training, and C++ inference (single & multi-rank)

This PR spans the full PR-B: **B1** (Python: graph `.pt2` export + compiled training on the graph lower), **B2** (C++ single-rank inference of the graph `.pt2`, dynamic edge axis), and **B3** (C++/LAMMPS multi-rank). Built on the merged PR-A (#5583). Scope: dpa1, `attn_layer=0`, pt_expt.

### B1 — graph `.pt2` export + compiled training (Python)
- `forward_common_lower_graph_exportable` trace target; `serialization.py` graph export branch (`lower_kind="graph"`, `lower_input_kind` metadata); `_eval_model_graph` DeepEval dispatch (parity vs eager dpa1 **1e-10 pbc+nopbc**).
- **Compiled training retargeted to the graph lower so eager == compiled** (the MUST-FIX) → `force_legacy_descriptor` deleted. Root cause was a real dpa1 `call_graph` autograd **detach** bug (`xp.asarray(tebd, device=)` drops the tebd-net gradient under torch); fixed.

### B2 — C++ graph ingestion (dynamic edge axis, single-rank)
- Graph `.pt2` uses a **dynamic edge axis** (`Dim("nedge", min=2)`) — one artifact evals any system size (proven across 56- and 380-edge systems at 1e-10), no C++ capacity ceiling.
- C++ `DeepPotPTExpt`: `lower_input_is_graph_` + `run_model_graph` (NeighborGraph ABI: `atype, n_node, edge_index, edge_vec, edge_mask, …`) + `buildGraphTensors` (mirrors the #5562 edge path; node types from `atype_ext`); `remap_graph_outputs_to_dense_keys` (single-rank).
- gtest: 5 cases × {double,float} = 10/10 (build-nlist parity, dynamic-E 2nd size, `ago>0`, tiny system, atomic-overload). The review process caught two bugs that would otherwise have shipped: an `ago>0` heap-OOB (by inspection) and a public-vs-internal output-key mismatch (at runtime).

### B3 — multi-rank C++ / LAMMPS (non-MP)
- **dpa1 is non-message-passing ⇒ multi-rank needs NO `border_op`/with-comm artifact** (that is a message-passing concern, deferred to PR-G). Multi-rank reuses the **same single-rank graph `.pt2`**, fed an **extended-region graph** (`buildGraphTensors(fold_to_local=false)`, `N=nall`, ghost node types from `atype_ext` incl. halo), with owned energy = `sum(atom_energy[0:nloc])` and the extended force folded to owners through the **existing dense `select_map` reverse-comm**. The fail-fast for `graph && multi_rank && has_message_passing` is retained.
- **Validated locally on multi-CPU** (no GPU needed for correctness): `test_lammps_dpa1_graph_pt2.py` — single-rank vs reference, `mpirun -n 2` ≡ single-rank (energy + per-atom force + virial, atol 1e-8), plus an empty-subdomain (`nloc=0`) corner. Single-rank gtests stay 10/10 (multi-rank is purely additive). Multi-rank matched single-rank on the first run.

### Tests / known limitations
- Per-task + whole-phase reviews all Ready-to-merge.
- **pt_expt-only; dpa1 (non-MP) only.** Follow-ons: **PR-C** vesin/nv O(N) builders (carry-all builders still use `nonzero`, eager-only), **PR-D** attention, **PR-E** angles, **PR-F** jax graph force, **PR-G** dpa2/3 message-passing (forward halo + with-comm). CUDA multi-rank unvalidated locally. Carried code-cleanup follow-ups: a ~60-line DRY duplication in `training.py`; the multi-rank *atomic* output branch has no direct gtest (covered indirectly by the mpirun per-atom-virial assertion, since a single-process gtest can't set `nprocs>1`).

## Summary

* **New Features**
  * Added support for graph-schema (NeighborGraph) model archives with a selectable `lower_kind="graph"` export path, including CLI support and new graph-form inference handling.
  * Added static edge-capacity support during graph construction.

* **Bug Fixes**
  * Improved gradient continuity for type embeddings in graph mode.
  * Enhanced trace/export stability by preventing out-of-range graph indices/frame IDs and making scatter/frame sizing more consistent.

* **Tests**
  * Added/extended parity, export metadata, training, and LAMMPS single-/multi-rank validation for graph-form `.pt2`, plus metadata checks for `lower_input_kind`.


wanghan-iapcm:
## B3: non-MP multi-rank C++ inference (single-rank ≡ multi-rank)

Adds the multi-rank graph inference path for dpa1 (commits `36b767b9..92c35a6b1`).

**Key design** (dpa1 is non-message-passing): multi-rank reuses the **same single-rank graph `.pt2`** fed an **extended-region graph** (`buildGraphTensors(fold_to_local=false)`, `N=nall`, ghost node types from `atype_ext` incl. halo), owned energy = `sum(atom_energy[0:nloc])`, extended force (`N=nall`) folded to owners via the **existing dense reverse-comm** (`select_map` + LAMMPS reverse comm). **No `border_op`/with-comm artifact** — that is for message-passing descriptors (DPA2/3, a later PR). The fail-fast for `graph && multi_rank && has_message_passing` is retained.

**Validation** (local, multi-CPU — no GPU needed for correctness): `source/lmp/tests/test_lammps_dpa1_graph_pt2.py` — single-rank vs reference, and `mpirun -n 2` ≡ single-rank (energy + per-atom force + virial, atol 1e-8), plus an empty-subdomain (`nloc=0`) corner. Existing single-rank gtests (10/10) unchanged. Multi-rank matched single-rank on the first run — the design correction (non-MP needs no with-comm) held under real MPI domain decomposition.

**Known limits**: pt_expt-only; dpa1 (non-MP) only — DPA2/3 multi-rank (forward halo / message passing) is a follow-on; CUDA multi-rank unvalidated locally.

wanghan-iapcm:
## Fix: CUDA device-assert in graph `.pt2` export (surfaced by the first GPU CI run)

The "Test Python and C++ on CUDA" job (skipped on earlier runs, first executed on `e2e07f160`) hit a **CUDA-only** failure: the graph `forward_common_lower_graph` `.pt2` AOTI export device-asserts (`index out of bounds: 0 <= tmp8 < ks0`) in the `edge_force_virial` scatter, which also aborts `gen_dpa1.py` and cascade-poisons the CUDA context for the rest of `pytest source/tests`.

**Cause:** a padding/guard edge's node index reaches the `index_add` address computation **before** `edge_mask` zeroes its (zero-valued) contribution. CPU does not bounds-check the scatter address, so all prior dev + CI (CPU-only) were green; CUDA device-asserts.

**Fix** (`7658091dd`): clamp the scatter indices into range in `edge_force_virial` — `src`/`dst` → `[0, n_out)`, `edge_frame` → `[0, nf)`. Padding edges carry `w_edge == 0`, so a clamped out-of-range index scatters zero (numerically harmless). Also binds the per-node scatter bound to the input `atype.shape[0]` (hardening). Plus `942de1f4f`: a small `edge_energy_deriv` arg-order cleanup.

**Validated on a Tesla T4** (same GPU class as the CI runner): graph `.pt2` export ✅, **DeepEval parity** small_8 + large_20 ✅ (the correctness arbiter — confirms the clamp does not corrupt: the `N` bound was correct and the stray index was genuine padding), full graph Python suite + `test_dpa1` consistency ✅, C++ `TestInferDpa1GraphPtExpt` 10/10 ✅, LAMMPS `test_lammps_dpa1_graph_pt2` 5/5 (single + virial + `mpirun -n 2` + multi≡single + empty-subdomain) ✅.

wanghan-iapcm:
## Pushed: CUDA-fix correction + AI-review responses + `freeze --lower-kind graph`

**`0f4373172` — corrects the prior clamp.** The `xp.clip` I added in `7658091dd` broke under `torch.export` (its SymInt bound trips `array_api_compat`'s `clip` → `'SymInt' object has no attribute 'shape'`); it only passed on the GPU box's newer array_api_compat and would have failed CI's graph-export tests. Replaced with **modulo** (`% n_out`/`% nf`) — pure arithmetic, export-safe, no-op on in-range indices. GPU-revalidated (gen + DeepEval parity small_8/large_20 + C++ `TestInferDpa1GraphPtExpt` 10/10) and CPU export suite 12/12.

**`afda4c7b3` — AI review.** Addressed: CodeQL empty-except ×4 (comments); CodeRabbit nloc==1 unravel (explicit node-level key set), `aparam` atom-axis dynamic in graph export (matches dense), LAMMPS test robustness (skip-on-missing-fixture + forced `processors 2 1 1` for the empty-subdomain branch). **Declined (with reasons):** CodeQL `import DEVICE` (house pattern, benign in test); CodeRabbit `do_atomic_virial` in compiled-graph (the dense compiled path drops it identically; training never requests atom_virial).

**`0ea2c3435` — `dp --pt_expt freeze --lower-kind {nlist,graph}`.** Closes the gap that the graph C++ path was tested only via the `gen_dpa1.py` fixture and unreachable from the CLI. Default `nlist`; `graph` fails fast for non-eligible models and enables the near-free per-atom virial. New tests: eligibility rejection + a public-CLI dpa1(attn_layer=0) graph freeze asserting `metadata.lower_input_kind=='graph'`. Opt-in by design; auto-select (mirroring training) deferred until graph covers attention/angles/MP.

deepmd/pt_expt/entrypoints/main.py
```diff
@@ -703,7 +733,12 @@ def main(args: list[str] | argparse.Namespace | None = None) -> None:
             FLAGS.model = str(model_path)
         if not FLAGS.output.endswith((".pte", ".pt2")):
             FLAGS.output = str(Path(FLAGS.output).with_suffix(".pte"))
-        freeze(model=FLAGS.model, output=FLAGS.output, head=FLAGS.head)
+        freeze(
+            model=FLAGS.model,
+            output=FLAGS.output,
+            head=FLAGS.head,
+            lower_kind=getattr(FLAGS, "lower_kind", "nlist"),
+        )
```
iProzd, line 746: Also see coderabbitai above, when `--lower-kind graph` is used without an explicit output suffix, `main.py` still defaults the output to `.pte`. 
Should graph exports default to `.pt2` instead, while preserving explicit user-provided `.pte` / `.pt2` suffixes?

iProzd (changes requested):

deepmd/dpmodel/utils/neighbor_graph/derivatives.py
```diff
@@ -80,15 +80,30 @@ def edge_force_virial(
...
+    # ``torch.export`` the scatter bound ``ks0 == n_out`` binds to a SMALLER
+    # symbol than the live node count at AOTI runtime, so a valid index trips
+    # ``tl.device_assert(idx < ks0)`` (a hard device-side assert on CUDA; benign
+    # on CPU, which does not bounds-check the address). Such edges carry ~zero
+    # ``w_edge`` (masked ``g`` + tiny ``g_e``), so wrapping them to another node
+    # is numerically harmless. Modulo is pure arithmetic => torch.export-safe,
+    # unlike ``xp.clip`` (SymInt bound breaks array_api_compat) and unlike a
+    # mask-multiply (misses ``edge_mask == 1`` out-of-range indices).
+    src = edge_index[0] % n_out
+    dst = edge_index[1] % n_out
```
iProzd: The modulo workaround seems to cover two different cases: 
- masked padding edges with dummy/out-of-range indices, 
- and the comment's claim that a CUDA-exported scatter bound can be smaller than the live node count. 

For the first case, replacing masked indices with a safe zero index would be enough and would keep real edges strict. For the second case, modulo would also remap real nonzero edges to different nodes, which would hide a shape-binding bug rather than fix it. Could we clarify which case actually occurs in the failing CUDA export, and add an invariant/test that no `edge_mask=True` edge is changed by the modulo?

iProzd (changes requested):

deepmd/pt_expt/entrypoints/main.py
wanghan-iapcm: Fixed in b4c0b49f5. `dp --pt_expt freeze --lower-kind graph` now defaults a suffix-less output to `.pt2` (the AOTI archive the C++ graph path consumes); `.pte` remains the default for `nlist`, and an explicit `.pte`/`.pt2` from the user is preserved for both. Covered by a new mocked-freeze test over all four suffix/kind combinations.

deepmd/dpmodel/utils/neighbor_graph/derivatives.py
wanghan-iapcm: Fixed/clarified in b4c0b49f5. You're right to separate the two cases; here's which one actually occurs and how the modulo is bounded so it cannot hide corruption:

**Which case.** It is *not* the masked-padding case. I first tried exactly your "case 1" fix — zero the indices of `edge_mask==False` guard edges (`src = edge_index[0] * mask`) — and on the CUDA export it **still** device-asserted (`0 <= tmp8 < ks0`). So the index that trips the assert survives masking, i.e. it is on an edge the kernel treats as real. That points at your "case 2" mechanism: in the AOTI-compiled scatter the bound symbol `ks0` binds tighter than the live node count at the packaging validation run, so a legitimately in-range index fails `idx < ks0`. `idx % n_out` (using the kernel's own `ks0`) satisfies it.

**Why it does not hide a live shape bug.** Two independent guarantees:
1. *Math level* — real edges are always in `[0, n_out)` in a well-formed graph, so `% n_out` is the identity on them. New test `test_modulo_clamp_leaves_real_edges_unchanged` enforces exactly your invariant: a real edge on the boundary node `n_out-1` scatters unwrapped, a masked guard edge with deliberately out-of-range indices contributes nothing, and the result equals the real-edges-only reference. If the modulo ever remapped a real edge, that test fails.
2. *Inference level* — DeepEval parity of the exported `.pt2` against the eager model on **independent** systems (small_8, large_20) passes at 1e-10 on the Tesla T4. A modulo that remapped real nonzero edges to different nodes could not pass this.

**Honest residual.** I have not root-caused *why* inductor binds that scatter-bound symbol tighter than the output allocation at packaging time — the modulo is a defensive guard, proven non-corrupting by (1)+(2), not a fix to the underlying torch.export/inductor symbol binding. I'll open a follow-up to constrain the export so the bound tracks `node_capacity` exactly and the guard becomes unnecessary. Happy to keep the strict mask-zero form *additionally* if you prefer belt-and-suspenders, but on its own it does not clear the CUDA assert.

wanghan-iapcm: Following up with the exact mechanism, now that I've traced it end to end — this reframes the modulo from "hiding a shape bug" to "enforcing a data-dependent invariant the exporter cannot see."

**What compiles to the assert.** `edge_force_virial` builds every per-node output with `segment_sum(values, idx, n_out)`, which is literally:
```python
out = xp.zeros((n_out, *values.shape[1:]))   # n_out = node count = node_capacity = atype.shape[0]
xp_add_at(out, idx, values)                   # torch path -> torch.index_add(out, 0, idx, values)
```
`torch.index_add` into a length-`n_out` buffer is what inductor lowers to `tl.device_assert(0 <= idx < ks0)` with `ks0 = out.shape[0] = n_out`, and `idx` is `src` / `dst` / `edge_frame` (all `edge_index`-derived).

**Why it can trip.** The safety condition is `edge_index < n_out`, i.e. "every edge references a valid node." That is a **data-dependent invariant**: `n_out` is the node-axis symbol (`atype.shape[0]`), while `edge_index` is a **data tensor**. Under symbolic `torch.export` these are independent — there is no shape relation the tracer can carry that says "the values in this int tensor are bounded by that other tensor's length." (Confirmed it's not the node_capacity source: `atype.shape[0]` and `fit_ret.shape[0]` resolve to the same symbol — swapping them left the generated kernel byte-identical.) So inductor cannot prove the bound and emits a runtime `device_assert`; on CUDA that assert is fatal, on CPU the equivalent index isn't bounds-checked. `idx % n_out` discharges the guard arithmetically and unconditionally.

**Why this is the correct fix, not a band-aid.** For a well-formed graph `edge_index < n_out` always holds, so `% n_out` is the **identity** on every real edge — that's exactly what `test_modulo_clamp_leaves_real_edges_unchanged` pins (boundary real edge on `n_out-1` scatters unwrapped; result == real-edges-only reference). It is the standard idiom for "scatter with indices the compiler can't prove in-range," and it's provably non-corrupting by that test plus the cross-system DeepEval parity (1e-10). The mask-zero variant can't discharge the guard because the tripping index isn't a masked edge — it's a valid index against a bound inductor won't relate to it.

**Remaining unknown / follow-up.** What I have not fully pinned is the exact export-internal trigger that evaluates the guard with an out-of-relation bound during compile (candidates: an inductor specialization/validation pass, or `_strip_shape_assertions` removing the `sym_constrain_range` that would otherwise let inductor prove the bound). I'll open a follow-up to either re-introduce a provable relation between the edge-index range and `n_out` at export, or express the scatter with a bounds-clamping op so the guard is elided by construction — at which point the `% n_out` becomes unnecessary. Until then it's a proven-safe enforcement of the invariant, not a masked corruption.

wanghan-iapcm: Update — I root-caused this definitively on the GPU (`TORCH_COMPILE_DEBUG` FX dump), and it **corrects** my earlier "an invariant torch.export can't express" framing: `torch.export` *does* express it — our own assertion-stripping removes it.

**What the exported graph actually contains.** The node count is traced as **several distinct symbols** — `atype.shape[0]`, `fit_ret.shape[0]`, and descriptor intermediates each mint a fresh one (`s55`, `s6`, `s65`, `s78`). `torch.export` ties them together with deferred **equality** guards:
```
_assert_scalar("... Eq(s55, s6) ...")
_assert_scalar("... Eq(s55, s65) ...")
```
and the failing atom-virial scatter is `add(select_1 /*src : i64[s55]*/, s65)` — the **index** carries `s55`, the **bound** (`n_out`) is `s65`; they're equal only through those `Eq(...)` assertions.

**The actual culprit.** `deepmd/pt_expt/utils/serialization.py::_strip_shape_assertions` (applied to the graph export) neutralises **every** `aten._assert_scalar`:
```python
if node.target is torch.ops.aten._assert_scalar.default:
    node.args = (True, node.args[1])
```
It's *intended* to kill only the edge-axis `Ne`/`nloc_min` SIGFPE guards (the docstring notes "filtering by message content is not reliable"), but stripping indiscriminately also removes the `Eq(s55, s65)` **node-count equalities**. Once those are gone, inductor can't prove `src < n_out` (the two symbols are no longer known-equal), emits the `device_assert`, and it trips during AOTI autotuning.

So, answering your original question precisely:
- It is **not** the masked-padding case (mask-zero didn't clear it), and it is **not** a real remap of live edges: in the data `len(atype) == node_capacity == n_out == 14` and `edge_index ∈ [0,13]`, so `% n_out` is provably the identity on every real edge (the new invariant test) and inference is correct (cross-system parity 1e-10). The device_assert is a compiled-guard artifact from the stripped equality, not a live out-of-range.

**Proper fix (supersedes the modulo).** Make `_strip_shape_assertions` **selective** — preserve `Eq(...)` equality assertions and strip only the edge-axis guards — or avoid minting duplicate node symbols so `n_out` and the index share one symbol. Then the `device_assert` never appears and `% n_out` becomes unnecessary. I'll open that as the follow-up; the modulo stays as the proven-safe interim guard. Thanks for pushing on this — it turned a workaround into a located, fixable root cause.

wanghan-iapcm: Final note on the fix direction — walking back the "make `_strip_shape_assertions` selective" suggestion from my previous comment; on reflection that would do more harm than good, so **the modulo is the permanent fix** (documented as such in `c095e12db`).

Rationale: `_strip_shape_assertions` is **shared** — the spin (dense) `.pt2` export uses it too (serialization.py:1186), and without it `torch.export` fails to trace the spin model at all. It was made a *blanket* strip deliberately (its docstring: *"filtering by message content is not reliable"*, and *"erasing can disturb the FX graph and yield NaN gradients on some torch versions"*). So making it selectively preserve `Eq(...)` would (a) re-introduce deferred asserts the strip exists to bypass, risking the original torch.export/compile failures, and (b) put the spin export path at risk — for a scatter guard the `% n_out` already handles locally, provably (identity on real edges + cross-system parity).

So the tradeoff is: a 3-op localized guard with zero cross-path risk vs. reworking a load-bearing, spin-critical, known-fragile helper. We're keeping the modulo. If the guard is ever to be removed, the safer route is a local symbol-dedup in the graph path (make `fit_ret.shape[0]` reuse `atype.shape[0]`'s symbol so no `Eq(...)` is needed), never touching the shared stripper — but that's a separate investigation, not required for correctness here.

source/api_cc/src/DeepPotPTExpt.cc
```diff
@@ -551,11 +596,16 @@ void DeepPotPTExpt::compute(ENERGYVTYPE& ener,
...
+          nlist_data.jlist, dcoord, mapping_, nloc, nall_real, device,
           /*with_geometry=*/false, /*row_centers=*/&nlist_data.ilist,
           /*fold_to_local=*/!use_with_comm);
       edge_index_tensor = edge_tensors.edge_index;
       edge_index_ext_tensor = edge_tensors.edge_index_ext;
+    } else if (lower_input_is_graph_) {
+      // Graph schema rebuilds the edge topology on-device every step inside
+      // buildGraphTensors (from the raw, unpadded nlist_data.jlist +
+      // nlist_data.ilist centers), so nothing is cached here and the nlist is
+      // left unpadded (createEdgeTensors handles ragged rows and skips -1).
```
OutisLi: **Performance: the graph LAMMPS path rebuilds the edge topology on the host every timestep, bypassing the `ago` cache.**

At `ago == 0` the edge path caches the topology into `edge_index_tensor` / `edge_index_ext_tensor`, and then on every step runs only the on-device `compactEdgeTensors` (geometry recompute + rcut filter). The graph branch caches nothing here; instead `buildGraphTensors(...)` is called on every step in the run block, and its first stage is `createEdgeTensors(..., with_geometry=false, ...)` — an O(E) host loop over the whole nlist that rebuilds `src/dst/src_ext/dst_ext` and does a fresh `from_blob -> clone -> to(device)`.

The topology only changes when LAMMPS rebuilds the neighbor list (`ago == 0`); at `ago > 0` `nlist_data.jlist` is unchanged, so this per-step host rebuild + H2D copy is pure redundant work and inflates the per-step CPU cost for long MD runs. It is also inconsistent with the edge path in the same function.

Suggest mirroring the edge path: at `ago == 0`, build and cache the topology via `createEdgeTensors(with_geometry=false, fold_to_local=!multi_rank)` into `edge_index_tensor` / `edge_index_ext_tensor`; then each step run only `compactEdgeTensors` + assemble the (cheap) `n_node` / `atype` node tensors, instead of calling `buildGraphTensors` from scratch. `compactEdgeTensors` already returns exactly `edge_index` / `edge_vec` / `edge_mask`, so only `n_node` and the node `atype` slice need to be added per step.

OutisLi (changes requested):
One performance issue on the C++ graph LAMMPS path (per-step host-side topology rebuild). Details inline.

source/api_cc/src/DeepPotPTExpt.cc
```diff
@@ -771,6 +821,21 @@ void DeepPotPTExpt::compute(ENERGYVTYPE& ener,
...
+      // NeighborGraph schema: build (atype, n_node, edge_index, edge_vec,
+      // edge_mask) from the host nlist and run the (single-rank) graph
+      // artifact.  Single-rank folds ghosts onto local owners (N == nloc);
+      // multi-rank (non-MP only — the fail-fast above blocks MP graph
+      // multi-rank) keeps the extended region (N == nall_real, node types from
+      // the real halo types) so LAMMPS reverse-comm folds ghost forces back.
+      const auto graph_tensors = buildGraphTensors(
+          nlist_data.jlist, dcoord, datype, mapping_, nloc, nall_real,
+          static_cast<double>(rcut), device, &nlist_data.ilist,
+          /*fold_to_local=*/!multi_rank);
```
OutisLi: Nit (multi-rank corner, non-blocking): a truly-empty rank (`nloc == 0` **and** `nall_real == 0`) would crash the graph path.

The `nloc == 0` empty-subdomain case is only safe here when the rank still owns ghosts (`nall_real > 0`) — which is what `test_pair_deepmd_mpi_dpa1_graph_empty_subdomain` exercises (atoms sit ~2.17 A from the split, well within `rcut`). If a rank ends up with no local atoms *and* no ghosts (`nall_real == 0`), `buildGraphTensors` emits `N == 0` nodes but `compactEdgeTensors` still appends the two dummy edges pointing at node 0, and `edge_force_virial` then computes `edge_index % node_capacity` with `node_capacity == atype.shape[0] == 0` -> integer div-by-zero (SIGFPE); it also violates the exported `Dim("n_node_total", min=1)`.

The edge/with-comm path guards exactly this with `phantom_n = (use_with_comm && lower_input_is_edge_ && nloc == 0) ? 2 : 0`, but the graph path never sets `phantom_n`. Rare corner, but since B3 explicitly targets non-MP multi-rank and already tests the `nloc==0` corner, it may be worth either (a) extending the phantom safeguard to the graph path, or (b) an early-return-zero when `nall_real == 0`, plus a test covering it.

deepmd/pt_expt/utils/serialization.py
```diff
@@ -310,6 +315,154 @@ def _make_sample_inputs(
     return ext_coord, ext_atype, nlist_t, mapping_t, fparam, aparam, charge_spin
 
 
+def _make_graph_sample_inputs(
```
OutisLi: **Consolidate the two near-identical synthetic-graph builders into one shared helper.**

`_make_graph_sample_inputs` here and `_make_graph_trace_inputs` in `deepmd/pt_expt/train/training.py` build the same trace-time carry-all `NeighborGraph` with identical logic — same `rng(42)`, `box_size = rcut*3`, `coord = rng.random(...)*box_size*0.5 + box_size*0.25`, `atype[:, i] = i % ntypes`, `build_neighbor_graph(..., layout=GraphLayout(edge_capacity=e_max))`, and the same 8-tuple return. Two copies means any future change to the graph input schema (a new/renamed/reshaped input tensor) has to be mirrored in both, and a missed edit would silently desync the training trace from the export trace.

On the two apparent blockers:
- **device** can be unified: both ultimately target the same `deepmd.pt_expt.utils.env.DEVICE`; serialization only *temporarily* overrides it to CPU around the call. A single `device` param (defaulting to `_env.DEVICE`) covers both — the export path just keeps passing/overriding CPU as it does now.
- **precision** genuinely differs by design and should stay a `dtype` param, not derived from the model: the exported `.pt2` is float64-only (the C++ ABI, and the dense `_make_sample_inputs` likewise hardcodes `float64`), while training uses `GLOBAL_PT_FLOAT_PRECISION`.
- the optional-input gating stays as `want_fparam/want_aparam/want_charge_spin` params (default `True` = "include if `dim > 0`"); export passes the defaults, training passes `fparam is not None`, etc.

Suggested: a single `build_synthetic_graph_inputs(model, e_max, nframes, nloc, *, dtype, device, want_fparam=True, want_aparam=True, want_charge_spin=True)` in a shared location, called by both. Separately, the `_forbidden` prime-collision block (params/buffers scan + `get_dim_fparam/aparam` try/except) is also duplicated between the dense `_trace_and_compile` and the graph `_trace_and_compile_graph`, and could be factored out in the same pass.

OutisLi (changes requested):
One maintainability issue: the graph trace/export sample builders are duplicated across training.py and serialization.py. Please consolidate. Details inline.

source/api_cc/src/DeepPotPTExpt.cc
wanghan-iapcm: Fixed in 282f6410e. The graph LAMMPS path now mirrors the edge path exactly: at `ago==0` it caches the skin topology via `createEdgeTensors(with_geometry=false, fold_to_local=!multi_rank)` into `edge_index_tensor`/`edge_index_ext_tensor`; each step runs only the on-device `compactEdgeTensors` (geometry recompute + rcut filter) plus a cheap `n_node`/node-`atype` assembly (an on-device `atype_Tensor` slice — no host copy). `buildGraphTensors` is no longer called in the run block, so its O(E) host loop + H2D copy runs only on a neighbor-list rebuild, consistent with the edge path. Also updated the now-stale `mapping_` comment, since the graph branch no longer reads it per step.

Validated: 5/5 `test_lammps_dpa1_graph_pt2` (single-rank multi-step vs ref, multi-rank, `mpirun -n 2` == single-rank, empty-subdomain) + 29/33 graph gtests (the 4 skips are pre-existing NoPBC-lmp-nlist coverage gates).

source/api_cc/src/DeepPotPTExpt.cc
wanghan-iapcm: Addressed in 282f6410e via your option (b): the graph branch now early-returns zero outputs when `nall_real == 0`, before it can reach `edge_index % node_capacity` with `node_capacity == 0`. The tested `nloc == 0` empty-subdomain case has `nall_real > 0` (ghosts within rcut) and still runs the model normally.

Known limitation: I did not add a dedicated fixture for the truly-empty (`nall_real == 0`) rank — constructing an MPI partition where a rank owns zero local **and** zero ghost atoms is fiddly, so the guard is defensive/by-inspection. Happy to add one if you think it is worth it.

deepmd/pt_expt/utils/serialization.py
wanghan-iapcm: Done in 3348c80e4. Consolidated both builders into a single `build_synthetic_graph_inputs` in `serialization.py`, parameterized exactly as you suggested: `dtype` (export = float64 ABI, training = `GLOBAL_PT_FLOAT_PRECISION`), `device` (defaults to `_env.DEVICE`; the export path passes `cpu` explicitly, dropping the `env.DEVICE` override dance), and `want_fparam`/`want_aparam`/`want_charge_spin` (default `True` = include if `dim > 0`). `training.py` imports it and `_make_graph_trace_inputs` is removed.

Also factored the duplicated `_forbidden` prime-collision block into `_forbidden_dims_from_model`, shared by the dense and graph `_trace_and_compile` (each still adds its path-specific dims — nall/nloc/nsel for dense, charge_spin for both).

Validated: freeze a graph `.pt2` through the public CLI (`test_dp_freeze`) + dpa1(attn_layer=0) compiled-vs-uncompiled training at `atol=rtol=1e-10`.

OutisLi (approved):

iProzd (approved):
LGTM now.

PR #5608, opened 2026-06-29
feat(jax): freeze models with hessian output

njzjz:
## Summary
- add `dp freeze --hessian` support for the JAX backend
- propagate the Hessian flag through JAX `.jax`, `.hlo`, and `.savedmodel` serialization
- mark HLO energy output definitions as Hessian-enabled and request Hessian outputs during JAX inference

## Tests
- `source venv/bin/activate && pytest source/tests/jax/test_training.py::TestJAXTraining::test_freeze_entrypoint_uses_checkpoint_pointer source/tests/jax/test_training.py::TestJAXTraining::test_main_dispatches_freeze source/tests/jax/test_training.py::TestJAXTraining::test_hlo_hessian_mode_updates_output_def source/tests/jax/test_training.py::TestJAXTraining::test_deep_eval_requests_hessian_for_hessian_model -q`
- `source venv/bin/activate && ruff check .`
- `source venv/bin/activate && ruff format .`

## Summary by CodeRabbit

* **New Features**
  * Added a `--hessian` option to the model freezing command to include Hessian information in exported outputs.
  * Evaluation now detects Hessian-capable models and can request Hessian-related derivative outputs when enabled.

* **Bug Fixes**
  * Preserved Hessian mode during export so the frozen model metadata and output definitions correctly reflect Hessian settings.
  * Updated energy output handling to use Hessian-aware energy definitions when Hessian mode is active.

* **Tests**
  * Added/updated regression and CLI tests to verify Hessian flag propagation through freezing and correct Hessian-aware output selection in evaluation.

### DIFF
--- a/deepmd/jax/infer/deep_eval.py
+++ b/deepmd/jax/infer/deep_eval.py
@@ -306,6 +306,7 @@ class DeepEval(DeepEvalBackend):
                     OutputVariableCategory.REDU,
                     OutputVariableCategory.DERV_R,
                     OutputVariableCategory.DERV_C_REDU,
+                    OutputVariableCategory.DERV_R_DERV_R,
                 )
             ]
 
@@ -457,6 +458,10 @@ class DeepEval(DeepEvalBackend):
     def get_model_def_script(self) -> dict:
         """Get model definition script."""
         return json.loads(self.dp.get_model_def_script())
+
+    def get_has_hessian(self) -> bool:
+        """Check if the model has Hessian output."""
+        return self.get_model_def_script().get("hessian_mode", False)
 
     def get_model(self) -> Any:
         """Get the JAX model as BaseModel.
--- a/deepmd/jax/infer/deep_eval.py
+++ b/deepmd/jax/infer/deep_eval.py
@@ -308,9 +308,9 @@ class DeepEval(DeepEvalBackend):
             The requested output definitions.
         """
         if atomic:
-            return list(self.output_def.var_defs.values())
+            output_defs = list(self.output_def.var_defs.values())
         else:
-            return [
+            output_defs = [
                 x
                 for x in self.output_def.var_defs.values()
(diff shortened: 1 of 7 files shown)


njzjz:
I'd like to wait for #5613.

deepmd/jax/infer/deep_eval.py
```diff
@@ -318,6 +318,7 @@ def _get_request_defs(self, atomic: bool) -> list[OutputVariableDef]:
                     OutputVariableCategory.REDU,
                     OutputVariableCategory.DERV_R,
                     OutputVariableCategory.DERV_C_REDU,
+                    OutputVariableCategory.DERV_R_DERV_R,
```
wanghan-iapcm, line 321: Adding `DERV_R_DERV_R` here without a `get_has_hessian()` guard makes **every** JAX model request `energy_derv_r_derv_r`, including ones frozen without hessian. `self.output_def` is the consumer def from `DeepPot.output_def`, which hard-codes `r_hessian=True`, so this category is always present (it is not gated on `hessian_mode`). For a non-hessian model the stablehlo/savedmodel never produces that key, so `_eval_model` falls into the fill branch and allocates `np.full([nframes, 3*natoms, 3*natoms], nan)` — an O(N^2) array (~72 MB at 1000 atoms) on the common non-atomic force/energy path — which `DeepPot.eval` then discards because `get_has_hessian()` is False.

This reintroduces the regression fixed for the PT backend in #5045 ("fix: remove hessian outdef if not necessary", 87cb6ef0d). The PR already adds `get_has_hessian()` (below), so the fix is to mirror pt (`deepmd/pt/infer/deep_eval.py`) and drop `DERV_R_DERV_R` from the request defs when `not self.get_has_hessian()`.

wanghan-iapcm (approved):

PR #5613, opened 2026-06-30
fix(jax): restore jax2tf savedmodel export

njzjz:
## Summary

- restore the JAX `.savedmodel` conversion path to use `jax2tf.convert(...)` instead of the TF2 SavedModel exporter
- keep the TF2 eager exporter on the `.savedmodeltf` suffix
- restore the graph-safe TensorFlow helper code used around the jax2tf-converted model body and document why these helpers should not be collapsed into TF2/ndtensorflow shims

## Background

`source/tests/infer/convert-models.sh` documents `.savedmodel` as the JAX/JAX2TF output suffix and `.savedmodeltf` as the TF2 output suffix. After #5598, the JAX `.savedmodel` branch delegated to the TF2 exporter, so freshly exported `.savedmodel` and `.savedmodeltf` artifacts had the same ordinary TF op structure and no `XlaCallModule` nodes.

This restores the historical contract: `.savedmodel` is a JAX/jax2tf artifact and should contain `XlaCallModule`; `.savedmodeltf` remains the TF2 eager SavedModel artifact.

## TF2 JIT note

I also checked why `DP_JIT=1` on the TF2 exporter does not create `XlaCallModule` nodes. A minimal `tf.function(jit_compile=True)` SavedModel in TF 2.21 stores `_XlaMustCompile: true` on the FunctionDef/PartitionedCall, but the serialized graph still contains ordinary TF ops and no `XlaCallModule`. In contrast, a minimal `jax2tf.convert(...)` SavedModel serializes `XlaCallModule` nodes. So `XlaCallModule` is a marker for the jax2tf native serialization path, not for generic TF2 `jit_compile=True`.

## Validation

- `dp convert-backend source/tests/infer/deeppot_sea.yaml /tmp/.../deeppot_sea.savedmodel`, parsed `saved_model.pb`: 8 `XlaCallModule` ops
- `dp convert-backend source/tests/infer/deeppot_dpa.yaml /tmp/.../deeppot_dpa.savedmodel`, parsed `saved_model.pb`: 8 `XlaCallModule` ops
- `dp convert-backend source/tests/infer/deeppot_sea.yaml /tmp/.../deeppot_sea.savedmodeltf`, parsed `saved_model.pb`: 0 XLA op names, as expected for TF2 eager export
- `ruff format .`
- `ruff check .`

Please review, @wanghan-iapcm.

## Summary by CodeRabbit

* **New Features**
  * Enhanced JAX-based SavedModel export with additional execution endpoints and metadata queries (including neighbor-list/selection and output configuration).
  * Added in-graph coordinate transforms and improved neighbor-list/periodic ghost handling for export.

* **Bug Fixes**
  * Corrected neighbor-list padding/truncation, cutoff masking, and stable type-distinguished ordering.
  * Improved handling of virtual atoms and empty-cell periodic extension cases.

* **Tests**
  * Migrated neighbor-list/region tests to pure TensorFlow ops.
  * Added coverage for model call behavior and SavedModel export contents.

### DIFF
--- a/deepmd/jax/jax2tf/nlist.py
+++ b/deepmd/jax/jax2tf/nlist.py
@@ -1,53 +1,171 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
-"""Compatibility wrappers for TensorFlow neighbor-list helpers."""
+"""Neighbor-list helpers for the JAX/jax2tf SavedModel wrapper.
 
-from typing import (
-    Any,
-)
+These routines mirror the backend-independent neighbor-list logic, but keep it
+expressed directly as TensorFlow graph code. During SavedModel export the input
+sizes are symbolic tensors; using ndtensorflow array-api helpers here can route
+dynamic ``shape``/``size`` checks through Python control flow before AutoGraph
+has converted them. That breaks tracing and also makes it easy to bypass the
+jax2tf/XlaCallModule export path.
+"""
 
 import tensorflow as tf
 
-from deepmd.tf2.common import (
-    to_tf_tensor,
-)
-from deepmd.tf2.utils._dpmodel import build_neighbor_list as tf2_build_neighbor_list
-from deepmd.tf2.utils._dpmodel import (
-    extend_coord_with_ghosts as tf2_extend_coord_with_ghosts,
+from .region import (
+    to_face_distance,
 )
 
-__all__ = [
-    "build_neighbor_list",
-    "extend_coord_with_ghosts",
-]
+
+def tf_take_along_axis(params: tf.Tensor, indices: tf.Tensor, axis: int) -> tf.Tensor:
+    return tf.gather(params, indices, batch_dims=axis)
 
 
 def build_neighbor_list(
-    coord: Any,
-    atype: Any,
+    coord: tf.Tensor,
+    atype: tf.Tensor,
     nloc: int,
     rcut: float,
     sel: int | list[int],
     distinguish_types: bool = True,
 ) -> tf.Tensor:
-    return to_tf_tensor(
-        tf2_build_neighbor_list(
-            coord,
-            atype,
-            nloc,
-            rcut,
-            sel,
-            distinguish_types=distinguish_types,
-        )
+    """Build neighbor list for a single frame. Keeps nsel neighbors."""
+    batch_size = tf.shape(coord)[0]
+    coord = tf.reshape(coord, (batch_size, -1))
+    nall = tf.shape(coord)[1] // 3
+    # Fill virtual atoms with large coords so they are not neighbors of any
+    # real atom.
+    if tf.size(coord) > 0:
+        xmax = tf.reduce_max(coord) + tf.cast(2.0 * rcut, coord.dtype)
+    else:
+        xmax = tf.cast(2.0 * rcut, coord.dtype)
+    is_vir = atype < 0
+    coord1 = tf.where(
+        is_vir[:, :, None], xmax, tf.reshape(coord, (batch_size, nall, 3))
     )
+    coord1 = tf.reshape(coord1, (batch_size, nall * 3))
+    if isinstance(sel, int):
+        sel = [sel]
+    nsel = sum(sel)
+    coord0 = coord1[:, : nloc * 3]
+    diff = (
+        tf.reshape(coord1, [batch_size, -1, 3])[:, None, :, :]
+        - tf.reshape(coord0, [batch_size, -1, 3])[:, :, None, :]
+    )
+    rr = tf.linalg.norm(diff, axis=-1)
+    # If central atom has two zero distances, sorting sometimes can not exclude
+    # itself.
+    rr -= tf.eye(nloc, nall, dtype=diff.dtype)[None, :, :]
+    nlist = tf.cast(tf.argsort(rr, axis=-1), tf.int64)
+    rr = tf.sort(rr, axis=-1)
+    rr = rr[:, :, 1:]
+    nlist = nlist[:, :, 1:]
+    nnei = tf.shape(rr)[2]
+    if nsel <= nnei:
+        rr = rr[:, :, :nsel]
+        nlist = nlist[:, :, :nsel]
+    else:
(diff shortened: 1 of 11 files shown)


deepmd/jax/jax2tf/region.py
```diff
@@ -1,33 +1,56 @@
...
+
+def b_to_face_distance(cell: tf.Tensor) -> tf.Tensor:
+    volume = tf.linalg.det(cell)
+    c_yz = tf.linalg.cross(cell[:, 1, ...], cell[:, 2, ...])
+    h2yz = volume / tf.linalg.norm(c_yz, axis=-1)
+    c_zx = tf.linalg.cross(cell[:, 2, ...], cell[:, 0, ...])
+    h2zx = volume / tf.linalg.norm(c_zx, axis=-1)
+    c_xy = tf.linalg.cross(cell[:, 0, ...], cell[:, 1, ...])
+    h2xy = volume / tf.linalg.norm(c_xy, axis=-1)
+    return tf.stack([h2yz, h2zx, h2xy], axis=1)
```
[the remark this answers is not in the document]
njzjz: Fixed in ecc6d204d. `b_to_face_distance` now uses `tf.abs(tf.linalg.det(cell))`, so left-handed cells produce positive face distances. I also checked a left-handed diagonal cell locally and got positive distances.

deepmd/jax/jax2tf/serialization.py
```diff
@@ -1,8 +1,320 @@
...
+                json.dumps(model_def_script, separators=(",", ":")), dtype=tf.string
+            )
+
+        tf_model.get_model_def_script = get_model_def_script
+
+        @tf.function
+        def has_message_passing() -> tf.Tensor:
+            return tf.constant(model.has_message_passing(), dtype=tf.bool)
+
+        tf_model.has_message_passing = has_message_passing
```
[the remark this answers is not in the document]
njzjz: Fixed in ecc6d204d. `tf_model.do_message_passing` now points to the same `tf.function` as `tf_model.has_message_passing`, matching the TF2 SavedModel exporter and keeping the legacy endpoint available.

njzjz (commented):
Addressed the CodeRabbit review comments in ecc6d204d.

deepmd/jax/jax2tf/nlist.py
```diff
@@ -1,53 +1,171 @@
...
-    return tuple(
-        to_tf_tensor(value)
-        for value in tf2_extend_coord_with_ghosts(coord, atype, cell, rcut)
+    """Extend atom coordinates by appending periodic images."""
+    atype_shape = tf.shape(atype)
+    nf, nloc = atype_shape[0], atype_shape[1]
+    # int64 for index
+    aidx = tf.range(nloc, dtype=tf.int64)
+    aidx = tf.tile(aidx[None, :], [nf, 1])
+    if tf.shape(cell)[-1] == 0:
```
wanghan-iapcm, line 136: Non-blocking (test coverage): the tnp -> plain-TF rewrite reintroduces several reachable branches that the updated `source/jax2tf_tests/` suite does not exercise:

- the non-PBC / empty-cell branch here (`if tf.shape(cell)[-1] == 0`) — reachable for nopbc inference, but every test passes a real 3x3 cell, so only the with-ghost branch runs;
- `build_neighbor_list`'s pad branch (`nsel > nnei`, the `else` of the `nsel <= nnei` check above) — reachable for sparse systems; the tests only hit the truncate branch;
- the entire `model_call_from_call_lower` wrapper in `make_model.py`, including its box vs. no-box branch (`if tf.shape(bb)[-1] != 0`) — no unit test;
- `region.normalize_coord` / `phys2inter` — `test_region.py` covers only `inter2phys` / `to_face_distance`.

A few small cases (an empty-cell call, an `nsel > nnei` call, and a `model_call_from_call_lower` round-trip) would close these.

deepmd/jax/jax2tf/serialization.py
```diff
@@ -1,8 +1,325 @@
...
+            # do not work, such as nall; nloc + nghost - 1.
+            if has_ghost_atoms:
+                nghost = "nghost"
+            else:
+                nghost = "0"
+            # The converted function is the part that carries the JAX model
+            # semantics into TensorFlow. Its SavedModel graph is expected to
+            # contain XlaCallModule ops; a graph made only of ordinary TF ops
+            # means this path has accidentally fallen back to the TF2 exporter.
+            return jax2tf.convert(
```
wanghan-iapcm, line 77: Non-blocking (regression guard): this PR exists because #5598 silently made the jax2tf path fall back to the TF2 eager exporter, so `.savedmodel` no longer contained `XlaCallModule` and behaved like an ordinary-TF graph. The restored path is verified manually (per the PR description), but there is no committed test asserting that a freshly-exported `.savedmodel` actually contains `XlaCallModule` ops — and `source/tests/consistent/io/test_io.py` skips the jax2tf savedmodel path. A small test that exports a tiny model and asserts the SavedModel graph contains an `XlaCallModule` node would guard against the exact regression this PR fixes from recurring.

deepmd/jax/jax2tf/nlist.py
njzjz: Added focused jax2tf tests covering the empty-cell/no-PBC extension path, neighbor-list padding when `nsel > nnei`, graph-mode `model_call_from_call_lower` with both box and no-box inputs, and `phys2inter`/`normalize_coord`. Verified with `venv/bin/python -m pytest source/jax2tf_tests -v` and `venv/bin/python -m ruff check .`.

deepmd/jax/jax2tf/serialization.py
njzjz: Added a regression test that exports a lightweight dummy JAX model through the actual jax2tf SavedModel exporter and parses `saved_model.pb` to assert an `XlaCallModule` op is present. Verified locally with full JAX deps via `venv/bin/python -m pytest source/jax2tf_tests -v`.

wanghan-iapcm (approved):

PR #5702, opened 2026-07-01
fix(tf): accept standard checkpoint inputs in change-bias

### LINKED ISSUE #5683: [Code scan] Accept standard TensorFlow checkpoint inputs in change-bias

## Problem

The CLI examples document TensorFlow `change-bias` with a checkpoint-style input:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/main.py#L824-L833

The frozen-model fallback message also recommends passing a checkpoint directory:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/tf/entrypoints/change_bias.py#L303-L309

But the dispatcher only accepts inputs ending in `.pb`, `.pbtxt`, `.ckpt`, `.meta`, `.data`, or `.index`:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/tf/entrypoints/change_bias.py#L92-L130

Real TensorFlow checkpoint prefixes commonly look like `model.ckpt-1000`, and checkpoint directories have no such suffix. Those inputs are rejected before `_change_bias_checkpoint_file()` can inspect the `checkpoint` state file:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/tf/entrypoints/change_bias.py#L149-L163

## Impact

Users following the documented workaround with a checkpoint directory, or passing a standard checkpoint prefix with a step suffix, get `RuntimeError("The model provided must be a checkpoint file or frozen model file (.pb)")` before checkpoint loading starts.

## Suggested fix

Accept checkpoint directories that contain a TensorFlow `checkpoint` state file, and accept checkpoint prefixes such as `model.ckpt-1000` by checking for sibling `.meta`/`.index` files or by asking TensorFlow for checkpoint state instead of relying only on filename suffixes.

wanghan-iapcm:
## Problem

The TensorFlow `change-bias` dispatcher only accepts inputs ending in `.pb`, `.pbtxt`, `.ckpt`, `.meta`, `.data`, or `.index`, and rejects everything else with `RuntimeError("The model provided must be a checkpoint file or frozen model file (.pb)")`.

However, the CLI examples and the frozen-model fallback message recommend passing a checkpoint directory, and real TensorFlow checkpoint prefixes commonly look like `model.ckpt-1000`. A checkpoint directory has no recognized suffix, and a `model.ckpt-1000` prefix ends in `-1000`, so both are rejected before `_change_bias_checkpoint_file()` — which already resolves the checkpoint via the `checkpoint` state file — can run.

## Fix

Route suffix-less inputs to the checkpoint handler when a TensorFlow `checkpoint` state file is present in the effective directory (the input itself if it is a directory, otherwise its parent). Inputs without such a state file still raise the original error. `_change_bias_checkpoint_file()` now also resolves `checkpoint_dir` correctly when the input is a directory.

## Test

`source/tests/tf/test_change_bias.py` previously exercised only `.pb`, `.ckpt`, and bad-suffix inputs, never a bare directory or a `ckpt-<step>` prefix. Two new tests mock `_change_bias_checkpoint_file` and assert it is invoked for a `model.ckpt-1000` prefix and for a checkpoint directory that contain a `checkpoint` state file. On the current dispatcher both raise the "checkpoint file or frozen model file" RuntimeError before reaching the handler; after the fix both route correctly. The existing rejection tests (e.g. a `model.xyz` file with no checkpoint state file) still raise.

Fix #5683

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of checkpoint inputs so bias changes now work reliably for both checkpoint file prefixes and checkpoint directories.
  * Unrecognized input paths are now treated as checkpoint-based inputs only when a valid checkpoint state is present, reducing incorrect routing.

* **Tests**
  * Added coverage for checkpoint prefix and directory cases to verify the correct bias-change path is used.

### DIFF
--- a/deepmd/tf/entrypoints/change_bias.py
+++ b/deepmd/tf/entrypoints/change_bias.py
@@ -126,6 +126,23 @@ def change_bias(
             log_level,
         )
     else:
+        # Checkpoint directory, or a checkpoint prefix such as "model.ckpt-1000"
+        # that carries no recognized suffix. Route these to the checkpoint
+        # handler when a TensorFlow "checkpoint" state file is present.
+        input_path = Path(INPUT)
+        checkpoint_dir = input_path if input_path.is_dir() else input_path.parent
+        if (checkpoint_dir / "checkpoint").is_file():
+            return _change_bias_checkpoint_file(
+                INPUT,
+                mode,
+                bias_value,
+                datafile,
+                system,
+                numb_batch,
+                model_branch,
+                output,
+                log_level,
+            )
         raise RuntimeError(
             "The model provided must be a checkpoint file or frozen model file (.pb)"
         )
@@ -147,7 +164,9 @@ def _change_bias_checkpoint_file(
     tf.reset_default_graph()
 
     checkpoint_path = Path(checkpoint_prefix)
-    checkpoint_dir = checkpoint_path.parent
+    checkpoint_dir = (
+        checkpoint_path if checkpoint_path.is_dir() else checkpoint_path.parent
+    )
 
     # Check for valid checkpoint and find the actual checkpoint path
     checkpoint_state_file = checkpoint_dir / "checkpoint"
--- a/source/tests/tf/test_change_bias.py
+++ b/source/tests/tf/test_change_bias.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
+import importlib
 import json
 import os
 import shutil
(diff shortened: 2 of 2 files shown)


njzjz (approved):
LGTM

PR #5714, opened 2026-07-02
feat(pt_expt): O(N) on-device NeighborGraph builders — vesin & nvalchemiops (PR-C)

wanghan-iapcm:
## NeighborGraph PR-C — O(N) on-device graph builders (vesin / nv)

Third PR of the NeighborGraph series (after #5581 foundation, #5583 PR-A dpa1 graph forward, #5604 PR-B `.pt2`/C++). Adds two **O(N) carry-all** NeighborGraph builders behind the World-2 `neighbor_graph_method` dispatcher, replacing PR-A's O(N²) `dense` search / per-frame ASE stopgap with on-device cell lists.

### What

- **`build_neighbor_graph_vesin`** (`deepmd/pt_expt/utils/vesin_graph_builder.py`) — `vesin.torch` cell list. **Device-following**: runs the search on the input tensor's device (CUDA kernel on CUDA input, CPU cell list otherwise).
- **`build_neighbor_graph_nv`** (`deepmd/pt_expt/utils/nv_graph_builder.py`) — nvalchemiops GPU cell list, **frame-batched** (`batch_idx`/`batch_ptr`, one kernel for all frames — no Python loop). CUDA-only.
- Both structurally clone `build_neighbor_graph_ase`: search → per-frame local `(i, j, S)` → `neighbor_graph_from_ijs(...)`, which recomputes `edge_vec` **differentiably** from the original grad-carrying coords.
- Wired into the pt_expt make_model graph dispatch (`neighbor_graph_method ∈ {"legacy","dense","ase","vesin","nv"}`) and DeepEval `.pt2` graph inference (new `neighbor_graph_method` kwarg, default `"dense"` → existing inference byte-identical). dpmodel/jax fail-fast on `vesin`/`nv` (torch/CUDA-only).

**Perf-only:** all builders emit the SAME neighbor set as `dense` (carry-all, `sel`=normalization-only), proven by exact set-equality; energy/force/virial are unchanged (parity 1e-12 CPU / 1e-10 CUDA).

### Layering

- `dpmodel` stays torch-free: vesin/nv builders live in `pt_expt`; the dpmodel dispatch only carries a fail-fast message.
- vesin/nv are **optional deps, NOT in pyproject** — lazy-imported, guarded by `is_vesin_torch_available()` / `is_nv_available()`, `ImportError` with an install hint on absence.
- nv decode is a faithful transcription of the tested `deepmd/pt/utils/nv_nlist.py:_matrix_to_extended_inputs` Step-1 extraction.

### Testing

- **Local (CPU):** 13 passed (vesin builder 5, vesin/reject dispatch 2, DeepEval graph 6), nv self-skips.
- **Remote GPU (Tesla T4), commit 78f6c24f:**
  - `test_nv_graph_builder.py`: 4 passed (set-equality vs dense periodic+non-periodic, frame-batch, differentiable `edge_vec`).
  - `test_graph_builder_dispatch.py`: 3 passed on CUDA (**vesin** parity 1e-10, dpmodel reject vesin+nv, **nv** parity 1e-10).
  - `test_graph_deepeval.py`: 6 passed on CUDA (`.pt2` graph dense parity + vesin, AOTI compile + `torch.as_tensor` extraction).

### Known limitations

- **vesin per-frame Python loop** — `vesin.torch.compute` is single-system (no batch dim), so multi-frame vesin loops over frames (each an O(N) search; `nf=1` inference has zero loop cost). **nv batches natively** — the loop-free path for batched training.
- vesin/nv not in `pyproject`; no `"auto"` selector (explicit strings only).
- nv `search_capacity = max(64, nloc)` initial heuristic + 1.25× grow loop (`.item()` host-sync per grow).
- nv passes the *normalized* coord to `from_ijs` (differentiable lattice translation, identity gradient; `edge_vec` is lattice-invariant) — consistent with the pt nv-nlist path.
- jax O(N) graph builders (matscipy/jax-md) = PR-F; attention/angles/MP = PR-D/E/G.

Implements `plan_neighbor_graph_prC_implementation`; design spec: discussion wanghan-iapcm/deepmd-kit#4.

## Summary by CodeRabbit

* **New Features**
  * Added experimental neighbor-graph builders supporting NVIDIA CUDA (NV) and Vesin (vesin), including periodic handling and frame-aware edge construction.
  * Enabled configurable `neighbor_graph_method` for graph-form `.pt2` inference and tightened dispatch support to `dense/ase/vesin/nv`.
* **Bug Fixes**
  * Excluded virtual atoms (`type < 0`) consistently from neighbor-graph edges for ASE, Vesin, and NV paths.
  * Improved robustness for empty/zero-neighbor cases and ensured edge vectors remain differentiable.
* **Tests**
  * Added backend parity, NV decode regression, gradient/device checks, virtual-atom coverage, and strengthened plugin import validation.

### DIFF
--- a/deepmd/pt_expt/infer/deep_eval.py
+++ b/deepmd/pt_expt/infer/deep_eval.py
@@ -65,6 +65,10 @@ from deepmd.pt_expt.utils.vesin_neighbor_list import (
 if TYPE_CHECKING:
     import ase.neighborlist
 
+    from deepmd.dpmodel.utils.neighbor_graph import (
+        NeighborGraph,
+    )
+
 
 # Public output keys emitted by the graph-form AOTI forward
 # (``forward_lower_graph_exportable``) keyed by the output-variable category that
@@ -137,11 +141,15 @@ class DeepEval(DeepEvalBackend):
         auto_batch_size: bool | int | AutoBatchSize = True,
         neighbor_list: Optional["ase.neighborlist.NewPrimitiveNeighborList"] = None,
         nlist_backend: str = "auto",
+        neighbor_graph_method: str = "dense",
         **kwargs: Any,
     ) -> None:
         self.output_def = output_def
         self.model_path = model_file
         self.neighbor_list = neighbor_list
+        # World-2 graph-form ``.pt2`` (lower_input_kind == "graph") builder select:
+        # "dense"/"ase" (backend-agnostic) or "vesin"/"nv" (on-device O(N)).
+        self._neighbor_graph_method = neighbor_graph_method
         self._is_pt2 = model_file.endswith(".pt2")
 
         if self._is_pt2:
@@ -1660,9 +1668,6 @@ class DeepEval(DeepEvalBackend):
         forward returns the LOCAL public keys directly, so results are reshaped
         without ``communicate_extended_output``.
         """
-        from deepmd.dpmodel.utils.neighbor_graph import (
-            build_neighbor_graph,
-        )
         from deepmd.pt_expt.utils.env import (
             DEVICE,
         )
@@ -1678,28 +1683,19 @@ class DeepEval(DeepEvalBackend):
         box_input = cells.reshape(nframes, 9) if cells is not None else None
         # Dynamic edge axis (B2.0): build the carry-all graph at its exact edge
         # count (no static padding); the AOTI artifact accepts any E.
-        graph = build_neighbor_graph(
-            coord_input,
-            atom_types,
-            box_input,
-            self._rcut,
-        )
+        graph = self._build_eval_graph(coord_input, atom_types, box_input, DEVICE)
 
         atype_t = torch.tensor(
             np.asarray(atom_types).reshape(-1), dtype=torch.int64, device=DEVICE
         )
-        n_node_t = torch.tensor(
-            np.asarray(graph.n_node), dtype=torch.int64, device=DEVICE
+        # graph fields may be numpy (dense/ase) or torch, possibly on CUDA
+        # (vesin/nv) -- torch.as_tensor handles both and moves to DEVICE.
+        n_node_t = torch.as_tensor(graph.n_node, dtype=torch.int64, device=DEVICE)
+        edge_index_t = torch.as_tensor(
+            graph.edge_index, dtype=torch.int64, device=DEVICE
         )
-        edge_index_t = torch.tensor(
-            np.asarray(graph.edge_index), dtype=torch.int64, device=DEVICE
-        )
-        edge_vec_t = torch.tensor(
-            np.asarray(graph.edge_vec), dtype=torch.float64, device=DEVICE
-        )
-        edge_mask_t = torch.tensor(
-            np.asarray(graph.edge_mask), dtype=torch.bool, device=DEVICE
-        )
+        edge_vec_t = torch.as_tensor(graph.edge_vec, dtype=torch.float64, device=DEVICE)
+        edge_mask_t = torch.as_tensor(graph.edge_mask, dtype=torch.bool, device=DEVICE)
 
         fparam_t, aparam_t = self._prepare_optional_lower_inputs(
             fparam, aparam, nframes, natoms, DEVICE
@@ -1733,6 +1729,61 @@ class DeepEval(DeepEvalBackend):
                     np.full(np.abs(shape), np.nan, dtype=GLOBAL_NP_FLOAT_PRECISION)
                 )
         return tuple(results)
+
+    def _build_eval_graph(
+        self,
+        coord_input: np.ndarray,
+        atom_types: np.ndarray,
+        box_input: np.ndarray | None,
+        device: "torch.device",
+    ) -> "NeighborGraph":
+        """Build the carry-all NeighborGraph for graph-form ``.pt2`` inference.
+
+        Dispatches on ``self._neighbor_graph_method``: ``dense``/``ase`` run
+        backend-agnostic (numpy); ``vesin``/``nv`` run on-device (torch, O(N)).
+        All backends emit the SAME neighbor set (carry-all, sel-free), so the
+        selection is a pure performance choice and results are unchanged.
+        """
+        method = self._neighbor_graph_method
+        if method == "dense":
+            from deepmd.dpmodel.utils.neighbor_graph import (
+                build_neighbor_graph,
+            )
+
+            return build_neighbor_graph(coord_input, atom_types, box_input, self._rcut)
+        if method == "ase":
+            from deepmd.dpmodel.utils.neighbor_graph import (
+                build_neighbor_graph_ase,
+            )
+
+            return build_neighbor_graph_ase(
+                coord_input, atom_types, box_input, self._rcut
+            )
+        if method in ("vesin", "nv"):
+            cc = torch.as_tensor(coord_input, dtype=torch.float64, device=device)
+            aa = torch.as_tensor(
+                np.asarray(atom_types), dtype=torch.int64, device=device
+            )
+            bb = (
+                torch.as_tensor(box_input, dtype=torch.float64, device=device)
+                if box_input is not None
+                else None
+            )
+            if method == "vesin":
+                from deepmd.pt_expt.utils.vesin_graph_builder import (
+                    build_neighbor_graph_vesin,
+                )
+
+                return build_neighbor_graph_vesin(cc, aa, bb, self._rcut)
+            from deepmd.pt_expt.utils.nv_graph_builder import (
+                build_neighbor_graph_nv,
+            )
+
+            return build_neighbor_graph_nv(cc, aa, bb, self._rcut)
+        raise ValueError(
+            f"unknown neighbor_graph_method {method!r}; "
+            "use 'dense', 'ase', 'vesin', or 'nv'"
+        )
 
     def _get_output_shape(
         self, odef: OutputVariableDef, nframes: int, natoms: int
--- a/deepmd/pt_expt/infer/deep_eval.py
+++ b/deepmd/pt_expt/infer/deep_eval.py
@@ -129,6 +129,19 @@ class DeepEval(DeepEvalBackend):
     neighbor_list : ase.neighborlist.NewPrimitiveNeighborList, optional
         The ASE neighbor list class to produce the neighbor list. If None, the
         neighbor list will be built natively in the model.
+    nlist_backend : str, default: "auto"
+        Neighbor-list builder for the NLIST/extended lower path (``.pte`` and
+        nlist-form ``.pt2``): ``"auto"`` / ``"vesin"`` / ``"native"``. Not
+        used by graph-form ``.pt2`` artifacts.
+    neighbor_graph_method : str, default: "dense"
+        Carry-all graph builder for GRAPH-FORM ``.pt2`` artifacts ONLY
+        (``metadata["lower_input_kind"] == "graph"``): ``"dense"`` / ``"ase"``
+        (backend-agnostic) or ``"vesin"`` / ``"nv"`` (on-device O(N)). A
+        non-default value on any other artifact raises at construction — the
+        knob would silently do nothing there; use ``nlist_backend`` for the
+        nlist path instead. All builders emit the same neighbor set, so the
+        choice is performance-only. Consolidating the two knobs into a single
+        backend-selection API is deferred to the dense-nlist deprecation.
     **kwargs : dict
         Keyword arguments.
     """
@@ -163,6 +176,20 @@ class DeepEval(DeepEvalBackend):
                 f"Unsupported model file '{model_file}' for the pt_expt "
                 "backend: expected `.pt2` / `.pte` (deployable archives) or "
                 "`.pt` (training checkpoint)."
+            )
+
+        # neighbor_graph_method is consumed ONLY by graph-form .pt2 eval
+        # (_eval_model_graph); fail fast instead of silently ignoring it on
+        # nlist-form artifacts (there, the builder knob is nlist_backend).
+        if (
+            neighbor_graph_method != "dense"
+            and getattr(self, "metadata", {}).get("lower_input_kind") != "graph"
+        ):
+            raise ValueError(
+                f"neighbor_graph_method={neighbor_graph_method!r} only applies to "
+                "graph-form .pt2 artifacts (lower_input_kind == 'graph'); this "
+                f"model is not graph-form. Use nlist_backend to select the "
+                "neighbor-list builder for the nlist path."
             )
 
         self._setup_nlist_backend(nlist_backend)
--- a/deepmd/pt_expt/model/make_model.py
+++ b/deepmd/pt_expt/model/make_model.py
@@ -456,6 +456,17 @@ def make_model(
                 build_neighbor_graph_ase,
             )
 
+            # mirror the dpmodel guard: _resolve_graph_method's eligibility
+            # check only protects the default (None) path; an EXPLICIT
+            # neighbor_graph_method would otherwise reach the builders for
+            # descriptors without a graph lower.
+            descriptor = getattr(self.atomic_model, "descriptor", None)
+            uses_graph_lower = getattr(descriptor, "uses_graph_lower", lambda: False)
+            if not (self.mixed_types() and uses_graph_lower()):
+                raise NotImplementedError(
+                    "neighbor_graph_method requires a mixed_types descriptor with a "
+                    "graph lower (e.g. dpa1 attn_layer=0)"
+                )
             rcut = self.get_rcut()
             if method == "dense":
                 ng = build_neighbor_graph(cc, atype, bb, rcut)
--- a/deepmd/pt_expt/model/make_model.py
+++ b/deepmd/pt_expt/model/make_model.py
@@ -461,9 +461,16 @@ def make_model(
                 ng = build_neighbor_graph(cc, atype, bb, rcut)
             elif method == "ase":
                 ng = build_neighbor_graph_ase(cc, atype, bb, rcut)
+            elif method == "vesin":
+                from deepmd.pt_expt.utils.vesin_graph_builder import (
+                    build_neighbor_graph_vesin,
+                )
+
+                ng = build_neighbor_graph_vesin(cc, atype, bb, rcut)
             else:
                 raise ValueError(
-                    f"unknown neighbor_graph_method {method!r}; use 'dense' or 'ase'"
+                    f"unknown neighbor_graph_method {method!r}; "
+                    "use 'dense', 'ase', 'vesin', or 'nv'"
                 )
             nf, nloc = atype.shape[:2]
             atype_flat = atype.reshape(nf * nloc)
--- a/deepmd/pt_expt/model/make_model.py
+++ b/deepmd/pt_expt/model/make_model.py
@@ -467,6 +467,12 @@ def make_model(
                 )
 
                 ng = build_neighbor_graph_vesin(cc, atype, bb, rcut)
+            elif method == "nv":
+                from deepmd.pt_expt.utils.nv_graph_builder import (
+                    build_neighbor_graph_nv,
+                )
+
+                ng = build_neighbor_graph_nv(cc, atype, bb, rcut)
             else:
                 raise ValueError(
                     f"unknown neighbor_graph_method {method!r}; "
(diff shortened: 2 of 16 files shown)


deepmd/pt_expt/utils/vesin_graph_builder.py
```diff
@@ -0,0 +1,115 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+"""Carry-all NeighborGraph builder backed by vesin.torch (O(N) cell list).
+
+World-2 counterpart of vesin_neighbor_list.py: instead of building the dense
+quartet, it returns per-frame local (i, j, S), then delegates to the array-API
+``neighbor_graph_from_ijs`` (which recomputes ``edge_vec`` differentiably from
+the ORIGINAL grad-carrying coords). torch-only => lives in pt_expt.
```
iProzd, line 7: vesin multi-frame loops over frames in Python while nv batches natively. Could we confirm that batched GPU training routes through `nv` (not `vesin`) so the per-frame loop is not on the training hot path, and maybe document that `vesin` is intended for `nf=1` inference / CPU?

deepmd/pt_expt/utils/nv_graph_builder.py
```diff
@@ -0,0 +1,176 @@
...
+        return neighbor_graph_from_ijs(
+            empty_i, empty_i, empty_S, coord, box, empty_i, nloc, layout=layout
+        )
+
+    # Mirror nv_nlist.build's search setup: normalize periodic coords (a
+    # differentiable lattice translation, identity gradient), flatten the batch,
+    # and search all frames in ONE kernel via batch_idx/batch_ptr.
+    if periodic:
+        cell = box.reshape(nf, 3, 3).to(device=device, dtype=coord.dtype)
+        coord = normalize_coord(coord, cell)
```
iProzd: Minor consistency note: `build_neighbor_graph_nv` normalizes coords before `neighbor_graph_from_ijs` while `build_neighbor_graph_vesin` passes the original unnormalized coords. Both are internally self-consistent (S matches the coords each search used), but a one-line comment on each explaining why they differ would help future readers.

deepmd/pt_expt/utils/nv_graph_builder.py
```diff
@@ -0,0 +1,176 @@
...
+    dst = edge_idx // max_neighbors  # flattened center
+    src = neighbor_matrix.reshape(-1).index_select(0, edge_idx).to(torch.int64)
+    shift = shifts.reshape(-1, 3).index_select(0, edge_idx).to(torch.int64)
+    frame_idx = (dst // nloc).to(torch.int64)  # frame of the edge
+    center_local = (dst % nloc).to(torch.int64)  # i = center
+    src_local = (src % nloc).to(torch.int64)  # j = neighbor
+
+    return neighbor_graph_from_ijs(
+        center_local, src_local, shift, coord, box_out, frame_idx, nloc, layout=layout
+    )
```
iProzd: This decode (`matrix -> (i, j, S)`) is pure torch and CPU-runnable, but it currently lives inside the CUDA-only `build_neighbor_graph_nv`, so it only runs under the opt-in CUDA suite. Could we extract it into a standalone helper (taking `neighbor_matrix`, `num_neighbors`, `shifts`, `nloc`, `coord`, `box`) and unit-test it on the default CPU CI with synthetic inputs? The GPU `neighbor_list` search stays opt-in, but the regression-prone arithmetic (`// max_neighbors`, `% nloc`, frame isolation, `neighbor_graph_from_ijs`) would then be behind the normal merge gate. Not blocking since `nv` is explicit opt-in.

source/tests/pt_expt/utils/test_nv_graph_builder.py
```diff
@@ -0,0 +1,106 @@
...
+
+nv_builder = pytest.importorskip("deepmd.pt_expt.utils.nv_graph_builder")
+from deepmd.pt.utils.nv_nlist import (
+    is_nv_available,
+)
+
+pytestmark = pytest.mark.skipif(
+    not (torch.cuda.is_available() and is_nv_available()),
+    reason="nvalchemiops requires CUDA + nvalchemi-toolkit-ops",
+)
```
iProzd, line 22: These CUDA integration tests correctly stay gated. The gap is separate: the CPU-runnable decode in `nv_graph_builder.py` has no default-CI test, so a shared-code refactor could regress it silently without the CUDA tag. Suggest a CPU unit test of the extracted decode (see my comment on the builder).

iProzd (changes requested):

deepmd/pt_expt/utils/vesin_graph_builder.py
wanghan-iapcm: Confirmed: batched training never routes through vesin by default — `neighbor_graph_method=None` resolves to the `"dense"` converter (make_model.py `_resolve_graph_method`), and vesin/nv are explicit opt-in only. Documented the scope in the module docstring in 690591849: vesin = nf==1 inference / CPU (per-frame loop ~1 ms/frame call overhead measured on a T4); batched multi-frame GPU work should opt into nv, which batches all frames in one kernel.

deepmd/pt_expt/utils/nv_graph_builder.py
wanghan-iapcm: Added cross-referencing comments in both builders in 690591849: nvalchemiops requires in-cell positions, so nv searches AND recomputes edge_vec on the normalized coords (S matches what the search saw; normalize_coord is a differentiable lattice translation with identity gradient); vesin's cell list handles unwrapped positions natively, so it keeps the original grad-carrying coords.

deepmd/pt_expt/utils/nv_graph_builder.py
wanghan-iapcm: Done in 690591849: extracted `nv_matrix_to_ijs(neighbor_matrix, num_neighbors, shifts, nloc) -> (i, j, S, frame_idx)` — pure torch, device-agnostic — and the builder now calls it. New default-CI CPU test `test_nv_matrix_decode.py` covers the regression-prone arithmetic with a hand-checked two-frame fixture (incl. stale-slot garbage that must be ignored), an all-empty matrix, and a random-vs-brute-force-oracle sweep. The GPU `neighbor_list` search stays behind the opt-in CUDA suite.

source/tests/pt_expt/utils/test_nv_graph_builder.py
wanghan-iapcm: Covered by the extraction in 690591849 — `nv_matrix_to_ijs` now has a default-CI CPU unit test (`source/tests/pt_expt/utils/test_nv_matrix_decode.py`: hand-checked, empty, random-vs-oracle), so a shared-code refactor of the decode can no longer regress silently without the CUDA tag.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -1734,6 +1730,61 @@ def _eval_model_graph(
...
+        coord_input: np.ndarray,
+        atom_types: np.ndarray,
+        box_input: np.ndarray | None,
+        device: "torch.device",
+    ) -> "NeighborGraph":
+        """Build the carry-all NeighborGraph for graph-form ``.pt2`` inference.
+
+        Dispatches on ``self._neighbor_graph_method``: ``dense``/``ase`` run
+        backend-agnostic (numpy); ``vesin``/``nv`` run on-device (torch, O(N)).
+        All backends emit the SAME neighbor set (carry-all, sel-free), so the
```
OutisLi, line 1771: The "SAME neighbor set" statement is accurate for normal real atoms, but it is worth documenting one boundary here: the dense builder filters `atype < 0` padding/placeholder atoms during graph construction, while the O(N) builders (`ase`/`vesin`/`nv`, matching the SeZM edge-schema style) build geometric edges first and rely on downstream masks/consumers to suppress non-real atoms. For inputs with mixed-type padding this is not literally the same edge set, so downstream graph consumers that need dense parity should explicitly mask/filter edges touching `atype < 0`, or this docstring should state that responsibility.

iProzd (commented):
Agree with @OutisLi, and I think this is a correctness gap for the opt-in O(N) builders, not just a docstring scope note. Confirmed in code:

- `build_neighbor_graph` (dense) drops virtual-atom edges via `keep_mask = ... & ~(extended_atype<0) & ~(atype<0)`.
- `vesin`/`nv`/`ase` build geometric edges from coords only and never look at `atype`.
- `forward_common_atomic_graph` masks edges only for `pair_excl`; for `atype<0` it only clamps the type and zeroes the *virtual atom's own* output — it does not drop edges from a virtual neighbor into a real center.

So for inputs with `atype<0` (e.g. mixed-natoms padding), a real center that has a virtual atom within `rcut` gets an extra type-0 edge under vesin/nv/ase but not under dense, polluting the real atom's descriptor/energy/force. The "SAME neighbor set / parity" claim then holds only for all-real-atom inputs, which is likely all the current set-equality fixtures cover.

Rather than only documenting the caller's responsibility, could the O(N) builders filter edges touching `atype<0` (drop where center or neighbor-owner is virtual) to match dense, so the parity claim stays literally true? A set-equality test with a virtual-atom fixture would lock it in. Default users are unaffected (default resolves to `dense`), so this is opt-in-only severity, but it undermines the stated dense parity.

deepmd/pt_expt/utils/nv_graph_builder.py
```diff
@@ -0,0 +1,222 @@
...
+    dst = edge_idx // max_neighbors  # flattened center
+    src = neighbor_matrix.reshape(-1).index_select(0, edge_idx).to(torch.int64)
+    shift = shifts.reshape(-1, 3).index_select(0, edge_idx).to(torch.int64)
+    frame_idx = (dst // nloc).to(torch.int64)  # frame of the edge
+    center_local = (dst % nloc).to(torch.int64)  # i = center
+    src_local = (src % nloc).to(torch.int64)  # j = neighbor
+    return center_local, src_local, shift, frame_idx
+
+
+def build_neighbor_graph_nv(
```
OutisLi, line 201: This should reuse NvNeighborList instead of introducing a parallel nvalchemiops implementation. NvNeighborList already owns the matrix search, capacity growth, device context, and extended/edges conversions used by the SeZM path. Adding build_neighbor_graph_nv separately means future fixes must be kept in sync manually, and this copy has already drifted by omitting the existing _input_device_context(device) guard. Please extend NvNeighborList with a graph return mode or a thin build_graph method that produces NeighborGraph, while preserving the current extended and edges return modes so existing SeZM behavior stays unchanged.

OutisLi (changes requested):
Please rework the nv graph path to reuse the existing NvNeighborList implementation rather than adding a second nvalchemiops search stack. The new graph builder duplicates the kernel setup, capacity loop, device handling, and matrix decode that NvNeighborList already owns, which increases maintenance cost and has already diverged in behavior. This can be fixed by extending NvNeighborList with a graph-oriented return mode or a thin build_graph method while preserving the existing extended/edges modes so current SeZM users are not affected.

iProzd:
Others LGTM.

deepmd/pt_expt/utils/vesin_graph_builder.py
```diff
@@ -0,0 +1,125 @@
...
+    GraphLayout,
+    NeighborGraph,
+    neighbor_graph_from_ijs,
+)
+from deepmd.pt_expt.utils.vesin_neighbor_list import (
+    is_vesin_torch_available,
+)
+
+
+def build_neighbor_graph_vesin(
```
OutisLi, line 86: Same structural issue as the nv path: this duplicates the existing VesinNeighborList search/conversion logic instead of extending the established NeighborList strategy. VesinNeighborList already owns the vesin.torch O(N) builder and supports extended/edges return modes for the SeZM path. Please make NeighborGraph another output contract of that shared strategy layer, or provide a thin build_graph method there, while preserving existing extended and edges behavior.

OutisLi (changes requested):
The same design concern applies to both nv and vesin: graph-form NeighborGraph construction should extend the existing NeighborList strategies instead of adding separate parallel builders. VesinNeighborList and NvNeighborList already provide the O(N) search backends used by SeZM through extended/edges return modes; the graph path should be another output contract on the same strategy layer, not a second copy of each backend. Please unify this design while preserving the existing extended and edges modes for current SeZM users.

deepmd/pt_expt/model/make_model.py
```diff
@@ -461,9 +461,22 @@ def _call_common_graph(
                 ng = build_neighbor_graph(cc, atype, bb, rcut)
             elif method == "ase":
                 ng = build_neighbor_graph_ase(cc, atype, bb, rcut)
+            elif method == "vesin":
```
OutisLi, line 475: This pt_expt override no longer performs the graph-eligibility check that dpmodel keeps before building a NeighborGraph. `_resolve_graph_method()` only protects the default `neighbor_graph_method=None` path; an explicit `neighbor_graph_method="dense"`, `"ase"`, `"vesin"`, or `"nv"` reaches this dispatch even for descriptors that do not support the graph lower (for example non-mixed descriptors, DPA1 with attention layers, or other configs where `uses_graph_lower()` is false). Please restore the same `mixed_types() and descriptor.uses_graph_lower()` fail-fast check here before dispatching to any builder, and add a regression test with a non-graph-eligible descriptor explicitly passing one of these methods.

OutisLi (changes requested):
Please also restore the graph-lower eligibility check in the pt_expt graph dispatch. Explicit neighbor_graph_method values should not bypass the same mixed_types/uses_graph_lower contract enforced by dpmodel and by graph export.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -137,11 +141,15 @@ def __init__(
         auto_batch_size: bool | int | AutoBatchSize = True,
         neighbor_list: Optional["ase.neighborlist.NewPrimitiveNeighborList"] = None,
         nlist_backend: str = "auto",
+        neighbor_graph_method: str = "dense",
```
OutisLi, line 157: Please document that `neighbor_graph_method` is graph-form only. For regular `.pte` / nlist-form `.pt2` inference this value is currently ignored because `_eval_model()` only dispatches to `_eval_model_graph()` when `metadata["lower_input_kind"] == "graph"`; users who want the nlist path O(N) builder must use `nlist_backend`, not `neighbor_graph_method`. It would be clearer to either fail fast / warn when `neighbor_graph_method != "dense"` on a non-graph artifact, or spell out this no-op behavior in the constructor docstring.

OutisLi, line 157: This adds a second public neighbor-builder knob next to `nlist_backend`, but the names and value sets do not make the distinction clear. `nlist_backend` controls the nlist/extended lower path and accepts `auto`/`vesin`/`native`; `neighbor_graph_method` controls only graph-form `.pt2` and accepts `dense`/`ase`/`vesin`/`nv`. A user who wants an O(N) backend cannot tell which knob to use, and `nv` oddly works only on the graph path even though NvNeighborList already exists for the nlist strategy layer. Please consolidate this into a single backend-selection API, or otherwise rename/fail-fast/document the two paths clearly, and make `nv` support consistent across both nlist and graph lower schemas without breaking the existing SeZM extended/edges behavior.

OutisLi (changes requested):
Please revisit the public neighbor-backend API design. The PR now exposes two knobs, nlist_backend and neighbor_graph_method, whose scopes are hard to infer from their names and whose supported values differ. From a user perspective both choose the neighbor construction backend, so the API should be made more natural and unambiguous. In particular, if nv is a supported O(N) backend, it should not work only for graph-form .pt2 while the nlist path has no nv option, especially since NvNeighborList already exists.

deepmd/pt_expt/infer/deep_eval.py
OutisLi, line 157: One performance note for the eventual unified/auto backend selection: in the SeZM path we found that CPU + `nf == 1` is better served by `vesin` than `nvalchemiops`, because vesin has the lighter single-system cell-list path while nvalchemiops/Warp has more setup overhead to amortize. The current SeZM selector therefore uses `vesin` for `device.type == "cpu" and nf == 1`, and prefers `nv` for CUDA or multi-frame batches when available. This may be worth carrying over when designing the common default/auto policy.

deepmd/pt_expt/infer/deep_eval.py
wanghan-iapcm: Valid — and rather than documenting the divergence, fixed the builders to honor the contract in 63960681a: ase/vesin/nv now post-filter the (i, j, S) edge list so virtual atoms (`atype < 0`) are excluded as both centers and neighbors, matching the dense reference builder (builder.py's documented contract). The "SAME neighbor set" docstring is therefore accurate again, including for mixed-type padding inputs, and downstream graph consumers need no extra masking. Set-equality tests vs the dense builder with a virtual atom added for all three: ase on the default CI, vesin (guarded), nv (CUDA-gated, validated on a T4: 11/11 with the vesin suite). Note the ase-builder half of this bug also exists on master (PR-A vintage) and is fixed here in passing.

deepmd/pt_expt/utils/nv_graph_builder.py
wanghan-iapcm: Agreed on single-sourcing — implemented in 2f7ea9ed1 with the dependency direction inverted relative to the suggestion (per project roadmap): the graph-builder module now OWNS the search (`nv_search_matrix`: device-context pinning — restoring the `_input_device_context` guard this copy had dropped, good catch — periodic normalization, batch setup, grow-until-fit loop) and `NvNeighborList.build` delegates to it via a lazy import. Rationale: the World-1 strategy layer is scheduled for deprecation with the dense-nlist path, so the long-lived graph module is the owner and the legacy class is the caller — when World-1 is removed, nothing needs to move again. extended/edges return modes are byte-unchanged; GPU-validated on a T4 (nv builder 18/18, NvNeighborList/nlist_backend suites 75 passed + 26 subtests).

deepmd/pt_expt/utils/vesin_graph_builder.py
wanghan-iapcm: Same resolution as the nv path, in 2f7ea9ed1: the raw vesin search is single-sourced as `vesin_graph_builder.vesin_search_ijs` (device pinning included) and BOTH `VesinNeighborList._build_single` and `_build_single_edges` now call it (lazy import to break the module cycle) — graph module owns, deprecation-bound strategy delegates. Extended/edges behavior byte-unchanged (`ss` float cast kept at the `_build_single` call site so `ss @ box` math is identical); full vesin + neighbor_list suites pass on CPU and T4.

deepmd/pt_expt/model/make_model.py
wanghan-iapcm: Valid — fixed in a6eb9bca8: `_call_common_graph` now carries the same `mixed_types() and descriptor.uses_graph_lower()` fail-fast as dpmodel (NotImplementedError), so an explicit `neighbor_graph_method` can no longer reach the builders for ineligible descriptors. Regression test drives se_e2_a through all four explicit methods and expects the raise.

deepmd/pt_expt/infer/deep_eval.py
wanghan-iapcm: Done in d2cf9205b: both knobs are now documented in the DeepEval docstring (which controls which path, and that they don't overlap), and a non-default `neighbor_graph_method` on a non-graph artifact now RAISES at construction instead of being silently ignored, pointing the user at `nlist_backend`. Regression test on a nlist-form .pt2.

wanghan-iapcm: Agreed the two-knob split is not a good end state. What this PR now does: documents the split explicitly and fail-fasts on misuse (d2cf9205b), so a user can no longer pick the wrong knob silently. Full consolidation into a single backend-selection API (and nv on the nlist path) is deferred deliberately: the dense-nlist lower is scheduled for total deprecation in favor of the graph schema (see the NeighborGraph design discussion), at which point `nlist_backend` disappears and `neighbor_graph_method` becomes THE selector — consolidating now would mean building API surface on the path being removed. Happy to reorder if you think the deprecation is too far out.

wanghan-iapcm: Thanks — carried into the vesin builder's scope note and it matches our benchmark data (T4, diamond C, rcut 6): vesin and nv are equal within noise at nf==1 on CUDA, and vesin's per-frame Python loop (~1 ms/frame) makes nv the clear choice for multi-frame batches; your CPU+nf==1 vesin preference is consistent with vesin's lighter single-system path. Will encode exactly this policy (cpu+nf==1 -> vesin; CUDA/multi-frame -> nv when available) when the auto selector lands with the unified API.

iProzd (approved):

OutisLi (approved):

PR #5715, opened 2026-07-02
feat(dpmodel): graph-native se_atten attention (NeighborGraph PR-D)

wanghan-iapcm:
Implements NeighborGraph PR-D: the graph path now supports `attn_layer > 0` for dpa1/se_atten, removing the attn_layer=0-only restriction shipped in #5583.

## What

- **Segment toolkit**: `segment_max` + numerically-stable, mask-aware `segment_softmax` (`deepmd/dpmodel/utils/neighbor_graph/segment.py`), built on the existing `xp_maximum_at`.
- **`center_edge_pairs`** (`neighbor_graph/pairs.py`): pairs of edges sharing a center — the edge-pair axis shared with the upcoming angle machinery (PR-E). Segment-based enumeration (a global `(E,E)` boolean is deliberately avoided: `O(N²·nnei²)` memory). Two forms: compact eager (dynamic `P`, carry-all graphs) and **shape-static** (`P = n_center·nnei²`, pure arange/reshape arithmetic, no `nonzero`) for the center-major static layout — this keeps the traced/compiled/export path traceable.
- **`DescrptBlockSeAtten._graph_attention`**: op-for-op ragged mirror of `GatedAttentionLayer`/`NeighborGatedAttention` — per-center `q@kᵀ` becomes per-pair `q_m·k_n`, softmax over keys becomes `segment_softmax` grouped by the query edge; head_dim QKV slicing, q/k/v normalize, temperature/scaling, smooth shift trick, post-softmax `sw` and `dotr` weighting, residual + LayerNorm per layer.
- `edge_env_mat(return_sw=True)` exposes the per-edge switch (zeroed on padding) for the smooth branch.
- `uses_graph_lower` widened: attention configs (concat tebd, no exclude_types) are now graph-eligible — pt_expt eager/compiled/exported paths route them through the graph lower by default.

## Numerical semantics (reviewed decision)

- **Shape-static adapter path** (the dense `call` adapter, `from_dense_quartet(compact=False)` + `static_nnei`): **bit-exact vs the dense body, rtol 1e-12**, full flag matrix (attn_layer 1/2 × dotr × smooth × normalize × temperature, binding AND non-binding sel).
- **Carry-all graphs**: exact for non-smooth attention. For `smooth_type_embedding=True`, the dense branch keeps sel-padding slots in the attention softmax **denominator** (weight `exp(-attnw_shift)`), which makes the dense output *depend on sel itself* (measured up to ~1e-4 with an identical physical neighbor set). The carry-all form **drops those phantom terms by design** — the sel-independent math. Pinned by a clean-divergence test; route-equivalence fixtures pin `smooth_type_embedding=False`.
- se_atten_v2 (`tebd_input_mode="strip"`) remains graph-ineligible (strip mode is a later PR) — pinned by test.

## Testing

- 38 new dpmodel tests (segment toolkit, pairs incl. random-vs-oracle + static-vs-compact equality, attention parity matrix, binding-sel divergence sanity).
- pt_expt: `test_make_fx_graph_attn` (graph forward + autograd at attn_layer=2 traces under make_fx, both smooth branches — required since compiled training uses the graph lower); model-level graph-vs-legacy force/virial/atom-virial parity parametrized over attn_layer {0,2}.
- Local CPU: common/dpmodel 583, consistent dpa1+se_atten_v2 209, pt_expt descriptor/model/utils 701 (2 failures: dpa4 export inductor error **pre-existing on upstream/master**, and a route-parity fixture fixed in-branch).
- **GPU-validated (Tesla T4, cuda:0)**: dpmodel suites 38, pt_expt graph-lower/make_fx/consistency 44 (CUDA 1e-10), route-parity 6, attention AOTI export pipeline + dpa1 cross-backend consistency 105 — all passed.

## Known limitations

- Strip-mode (se_atten_v2) attention stays on the dense path.
- Carry-all smooth attention diverges from dense by design (see above); old behavior reachable via `neighbor_graph_method="legacy"` / explicit World-1 builders.
- `num_heads == 1` assumed (dpa1 never exposes num_heads); fail-fast otherwise.
- Compact `center_edge_pairs` is eager-only (`nonzero`); traced paths use the shape-static form.
- 3-body angles (PR-E), jax graph force (PR-F), dpa2/3 MP (PR-G) unchanged.

## Summary by CodeRabbit

* **New Features**
  * Expanded graph-native attention support for additional DPA1/se_atten configurations, enabling transformer-style graph execution suitable for tracing/export.
  * Added center-based neighbor edge-pair enumeration with shape-static control to improve graph layout consistency.
  * Improved graph tracing/export with optional dynamic-shape hinting.
* **Bug Fixes**
  * Stabilized graph attention softmax under masking/padding and ensured correct behavior for empty/no-edge cases.
* **Tests**
  * Added/updated parity, eligibility, FX traceability, export/graph-lower, and single-atom (no edges) coverage across attention settings.

### DIFF
--- a/deepmd/dpmodel/descriptor/dpa1.py
+++ b/deepmd/dpmodel/descriptor/dpa1.py
@@ -434,6 +434,12 @@ class DescrptDPA1(NativeOP, BaseDescriptor):
         with concat type-embedding and no type exclusion. Remaining ineligible
         configs (``tebd_input_mode == "strip"``, ``exclude_types``) fall back
         to the legacy dense path, so those models keep working unchanged.
+
+        Eligibility does NOT imply numerical interchangeability with the
+        dense route for every config: with ``smooth_type_embedding=True``
+        the carry-all graph attention is sel-independent by design and
+        differs from the dense lower by up to ~1e-4 (see the Notes of
+        :meth:`call_graph`).
         """
         return (
             self.se_atten.tebd_input_mode == "concat"
@@ -743,6 +749,21 @@ class DescrptDPA1(NativeOP, BaseDescriptor):
         This method is graph-native: it takes no dense quartet inputs and does
         not produce the dense ``sw`` (that lives in the dense :meth:`call`
         adapter, which has the ``nlist``/``coord_ext`` needed to build it).
+
+        Notes
+        -----
+        **Smooth attention is intentionally sel-independent on the graph
+        path.** For ``smooth_type_embedding=True`` the legacy dense attention
+        keeps the sel-padding slots in its softmax DENOMINATOR (phantom
+        ``exp(-attnw_shift)`` terms), which makes dense output depend on the
+        ``sel`` setting by up to ~1e-4 even for identical physical neighbors.
+        A carry-all graph has no padding slots, so its softmax runs over the
+        real neighbor pairs only: cleaner, sel-independent semantics that
+        deliberately DIFFER from the dense route for smooth models. The two
+        routes agree bit-tight only for ``smooth_type_embedding=False`` (at
+        non-binding ``sel``), or when this kernel is realized on a dense
+        layout via ``static_nnei`` (the dense :meth:`call` adapter), which
+        reproduces the phantom terms for exact backward compatibility.
 
         Parameters
         ----------
--- a/deepmd/dpmodel/descriptor/dpa1.py
+++ b/deepmd/dpmodel/descriptor/dpa1.py
@@ -26,6 +26,7 @@ from deepmd.dpmodel.array_api import (
 )
 from deepmd.dpmodel.common import (
     cast_precision,
+    get_xp_precision,
     to_numpy_array,
 )
 from deepmd.dpmodel.utils import (
@@ -760,8 +761,19 @@ class DescrptDPA1(NativeOP, BaseDescriptor):
             (N, ng, 3) equivariant single-particle representation, flat node
             axis.
         """
+        import dataclasses
+
         xp = array_api_compat.array_namespace(graph.edge_vec)
         dev = array_api_compat.device(graph.edge_vec)
+        # manual @cast_precision: the decorator casts array ARGUMENTS, but the
+        # graph's only float input (edge_vec) is inside the NeighborGraph
+        # dataclass, invisible to it. Cast edge_vec down to the descriptor
+        # precision on entry and the outputs back to the caller's dtype on
+        # exit (differentiable: grad still flows to the caller's edge_vec leaf).
+        in_dtype = graph.edge_vec.dtype
+        prec = get_xp_precision(xp, self.precision)
+        if in_dtype != prec:
+            graph = dataclasses.replace(graph, edge_vec=xp.astype(graph.edge_vec, prec))
         grrg, rot_mat = self.se_atten.call_graph(
             graph, atype, type_embedding=type_embedding, static_nnei=static_nnei
         )
@@ -775,6 +787,9 @@ class DescrptDPA1(NativeOP, BaseDescriptor):
             atype_local = xp.asarray(atype, device=dev)
             atype_embd = xp.take(type_embedding, atype_local, axis=0)  # (N, tebd_dim)
             grrg = xp.concat([grrg, atype_embd], axis=-1)
+        if in_dtype != prec:
+            grrg = xp.astype(grrg, in_dtype)
+            rot_mat = xp.astype(rot_mat, in_dtype)
         return grrg, rot_mat
 
     def enable_compression(
--- a/deepmd/dpmodel/descriptor/dpa1.py
+++ b/deepmd/dpmodel/descriptor/dpa1.py
@@ -428,15 +428,14 @@ class DescrptDPA1(NativeOP, BaseDescriptor):
     def uses_graph_lower(self) -> bool:
         """Returns whether this descriptor supports the graph-native lower.
 
-        The graph-native energy lower (``call_graph``) currently covers only the
-        non-attention (``attn_layer == 0``) factorizable path with concat
-        type-embedding and no type exclusion. Any other config (attention,
-        ``tebd_input_mode == "strip"``, ``exclude_types``) falls back to the
-        legacy dense path, so those models keep working unchanged.
+        The graph-native lower (``call_graph``) covers the factorizable path
+        AND transformer attention (``attn_layer >= 0``, NeighborGraph PR-D)
+        with concat type-embedding and no type exclusion. Remaining ineligible
+        configs (``tebd_input_mode == "strip"``, ``exclude_types``) fall back
+        to the legacy dense path, so those models keep working unchanged.
         """
         return (
-            self.se_atten.attn_layer == 0
-            and self.se_atten.tebd_input_mode == "concat"
+            self.se_atten.tebd_input_mode == "concat"
             and not self.se_atten.exclude_types
         )
 
@@ -643,6 +642,9 @@ class DescrptDPA1(NativeOP, BaseDescriptor):
             graph,
             atype_local,
             type_embedding=self.type_embedding.call(),
+            # the adapter graph is shape-static center-major (compact=False):
+            # keep the attention pair enumeration nonzero-free (traceable)
+            static_nnei=nnei,
         )
         # call_graph returns flat (N, ...) node axis; reshape to (nf, nloc, ...)
         # for the dense 5-tuple ABI -- this reshape is LOCAL to the adapter shim.
@@ -727,8 +729,9 @@ class DescrptDPA1(NativeOP, BaseDescriptor):
         graph: Any,
         atype: Array,
         type_embedding: Array | None = None,
+        static_nnei: int | None = None,
     ) -> tuple[Array, Array]:
-        """Descriptor-level graph-native forward (``attn_layer == 0``).
+        """Descriptor-level graph-native forward.
 
         Wraps the block kernel
         :meth:`DescrptBlockSeAtten.call_graph`, adds the descriptor-level
@@ -760,7 +763,7 @@ class DescrptDPA1(NativeOP, BaseDescriptor):
         xp = array_api_compat.array_namespace(graph.edge_vec)
         dev = array_api_compat.device(graph.edge_vec)
         grrg, rot_mat = self.se_atten.call_graph(
-            graph, atype, type_embedding=type_embedding
+            graph, atype, type_embedding=type_embedding, static_nnei=static_nnei
         )
         # FLAT node axis (N, ...): no (nf, nloc) reshape -- ragged-native, spec.
         if self.concat_output_tebd:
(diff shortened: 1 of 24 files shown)


wanghan-iapcm:
Pushed two additional commits that remove the "attention graph-form export deferred" limitation:

- **`feat(dpmodel): make compact center_edge_pairs traceable via unbacked SymInts`** — the carry-all pair enumeration (`nonzero` + tensor-`repeat`) now registers its data-dependent sizes via a new torch-free `xp_hint_dynamic_size` shim (no-op for numpy/jax), takes empty-input fast paths only on concrete int shapes, builds iotas as `cumsum(ones)-1` (the array_api_compat `arange` wrapper branches on the length in Python), and skips the policy-compression `nonzero` when no filter applies (the attention default). Eager numpy/torch results are unchanged (bit-identical; full eager suites re-run green).
- **`feat(pt_expt): graph-form .pt2 export for dpa1 attention (attn_layer > 0)`** — with the above, `lower_kind="graph"` now exports attention models **unchanged in ABI** (same 5-tensor NeighborGraph schema, dynamic edge axis) and with **carry-all semantics preserved** (no sel truncation, unlike the dense-adapter nlist-form export). Adds a symbolic-trace merge gate (attn_layer 0/2), parametrizes the graph-`.pt2` DeepEval fixture over attn_layer (dynamic sizes, PBC/non-PBC, 1e-10 vs the sel-capped dense reference at non-binding sel), a single-atom zero-edge runtime test, and fixes the stale freeze-gate message.

AOTI parity vs eager carry-all measured at ≤5e-18 across system sizes. Benchmark on a Tesla T4 (fp64, diamond C at experimental density, rcut 6 / sel 180, eager): the graph path is flat ~100 µs/atom (O(N)) and still runs at 4096 atoms where the dense path OOMs; with attention the graph is consistently faster at every size that fits.

Known limitations: relies on torch unbacked-SymInt maturity (validated on 2.10; CPU AOTI); jax.jit of the compact path still needs a static realization (PR-F); C++ gtest of an attention graph `.pt2` not added (ABI unchanged from the attn=0 artifact).

deepmd/dpmodel/utils/neighbor_graph/segment.py
```diff
@@ -37,3 +38,52 @@ def segment_mean(data: Array, segment_ids: Array, num_segments: int) -> Array:
...
+    # guard -inf (empty / fully-masked segments) so gather doesn't yield inf-inf
+    seg_max = xp.where(xp.isinf(seg_max), xp.zeros_like(seg_max), seg_max)
+    shifted = data - xp.take(seg_max, segment_ids, axis=0)
+    ex = xp.exp(shifted)
+    if mask is not None:
+        ex = ex * xp.astype(mask, ex.dtype)
+    denom = segment_sum(ex, segment_ids, num_segments)
+    denom_e = xp.take(denom, segment_ids, axis=0)
+    safe = xp.where(denom_e > 0, denom_e, xp.ones_like(denom_e))
+    return ex / safe
```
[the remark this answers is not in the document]
wanghan-iapcm: Valid — fixed in 859399d83. `shifted` now uses `data_for_max` (masked entries already -inf, so they exp() to exactly 0) instead of the raw data, eliminating the inf * 0 = nan path; the `ex * mask` multiply is kept as a defensive no-op. Added a regression test with a masked logit 1e5 above the unmasked max (well past the fp64 exp-overflow threshold), asserting finite output, exact parity of the unmasked entries, and zero weight on the masked one. Worth noting the scenario became more reachable with this PR's fp32 support (overflow threshold ~88 instead of ~709), so good catch.

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -1699,8 +1726,7 @@ def call_graph(
 
         Notes
```
OutisLi, line 1748: The carry-all smooth-attention behavior here looks like the cleaner semantics to me: dropping phantom sel-padding terms from the softmax denominator is more natural than preserving a dense-layout artifact. The remaining issue is documentation/API visibility. With this PR, attention configs become graph-eligible and pt_expt can route them through the carry-all graph by default, so users may reasonably expect dense/nlist and graph routes to be numerically interchangeable. For `smooth_type_embedding=True`, they are intentionally not: dense keeps padding slots in the denominator while carry-all has no such phantom keys. Please document this near `call_graph` / the graph-export path (and ideally user-facing docs) as an intentional sel-independent graph semantics, not just in tests/PR text.

deepmd/pt_expt/entrypoints/main.py
```diff
@@ -563,10 +563,12 @@ def freeze(
 
     m.eval()
 
-    # The graph lower is opt-in and only valid for graph-eligible models (dpa1
-    # attn_layer==0 today). Fail fast with a clear message rather than emitting a
-    # broken .pt2. Enable the per-atom virial for the graph form -- it is
-    # near-free there (one extra scatter off the single shared backward).
+    # The graph lower is opt-in and only valid for graph-eligible models
+    # (dpa1 with concat tebd and no type exclusion; attention layers included
```
OutisLi, line 573: The stale text is in the `freeze()` parameter docstring above this hunk: the `lower_kind` docs still say graph is currently dpa1 with `attn_layer == 0` (around lines 498-504). GitHub does not allow anchoring this directly on that unchanged docstring line, so I am anchoring it on the nearby updated eligibility hunk. Please update that docstring, and the remaining old PR-A wording in the graph lower/training path, to match the new eligibility contract: mixed-types DPA1/se_atten with concat type embedding and no `exclude_types`, with attention layers included. Also avoid implying unconditional dense parity, since carry-all smooth attention intentionally differs from legacy dense.

deepmd/dpmodel/descriptor/dpa1.py
wanghan-iapcm: Agreed and documented in 89ce6b77b, at the three layers a reader might arrive from: (1) a Notes block on `DescrptDPA1.call_graph` — the authoritative statement (what the dense layout does with sel-padding in the softmax denominator, why carry-all has no phantom keys, the observed ~1e-4 magnitude, and that the `static_nnei` dense-adapter realization reproduces the phantom terms for exact backward compatibility); (2) `uses_graph_lower` now states explicitly that eligibility does NOT imply numerical interchangeability for `smooth_type_embedding=True`; (3) the user-facing `freeze(lower_kind=...)` docstring carries the same note for the graph-export path. A docs-site page for graph-form export doesn't exist yet — when it lands (PR-G-era), the semantics note should be lifted there as well.

deepmd/pt_expt/entrypoints/main.py
wanghan-iapcm: Fixed in 89ce6b77b — thanks for the anchor workaround. Updated the `freeze()` `lower_kind` docstring and the remaining PR-A-era wording (training-path predicate `_model_uses_graph_lower` docstring + compiled-lower comment, and the dpmodel/pt_expt `forward/call_common_lower_graph` docstrings) to the new contract: mixed-types dpa1/se_atten with concat type embedding and no `exclude_types`, attention layers included. The `call_common` graph-route doc also no longer implies unconditional dense parity: non-binding-sel exact parity is now scoped to the non-smooth branch, with the smooth divergence called out as intentional sel-independent semantics.

deepmd/dpmodel/utils/neighbor_graph/pairs.py
```diff
@@ -0,0 +1,198 @@
...
+    # under make_fx/torch.export symbolic tracing shape[0] is a SymInt (the
+    # nonzero output size is UNBACKED), and branching on it raises
+    # GuardOnDataDependentSymNode. The general code below handles the empty
+    # case correctly anyway (all downstream arrays come out empty).
+    if isinstance(dst.shape[0], int) and dst.shape[0] == 0:
+        return empty
+    # real edges only, grouped by center (stable sort keeps original order
+    # within a center — irrelevant for correctness, deterministic for tests)
+    (real_idx,) = xp.nonzero(edge_mask)
+    xp_hint_dynamic_size(real_idx)  # unbacked size R: register >= 0 size hint
```
iProzd, line 151: The compact `center_edge_pairs` export path relies on unbacked-SymInt support ("torch >= 2.6" per the pairs.py docstring), but there is no version guard: on older torch, exporting a carry-all attention graph would fail deep inside make_fx/AOTI with an obscure error. If deepmd's minimum supported torch is below 2.6, could the graph export fail fast with a clear message?

iProzd (changes requested):
Two follow-ups now that we are shipping the default flip:

1. Since `smooth_type_embedding=True` + `attn_layer=2` are the dpa1 constructor defaults, existing checkpoints trained under the dense semantics will shift up to ~1e-4 when evaluated on pt_expt after this PR. Could we record this as an explicit user-facing behavior change (changelog / release notes / migration note), including the escape hatch (`neighbor_graph_method="legacy"` restores the old numbers)? Users hitting reference-value regressions should be able to find the explanation without reading `call_graph` docstrings.

2. The pt_expt-graph vs dense end-to-end divergence is currently invisible to the consistent suites (they exercise the bit-exact shape-static adapter). Could we add one end-to-end test pinning the expected divergence — nonzero and bounded by the documented ~1e-4 magnitude — so a future refactor cannot silently change the carry-all smooth semantics?

OutisLi (approved):

deepmd/dpmodel/utils/neighbor_graph/pairs.py
wanghan-iapcm: Valid — fixed in da46ed2a3. deepmd's torch floor is indeed below 2.6 (build floor is 2.1), so added `check_graph_trace_torch_version` in `serialization.py`: when the descriptor has `attn_layer > 0` and torch < 2.6, both graph trace sites (the .pt2 export branch and the training compile in `_trace_and_compile_graph`) now raise a clear RuntimeError up front ("requires torch >= 2.6 ... upgrade torch, set attn_layer: 0, or use the dense path") instead of dying inside make_fx/AOTI. Factorizable models (attn_layer=0) trace with backed symbols only, so they are deliberately not restricted. Unit tests cover reject/pass/no-descriptor branches with monkeypatched versions, incl. local-suffix strings like 2.5.1+cu124.

wanghan-iapcm:
Both review-body items addressed in fc30ee44a (the inline torch<2.6 guard was addressed earlier in da46ed2a3/60c5e8377):

1. **User-facing migration note** — added to `doc/model/train-se-atten.md` ("Difference among different backends"): documents the pt_expt carry-all graph default, the up-to-~1e-4 shift for `smooth_type_embedding=true` + `attn_layer>0` checkpoints trained under dense semantics, sel-independence of the graph path, and the `neighbor_graph_method="legacy"` escape hatch.

2. **End-to-end divergence pin** — new `test_smooth_attention_divergence_pinned` in `source/tests/pt_expt/model/test_dpa1_graph_lower.py`: public `call_common` default route vs `legacy`, smooth=True/attn_layer=2, asserts the energy divergence is nonzero (>1e-10, measured ~4.5e-6 on the small fixture) and bounded (<1e-3), so a refactor cannot silently change the carry-all smooth semantics.

iProzd (approved):

PR #5717, opened 2026-07-03
feat(dpmodel): NeighborGraph 3-body angle machinery (PR-E)

wanghan-iapcm:
## Summary

NeighborGraph PR-E: the optional 3-body **angle** extension of the edge-graph neighbor-list contract (design discussion wanghan-iapcm/deepmd-kit#4). An angle is a pair of edges sharing a center (`dst(edge_a) == dst(edge_b)`), stored as `angle_index (2, A)` into `[0, E)` + `angle_mask (A,)` — `edge_vec` stays the ONLY geometry leaf, so force/virial assembly is untouched (proven by an invariance test).

> **Stacked on #5715 (PR-D)** — reuses its `center_edge_pairs`. Please merge #5715 first; this branch then rebases clean onto master. Only the last 10 commits belong to this PR.

### What's added (all in `deepmd/dpmodel/utils/neighbor_graph/`)

- `pad_and_guard_angles` (graph.py) — angle-axis padder mirroring `pad_and_guard_edges` (dynamic guard append / static `angle_capacity` with overflow ValueError).
- `angles.py` (new):
  - `build_angle_index(edge_index, edge_vec, edge_mask, n_total, a_rcut, *, ordered=False, include_self=False, layout=None)` — unordered, no-self pairs of edges sharing a center where BOTH edges are within `a_rcut`; built on PR-D's `center_edge_pairs`; `pair_mask` folded into `angle_mask` (never discarded).
  - `attach_angles(graph, a_rcut, ...)` — post-hoc: edge graph in, graph with angle fields out (`dataclasses.replace`); default builders keep angles `None`.
  - `angle_to_edge_sum` / `angle_to_node_sum` — segment-sum aggregation to the query edge / shared center.
  - `graph_angle_cos(angle_index, edge_vec, eps=1e-6)` — per-angle cos θ mirroring dpa3 repflows `cosine_ij` eps placement exactly (`+eps` in norm denominators, `*(1-eps)` on the product).
  - `angle_padding_fraction(graph)` — mask-derived padding-waste report for static capacities.

### Semantics / decisions

- **Unordered, no-self by default**: dpa3's dense angle tensor is the redundant ordered `a_sel x a_sel` square including the `j==k` diagonal; the graph set keeps one entry per unordered `{j,k}` pair and moves the degenerate diagonal to the (a_rcut-filtered) edge channel. Dense parity is therefore asserted against the OFF-DIAGONAL `cosine_ij[j,k]` (j != k) at rtol/atol **1e-12** (same-math fp64), at non-binding `a_sel`. The ordered+self full square stays available via flags.
- **`a_sel` = normalization-only** (carry-all within `a_rcut`), consistent with the edge-`sel` decision.
- Second oracle: se_t dot-product convention cross-checked from coordinates in the `sw == 1` regime (rtol 1e-12).

### Tests

`source/tests/common/dpmodel/test_angle_builder.py` (21) + `test_graph_angle_cos_parity.py` (6): brute-force triplet oracle (all flag combinations, multi-center, static layout, `node_capacity` branch), dpa3 dense-parity + no-self-angle assertion, se_t coordinate oracle, force/virial bit-exact invariance with/without angles, padding-fraction (incl. `total==0`), torch-namespace smoke tests for every new function. Full neighbor-graph suite: 54 passed.

### Known limitations

- **Machinery + angle-channel math only** — no dpa3 graph descriptor here (dpa3 is message-passing; wiring = PR-G). se_t/se_t_tebd are not migrated (`mixed_types=False`), used as oracle only.
- Angle enumeration is the compact **eager** form (`nonzero` in `center_edge_pairs`) even when a static `layout` is passed — `angle_capacity` fixes the output shape only; shape-static enumeration for export is deferred to PR-G.
- Not bit-parity with dense dpa3 **by construction** (unordered/no-self reformulation; recoverable in PR-G via symmetric angle→edge 2x + edge-channel diagonal).
- Aggregation helpers follow the mask-then-reduce convention: callers mask per-angle data by `angle_mask` before summing (padding angles point at edge 0).
- numpy/torch validated; jax rides the array-API surface (no jax-specific test here); `A ~ sum(deg^2)` capacity overhead mitigated by `a_rcut < rcut` and reported by `angle_padding_fraction`.

## Summary by CodeRabbit

* **New Features**
  * Added angle-graph utilities to build and attach 3-body angle relationships on top of neighbor graphs.
  * Introduced angle padding/guarding for fixed-capacity layouts.
  * Added helpers to compute angle cosine values, reduce angle data back to edges/nodes, and report angle padding coverage.
  * Expanded publicly available exports for angle/edge-pair utilities and additional segment reductions (max/softmax).

* **Tests**
  * Added extensive unit tests covering index building, attachment, masking/aggregation correctness, padding behavior, and cosine parity across NumPy/Torch.

### DIFF
--- a/deepmd/dpmodel/utils/neighbor_graph/angles.py
+++ b/deepmd/dpmodel/utils/neighbor_graph/angles.py
@@ -141,6 +141,35 @@ def attach_angles(
     return dataclasses.replace(graph, angle_index=ai, angle_mask=am)
 
 
+def graph_angle_cos(angle_index: Array, edge_vec: Array, eps: float = 1e-6) -> Array:
+    """Per-angle cosine, mirroring dpa3 ``cosine_ij`` (repflows.py:632-644).
+
+    Parameters
+    ----------
+    angle_index : Array
+        Shape (2, A) index pairs into edge list. ``angle_index[0, a]`` is
+        edge_a and ``angle_index[1, a]`` is edge_b for angle ``a``.
+    edge_vec : Array
+        Shape (E, 3) edge vectors (r_src - r_dst, i.e. neighbor - center).
+    eps : float, optional
+        Numerical stabiliser: norm denominators use ``||v|| + eps`` and the
+        dot product is scaled by ``(1 - eps)``.  Mirrors the dpa3 dense
+        channel exactly (repflows.py:643-649).
+
+    Returns
+    -------
+    Array
+        Shape (A,) cosine values, one per angle slot (valid and padding).
+        Padding slots carry arbitrary values; mask with angle_mask before use.
+    """
+    xp = array_api_compat.array_namespace(edge_vec)
+    va = xp.take(edge_vec, angle_index[0, :], axis=0)  # (A, 3)
+    vb = xp.take(edge_vec, angle_index[1, :], axis=0)  # (A, 3)
+    na = va / (xp.linalg.vector_norm(va, axis=-1, keepdims=True) + eps)
+    nb = vb / (xp.linalg.vector_norm(vb, axis=-1, keepdims=True) + eps)
+    return xp.sum(na * nb, axis=-1) * (1.0 - eps)
+
+
 def angle_to_edge_sum(data: Array, angle_index: Array, num_edges: int) -> Array:
     """Aggregate per-angle data to the angle's query edge (edge_a).
 
--- a/deepmd/dpmodel/utils/neighbor_graph/angles.py
+++ b/deepmd/dpmodel/utils/neighbor_graph/angles.py
@@ -21,6 +21,10 @@ if TYPE_CHECKING:
 
 import dataclasses
 
+from deepmd.dpmodel.utils.safe_gradient import (
+    safe_for_vector_norm,
+)
+
 from .graph import (
     GraphLayout,
     NeighborGraph,
@@ -77,10 +81,13 @@ def build_angle_index(
     # a_rcut edge gate: only edges within a_rcut may participate in an angle.
     # Strict `<` (not the edge channel's `<= rcut`, builder.py:284) is
     # intentional: it mirrors dpa3's dense angle gate exactly
-    # (`a_dist_mask = dist < self.a_rcut`, repflows.py:598), so an edge
-    # sitting exactly at a_rcut is excluded from angles the same way dense
-    # excludes it, even though it would still be kept as an edge.
-    dist = xp.linalg.vector_norm(edge_vec, axis=-1)  # (E,)
+    # (`a_dist_mask = safe_for_vector_norm(diff, axis=-1) < self.a_rcut`,
+    # repflows.py:598), so an edge sitting exactly at a_rcut is excluded from
+    # angles the same way dense excludes it, even though it would still be kept
+    # as an edge. `safe_for_vector_norm` (not plain `vector_norm`) matches dpa3
+    # value-for-value and keeps the gate off the NaN-gradient path shared with
+    # the normalization below.
+    dist = safe_for_vector_norm(edge_vec, axis=-1)  # (E,)
     a_edge_mask = xp.astype(edge_mask, xp.bool) & (dist < a_rcut)
     # compact eager form only (static_nnei not exposed until angle export is
     # needed, PR-G). dst = edge_index[1, :] per the [src, dst] SoA convention.
@@ -182,8 +189,13 @@ def graph_angle_cos(angle_index: Array, edge_vec: Array, eps: float = 1e-6) -> A
     xp = array_api_compat.array_namespace(edge_vec)
     va = xp.take(edge_vec, angle_index[0, :], axis=0)  # (A, 3)
     vb = xp.take(edge_vec, angle_index[1, :], axis=0)  # (A, 3)
-    na = va / (xp.linalg.vector_norm(va, axis=-1, keepdims=True) + eps)
-    nb = vb / (xp.linalg.vector_norm(vb, axis=-1, keepdims=True) + eps)
+    # safe_for_vector_norm (not plain vector_norm) mirrors dpa3 exactly
+    # (repflows.py:642-643) and gives a 0 gradient at ||v||==0 instead of NaN;
+    # edge_vec is the sole autograd leaf, so this removes a latent NaN-gradient
+    # landmine on the geometry path (values are identical for real, non-zero
+    # edges, so fp64 dense/se_t parity is unchanged).
+    na = va / (safe_for_vector_norm(va, axis=-1, keepdims=True) + eps)
+    nb = vb / (safe_for_vector_norm(vb, axis=-1, keepdims=True) + eps)
     return xp.sum(na * nb, axis=-1) * (1.0 - eps)
 
 
--- a/deepmd/dpmodel/utils/neighbor_graph/angles.py
+++ b/deepmd/dpmodel/utils/neighbor_graph/angles.py
@@ -1,0 +1,91 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+"""3-body angle graph: pairs of edges sharing a center within a_rcut.
+
+Angles reference EDGES (angle_index into [0,E)); edge_vec stays the only
+geometry leaf. a_sel is normalization-only (not a truncation). Reuses PR-D's
+center_edge_pairs; a_rcut filters the participating edges.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import array_api_compat
+
+if TYPE_CHECKING:
+    from deepmd.dpmodel.array_api import Array
+
+from .graph import GraphLayout, pad_and_guard_angles
+from .pairs import center_edge_pairs
+
+
+def build_angle_index(
+    edge_index: Array,
+    edge_vec: Array,
+    edge_mask: Array,
+    n_total: int,
+    a_rcut: float,
+    *,
+    ordered: bool = False,
+    include_self: bool = False,
+    layout: GraphLayout | None = None,
+) -> tuple[Array, Array]:
+    """Build angle index for 3-body terms.
+
+    Parameters
+    ----------
+    edge_index : Array
+        Shape (2, E) [src, dst] SoA edge indices.
+    edge_vec : Array
+        Shape (E, 3) edge vectors (neighbor - center).
+    edge_mask : Array
+        Shape (E,) boolean validity mask for edges.
+    n_total : int
+        Total number of nodes.
+    a_rcut : float
+        Angle cutoff. Only edges with norm < a_rcut participate in angles.
+    ordered : bool, optional
+        If True, include both (a, b) and (b, a) pairs (ordered pairs).
+    include_self : bool, optional
+        If True, include self-angle pairs (a, a).
+    layout : GraphLayout or None, optional
+        If provided, uses layout.angle_capacity as static padding capacity.
+
+    Returns
+    -------
+    angle_index : Array
+        Shape (2, A) index pairs into the edge list.
+    angle_mask : Array
+        Shape (A,) boolean mask for valid angles.
+    """
+    xp = array_api_compat.array_namespace(edge_index)
+    # a_rcut edge gate: only edges within a_rcut may participate in an angle
+    dist = xp.linalg.vector_norm(edge_vec, axis=-1)  # (E,)
+    a_edge_mask = xp.astype(edge_mask, xp.bool) & (dist < a_rcut)
+    # compact eager form only (static_nnei not exposed until angle export is
+    # needed, PR-G). dst = edge_index[1, :] per the [src, dst] SoA convention.
+    q_e, k_e, pair_mask = center_edge_pairs(
+        edge_index[1, :],
+        a_edge_mask,
+        n_total,
+        include_self=include_self,
+        ordered=ordered,
+    )
+    # compact form returns all-True pair_mask, but NEVER discard it: the
+    # shape-static form keeps filtered pairs and invalidates them only here.
+    angle_index = xp.stack([q_e, k_e], axis=0)  # (2, A_real)
+    cap = layout.angle_capacity if layout is not None else None
+    ai, am = pad_and_guard_angles(angle_index, cap, min_angles=2)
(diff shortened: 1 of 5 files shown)


deepmd/dpmodel/utils/neighbor_graph/angles.py
```diff
@@ -0,0 +1,254 @@
...
+    -------
+    angle_index : Array
+        Shape (2, A) index pairs into the edge list.
+    angle_mask : Array
+        Shape (A,) boolean mask for valid angles.
+    """
+    xp = array_api_compat.array_namespace(edge_index)
+    # a_rcut edge gate: only edges within a_rcut may participate in an angle
+    dist = xp.linalg.vector_norm(edge_vec, axis=-1)  # (E,)
+    a_edge_mask = xp.astype(edge_mask, xp.bool) & (dist < a_rcut)
```
iProzd, line 91: The angle participation gate here uses a strict `dist < a_rcut`, whereas the edge channel (`build_neighbor_graph`) keeps neighbors with `dist <= rcut`. An edge sitting exactly at `a_rcut` is therefore excluded from angles but kept as an edge. The dense-parity tests run at non-binding `a_sel`, so no atom lands on the boundary and this difference isn't exercised. Is the strict `<` intentional, and does it match the `a_rcut` gate operator in dpa3 repflows? If dpa3 uses `<=`, this would be a boundary-only parity gap once the dpa3 graph descriptor is wired in PR-G.

deepmd/dpmodel/utils/neighbor_graph/angles.py
```diff
@@ -0,0 +1,254 @@
...
+        Shape (2, A) angle index pairs into edges.
+    num_edges : int
+        Total number of edges (E).
+
+    Returns
+    -------
+    Array
+        Shape (E,) or (E, ...) aggregated per-edge data.
+    """
+    return segment_sum(data, angle_index[0, :], num_edges)
```
iProzd, line 226: These angle aggregation helpers (`angle_to_edge_sum` / `angle_to_node_sum`) are thin `segment_sum` wrappers that do not take `angle_mask` and rely on the caller to zero padding angles first (padding angles point at edge 0 / node 0 via `pad_value=0`). That differs from `edge_force_virial`, which masks internally before scattering. With no consumer yet (dpa3 wiring is PR-G), this is just an ergonomics footgun: a caller that forgets the mask silently accumulates padding into edge 0 / node 0. Could we either take `angle_mask` and mask internally (consistent with the edge force/virial path), or make the "caller must mask first" contract explicit in the signature/docstring?

deepmd/dpmodel/utils/neighbor_graph/angles.py
wanghan-iapcm: Confirmed intentional — dpa3's dense angle gate uses the same strict `<` (`repflows.py:598`: `a_dist_mask = (safe_for_vector_norm(diff, axis=-1) < self.a_rcut)`), so this matches dense parity rather than diverging from it. The edge channel's `<= rcut` and the angle channel's `< a_rcut` are two independently-defined gates by design, not a single gate reused twice, so there's no requirement they share an operator. Added a comment at the gate site (df4767707) citing the dpa3 line so this doesn't need re-asking once PR-G wires up the dpa3 graph descriptor.

deepmd/dpmodel/utils/neighbor_graph/angles.py
wanghan-iapcm: Went with the second option: made the "caller must mask first" contract explicit rather than folding `angle_mask` into the functions. Reasoning: unlike `edge_force_virial`, these are thin single-line `segment_sum` wrappers used in more places than just the force/virial path (energy accumulation too), so an internal mask-and-multiply would add a branch and an extra elementwise pass to every call site, most of which already have masked data on hand from an upstream step. Rather than silently re-deriving that from a raw `angle_mask` here. Docstrings for both functions now state the precondition and the required action (`data * angle_mask` before calling) — see df4767707. The existing `test_angle_aggregation_ignores_padding_when_masked` test (added earlier in this PR) already pins the failure mode if a caller forgets. Happy to revisit if PR-G's dpa3 wiring finds this awkward in practice.

deepmd/dpmodel/utils/neighbor_graph/angles.py
```diff
@@ -0,0 +1,273 @@
...
+    -------
+    Array
+        Shape (A,) cosine values, one per angle slot (valid and padding).
+        Padding slots carry arbitrary values; mask with angle_mask before use.
+    """
+    xp = array_api_compat.array_namespace(edge_vec)
+    va = xp.take(edge_vec, angle_index[0, :], axis=0)  # (A, 3)
+    vb = xp.take(edge_vec, angle_index[1, :], axis=0)  # (A, 3)
+    na = va / (xp.linalg.vector_norm(va, axis=-1, keepdims=True) + eps)
+    nb = vb / (xp.linalg.vector_norm(vb, axis=-1, keepdims=True) + eps)
```
OutisLi: `graph_angle_cos` normalizes with `xp.linalg.vector_norm`, but the dpa3 channel this docstring claims to mirror exactly uses `safe_for_vector_norm` for this very step (`deepmd/dpmodel/descriptor/repflows.py:642-643`, and the `a_rcut` gate at `:598`).

The two agree in value for non-zero vectors — `safe_for_vector_norm` is literally `where(||x||>0, vector_norm(x), 0)` — which is why the fp64 dense/se_t parity passes at 1e-12. They differ only in the gradient at `||v|| == 0`: `v / (||v|| + eps)` with `vector_norm` back-props to `NaN` there, whereas `safe_for_vector_norm` exists precisely to yield a 0 gradient.

Since `edge_vec` is the sole autograd leaf, this is a latent `NaN`-gradient landmine on the geometry path. It is not triggered by the current machinery (real edges are non-zero; padding angles point at `pad_value=0`, an in-range real edge) and no test exercises gradients, so it would only surface once the dpa3 graph descriptor is wired (PR-G) and gradients flow, or if the `angle_index` construction contract changes.

Suggest normalizing with `safe_for_vector_norm` (from `deepmd.dpmodel.utils.safe_gradient`) here, and using it for the `a_rcut` gate at line 83 too (value-identical, non-differentiated, but keeps the dpa3 parity claim honest and avoids confusing future readers). That gives exact parity as claimed and removes the `NaN`-gradient edge case on the autograd leaf.

OutisLi (changes requested):
Thanks for the careful machinery and the thorough oracle/parity tests. The forward parity is solid; my one change request is about autograd safety of the norm used in `graph_angle_cos` (details inline). A couple of minor, non-blocking notes will follow separately.

deepmd/dpmodel/utils/neighbor_graph/angles.py
```diff
@@ -0,0 +1,273 @@
...
+    a_edge_mask = xp.astype(edge_mask, xp.bool) & (dist < a_rcut)
+    # compact eager form only (static_nnei not exposed until angle export is
+    # needed, PR-G). dst = edge_index[1, :] per the [src, dst] SoA convention.
+    q_e, k_e, pair_mask = center_edge_pairs(
+        edge_index[1, :],
+        a_edge_mask,
+        n_total,
+        include_self=include_self,
+        ordered=ordered,
+    )
```
OutisLi, line 100: **(non-blocking, perf follow-up)** For the angle path (`ordered=False, include_self=False`), the `center_edge_pairs` compact form materializes each center's full `d×d` square (`P = Σ deg²`) and then `nonzero`-compacts it down to the `Σ C(deg, 2)` upper triangle. The intermediate is ~2× the output, and the `nonzero` adds one more data-dependent op (an extra unbacked SymInt under `torch.export`). The `A ~ sum(deg²)` cost listed under "known limitations" is really this over-materialization — the upper triangle can be emitted directly.

Because edges are grouped by a **stable** sort of `nonzero(edge_mask)` (increasing edge ids), within each center block the edge ids increase with sorted position. So emitting `(earlier_pos, later_pos)` reproduces the oracle's `(m ≤ n)` orientation exactly. The `ordered=False` branch can be:

```python
# each query edge emits only its in-center upper-triangle partners
localpos = iota_r - xp.take(start, ds, axis=0)   # rank within its center block
self_off = 0 if include_self else 1              # >= (self) vs > (no-self)
n_emit = deg - localpos - self_off               # keys emitted per query edge (>= 0)
query_sorted = xp.repeat(iota_r, n_emit)
xp_hint_dynamic_size(query_sorted)
pair_off = xp.cumulative_sum(n_emit) - n_emit
iota_t = xp.cumulative_sum(xp.ones_like(query_sorted)) - 1
ramp = iota_t - xp.take(pair_off, query_sorted, axis=0)
key_sorted = query_sorted + self_off + ramp      # max = start_c + deg_c - 1 -> stays in-block
query_edge = xp.take(eid, query_sorted, axis=0)
key_edge = xp.take(eid, key_sorted, axis=0)
return query_edge, key_edge, xp.ones((query_sorted.shape[0],), dtype=xp.bool, device=dev)
```

`key = query + self_off + ramp` is capped by `n_emit` at the center block's last slot, so "same center" holds by construction — no cross-center check and no `nonzero`. The `ordered=True` branches keep the full-square path unchanged.

I transcribed both the current and this version in numpy and checked against the brute-force oracle over 3000 random `(dst, mask)` inputs × all 4 flag combos: **0 mismatches, orientation-exact**, so `test_center_edge_pairs.py` passes untouched. Measured unordered/no-self materialization dropped from `Σ deg² = 317601` to `Σ C(deg, 2) = 138051` (0.435×), and it removes the `nonzero`.

This lives in `center_edge_pairs` (`pairs.py`, the shared primitive), and PR-E is the first live consumer of the unordered path — so flagging it here; feel free to defer to when the dpa3 graph descriptor lands in PR-G.

OutisLi (commented):
One non-blocking performance follow-up on the unordered angle enumeration (details inline). The change itself lands in the shared `center_edge_pairs` primitive.

deepmd/dpmodel/utils/neighbor_graph/angles.py
```diff
@@ -0,0 +1,273 @@
...
+            pair_mask,
+            xp.zeros(
+                (am.shape[0] - pair_mask.shape[0],),
+                dtype=xp.bool,
+                device=array_api_compat.device(pair_mask),
+            ),
+        ],
+        axis=0,
+    )
+    return ai, am & pm_padded
```
OutisLi, line 118: **(non-blocking, readability)** In the compact form — the only one reachable here, since `build_angle_index` never passes `static_nnei` — `center_edge_pairs` returns an all-`True` `pair_mask` over the already-filtered pairs, and its length equals the real-angle prefix. So `pm_padded` is `True` exactly on `[0, A_real)` and `False` afterwards, which is identical to `am`; `am & pm_padded` is thus an identity, and this fold is a no-op in this PR.

The comment correctly notes it is kept for the future shape-static form (PR-G), where `pair_mask` will carry `False` entries. Since that path does not exist yet, consider dropping the fold for now and returning `ai, am` directly, then reintroducing it together with the shape-static enumeration in PR-G. That keeps this PR free of an identity transform the reader has to reason about. Not blocking either way.

OutisLi (commented):
One more non-blocking readability note on the pair_mask fold (inline).

deepmd/dpmodel/utils/neighbor_graph/angles.py
wanghan-iapcm: Fixed in [`10e42a35`](https://github.com/deepmodeling/deepmd-kit/pull/5717/commits/10e42a351cfaf5dd0ae47b8d6eb922fd560e72d7). You're right on all counts — verified against the dense code: `repflows.py:598` gates with `safe_for_vector_norm(diff, axis=-1) < self.a_rcut` and `repflows.py:642-643` normalizes with `safe_for_vector_norm`, while both `angles.py` sites used plain `vector_norm`. Swapped both (the normalization in `graph_angle_cos` and the `a_rcut` gate) to `safe_for_vector_norm`, and updated the two "mirrors dpa3 exactly" comments to name the function so the parity claim is honest.

Values are unchanged for real non-zero edges (fp64 dense/se_t parity still passes at 1e-12). Added `test_graph_angle_cos_zero_edge_grad_is_finite`: a jax-autograd test on a zero-length `edge_vec` asserting the geometry-path gradient stays finite, plus a discrimination check that plain `vector_norm` DOES back-prop NaN under jax on the same construction — so the test actually exercises the safe path. (Confirmed torch's `vector_norm` already defines a 0 subgradient at zero, so jax is the discriminating backend; the test is `pytest.importorskip('jax')`-guarded.)

deepmd/dpmodel/utils/neighbor_graph/angles.py
wanghan-iapcm: Agreed the direct upper-triangle emission is the right shape for the unordered path, and thanks for transcribing + oracle-checking both versions (the 0.435× materialization drop and dropping the data-dependent `nonzero` are both worth having). Deferring to PR-G as you suggested: the change lands in the shared `center_edge_pairs` primitive rather than this PR's angle code, and PR-G is where the shape-static enumeration (the branch that makes this path hot and adds the export-time unbacked SymInt) actually lands — so it's cleaner to do the rewrite together with its first shape-static consumer and re-verify `test_center_edge_pairs.py` against that. Tracking it for PR-G.

deepmd/dpmodel/utils/neighbor_graph/angles.py
wanghan-iapcm: You're right that `am & pm_padded` is an identity in the compact form — in this PR `pair_mask` is all-`True` over the already-filtered pairs, so the fold is a no-op. I'd prefer to keep it (with the existing comment explaining why) rather than drop-and-reintroduce: the shape-static form in PR-G is the only path that emits `False` entries in `pair_mask`, and this is the single site that folds them into `angle_mask`. Keeping the sink in place now means PR-G only has to flip `center_edge_pairs` to the shape-static branch without also re-threading the invalidation back through `build_angle_index` — and the comment already flags it as forward-compat scaffolding for the reader. Happy to revisit if you feel the identity transform is more confusing than the churn it avoids.

iProzd (approved):

OutisLi (approved):

PR #5722, opened 2026-07-03
fix: make global DOS inference work on the dpmodel and JAX backends

### LINKED ISSUE #5674: [Code scan] Use reduced DOS outputs for non-atomic DeepDOS inference

## Problem

`DeepDOS` declares a reducible atomic `dos` output, so `ModelOutputDef` also creates the reduced `dos_redu` output:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/infer/deep_dos.py#L38-L52

However, `DeepDOS.eval()` always reads `results["dos"]` and recomputes the global DOS by summing atomic DOS, even when `atomic=False`:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/infer/deep_dos.py#L127-L138

That is not compatible with all backends. The dpmodel backend only requests reduced, coordinate-derivative, and cell-derivative outputs when `atomic=False`, so the atomic OUT variable `dos` is not included:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/infer/deep_eval.py#L266-L278

The JAX backend uses the same non-atomic request filtering:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/jax/infer/deep_eval.py#L273-L285

`dp test` passes `atomic=has_atom_dos`, so a normal global-DOS-only test path calls `DeepDOS.eval(..., atomic=False)`:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/entrypoints/test.py#L1020-L1028

## Impact

dpmodel and JAX DOS inference can fail with a missing `results["dos"]` key when the caller only needs the global DOS. This affects global-DOS-only workflows even though the backend already has the reduced `dos_redu` output available.

## Suggested fix

Use `results["dos_redu"]` for the global DOS when it is present, and only read/sum `results["dos"]` when atomic DOS was requested or when a backend does not provide the reduced output. This also avoids requesting atomic DOS just to compute a global quantity.

wanghan-iapcm:
## Problem

Fixes #5674. Three related defects in global DOS inference, uncovered in sequence:

1. `DeepDOS.eval` unconditionally read the atomic `dos` output and summed it, even for `atomic=False`. The dpmodel and JAX backends only return the atomic output when `atomic=True`; on the global-DOS-only path (e.g. `dp test` without atomic DOS labels) `results["dos"]` raised `KeyError`. TF and PyTorch masked this because they always include the atomic output.

2. Fixing the `KeyError` exposed that dpmodel and JAX DOS inference was broken more deeply: both `DeepEval.get_numb_dos` implementations hard-returned `0`, so the DOS reshape target was `(nframes, 0)` and inference failed on every path, not just the missing-key case.

3. On TF, the global DOS did not equal the sum of the atomic DOS for multi-frame inputs — even though, by construction of the model, it must. `deepmd/tf/model/dos.py` reduced the atomic DOS with `reshape([natoms[0], -1])` + `reduce_sum(axis=0)`, which sums across the wrong axis and mixes atoms from different frames together. Single-frame inputs happened to give the right answer, so no test caught it.

## Fix

Backend-agnostic (`deep_dos.py`): prefer the atomic `dos` output and sum it whenever the backend returns it (this is the exact global DOS on TF/PT, whose reduced output is not necessarily the plain sum), and fall back to the reduced `dos_redu` only when the atomic output is absent (dpmodel/JAX at `atomic=False`). Reading `dos` unconditionally is what raised the original `KeyError`.

dpmodel: add `get_numb_dos` to the dpmodel `DOSModel` (mirroring the PyTorch model), add a default `get_numb_dos` returning 0 on the shared base model so non-DOS models can still be serialized, and delegate `dpmodel/infer/deep_eval.py:get_numb_dos` to the model.

JAX: the evaluator wraps a deserialized `HLO` object with no live model, so `numb_dos` is now persisted into the StableHLO export constants and exposed via `HLO.get_numb_dos`; the `dos` output is registered in the HLO `OUTPUT_DEFS` table; and `jax/infer/deep_eval.py:get_numb_dos` delegates to the model. With these, JAX DOS inference works end to end.

TF: reduce the atomic DOS per frame — `reshape([-1, natoms[0], numb_dos])` + `reduce_sum(axis=1)`, mirroring the energy model — so the global DOS equals the atomic sum for multi-frame inputs.

## Test

- `source/tests/common/dpmodel/test_deep_dos.py`: builds a dpmodel DOS model and evaluates it — `atomic=False` returns the global DOS (`KeyError` on master), and the global DOS equals the sum of the atomic DOS (guarding the `dos_redu == sum(dos)` invariant relied on by all backends).
- `source/tests/jax/test_deep_dos.py`: exports a DOS model to `.hlo`, checks `numb_dos` survives the round trip, and evaluates the global DOS.
- `source/tests/tf/test_model_dos.py`: adds `test_multiframe_global_equals_atomic_sum`, which builds a two-frame DOS graph and asserts the global DOS equals the per-frame atomic sum — this fails on the old axis-0 reduction and passes after the per-frame fix. The existing single-frame assertions were updated to the corrected output shapes.

dpmodel and JAX DOS inference previously had no test, and the TF path had only single-frame coverage; DOS was effectively exercised only where the bugs were masked.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected DOS inference so global DOS is reported properly instead of using a fixed default.
  * Fixed multi-frame DOS aggregation to keep results separated by frame and sum across atoms correctly.
  * Improved consistency when using the model in different backends and after export, so DOS output counts are preserved.

* **New Features**
  * Added support for exposing DOS output counts in model inference and export workflows.


deepmd/infer/deep_dos.py
```diff
@@ -124,18 +124,20 @@ def eval(
...
+        # the original global DOS for backends that always return it (TF, PT),
+        # whose reduced output is not necessarily the plain sum of the atomic
+        # DOS. The dpmodel and JAX backends omit the atomic `dos` when
+        # atomic=False, so fall back to the reduced `dos_redu` there (reading
+        # `dos` unconditionally would raise KeyError).
+        if "dos" in results:
+            atomic_energy = results["dos"].reshape(nframes, natoms, self.get_numb_dos())
+            energy = np.sum(atomic_energy, axis=1)
+            if atomic:
+                return (energy, atomic_energy)
```
njzjz: We may need to review the implementation of dpmodel and fix it, instead of adding a workaround here.

wanghan-iapcm: Good catch. dpmodel/JAX weren't actually deficient — the real issue was that `deep_dos.py`'s eval() didn't follow the canonical pattern used by `deep_pot.py` (always read the reduced `*_redu` output for the global value, only touch the atomic output when `atomic=True`). `dos_redu` is requested by every backend regardless of the `atomic` flag, so summing the atomic output manually was never necessary — that only existed because TF's `dos_redu` was computed with a buggy per-frame reduction (fixed elsewhere in this PR).

Pushed a follow-up (e85befe) that removes the workaround and mirrors `deep_pot.py`'s pattern directly. This exposed that the checked-in `deepdos.pbtxt` test fixture still embedded the pre-fix TF reduction graph (predates this PR), so I regenerated it via `serialize_from_file`/`deserialize_to_file` (preserves the trained weights, rebuilds the graph with current code) — all existing expected values in `test_deepdos.py` still pass unchanged.

njzjz (approved):

PR #5724, opened 2026-07-03
fix(pt_expt): dispatch property models to DeepProperty inference

### LINKED ISSUE #5671: [Code scan] Dispatch pt_expt property models to DeepProperty

## Problem

The experimental PyTorch backend registers and constructs property models:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/pt_expt/model/property_model.py#L27-L43

The training path also has property-specific handling that expects `get_task_dim()`, `get_var_name()`, and `get_intensive()`:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/pt_expt/train/training.py#L216-L219

But `pt_expt` `DeepEval.model_type` only dispatches energy, DOS, dipole, polar, and WFC outputs, then raises `RuntimeError("Unknown model type")`:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/pt_expt/infer/deep_eval.py#L671-L695

The stable PyTorch and Paddle inference adapters both import `DeepProperty`, expose the property metadata APIs, and dispatch to `DeepProperty` when the model output contains the property variable name:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/pt/infer/deep_eval.py#L406-L456

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/pd/infer/deep_eval.py#L239-L289

## Impact

A `pt_expt` property checkpoint can be constructed and trained by the backend stack but fail user-facing inference wrapper selection with `RuntimeError("Unknown model type")`. Even if dispatch were bypassed, the backend evaluator does not expose the property metadata APIs expected by `DeepProperty`.

## Suggested fix

Mirror the stable PyTorch/Paddle property support in `deepmd/pt_expt/infer/deep_eval.py`: import `DeepProperty`, expose `get_intensive()`, `get_var_name()`, and `get_task_dim()`, and dispatch to `DeepProperty` when the property variable appears in `model_output_type`.

wanghan-iapcm:
## Problem

Fixes #5671. A pt_expt property checkpoint could be constructed and trained by the backend stack but not evaluated. `pt_expt` `DeepEval.model_type` dispatched only energy, DOS, dipole, polar, and WFC outputs and then raised `RuntimeError("Unknown model type")` for a property model, and the evaluator did not expose the `get_intensive` / `get_var_name` / `get_task_dim` getters that `DeepProperty` needs. Underneath, the pt_expt `PropertyModel` itself only implemented `get_var_name`, not `get_task_dim` or `get_intensive`, so simply mirroring the dispatch would not have been enough.

## Fix

- `pt_expt/model/property_model.py`: add `get_task_dim` (fitting output dimension) and `get_intensive` (from the output def), mirroring the PyTorch property model.
- `pt_expt/infer/deep_eval.py`: import `DeepProperty`, dispatch to it when the property variable name appears in the model output, and expose `get_var_name` / `get_task_dim` / `get_intensive` that delegate to the reconstructed model.

The dispatch branch and the getters are guarded by `hasattr` (and by `self._dpmodel is not None`), so:
- energy/DOS/dipole/polar/WFC models keep matching their own branches first;
- genuinely unknown model types still fall through to `"Unknown model type"` rather than raising an `AttributeError`;
- metadata-only mode (no reconstructed dpmodel, used by the C++ AOTI path) raises a clear `NotImplementedError` for the property getters instead of mis-dispatching.

## Test

Adds `source/tests/pt_expt/infer/test_deep_eval_property.py`, a full `serialize -> .pte -> DeepEval` round trip asserting `model_type is DeepProperty`, the three getters, and the eval output shape. Without the fix, constructing the `DeepEval` raises `RuntimeError("Unknown model type")`.

Verification for the reviewer's peace of mind: pt-backend property inference works end to end via the same mechanism (dispatch + getters + eval), and the full pt_expt energy inference suite (`source/tests/pt_expt/infer/test_deep_eval.py`, 92 passed / 2 skipped) is unaffected by the added branch.

## Summary by CodeRabbit

* **New Features**
  * Added support for property-style models across inference backends, including correct model dispatch to property evaluation.
  * Exposed property metadata (variable name, task dimension, and intensive/extensive flag) via new public getters and preserved it through export/import.
* **Bug Fixes**
  * Fixed single-output evaluation so results are wrapped consistently and mapped correctly to requested outputs.
  * Corrected atomic property tensor reshaping behavior for non-atomic evaluation.
* **Tests**
  * Added end-to-end unit, JAX, and TF2 consistency tests covering dispatch, metadata getters, and output shapes/values.


njzjz (commented):
See above

wanghan-iapcm:
Addressed in 8198ae381 — extended property inference to all three backends (dpmodel, jax, tf2), each with a full serialize → artifact → DeepEval round-trip test asserting the DeepProperty dispatch, the three getters, and that the multi-frame global property equals the atomic sum.

Rather than copy the dispatch and getters into each backend, I centralized them in the shared `DeepEvalBackend` base: `model_type` is now a concrete property dispatching on the model output types (and the property var name via `get_model()`), and `get_var_name`/`get_task_dim`/`get_intensive` delegate to `get_model()`. dpmodel/jax/tf2 drop their duplicated `model_type` overrides and inherit this; pt/pt_expt keep theirs (extra branches / different model access). The property getters live once on the dpmodel `PropertyModel`, which the jax and tf2 model classes already subclass, so they inherit them for free. The jax `HLO` and tf2 SavedModel wrappers hold an exported artifact with no live model, so var_name/task_dim/intensive are persisted into the StableHLO constants / SavedModel and read back.

Enabling the dispatch surfaced two latent bugs that single-frame tests had masked, also fixed here: `DeepProperty.eval` read the atomic output unconditionally (KeyError at atomic=False on dpmodel/jax — now reads the reduced output for the global value, atomic only when atomic=True, mirroring `DeepPot.eval`); and `AutoBatchSize.execute_all` unwraps a single-output result out of its tuple, making the eval zip iterate over the frame axis (this also fixes multi-frame global-only DOS inference on these backends).

njzjz (approved):

PR #5732, opened 2026-07-04
fix(lmp): avoid double qe2f in DPLR efield

### LINKED ISSUE #5646: [Code scan] Apply qe2f only once in fix dplr efield forces

## Problem

`fix dplr efield` applies the LAMMPS charge-to-force conversion factor twice.

Evidence:

- The fix stores `force->qe2f` in `qe2f`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/lmp/fix_dplr.cpp#L59-L65
- Constant `efield` values are stored as `qe2f * value`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/lmp/fix_dplr.cpp#L94-L112
- Equal-style variable `efield` values are also stored as `qe2f * value`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/lmp/fix_dplr.cpp#L834-L845
- The force is later computed as `q[ii] * efield[dd] * force->qe2f`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/lmp/fix_dplr.cpp#L646-L659

## Impact

The external-field force and related energy/virial bookkeeping are scaled by an extra `qe2f`. In unit systems where `qe2f != 1`, `fix dplr efield` produces incorrectly scaled electric-field contributions.

## Suggested Fix

Store raw electric-field values in `efield` and apply `qe2f` once when forming forces, or keep converted values and remove the second multiply. Add a one-atom charged DPLR regression comparing the force against `q * E * qe2f` for constant and equal-style variable fields.

Yi-FanLi:
## Summary
- remove the second qe2f multiplication when forming fix dplr efield forces
- keep the existing conversion when storing constant and equal-style efield values
- add real-unit regressions for constant and equal-style efield paths so qe2f != 1 is covered

Closes #5646.

## Tests
- git diff --check
- /tmp/deepmd-check-venv/bin/ruff check .
- /tmp/deepmd-check-venv/bin/ruff format --check .

## Not run
- Targeted LAMMPS DPLR pytest tests could not run in the local ad hoc venv: the lammps Python module is not installed there, and pytest startup segfaulted in the temporary environment before collection.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected electric-field force calculations so results are accumulated consistently in real-unit simulations.
  * Fixed related force and virial contributions to match expected values.

* **Tests**
  * Added coverage for electric-field behavior in real units.
  * Included checks for both constant and variable field settings to confirm force results remain accurate.

### DIFF
--- a/source/lmp/fix_dplr.cpp
+++ b/source/lmp/fix_dplr.cpp
@@ -652,7 +652,7 @@ void FixDPLR::post_force(int vflag) {
     for (int ii = 0; ii < nlocal; ++ii) {
       double tmpf[3];
       for (int dd = 0; dd < 3; ++dd) {
-        tmpf[dd] = q[ii] * efield[dd] * force->qe2f;
+        tmpf[dd] = q[ii] * efield[dd];
       }
       for (int dd = 0; dd < 3; ++dd) {
         dfele[ii * 3 + dd] += tmpf[dd];
--- a/source/lmp/tests/test_dplr.py
+++ b/source/lmp/tests/test_dplr.py
@@ -355,6 +355,13 @@ def lammps_si():
     lmp.close()
 
 
+@pytest.fixture
+def lammps_real():
+    lmp = _lammps(data_file=data_file, units="real")
+    yield lmp
+    lmp.close()
+
+
 def test_pair_deepmd_sr(lammps) -> None:
     lammps.pair_style(f"deepmd {pb_file.resolve()}")
     lammps.pair_coeff("* *")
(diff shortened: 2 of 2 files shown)


njzjz (approved):
LGTM

Issue #18, opened 2019-02-24
Exclude certain atoms from neighbourlist


Jimbo994:
Dear DeePMD developers,

My name is Jim, I am a Msc chemistry student at the University of Amsterdam. I am currently doing my master research internship on the topic Neural network potentials.

I have read the DeePMD papers and installed your packages and I think it is really exciting stuff. However, I would like to test some things out but I am not sure how to do this.

I want to be able to take only certain "interactions" into account. With that I mean the following: say that I have a system containing water molecules, Now I want oxygen atoms to only feel hydrogen atoms instead of feeling both hydrogen atoms and oxygen atoms. The way I was thinking of achieving this is by adjusting the way the Dij matrix is created.

I want to be able to specify what atoms I want in the neighbour list, so say that I am looking at an oxygen atom, then I want its neighbour list or Dij matrix only to contain hydrogen atoms. In this way, I am forcing the oxygen atom to only feel the hydrogen atoms. 

In this way, it would hopefully be possible to reduce computational cost but still obtain reliable force fields. Would this be possible to do by making adjustments to your code?

Many thanks in advance,

Jim Boelrijk

jameswind:
Hi Jim,

This is definitely possible but it requires some modifications. We now have
two major versions of Deep Potential, the original one and the DeepPot-SE
version. Which one would you like to work on?  Let me answer your questions
in more details after hearing more from you.
Besides reducing the number of inputs, you may also consider reducing the
number of atomic net. For example, when we consider a molecular system like
water, maybe the construction could be the summation of atomic energies of
only oxygen. This may have some limitations and we did not do any tests,
but it will definitely reduce computational cost if it works.

Best,
Linfeng

Jimbo994:
Hi Linfeng,

After having studied the paper on DeepPot-SE, I am aware that this version should in principle give better results due to it being smooth and continuously differentiable.  In that regard it seems wise to use DeepPot-SE. However, I cannot yet fully grasp the construction of symmetry preserving functions in the way that it is described in the paper.

I do have a reasonable feeling of how the original version works, so perhaps it is best to (first) work on that one. 

I am not sure if reducing the number of atomic net will achieve what I want. In the example you have given of water, where you only obtain atomic energies of oxygen, would you still be able to perform molecular dynamics of water with the neural net? It probably would not be able to obtain energies of hydrogen atoms right? 

Something that I would like to check for example is the following:
[...]

PR #5733, opened 2026-07-05
refactor: pair exclude_types as canonical NeighborGraph transform; dpa1 graph path supports exclude_types (decision #18)

wanghan-iapcm:
## Pair `exclude_types` as a canonical NeighborGraph transform 

> **Stacked on #5715 (NeighborGraph PR-D).** Only the commits after `6fc45bd26` belong to this PR; will rebase onto master once #5715 merges.

Makes pair-type exclusion a single canonical transform applied once at the neighbor-list/graph build seam, and uses it to add descriptor-level `exclude_types` support to the dpa1 graph path (removing that eligibility gate), consistently across dpmodel, pt_expt, jax, and the C++ inference path.

### What changed
- **Canonical graph transform** `apply_pair_exclusion(graph, atype, pair_excl, *, compact=False)` in `deepmd/dpmodel/utils/neighbor_graph/`: ANDs `PairExcludeMask.build_edge_exclude_mask` into `graph.edge_mask`. `compact=False` is mask-only (shape-static, export/AOTI-safe); `compact=True` drops masked edges (eager-only; raises on angle-carrying graphs). Idempotent.
- **Atomic-model seam** (`base_atomic_model`): model-level `pair_exclude_types` is applied at BUILD time, so `forward_common_atomic{,_graph}` no longer re-apply it — they consume a pre-excluded nlist/graph (single-owner, **not** a backstop). To keep that from being fail-open, an eager (numpy) fail-safe assertion (`_assert_nlist_pair_excluded` / `_assert_graph_pair_excluded`) rejects a non-excluded input at the seam with a clear error; it is a no-op under `torch.export` / jax `jit` and in compiled production (the exported-`.pt2` / C++ paths are covered by ingestion-site tests). See the ingestion-path inventory below.
- **dpa1 graph path supports descriptor-level `exclude_types`**: the `NotImplementedError` and the `uses_graph_lower()` exclude condition are removed; exclusion applied inside `DescrptBlockSeAtten.call_graph` before the segment sums. Graph-vs-dense parity at non-binding sel is exact (rtol=atol=1e-12, attn_layer 0 and 2, type_one_side both).
- **Build-time exclusion (dispatcher)**: `build_neighbor_graph` and the pt_expt graph builders (dense/ase/vesin/nv) gain optional `pair_excl`/`compact` with default post-search application; `_call_common_graph` passes model-level excludes at build time; oracle set-equality tests per available builder.
- **Dense-nlist port**: `apply_pair_exclusion_nlist(nlist, atype_ext, pair_excl)` extracted from the inline seam code; `build_neighbor_list` + Vesin/Nv/Default strategies gain `pair_excl`; `return_mode='edges'` + `pair_excl` fails fast.
- **C++ twin**: `buildPairExcludeTable` / `applyPairExclusion` / `applyPairExclusionNlist` in `source/api_cc/include/commonPT.h`, mirroring the Python transforms (same arg order/variable names, cross-referenced docs); `pair_exclude_types` serialized into `.pt2` `metadata.json` and rebuilt once in `DeepPotPTExpt::init` (device-resident table, uploaded once — no per-step H2D). Exclusion is applied at the C++ ingestion seam — the **single owner** on every run path; the exported lower consumes a pre-excluded input and never re-applies it. New gtest (8 tests) vs Python DeepEval reference at 1e-10, plus multi-rank LAMMPS exclusion tests (below).
- **Fix**: `apply_pair_exclusion` uses `logical_and` + bool cast (array_api_strict rejected `bool*bool`), caught by the jax/strict consistency rows now traversing the graph path.

### Ingestion-path inventory (exclusion coverage)

Because the seam is now the single owner (fail-open if skipped), here is every entry point that builds a nlist/graph reaching the exclusion-owning lower, and how each applies exclusion:

| Entry point | Applies exclusion at | Coverage |
|---|---|---|
| dpmodel `_call_common` (dense) | `build_neighbor_list(pair_excl=)` | eager guard + consistency rows |
| dpmodel `_call_common_graph` | `build_neighbor_graph(pair_excl=)` | eager guard + graph/dense parity |
| dpmodel descriptor-level `exclude_types` | `apply_pair_exclusion` in `call_graph` | dpa1 graph parity |
| pt_expt DeepEval — nlist | `apply_pair_exclusion_nlist` / vesin `pair_excl` | parity vs dpmodel |
| pt_expt DeepEval — graph | `_build_eval_graph(pair_excl=)` (dense/ase/vesin/nv) | parity vs dpmodel |
| pt_expt/jax/pd training | `build_neighbor_graph(pair_excl=)` in stat/forward | training e2e |
| input statistics | `build_neighbor_list(pair_excl=)` in `EnvMatStatSe.iter` | stat tests (hash key = follow-up, pre-existing) |
| **jax-md `call_lower`** | `apply_pair_exclusion_nlist` at the seam (**fixed here**) | `test_dense_neighbor_applies_model_pair_exclusion` |
| C++ SP dense | `applyPairExclusionNlist` | gtest |
| C++ MP dense (with-comm) | `applyPairExclusionNlist` (**fixed here**) | DPA3 MP≡SP + active-vs-baseline |
| C++ SP graph | `applyPairExclusion` | gtest |
| C++ MP graph (non-MP, extended) | `applyPairExclusion` | dpa1 MP≡SP + active-vs-baseline |
| C++ MP graph (message-passing) | — | **fail-fast** (PR-G) |
| C++ / DeepEval edge lower (SeZM/DPA4) | baked into the exported graph by the **pt backend** (unchanged) | out of scope — pt-backend export |

Guard: `forward_common_atomic{,_graph}` additionally carry an eager fail-safe assertion (numpy only) that rejects a non-excluded input, so any *future* dpmodel/jax ingestion miss fails loudly instead of silently including excluded pairs.

### Known limitations
- nv builder's `pair_excl` path has no local oracle test (CUDA-only); to be validated on a GPU box.
- Input statistics remain on the dense path (graph-native stats is a separate follow-on); the stat-cache **hash key** does not yet include `pair_exclude_types` — pre-existing (predates this PR) and tracked as a separate fix.
- smooth_type_embedding + exclude parity untestable at 1e-12 (pre-existing dense sel-padding divergence, #5715).
- `build_edge_exclude_mask` still returns int32 (bool cast at call sites; follow-up).

### Spin routing
Spin models auto-inject `exclude_types` (virtual/placeholder types) into their backbone descriptor; before this PR that condition *accidentally* kept spin on the dense path. With exclude_types now graph-eligible, spin backbones flipped onto the carry-all graph route, which (a) diverges from the sel-capped reference on sel-binding spin systems and (b) trips a torch-inductor scatter codegen assertion during spin `.pt2` export. Fixed explicitly: `DescrptDPA1.disable_graph_lower()` (not serialized; re-derived structurally) is set in `SpinModel.__init__` — the single choke point covering `get_spin_model`, `SpinModel.deserialize`, and the pt_expt spin classes — plus a belt-and-braces `neighbor_graph_method="legacy"` at `SpinModel.call_common`. Regression tests pin the routing and its serialize→deserialize survival; the full spin export suite (23) and spin checkpoint-interop suite (12) are green.

### Verification
Full pt_expt suite: 1196 passed / 39 skipped / 3 failed — the 3 failures (`test_dpa4_freeze_to_pt2`, `test_dpa4_deep_eval_*`) are byte-identical on the base commit (pre-existing torch-inductor dpa4 export issue on this box, unrelated). dpmodel exclusion suites 69 passed; consistency dpa1 99 passed/63 skipped (incl. jax + array_api_strict exclude rows); C++ `Dpa1PairExcl` gtest 8/8.

## Summary by CodeRabbit

* **New Features**
  * Added model-level pair-type exclusion across neighbor lists and neighbor graphs, including propagation into exported `.pt2` metadata and enforcement at inference ingestion.
  * Expanded DPA1 graph-native attention support (including higher attention layers) and introduced stable segmented reductions for graph attention.
  * Added export-time guards to block unsupported graph tracing on older torch versions.
* **Bug Fixes**
  * Improved dense vs graph consistency for exclusion/masking behavior.
  * Preserved spin model legacy neighbor routing for sel-binding cases.
* **Documentation**
  * Refreshed graph-export eligibility and backend behavior notes for attention/smooth differences.
* **Tests**
  * Added/expanded parity, exclusion, compaction, and tracing regression coverage.


source/api_cc/src/DeepPotPTExpt.cc
```diff
@@ -875,14 +892,27 @@ void DeepPotPTExpt::compute(ENERGYVTYPE& ener,
...
+                          aparam_tensor, charge_spin_tensor);
     } else {
-      flat_outputs = run_model(coord_Tensor, atype_Tensor, firstneigh_tensor,
-                               mapping_tensor, fparam_tensor, aparam_tensor,
-                               charge_spin_tensor);
+      // Model-level pair exclusion is a BUILD-time transform (decision
+      // #18/A4): the exported dense lower consumes a pre-excluded nlist and
+      // never re-applies it; this is the single application site on the C++
+      // dense route.
+      const at::Tensor excl_nlist = deepmd::applyPairExclusionNlist(
```
OutisLi, line 939: Single-rank dense applies exclusion here, but its siblings do not: `run_model_with_comm` (L842, multi-rank dense) and `run_model_edges` / `run_model_edges_with_comm` (L822 / L835 / L851 / L1265, edge schema) pass the raw tensors. Since the exported `.pt2` no longer applies model-level exclusion (removed from `forward_common_atomic`), those paths silently drop it, so multi-rank ≠ single-rank for message-passing models and edge-schema models lose it entirely. Please thread `applyPairExclusionNlist` / `applyPairExclusion` through them too (or fail-fast).

OutisLi (changes requested):
## Model-level `pair_exclude_types` is silently dropped on the C++ multi-rank (message-passing) dense path and the edge-schema paths

This PR moves model-level `pair_exclude_types` from a consume-time transform to a build-time (ingestion-seam) transform: `BaseAtomicModel.forward_common_atomic` / `forward_common_atomic_graph` no longer apply it, so the exported `.pt2` lower no longer carries it and the C++ ingestion seam becomes the **only** application site.

But the seam was added to only two of the run paths in `DeepPotPTExpt::compute`:
- ✅ single-rank graph → `run_model_graph` (`applyPairExclusion` on `edge_mask`)
- ✅ single-rank dense → `run_model` (`applyPairExclusionNlist` on the nlist)

The following paths pass the raw tensors and therefore **silently ignore** model-level exclusion:
- ❌ `run_model_with_comm` — multi-rank dense (`DeepPotPTExpt.cc:842`). `use_with_comm = has_comm_artifact_ && multi_rank` (`:521`), i.e. exactly the message-passing (DPA2/DPA3) path under `mpirun`.
- ❌ `run_model_edges` / `run_model_edges_with_comm` — edge schema, e.g. SeZM (`:822` / `:835` / `:851` / `:1265`).

### Consequences
- A message-passing model with `pair_exclude_types` gives multi-rank ≠ single-rank, and it is a **regression** vs. before this PR (exclusion used to be baked into the `.pt2`, covering every path).
- Edge-schema models lose exclusion on every rank.
- No fail-fast — the output is silently wrong.

dpa1 (this PR's focus) is unaffected: it has no comm artifact, so multi-rank still routes through `run_model`. The gap is latent for the model types this PR doesn't target, but the shared base-class change degrades them with no guard.

### Why it's easy to miss — the narrative contradicts the code
- The PR description says the atomic-model seam "stays as idempotent backstop" and "Multi-rank (mpirun) exclusion shares the same seam"; neither holds now.
- `vesin_neighbor_list.py` docstring and `pt_expt/model/make_model.py` (~L472) still describe an "idempotent backstop", contradicting the "NOT applied here" comments in `base_atomic_model.py`.

### Requested fix (keeping decision #18's single-build-seam design)
Apply the transform on **every** C++ run path — `applyPairExclusionNlist` on the `run_model_with_comm` nlist, and `applyPairExclusion` on the `edge_mask` for the edge / edge-with-comm paths — and add a fail-fast for any `pair_exclude_types` × path combination that genuinely cannot be covered. Please also reconcile the stale "backstop" comments/docstrings so the single-owner contract reads consistently.

source/api_cc/include/commonPT.h
```diff
@@ -462,6 +464,151 @@ inline GraphTensorPack buildGraphTensors(
...
+  }
+  const auto device = edge_mask.device();
+  const auto src = edge_index.index({0});  // (E,) neighbour
+  const auto dst = edge_index.index({1});  // (E,) center
+  const auto src_t = atype.index_select(0, src);
+  const auto dst_t = atype.index_select(0, dst);
+  // type_ij = atype[dst] * (ntypes + 1) + atype[src]  (matches Python)
+  const auto type_ij = dst_t * (ntypes + 1) + src_t;
+  const auto table =
+      torch::from_blob(const_cast<int*>(type_mask_table.data()),
```
OutisLi: This `from_blob` + `clone` + `to(device)` runs on every `compute()` step, uploading a constant table to the device each MD step (plus a possible CUDA stream sync). Build `pair_exclude_table_` as a device `torch::Tensor` once in `init` (device is known there from `gpu_id`/`gpu_enabled`), change this helper to take `const torch::Tensor& type_mask_table` and early-return on `!type_mask_table.defined()`, then `index_select` directly. Same for `applyPairExclusionNlist` (:602-606). Full sketch in the review summary.

OutisLi (changes requested):
## C++ hot path uploads the constant pair-exclusion table to the device every step

`applyPairExclusion` (`commonPT.h:547-551`) and `applyPairExclusionNlist` (`:602-606`) rebuild the keep table from the `std::vector` and copy it to the device on **every `compute()` call** (i.e. every MD step):

```cpp
const auto table =
    torch::from_blob(const_cast<int*>(type_mask_table.data()),
                     {static_cast<std::int64_t>(type_mask_table.size())},
                     torch::TensorOptions().dtype(torch::kInt32))
        .clone()
        .to(device);
```

`pair_exclude_table_` is a compile-time constant fixed in `init()`, and the device is fully determined at `init` (`gpu_id` / `gpu_enabled`), so this is a per-step CPU clone + H2D upload (plus a stream op / possible sync on CUDA) of a value that never changes. It is guarded by the empty-table early-return, so only exclusion-enabled models pay it — but for those it is pure per-step waste on the hot path.

### Proposed fix — upload once at `init`, pass the device tensor in

Store the table as a device-resident `torch::Tensor` built once, and let the two helpers consume it directly (an undefined tensor => identity, mirroring the current empty-vector early-exit). `buildPairExcludeTable` stays a pure, unit-testable vector builder.

**1. Member (`DeepPotPTExpt.h`)** — change the type:
```cpp
// Device-resident (ntypes+1)^2 keep table, uploaded once in init.
// Undefined tensor => no model-level exclusion.
torch::Tensor pair_exclude_table_;
```

**2. One-time upload in `init`** (device known here; same construction as `compute()` at :463-466):
```cpp
std::vector<int> tbl = deepmd::buildPairExcludeTable(ntypes, pair_exclude_types);
if (!tbl.empty()) {
  torch::Device device(torch::kCUDA, gpu_id);
  if (!gpu_enabled) {
    device = torch::Device(torch::kCPU);
  }
  pair_exclude_table_ =
      torch::from_blob(tbl.data(),
                       {static_cast<std::int64_t>(tbl.size())}, torch::kInt32)
          .clone()
          .to(device);
}
```

**3. Helpers take the prebuilt tensor** and drop the per-call `from_blob/clone/to`:
```cpp
inline torch::Tensor applyPairExclusion(const torch::Tensor& edge_index,
                                        const torch::Tensor& edge_mask,
                                        const torch::Tensor& atype,
                                        const torch::Tensor& type_mask_table,
                                        const int ntypes) {
  if (!type_mask_table.defined()) {
    return edge_mask;  // identity
  }
  const auto src = edge_index.index({0});
  const auto dst = edge_index.index({1});
  const auto type_ij =
      atype.index_select(0, dst) * (ntypes + 1) + atype.index_select(0, src);
  const auto keep = type_mask_table.index_select(0, type_ij).to(torch::kBool);
  return torch::logical_and(edge_mask, keep);
}
```
`applyPairExclusionNlist` mirrors this (`keep = type_mask_table.index_select(0, type_ij.reshape({-1})).reshape({nf, nloc, nnei})`). The table already lives on the model device (== `edge_mask` / `nlist` device), so `index_select` needs no per-call transfer.

This keeps the single-owner design intact, removes the per-step allocation + H2D entirely, and is self-contained: the helpers are called only inside `DeepPotPTExpt::compute` (the gtest drives them through the public API), so no other call site changes.

iProzd (changes requested):
The one canonical exclusion transform is a sound design, but moving it to the build seam and dropping the `forward_common_atomic*` backstop makes exclusion **fail-open**: those methods now document a "nlist/graph is already pre-excluded" contract, so any ingestion entry point that skips the build seam silently *includes* excluded pairs instead of erroring. For a security-relevant feature (exclusion forces zero interaction between types), silent inclusion is the dangerous failure mode — and the leaks are being found one at a time (jax_md `call_lower`, DeepEval `return_mode="edges"`, the C++ with-comm / edge-schema paths, the env_mat_stat hash key already flagged by njzjz-bot / OutisLi), which suggests there's no enumerated inventory of entry points.

Requesting changes on that architectural point: could we either keep a cheap fail-safe guard (an eager/debug assertion that inputs entering `forward_common_atomic*` are exclusion-applied), or enumerate every ingestion path in this PR with one regression test each, so coverage is complete-by-construction? The contract boundary is also currently untested (all exclusion tests go through the pre-excluding build seam), and the PR description's "stays as idempotent backstop" wording is now inaccurate at HEAD ("never re-applied here") — worth correcting so the blast radius reads right.

wanghan-iapcm:
## Pair-exclusion coverage across the C++ run paths

Thanks both — this was the right thing to flag. The build-seam refactor (decision #18/A4) made the C++ ingestion seam the single owner of model-level `pair_exclude_types`, and the seam was only wired into two of the run paths. Here is the full SP/MP × dense/graph matrix (`use_with_comm = has_comm_artifact_ && multi_rank`; **"comm" = message-passing DPA2/DPA3**):

| # | Cell | Code path | Exclusion | Status |
|---|------|-----------|-----------|--------|
| 1 | SP dense | `else→dense`, `applyPairExclusionNlist` | ✅ applied | works |
| 2 | MP + dpa1 dense | non-MP ⇒ no comm ⇒ `else→dense` | ✅ applied | works |
| 3 | **MP + comm dense** | `run_model_with_comm` | ❌ → ✅ **fixed** | now applied |
| 4 | SP graph | `else→graph`, `applyPairExclusion` | ✅ applied | works |
| 5 | MP + dpa1 graph | non-MP ⇒ `else→graph`, `n_node=nall_real` | ✅ applied | **now tested** |
| 6 | MP + comm graph | fail-fast (PR-G) | n/a | out of scope |

Cell 6 is not a silent drop — it hard-throws (`"Multi-rank message-passing graph … not yet supported (PR-G)"`), so exclusion is moot there.

### Fixed — cells 3 & 5

- **Cell 3** (`run_model_with_comm`, njzjz-bot #1 / OutisLi): now applies `applyPairExclusionNlist` on the nlist before the with-comm dense call, mirroring the single-rank site exactly (the cross-rank ghost exchange happens *inside* `run_model_with_comm` and does not change the nlist's meaning, so per-rank pre-exclusion is correct). Verified end-to-end on a **message-passing DPA3** model (`use_loc_mapping=False`, with-comm): `test_pair_deepmd_mpi_dpa3_pairexcl_matches_single_rank` asserts MP(`-n 2`) ≡ SP(`-n 1`) **and** that the excluded run differs from the same-weights no-exclusion baseline (so a silent drop on both ranks can't pass trivially).
- **Cell 5** (dpa1 MP graph): already correct — it shares the SP graph seam (line `:899`, `n_node = nall_real`) — but was untested with a non-empty exclusion set. Added `test_pair_deepmd_mpi_dpa1_pairexcl_graph_matches_single_rank` (same MP≡SP + active-vs-baseline shape).

### Perf (OutisLi)

The per-step table rebuild + H2D upload is removed: `pair_exclude_table_` is now a **device `torch::Tensor` uploaded once in `init`**; the two helpers take the prebuilt tensor and `index_select` it directly (undefined tensor ⇒ identity, mirroring the old empty-vector early-exit). `buildPairExcludeTable` stays a pure vector builder.

### Out of scope — the edge lower (njzjz-bot #2, #3)

The edge-schema lower (`run_model_edges*` and the `deep_eval` `return_mode="edges"` fast path) is a **pt-backend** export — SeZM/DPA4 via `deepmd/pt/entrypoints/freeze_pt2.py` (`lower_input_kind="edge_vec"`); pt_expt serialization only ever emits `"graph"`/`"nlist"`. This PR's entire **pt-backend** footprint is one file (`deepmd/pt/utils/nv_nlist.py`, +28), and the pt backend **still applies exclusion at consume time** (`deepmd/pt/model/atomic_model/base_atomic_model.py:363-366`), so a SeZM/DPA4 edge `.pt2` bakes exclusion into the exported graph — the C++/Python edge consumers correctly do **not** re-apply it (doing so would double-apply). This refactor only moves the ownership seam for the **dpmodel-derived backends** (pt_expt/jax/tf2); the pt-backend edge path is unchanged and unaffected, so nothing is dropped there.

### Deferred — stat-cache hash (njzjz-bot #4)

Valid, but **pre-existing**: input statistics depended on model-level `pair_exclude_types` before this PR too (via the accumulation-deselect that this PR replaced with a build-time nlist mask), and `EnvMatStatSe.get_hash()` never included it. It's also awkward to fix here — the model-level exclusion isn't visible at `get_hash` time (it arrives in the sample data, and `path / get_hash()` is computed before sampling). Tracking it as a separate fix so this PR stays scoped to the regressions it actually introduced.

### Also (OutisLi): stale "idempotent backstop" wording in `vesin_neighbor_list.py` and `pt_expt/model/make_model.py` reconciled with the single-owner design (both dpmodel seams now read "NOT applied here").

---

Commits: [`ff28370`](https://github.com/deepmodeling/deepmd-kit/pull/5733/commits/ff283707999a2d7528c2a18e504ef0d86439c908) (cell 3 fix + perf + docs + cell 5 test) and [`0ebe3e2`](https://github.com/deepmodeling/deepmd-kit/pull/5733/commits/0ebe3e23ce12dd5467daaacad6698aa19b9ad195) (DPA3 cell-3 regression test). The DPA3 multi-rank exclusion test was validated on a clean GPU box (the fixture is regenerated by `gen_dpa3.py` in CI).

source/api_cc/src/DeepPotPTExpt.cc
```diff
@@ -875,14 +892,27 @@ void DeepPotPTExpt::compute(ENERGYVTYPE& ener,
...
+          firstneigh_tensor, atype_Tensor, pair_exclude_table_, ntypes);
```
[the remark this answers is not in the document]
wanghan-iapcm: Fixed in [`ff28370`](https://github.com/deepmodeling/deepmd-kit/pull/5733/commits/ff283707999a2d7528c2a18e504ef0d86439c908): `applyPairExclusionNlist` is now applied on `firstneigh_tensor` before `run_model_with_comm`, mirroring the single-rank dense site. Added a with-comm regression test on a message-passing DPA3 model (`test_pair_deepmd_mpi_dpa3_pairexcl_matches_single_rank`) — MP(-n 2) ≡ SP(-n 1) AND excluded ≠ no-exclusion baseline — validated on a clean GPU box. Full path-by-path matrix + rationale in the [summary](https://github.com/deepmodeling/deepmd-kit/pull/5733#issuecomment-4913935637) (cell 3).

source/api_cc/src/DeepPotPTExpt.cc
```diff
@@ -1238,13 +1268,26 @@ void DeepPotPTExpt::compute(ENERGYVTYPE& ener,
                         edge_tensors.edge_index_ext, edge_tensors.edge_mask,
```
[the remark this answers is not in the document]
wanghan-iapcm: Out of scope — the edge lower is a **pt-backend** export (SeZM/DPA4 via `deepmd/pt/entrypoints/freeze_pt2.py`, `lower_input_kind="edge_vec"`); pt_expt serialization only emits `"graph"`/`"nlist"`. This PR does not touch the pt backend (its whole pt footprint is `nv_nlist.py`, +28), and the pt backend still applies exclusion at consume time (`pt/model/atomic_model/base_atomic_model.py:363-366`), so the edge `.pt2` bakes exclusion into the exported graph — the C++ edge consumers correctly do NOT re-apply it (that would double-apply). So exclusion isn't dropped here. Details in the [summary](https://github.com/deepmodeling/deepmd-kit/pull/5733#issuecomment-4913935637) ("Out of scope — the edge lower").

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -1772,19 +1791,26 @@ def _build_eval_graph(
         selection is a pure performance choice and results are unchanged.
         """
         method = self._neighbor_graph_method
+        # Model-level ``pair_exclude_types`` is a graph-BUILD transform
```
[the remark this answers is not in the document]
wanghan-iapcm: Same as the C++ edge path: this consumes a pt-backend edge `.pt2` (SeZM/DPA4) whose exported graph already bakes in exclusion (the pt backend still applies it at consume time), so the `return_mode="edges"` fast path correctly does not post-filter. Out of scope for this dpmodel-derived-backend refactor. See the [summary](https://github.com/deepmodeling/deepmd-kit/pull/5733#issuecomment-4913935637).

deepmd/dpmodel/utils/env_mat_stat.py
```diff
@@ -215,20 +281,34 @@ def iter(
                 self.descriptor.get_sel(),
                 mixed_types=self.descriptor.mixed_types(),
                 box=box,
+                # Model-level pair exclusion is a nlist-BUILD transform
+                # (decision #18/A4): fold it in here so the input stat matches
+                # the model forward, which feeds the descriptor a pre-excluded
+                # nlist. Excluded pairs then behave exactly like empty slots
+                # (env_mat 0, still counted) -- identical to descriptor-level
+                # exclude_types, replacing the previous accumulation-deselect.
+                pair_excl=pair_excl,
```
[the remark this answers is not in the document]
wanghan-iapcm: Valid, but **pre-existing** and deferred: stats depended on `pair_exclude_types` before this PR too (via the accumulation-deselect this PR replaced with a build-time nlist mask), and `get_hash()` never included it. It's also not cleanly fixable here — the model-level exclusion isn't visible at `get_hash` time (it arrives in sample data; `path / get_hash()` is computed before sampling). Tracking as a separate fix so this PR stays scoped to the regressions it introduced. See the [summary](https://github.com/deepmodeling/deepmd-kit/pull/5733#issuecomment-4913935637) ("Deferred — stat-cache hash").

source/api_cc/src/DeepPotPTExpt.cc
wanghan-iapcm: Addressed in [`ff28370`](https://github.com/deepmodeling/deepmd-kit/pull/5733/commits/ff283707999a2d7528c2a18e504ef0d86439c908). The dense multi-rank path (`run_model_with_comm`) now applies `applyPairExclusionNlist` — verified on a message-passing DPA3 with-comm model (MP≡SP + active-vs-baseline, clean-env). The edge-schema paths are intentionally left alone: they're a pt-backend lower (SeZM/DPA4) that still bakes exclusion into the exported graph, so re-applying in C++ would double-apply. Stale "backstop" docstrings reconciled. Full SP/MP × dense/graph matrix + the fail-fast reasoning for cell 6 in the [summary](https://github.com/deepmodeling/deepmd-kit/pull/5733#issuecomment-4913935637).

source/api_cc/include/commonPT.h
wanghan-iapcm: Done in [`ff28370`](https://github.com/deepmodeling/deepmd-kit/pull/5733/commits/ff283707999a2d7528c2a18e504ef0d86439c908), essentially your sketch: `pair_exclude_table_` is now a device `torch::Tensor` uploaded once in `init` (device known from `gpu_id`/`gpu_enabled`); `applyPairExclusion`/`applyPairExclusionNlist` take `const torch::Tensor&` and `index_select` it directly, with `!defined()` ⇒ identity replacing the empty-vector early-exit. No per-step CPU clone / H2D upload. `buildPairExcludeTable` stays a pure vector builder. Single-rank pairexcl gtests still pass after the change.

wanghan-iapcm:
@iProzd — thanks, this is the right framing and I agree: dropping the consume-time backstop made exclusion **fail-open**, which is why the leaks were surfacing one at a time. I've taken both remedies you offered — a fail-safe guard *and* an enumerated inventory — plus the doc fix.

### 1. Eager fail-safe guard ([`5ad1f05`](https://github.com/deepmodeling/deepmd-kit/pull/5733/commits/5ad1f051c5ccc248fe52345ea04b0cfe6faca389))

`forward_common_atomic{,_graph}` now carry `_assert_nlist_pair_excluded` / `_assert_graph_pair_excluded`: when `pair_excl` is set, a **real** neighbour (nlist ≥ 0) / **active** edge (edge_mask True) whose type pair is excluded is a leak → loud `AssertionError`. It's **numpy-eager only** (gated on `array_api_compat.is_numpy_array`), so it's a no-op under `torch.export` / jax `jit` (the check is data-dependent and can't be traced) and in compiled production — verified pt_expt `make_fx`/export still traces. This makes the whole dpmodel/jax **class** fail-*closed* by construction, not just the known instances.

The contract boundary you noted as untested is now tested: the two `test_graph_atomic_parity` seam tests were rewritten to assert the guard **rejects** a non-excluded input (previously they asserted the old fail-open pass-through).

### 2. Enumerated ingestion inventory

Added to the PR description — every entry point that builds a nlist/graph reaching the exclusion-owning lower, where each applies exclusion, and its coverage. It surfaced exactly the class of leak you predicted: the **jax-md `call_lower`** path built the lower nlist without exclusion. Fixed in [`5c5889c`](https://github.com/deepmodeling/deepmd-kit/pull/5733/commits/5c5889c212b4a1d135e6808a49e6cd98901138f2) (fold `apply_pair_exclusion_nlist` at the seam) with a regression test that runs without `jax_md` (via the `DenseNeighbor` path): an excluded (0,1) pair drops the only edge → energy 0 vs 0.04.

The other paths you cited are covered: the C++ with-comm dense (cell 3) and DeepEval graph/nlist paths apply the seam and now have tests (incl. a multi-rank DPA3 MP≡SP exclusion test on a message-passing model); the DeepEval/C++ **edge** (`return_mode="edges"` / `run_model_edges*`) is a **pt-backend** lower (SeZM/DPA4) whose exported graph still bakes exclusion in at consume time, so it is intentionally out of scope for this dpmodel-derived-backend refactor (re-applying would double-apply); the `env_mat_stat` **hash key** is pre-existing (stats already depended on exclusion before this PR; the key never included it) and tracked as a separate fix.

### 3. Doc fix

Corrected the PR description: the `forward_common_atomic*` / C++ seam is the **single owner**, not an "idempotent backstop" — the wording now matches HEAD's "NOT applied here" and the fail-safe guard, so the blast radius reads right.

Net: exclusion is fail-*closed* for the dpmodel/jax paths (guard) and complete-by-inventory for the compiled/C++ paths (tests), with the one genuine gap the inventory found (jax-md) now closed.

OutisLi (dismissed):
Retracted — posted prematurely and partly out of scope. Please disregard. Apologies for the noise.

OutisLi (approved):

deepmd/tf2/make_model.py
```diff
@@ -124,6 +132,13 @@ def model_call_from_call_lower(
         charge_spin=charge_spin,
         neighbor_list=neighbor_list,
     )
+    if pair_excl is not None:
```
iProzd: **[blocking] The new TF2 build-seam transform is not reached by the live model or trainer.**

`pair_excl` defaults to `None` here, but `deepmd/tf2/model/dp_model.py::call_common` invokes this helper without passing `model.atomic_model.pair_excl`; the compiled trainer bypasses this helper and calls `prepare_lower_inputs`, whose signature and builders also have no `pair_excl`. Since the inherited dpmodel lower no longer reapplies exclusion and its fail-safe is NumPy-only, TF2 tensors silently retain excluded pairs in direct inference and both eager/compiled training. The current IO test only exercises `tf2/utils/serialization.py`, which does pass this argument, so it cannot catch the gap. Could we apply the canonical transform inside `prepare_lower_inputs(..., pair_excl=...)`, thread it from both live `call_common` and the trainer, and add a live TF2/XLA non-vacuity test?

deepmd/jax/jax2tf/serialization.py
```diff
@@ -176,6 +176,12 @@ def call(
                     fparam=fparam,
                     aparam=aparam,
                     do_atomic_virial=do_atomic_virial,
+                    # exclusion is a nlist-BUILD transform (decision #18/A4);
+                    # the traced lower consumes a pre-excluded nlist. Guard
+                    # atomic_model too: test doubles (DummyModel) lack it.
+                    pair_excl=getattr(
```
iProzd, line 182: **[blocking] The JAX C++/LAMMPS `InputNlist` path bypasses this pre-exclusion.**

This applies `pair_excl` only in the exported high-level `call_*` wrappers, while `DeepPotJAX::compute(..., InputNlist, ...)` invokes the SavedModel `call_lower_*` functions directly (`source/api_cc/src/DeepPotJAX.cc:830-868`). Those wrappers currently only format the nlist before entering the serialized JAX lower, which no longer reapplies model-level exclusion, and the NumPy-only guard is skipped during tracing. Consequently normal JAX evaluation is correct while the same model can silently include excluded pairs under JAX LAMMPS inference. Could we apply the canonical transform in the exported `call_lower_*` wrappers or at the C++ ingestion seam, and add a non-vacuous parity test between high-level `call` and raw `call_lower`/C++ `InputNlist`?

iProzd (changes requested):
The NumPy fail-safe guards, JAX-MD fix, C++ graph/dense with-comm coverage, one-time device-table upload, and associated non-vacuity tests are all verified at HEAD. However, the ingestion inventory is still incomplete: live/compiled TF2 paths do not pass `pair_excl`, and `DeepPotJAX` invokes the exported `call_lower_*` functions directly for `InputNlist` without applying the new build-time transform. Requesting changes until these two production paths are wired and covered. The branch also currently conflicts with master, so the pair-exclusion changes need to be retained when resolving the NeighborGraph-angle, HLO-property, and IO-test conflicts.

deepmd/tf2/make_model.py
```diff
@@ -124,6 +132,13 @@ def model_call_from_call_lower(
...
+        # Model-level pair exclusion is a nlist-BUILD transform (decision
+        # #18/A4): fold it into the freshly built nlist here (already an
+        # ndtensorflow array, so no wrap/unwrap) via the canonical dpmodel
+        # helper, before the lower consumes it. Identity when nothing is
+        # excluded.
+        nlist = apply_pair_exclusion_nlist(nlist, extended_atype, pair_excl)
```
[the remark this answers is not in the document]
wanghan-iapcm: Fixed. `pair_excl` is now threaded into `prepare_lower_inputs` as the single owner, and the trainer passes `model.atomic_model.pair_excl` at the compiled-training prepare seam:

- [9bdda5d](https://github.com/deepmodeling/deepmd-kit/commit/9bdda5d6345b40a206c81f12d7cde6de661eee3f) — `prepare_lower_inputs` folds exclusion into the built nlist (the redundant application in `model_call_from_call_lower` is removed), and `trainer._make_compiled_prepare_lower_batch` passes `pair_excl`. Includes the requested non-vacuity regression test (`test_prepare_lower_inputs_folds_in_pair_exclusion`): a type-[0,1] pair within cutoff yields a real neighbour without exclusion and is erased to −1 with it.
- [4856eed](https://github.com/deepmodeling/deepmd-kit/commit/4856eed5dffc7a9d1d076d71ad138989543cb6bf) — follow-up: `tf2/make_model.py` moved to `tf2/model/make_model.py` (matching the other backends' layout), and the exclusion is applied at BUILD time inside the builder dispatch (custom strategies receive `pair_excl` via `build()`, the native inline builder applies `apply_pair_exclusion_nlist` directly), mirroring the dpmodel/pt_expt convention instead of a post-hoc application.

Verified on a TF box (`DP_TEST_TF2_ONLY=1`): `source/tests/tf2/test_training.py` 30 passed; `source/tests/consistent/model/test_ener.py` 120 passed including the tf2 pair/atom-exclusion cases (which are no longer skipped as of [c5b5345](https://github.com/deepmodeling/deepmd-kit/commit/c5b53451a391547cdc798c2fd1a5104b4a9f164b)).

deepmd/dpmodel/model/make_model.py
```diff
@@ -122,7 +131,7 @@ def model_call_from_call_lower(
     del coord, box, fparam, aparam
     builder = neighbor_list if neighbor_list is not None else DefaultNeighborList()
     extended_coord, extended_atype, nlist, mapping = builder.build(
-        cc, atype, bb, rcut, sel
+        cc, atype, bb, rcut, sel, pair_excl=pair_excl
```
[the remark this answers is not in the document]
wanghan-iapcm: Fixed, with a variation on the suggestion. Applying `apply_pair_exclusion_nlist` centrally on the returned quartet would put the transform outside the builder, inconsistent with the graph path where `build_neighbor_graph(..., pair_excl=)` owns exclusion — so the builder keeps ownership, and the keyword is passed **only when set**:

- [8988c05](https://github.com/deepmodeling/deepmd-kit/commit/8988c05594273052655d6f473208cac0a7fb5bad) / [a947793](https://github.com/deepmodeling/deepmd-kit/commit/a947793888cd88f879b1d02195cdc40ddbb2fcd1) — `builder.build(cc, atype, bb, rcut, sel, **excl_kwargs)` with `excl_kwargs = {} if pair_excl is None else {"pair_excl": pair_excl}`.

Behavior split:
- no exclusion (every existing custom strategy): the old published `build(coord, atype, box, rcut, sel, return_mode=...)` signature is called unchanged — no `TypeError`;
- non-empty exclusion + legacy builder without the keyword: loud `TypeError` instead of silently including excluded pairs (fail-closed, which is the failure mode we want for this feature);
- updated builders: exclusion applied at build, same as the graph path.

wanghan-iapcm:
@iProzd All three items are addressed at HEAD (4856eed5d).

**1. TF2 live/compiled paths now pass `pair_excl`.**
- live/eager upper (`dp_model.call_common`): [6535cbd](https://github.com/deepmodeling/deepmd-kit/commit/6535cbdb827b589ac70b0bf183dc3c57c7206fc7) (+ guard for test doubles, [79733df](https://github.com/deepmodeling/deepmd-kit/commit/79733df9471103a6c5f4933f845080bdd4665b19));
- compiled training (`trainer._make_compiled_prepare_lower_batch` → `prepare_lower_inputs`): [9bdda5d](https://github.com/deepmodeling/deepmd-kit/commit/9bdda5d6345b40a206c81f12d7cde6de661eee3f), with a non-vacuity regression test; `prepare_lower_inputs` is now the single owner and the builder applies exclusion at BUILD time ([4856eed](https://github.com/deepmodeling/deepmd-kit/commit/4856eed5dffc7a9d1d076d71ad138989543cb6bf), which also moves `tf2/make_model.py` under `tf2/model/` to match the other backends);
- regression guard: the consistency `test_ener` no longer skips tf2 for pair/atom exclusion ([c5b5345](https://github.com/deepmodeling/deepmd-kit/commit/c5b53451a391547cdc798c2fd1a5104b4a9f164b)) — this skip (added together with the eager tf2 backend) is what had been hiding the gap. Verified on a TF box (`DP_TEST_TF2_ONLY=1`): `test_training.py` 30 passed, `test_ener.py` 120 passed incl. the tf2 exclusion cases.

**2. DeepPotJAX InputNlist path now applies the build-time transform.**
- both jax2tf and tf2 SavedModels export a flat `get_pair_exclude_types` getter; `DeepPotJAX::init` builds the keep table once and the LAMMPS nlist is filtered before `call_lower_*` ([33318f4](https://github.com/deepmodeling/deepmd-kit/commit/33318f4c63cfe4ed8530de6e9cca6db33ff5d87b)). Models exported before the getter baked exclusion into the graph, so a missing getter ⇒ identity is correct for them (no fail-open for old artifacts).
- the filter is a torch-free twin of `applyPairExclusionNlist` (`applyPairExcludeNlistVec` in `common.h`, shared table builder hoisted there), unit-tested in every C++ build ([4846ec7](https://github.com/deepmodeling/deepmd-kit/commit/4846ec7837df5e08f1a12e3f227dada4485d6a24));
- end-to-end gtest on the InputNlist route for both SavedModel flavours (excluded ≠ baseline, and InputNlist path == coord-only path), with fixtures derived from `deeppot_sea` at identical weights ([4acf4de](https://github.com/deepmodeling/deepmd-kit/commit/4acf4de0879dec7fcd6d1e58d8699eecc17f14b4)); it runtime-skips when the JAX backend isn't built, so the assertions run under the TF-enabled CI build.

**3. Conflicts with master are resolved** ([15858b6](https://github.com/deepmodeling/deepmd-kit/commit/15858b6a26fecc96f0492fd10f07b55b7f0b8b51)): the pair-exclusion changes are retained; the NeighborGraph-angle `__all__`, `hlo.py` property-model attributes, and the io-test classes (`TestDeepPotPairExclude` + `TestDeepProperty` as two separate classes) are merged from both sides.

Also addressed from njzjz-bot's latest round (see inline replies): the `builder.build()` compatibility issue is resolved by passing `pair_excl` only when set, so custom `NeighborList` strategies keep the published signature, while a legacy builder combined with a non-empty exclusion fails loudly instead of silently including pairs.

deepmd/dpmodel/model/make_model.py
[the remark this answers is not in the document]
wanghan-iapcm: Follow-up — the design settled differently from my previous reply: the conditional-kwargs indirection is dropped in [a33ff79](https://github.com/deepmodeling/deepmd-kit/commit/a33ff7905ffa1747a5d43c56ea4e9b13b25a5f37), and `pair_excl=pair_excl` is now passed to `build()` unconditionally and uniformly (dpmodel/tf2 `make_model`, pt_expt `deep_eval`, and the graph builders all use the same direct form).

Rationale: `pair_excl` is part of the `NeighborList.build()` contract — the base-class signature carries it with a `None` default. A custom strategy written against the pre-PR signature therefore fails loudly with `TypeError` rather than silently including excluded pairs; for a feature whose dangerous failure mode is silent inclusion, fail-loud is the behavior we want, and updating a custom `build()` to accept (or ignore via `**kwargs`) the new keyword is a one-line change on the implementer's side.

njzjz (approved):
LGTM

deepmd/tf2/make_model.py
wanghan-iapcm: Fixed at HEAD (4856eed5d). `pair_excl` is now threaded into `prepare_lower_inputs` as the single owner, so both live and compiled paths reach the transform:
- live/eager upper (`dp_model.call_common`): [6535cbd](https://github.com/deepmodeling/deepmd-kit/commit/6535cbdb827b589ac70b0bf183dc3c57c7206fc7) (+ test-double guard [79733df](https://github.com/deepmodeling/deepmd-kit/commit/79733df9471103a6c5f4933f845080bdd4665b19));
- compiled training (`trainer._make_compiled_prepare_lower_batch`): [9bdda5d](https://github.com/deepmodeling/deepmd-kit/commit/9bdda5d6345b40a206c81f12d7cde6de661eee3f), with a non-vacuity regression test;
- the consistency `test_ener` no longer skips tf2 for pair/atom exclusion ([c5b5345](https://github.com/deepmodeling/deepmd-kit/commit/c5b53451a391547cdc798c2fd1a5104b4a9f164b)) — that skip was hiding the gap. `4856eed` also moves `tf2/make_model.py` under `tf2/model/` to match the other backends. See the [top-level summary](https://github.com/deepmodeling/deepmd-kit/pull/5733#issuecomment-4941518662) for the full walk-through.

deepmd/jax/jax2tf/serialization.py
wanghan-iapcm: Fixed at HEAD (4856eed5d). The `DeepPotJAX::compute(..., InputNlist ...)` route now applies the build-time transform:
- both jax2tf and tf2 SavedModels export a flat `get_pair_exclude_types` getter; `DeepPotJAX::init` builds the keep table once and the LAMMPS nlist is filtered before `call_lower_*` ([33318f4](https://github.com/deepmodeling/deepmd-kit/commit/33318f4c63cfe4ed8530de6e9cca6db33ff5d87b)). A missing getter ⇒ identity, which is correct for models exported before the getter (they baked exclusion into the graph), so no fail-open for old artifacts;
- the filter is a torch-free twin of `applyPairExclusionNlist` (`applyPairExcludeNlistVec` in `common.h`), unit-tested in every C++ build ([4846ec7](https://github.com/deepmodeling/deepmd-kit/commit/4846ec7837df5e08f1a12e3f227dada4485d6a24));
- end-to-end gtest on the InputNlist route for both SavedModel flavours (excluded ≠ baseline, InputNlist path == coord-only path) ([4acf4de](https://github.com/deepmodeling/deepmd-kit/commit/4acf4de0879dec7fcd6d1e58d8699eecc17f14b4)); it runtime-skips when the JAX backend is not built, so assertions run under the TF-enabled CI build. See the [top-level summary](https://github.com/deepmodeling/deepmd-kit/pull/5733#issuecomment-4941518662).

iProzd (approved):
LGTM.

PR #5736, opened 2026-07-05
fix(jax): avoid device lookup during type embedding tracing

njzjz:
## Summary

Fix a JAX/Flax NNX tracing failure in `TypeEmbedNet.call()` by avoiding a hard requirement that traced values expose a readable `.device` attribute.

## Background

JAX training can call `TypeEmbedNet.call()` from inside `nnx.jit` / `nnx.grad`. In that context the type-embedding weights are represented by NNX/JAX traced values such as `DynamicJaxprTracer`. The previous code passed

```python
device=array_api_compat.device(sample_array)
```

to `xp.eye`, and did the same for the padding `xp.zeros`. For an NNX traced parameter, `array_api_compat.device(...)` eventually tries to read `.device`; the tracer does not provide that attribute during tracing, so the training step fails with an error like:

```text
AttributeError: DynamicJaxprTracer has no attribute device
```

or, through JAX core wrapping:

```text
AttributeError: 'ShapedArray' object has no attribute 'device'
```

## Change

Add a small local helper in `deepmd/dpmodel/utils/type_embed.py` that returns `array_api_compat.device(array)` when available, and falls back to `None` when the backend value has no readable device during tracing. Passing `device=None` lets JAX create the constants on its default traced device, while preserving the existing explicit-device behavior for backends that expose it normally.

This keeps the change limited to the type-embedding constants that triggered the crash, rather than changing global array API device handling.

## Tests

Added `source/tests/jax/test_type_embed.py` to cover both:

- `TypeEmbedNet.call()` under `nnx.jit`
- `TypeEmbedNet.call()` under `nnx.jit` + `nnx.grad`

The old implementation fails this test during tracing before producing an output.

Validation run locally:

```bash
pytest source/tests/jax/test_type_embed.py -v
ruff check .
ruff format .
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for model execution on array backends that do not support device detection in all cases.
  * Reduced failures when padding or initializing arrays during type embedding operations.

* **Tests**
  * Added JAX coverage to verify the type embedding model runs under JIT, supports gradient tracing, returns the expected output shape, and produces valid numeric results.

### DIFF
--- a/deepmd/dpmodel/utils/type_embed.py
+++ b/deepmd/dpmodel/utils/type_embed.py
@@ -24,6 +24,13 @@ from deepmd.utils.version import (
 )
 
 
+def _array_device_or_none(array: Array) -> Any:
+    try:
+        return array_api_compat.device(array)
+    except AttributeError:
+        return None
+
+
 class TypeEmbedNet(NativeOP):
     r"""Type embedding network.
 
@@ -104,7 +111,7 @@ class TypeEmbedNet(NativeOP):
                 xp.eye(
                     self.ntypes,
                     dtype=sample_array.dtype,
-                    device=array_api_compat.device(sample_array),
+                    device=_array_device_or_none(sample_array),
                 )
             )
         else:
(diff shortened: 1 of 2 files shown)


wanghan-iapcm:
### Code review

No issues found.

wanghan-iapcm (approved):

PR #5737, opened 2026-07-05
fix(jax): support charge-spin savedmodel export

njzjz:
## Summary
- export charge_spin-aware JAX SavedModel signatures only for models that need charge/spin conditioning
- add Python and C++ JAX runtime fallback to stored default_chg_spin, while keeping old SavedModel signatures compatible
- allow C++ header wrapper to broadcast one-frame charge_spin input safely and add focused jax2tf regressions

## Tests
- ruff format .
- ruff check .
- pytest source/jax2tf_tests/test_make_model.py source/jax2tf_tests/test_serialization.py source/jax2tf_tests/test_tfmodel.py -q
- c++ -std=c++17 -Isource/api_c/include -fsyntax-only -x c++ ...
- c++ -std=c++17 -fsyntax-only ... source/api_cc/src/DeepPotJAX.cc

## Notes
- `cmake --build source/build --target deepmd_cc -j2` still fails before compiling this change because local CMake TensorFlow detection stops at `cmake/Findtensorflow.cmake:234 (file): file unknown error`.
- Review requested from @anyangml for charge_spin C++ runtime context and @iProzd for charge_spin/DPA3 behavior.

## Summary by CodeRabbit

* **New Features**
  * Added optional charge/spin conditioning across JAX evaluation, JAX→TF SavedModel export, TensorFlow model wrapper, and C++/C-API interfaces.
  * SavedModels now expose charge/spin support metadata (supported flag, dimension, default values).
  * Charge/spin inputs can be provided per frame or broadcast from a single vector.
* **Bug Fixes**
  * When charge/spin conditioning isn’t supported, provided inputs are ignored with a warning.
  * Improved handling of empty/unsupported charge/spin and clearer errors for invalid shapes.
* **Tests**
  * Added/extended tests covering forwarding, SavedModel export/import, defaults, broadcasting, and validation across JAX/TF/C++.


njzjz:
Addressed the review comments in c3ea9ba54:\n\n- reject explicit charge_spin when the JAX model does not support charge/spin conditioning in both Python inference and the C++ JAX backend;\n- remove the stale C++ header comment and unused validate_charge_spin overload;\n- update charge_spin docs/type hints;\n- add tests for unsupported charge_spin, invalid TFModelWrapper charge_spin shapes, and the charge-spin SavedModel export/call path.

deepmd/jax/infer/deep_eval.py
```diff
@@ -394,17 +400,32 @@ def _eval_model(
             aparam_input = aparam.reshape(nframes, natoms, self.get_dim_aparam())
         else:
             aparam_input = None
+        if charge_spin is not None and not self.has_chg_spin_ebd():
+            raise ValueError(
+                "charge_spin was provided, but this model does not support "
```
anyangml, line 418: I don't think this will affect the correctness of the prediction, so may be a warning is enough?

njzjz: I am not sure. Maybe you are right.

source/api_cc/src/DeepPotJAX.cc
```diff
@@ -459,6 +460,53 @@ inline TFE_TensorHandle* add_input(TFE_Op* op,
   return handle;
 }
 
+inline std::vector<double> make_charge_spin_input(
```
wanghan-iapcm, line 463: Non-blocking (test coverage): `make_charge_spin_input` here and the broadcast/fallback branch in `validate_charge_spin` (`source/api_c/include/deepmd.hpp`) add several reachable paths — `dchgspin==0` with a non-empty input (throw), value-from-`charge_spin` vs value-from-`default_chg_spin` vs throw, and `size==nframes*dim` (exact) vs `size==dim` (single-frame tile) vs throw — none of which are exercised by a C++ unit test (the PR only adds Python `source/jax2tf_tests/`). `validate_charge_spin` is a pure vector helper and can be unit-tested without a model fixture; a small test over the tile and default-fallback branches would cover them.

deepmd/jax/infer/deep_eval.py
```diff
@@ -394,17 +400,32 @@ def _eval_model(
...
         do_atomic_virial = any(
             x.category == OutputVariableCategory.DERV_C_REDU for x in request_defs
         )
+        model_kwargs = {
+            "box": to_jax_array(box_input),
+            "fparam": to_jax_array(fparam_input),
+            "aparam": to_jax_array(aparam_input),
+            "do_atomic_virial": do_atomic_virial,
+        }
+        if self.has_chg_spin_ebd():
```
wanghan-iapcm, line 438: Non-blocking (both-branches coverage): this `has_chg_spin_ebd()==True` path (and L409-410, which forwards `charge_spin` to the model, including the `True` + `charge_spin is None` sub-case) is never exercised. The only new `_eval_model` test uses a model whose `has_chg_spin_ebd()` returns `False`, so it covers just the `False` + provided -> raise path. A test that drives `_eval_model` with a charge-spin model would close the positive branch.

source/api_c/include/deepmd.hpp
```diff
@@ -1232,8 +1248,9 @@ class DeepPot : public DeepBaseModel {
     const VALUETYPE* aparam__ = !aparam_.empty() ? &aparam_[0] : nullptr;
     // charge_spin routes to the version-3 C API; nullptr keeps version-2 so
     // non-charge_spin models still work against an older libdeepmd_c.
-    const VALUETYPE* charge_spin__ =
-        validate_charge_spin(charge_spin, dchgspin, nframes);
+    std::vector<VALUETYPE> charge_spin_tiled_;
+    const VALUETYPE* charge_spin__ = validate_charge_spin(
```
wanghan-iapcm, line 1255: Non-blocking (doc accuracy): `validate_charge_spin` now accepts both `nframes*dim_chg_spin` and a single `dim_chg_spin`-sized vector (tiled to all frames), but the public `@param[in] charge_spin` doxygen on the `compute()` overloads (e.g. ~L1212-1215, and the `DeepPotModelDevi` copies) still lists only `nframes x dim_chg_spin` and attaches the "all frames same" clause to it — which actually describes the single-`dim_chg_spin` case that is left undocumented. Consider mirroring the sibling `fparam` doxygen just above, which correctly lists both sizes.

wanghan-iapcm (commented):
Three non-blocking observations from a review pass (each inline below). None blocks merge.

source/api_cc/src/DeepPotJAX.cc
```diff
@@ -459,6 +460,43 @@ inline TFE_TensorHandle* add_input(TFE_Op* op,
...
+    }
+    return result;
+  }
+  throw deepmd::deepmd_exception(
+      "charge_spin has " + std::to_string(source.size()) +
+      " values but the model expects dim_chg_spin=" + std::to_string(dchgspin) +
+      " (per frame) or " + std::to_string(expected) + " (for " +
+      std::to_string(nframes) + " frames).");
+}
+
```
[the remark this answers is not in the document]
njzjz: Addressed in c3ea9ba. Sorry, I missed the inline reply when resolving this thread.

deepmd/jax/infer/deep_eval.py
```diff
@@ -394,17 +396,27 @@ def _eval_model(
...
+            model_kwargs["charge_spin"] = to_jax_array(charge_spin_input)
         batch_output = model(
             to_jax_array(coord_input),
             to_jax_array(type_input),
-            box=to_jax_array(box_input),
-            fparam=to_jax_array(fparam_input),
-            aparam=to_jax_array(aparam_input),
-            do_atomic_virial=do_atomic_virial,
+            **model_kwargs,
         )
```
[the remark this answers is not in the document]
njzjz: Addressed in c3ea9ba. Sorry, I missed the inline reply when resolving this thread.

deepmd/jax/infer/deep_eval.py
njzjz: Updated in 22263c257 to follow this suggestion: explicit charge_spin on a model without charge-spin embedding now emits a warning and is ignored, instead of raising.

source/api_cc/src/DeepPotJAX.cc
njzjz: Addressed in 22263c257 by adding C++ unit coverage for validate_charge_spin exact-size input, single-frame tiling, empty/default path, unsupported-model rejection, and invalid-size rejection.

source/api_c/include/deepmd.hpp
njzjz: Addressed in 22263c257 by updating the charge_spin doxygen to document both accepted shapes: nframes x dim_chg_spin and dim_chg_spin broadcast to all frames.

deepmd/jax/infer/deep_eval.py
njzjz: Addressed in 22263c257 by adding DeepEval tests for the charge-spin-supported path with an explicit charge_spin value, plus the supported path with charge_spin=None.

anyangml (approved):

deepmd/jax/infer/deep_eval.py
```diff
@@ -394,17 +401,34 @@ def _eval_model(
             aparam_input = aparam.reshape(nframes, natoms, self.get_dim_aparam())
         else:
             aparam_input = None
+        if charge_spin is not None and not self.has_chg_spin_ebd():
```
iProzd, line 416: Cross-binding consistency: for the same misuse (passing `charge_spin` to a model without charge/spin support), the Python JAX path here warns and ignores, while the C++ JAX backend (`make_charge_spin_input` in `DeepPotJAX.cc`, and `validate_charge_spin` in `deepmd.hpp`) throws. This split came from softening only the Python side to a warning per anyangml's suggestion, leaving C++ on the original throw. anyangml's point (charge_spin is a genuine no-op on an unsupported model) argues for warn-and-ignore, but then C++ should match; conversely if the compiled path should stay strict, Python should throw too. Which contract is intended across bindings? Right now the same script switching from Python to C++ inference would change behavior on this input.

source/api_cc/src/DeepPotJAX.cc
wanghan-iapcm, line 463: **Non-blocking (test coverage):** the new C++ runtime `charge_spin` path for the JAX backend is untested. `make_charge_spin_input` (default-value fallback, single-frame broadcast/tiling, size-mismatch throw) and the new `computew` `charge_spin` overloads have no C++ coverage — the added `test_deepmd_exception.cc` only exercises the header helper `validate_charge_spin`, which does not handle `default_chg_spin`. There is also no `.savedmodel` charge_spin fixture (`source/tests/infer/gen_chg_spin.py` exports only `.pt2`/`.pth`).

This mirrors the coverage that #5509 added for the pt/pt_expt backends at reviewers' request (`source/api_cc/tests/test_deeppot_chg_spin_pt.cc`, `test_deeppot_chg_spin_ptexpt.cc`). The Python-level paths are tested here, so this is a coverage gap rather than a known bug — a JAX `test_deeppot_chg_spin_jax.cc` + savedmodel fixture would close it.

wanghan-iapcm: To connect this with the earlier thread: commit 22263c257 added coverage for `validate_charge_spin` (the `deepmd.hpp` header helper), but `make_charge_spin_input` here in `DeepPotJAX.cc` is a distinct function — it holds the `default_chg_spin` fallback that `validate_charge_spin` never sees. It is reachable only through a full `DeepPotJAX::compute()` run, which needs a `.savedmodel` charge_spin fixture (still absent from `source/tests/infer/gen_chg_spin.py`). So the new unit tests, thorough as they are, cannot reach this path or the `computew` charge_spin overloads. Not a blocker — just flagging that this specific gap is still open.

wanghan-iapcm (approved):

iProzd (approved):

PR #5747, opened 2026-07-07
feat(dpmodel): NeighborGraph support for dpa1 tebd_input_mode="strip"

wanghan-iapcm:
## Summary

Adds NeighborGraph (graph-native lower) support for the dpa1 descriptor with `tebd_input_mode="strip"`, closing the last descriptor-level gap that forced strip-mode models (and `se_atten_v2`, which is strip-by-construction) onto the legacy dense path.

The dense strip branch factorizes the per-neighbor feature as `gg = gg_s*gg_t + gg_s` — a radial-only geometric embedding times a type-pair strip embedding (optionally switch-smoothed). Because this has **no neighbor-axis coupling**, it maps to the graph path edge-for-edge. The change is:

- **Kernel** (`DescrptBlockSeAtten`): a new per-edge helper `_graph_edge_gg_strip` (op-for-op mirror of the dense strip branch, including the `center*ntypes_pad + nei` nei-fastest two-side table layout and `int64` gather indices), selected by a `concat`/`strip` branch in `call_graph`.
- **Routing** (`DescrptDPA1.uses_graph_lower`): admits `strip`, while keeping compressed descriptors and `exclude_types` on the dense path (they have no graph kernel here). `se_atten_v2` inherits this and becomes graph-eligible for free.
- **pt_expt**: the two graph `make_fx` export tests are parametrized over `tebd_input_mode` to prove the strip kernel is fx-traceable.

No new op, no attention/`segment_sum` change, no C++/serialization change.

## Scope

This PR is deliberately **independent of #5733** (graph `exclude_types`). It does **not** change `exclude_types` eligibility — the `uses_graph_lower` `and not exclude_types` gate and the `call_graph` `exclude_types` raise are both kept. When both land, whichever merges second resolves a small (2–3 line) mechanical conflict at `uses_graph_lower` / the `call_graph` guard.

## Test plan

- Block-level graph-vs-dense strip parity at `rtol=atol=1e-12` over `type_one_side × smooth` (attn=0) and `type_one_side` (attn=2, non-smooth).
- Descriptor-level routed-`call` vs `_call_dense` parity over `type_one_side × smooth × attn_layer` (incl. attn=2 + smooth=True, bit-exact via the `static_nnei` adapter), plus a negative-contract gate test (compressed → dense, strip+`exclude_types` → dense).
- `se_atten_v2` eligibility + graph-vs-dense parity (replaces the obsolete "strip stays dense" test).
- pt_expt strip `make_fx` export; cross-backend consistency strip cases now route pt_expt through the graph adapter.

Validated on **CPU** and on **GPU (Tesla T4, cuda:0)**: pt_expt dpa1 50 passed, consistency strip 22 passed + `se_atten_v2` 110 passed, dpmodel strip suites 46 passed. No tolerances relaxed, no tests skipped.

## Known limitations

- **Compression** stays on the dense path by design (strip-only tabulation has no graph kernel); the gate excludes `self.compress`.
- **`exclude_types`** stays dense (out of scope — owned by #5733).
- **jax** graph lower remains energy-only (analytical force on the graph route is a separate follow-up).
- The graph path's `segment_sum`→`index_add` is atomic/non-deterministic on CUDA (1–2 fp64 ULP), inherent to atomic scatter; GPU parity validated within tolerance.
- Pre-existing (not introduced here): a softmax `RuntimeWarning` on the shared attention path (max over fully-masked segments), also present on the concat path.

## Summary

* **New Features**
  * Added graph-based execution support for the additional stripped type-embedding mode.

* **Bug Fixes**
  * Updated graph routing eligibility: compressed descriptors and excluded-type configurations now reliably fall back to dense execution; graph routing is disabled when compression is enabled.

* **Tests**
  * Added bit-exact parity tests between graph and dense paths for the new stripped mode (including routing eligibility checks).
  * Expanded FX graph export/trace coverage for both embedding modes.
  * Adjusted neighbor-list fallback validation with model-specific tolerance handling and added a new smooth variant.

### DIFF
--- a/deepmd/dpmodel/descriptor/dpa1.py
+++ b/deepmd/dpmodel/descriptor/dpa1.py
@@ -1749,7 +1749,7 @@ class DescrptBlockSeAtten(NativeOP, DescriptorBlock):
         Notes
         -----
         Known limitations:
-        - ``tebd_input_mode == "concat"`` only (strip mode lands later);
+        - ``tebd_input_mode`` in {"concat", "strip"}; compressed descriptors stay dense;
         - ``exclude_types`` is not yet supported and raises (lands in a later PR).
         """
         from deepmd.dpmodel.utils.neighbor_graph import (
@@ -1757,9 +1757,9 @@ class DescrptBlockSeAtten(NativeOP, DescriptorBlock):
             segment_sum,
         )
 
-        if self.tebd_input_mode not in ["concat"]:
+        if self.tebd_input_mode not in ["concat", "strip"]:
             raise NotImplementedError(
-                "graph path supports tebd_input_mode='concat' only (NeighborGraph PR-A)"
+                f"graph path does not support tebd_input_mode={self.tebd_input_mode!r}"
             )
         if self.exclude_types:
             raise NotImplementedError(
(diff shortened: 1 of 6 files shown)


deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -1749,17 +1755,17 @@ def call_graph(
...
         from deepmd.dpmodel.utils.neighbor_graph import (
             edge_env_mat,
             segment_sum,
         )
 
-        if self.tebd_input_mode not in ["concat"]:
+        if self.tebd_input_mode not in ["concat", "strip"]:
             raise NotImplementedError(
-                "graph path supports tebd_input_mode='concat' only (NeighborGraph PR-A)"
+                f"graph path does not support tebd_input_mode={self.tebd_input_mode!r}"
```
OutisLi, line 1768: if tebd_input_mode not in ["concat", "strip"] it will not pass argcheck?

OutisLi (approved):

deepmd/dpmodel/descriptor/dpa1.py
wanghan-iapcm: Good catch — confirmed no, it will pass argcheck. `tebd_input_mode` is declared as a plain `str` Argument (no `choices=`) in `argcheck.py`, so any string value is accepted at input-validation time; the docstrings there just prose-describe 'concat'/'strip' as the supported modes without enforcing it. So this `NotImplementedError` guard is genuinely reachable, not dead code — it's needed because an invalid `tebd_input_mode` can reach `call_graph` at runtime. For reference, the dense `call()` path (`dpa1.py` around line 1575, `if ... in ["concat"]: ... elif ... in ["strip"]: ...` with no `else`) has no equivalent guard and would fail with an unclear `UnboundLocalError` on `gg` for the same bad input — that's a pre-existing gap outside this PR's scope, so I've left it as-is here rather than expanding scope. Tightening `argcheck.py` with an explicit `choices=["concat", "strip"]` would be the real fix for this whole class of issue; happy to file that separately if useful.

PR #5748, opened 2026-07-07
feat(jax): support DPA4 descriptor

njzjz:
## Summary

- Add and register the JAX DPA4/SeZM descriptor adapter and its JAX/NNX submodule mappings.
- Preserve trainable and frozen descriptor parameter state, and reject DPA4 options that are not supported by the JAX descriptor backend.
- Enable JAX coverage in the existing DPA4 descriptor consistency tests and add focused descriptor parameter-state tests.
- Keep this pull request descriptor-only; fitting, model factory, trainer, checkpoint, and model-conversion changes are excluded.

## Validation

- ruff format .
- ruff check .
- python -m py_compile deepmd/jax/descriptor/dpa4.py source/tests/consistent/descriptor/test_dpa4.py source/tests/jax/test_dpa4.py
- pytest source/tests/jax/test_dpa4.py -v — 2 passed

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh


njzjz:
Pushed follow-up commit `590823bb2` for the JAX CI/review feedback.\n\nChanges:\n- Avoid direct `nnx.List` usage when the installed Flax NNX does not provide it.\n- Remove the no-op `SeZMEnergyFittingNet.__setattr__` override.\n- Normalize explicit `null` `descriptor` / `fitting_net` blocks before setting DPA4 defaults.\n- Reinsert missing zero-size state leaves during JAX checkpoint restore, matching the save-side filtering.\n- Reverted the extra `source/tests/consistent` churn from this follow-up, per request.\n\nLocal validation:\n- JAX smoke for DPA4 descriptor construction, fitting construction, DPA4 model construction with `fitting_net: null`, and zero-size state reinsertion passed.\n- `ruff format . && ruff check . && git diff --check` passed.\n\nCI is rerunning.

deepmd/jax/descriptor/dpa4.py
```diff
@@ -0,0 +1,283 @@
...
+    pass
+
+
+register_dpmodel_mapping(
+    WignerDCalculatorDP,
+    lambda v: WignerDCalculator(v.lmax, eps=v.eps, precision=v.precision),
+)
+
+
+_TRAINABLE_ATTRS: dict[str, tuple[str, ...]] = {
```
wanghan-iapcm, line 107: **Non-blocking (correctness — silently frozen params):** promotion here is strictly allowlist-gated — `_promote_trainable_tree` promotes only the attributes named in `_TRAINABLE_ATTRS` / `_TRAINABLE_LIST_ATTRS`; anything else stays an `ArrayAPIVariable` (not an `nnx.Param`), so `nnx.Optimizer(..., wrt=nnx.Param)` / `nnx.grad` never update it — it is silently frozen during training. A few trainable dpmodel params appear to be missing from the maps:

- `FrameContract.weight` / `FrameExpand.weight` — trainable in `deepmd/dpmodel/descriptor/dpa4_nn/grid_net.py` (pt registers them as `nn.Parameter`), reachable when `node_wise_so3` / `message_node_so3` instantiate a `mode=="cross"` grid net. Neither class is a key in the maps.
- `SeZMInteractionBlock.adam_ffn_layer_scales` — a trainable LayerScale list when `layer_scale=True` (`dpa4_nn/block.py`), also listed in `dpa4_nn/lora.py`'s unfreeze registry. The maps already cover the analogous `SO2Convolution.adam_so2_layer_scales`, but not this one.

These are opt-in (non-default) features, so a stock DPA4 config is unaffected — but when a user enables them, training silently skips those weights, and the forward-only consistency tests cannot catch it (a leaf gives identical output whether it is a `Param` or a `Variable`). Consider driving promotion from the dpmodel `trainable` flag / mirroring pt's `nn.Parameter` set rather than a hand-maintained allowlist (which will keep silently dropping future params), and adding a test that asserts the differentiated leaf set equals the trainable dpmodel param set.

deepmd/jax/utils/serialization.py
```diff
@@ -50,6 +50,40 @@ def _normalize_restored_state_keys(
...
+            subtree = _zero_size_subtree(item)
+            if subtree is not _NO_ZERO_SIZE_LEAF:
+                restored[key] = subtree
+        return restored if restored else _NO_ZERO_SIZE_LEAF
+    if getattr(value, "size", None) == 0:
+        return value
+    return _NO_ZERO_SIZE_LEAF
+
+
+def _restore_missing_zero_size_leaves(template: Any, restored: Any) -> Any:
```
wanghan-iapcm: **Non-blocking (test coverage):** the new stateful glue in this PR ships without dedicated tests — only the two forward-consistency tests were added and `source/tests/jax/` was not touched. Per CLAUDE.md ("UTs should cover all code paths, including both branches of boolean conditions"), worth covering:

- the zero-size drop -> restore round-trip: `_drop_zero_size_array_leaves` (`jax/train/trainer.py`) paired with `_restore_missing_zero_size_leaves` / `_zero_size_subtree` here. This mutates checkpoint state on save and reconstructs it on load, so a mismatch would silently corrupt a restored model — and it is trivially testable as a pure round-trip assertion.
- `get_sezm_model`'s guard branches (spin / bridging / lora / use_compile / preset_out_bias, plus the descriptor/fitting/exclude_types mismatches) — `source/tests/jax/test_model_factory.py` is the established home for routing/validation tests.
- `_match_label_shapes` and the force-reshape branch in `_evaluate_model_dict` (both the reshape and no-op branches).

njzjz:
Pushed follow-up commit 68cb12e0d for the remaining review feedback.

Changes:
- Promote optional cross-grid FrameContract/FrameExpand weights and FFN LayerScale leaves to JAX parameters.
- Remove the redundant imports reported by CodeQL.
- Add focused coverage for parameter promotion, zero-size checkpoint restore, DPA4 factory validation, and force-shape normalization.

Validation:
- Focused JAX tests: 9 passed, 5 subtests passed.
- DPA4 fitting consistency: 28 passed, 36 skipped.
- ruff format . and ruff check . passed.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/jax/fitting/dpa4_ener.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@flax_module
+class GLUFittingNet(GLUFittingNetDP):
+    pass
+
+
+register_dpmodel_mapping(
+    GLUFittingNetDP,
+    lambda v: GLUFittingNet.deserialize(v.serialize()),
```
OutisLi: [P1] Preserve the fitting layers' trainable flags during conversion

This round-trip loses `trainable`: `GLUFittingNetDP.serialize()` does not include that field, so `GLUFittingNet.deserialize()` rebuilds every `NativeLayer` with the default `trainable=True`. Consequently a supported JAX DPA4 config with `fitting_net.trainable: false` still exposes all hidden/output weights as `nnx.Param`; since the trainer optimizes `wrt=nnx.Param`, an SGD step actually changes those supposedly frozen weights (while the source dpmodel layers remain `trainable=False`). Please preserve the source layers' trainability when converting the network, and cover an all-false fitting config with an optimizer-step test.

OutisLi (changes requested):
Requesting changes for the blocking DPA4 fitting-trainability correctness issue described inline.

deepmd/jax/train/trainer.py
```diff
@@ -831,6 +833,7 @@ def _write_checkpoint(self, ckpt_path: Path, *, step: int) -> None:
         else:
             _, single_state = nnx.split(self.models[DEFAULT_TASK_KEY])
             state = single_state.to_pure_dict()
+        state = _drop_zero_size_array_leaves(state)
```
OutisLi: [P1] Apply zero-size filtering to every `.jax` save path

This filtering only protects checkpoints produced by `DPTrainer._write_checkpoint`. The other supported `.jax` producer, `deserialize_to_file()` in `deepmd/jax/utils/serialization.py`, still passes the unfiltered NNX state directly to `StandardSave`. A minimal DPA4 model contains zero-size leaves such as `atomic_model.descriptor.mean`, `stddev`, and `output_ffn.act.expand_index`, so `deserialize_to_file("model.jax", ...)` — and therefore `dp convert-backend ... model.jax` through `JAXBackend.deserialize_hook` — fails with `ValueError: Cannot save arrays with zero size`. Please move/share the filtering helper with the serialization layer, apply it to both `.jax` writers, and cover this with a real Orbax DPA4 save/restore test rather than only a pure-dict helper round trip.

OutisLi (commented):
Adding a second blocking DPA4 correctness finding; the existing changes-requested state remains applicable.

deepmd/jax/model/model.py
```diff
@@ -109,6 +109,53 @@ def get_zbl_model(data: dict) -> DPZBLModel:
...
+    if data["fitting_net"]["type"] not in ("dpa4_ener", "sezm_ener"):
+        raise ValueError(
+            "Model type 'dpa4' requires the DPA4/SeZM energy fitting net, but got "
+            f"fitting_net type '{data['fitting_net']['type']}'."
+        )
+
+    descriptor_exclude_types = [
+        list(pair) for pair in (data["descriptor"].get("exclude_types") or [])
+    ]
+    if "pair_exclude_types" in data:
```
OutisLi: [P2] Distinguish the normalized default from an explicit pair exclusion

This presence check does not survive the real training configuration path: `model_args().normalize_value()` always inserts the model-level default `pair_exclude_types=[]`. Thus a user who specifies only `descriptor.exclude_types=[[0, 1]]` reaches this branch with an apparently “provided” empty model-level list and gets the mismatch `ValueError`; the fallback below that copies descriptor exclusions is reachable only from raw, unnormalized dictionaries. The new factory test uses such a raw dict, so it misses the training failure. Please canonicalize these fields in a way that accounts for the normalized default (or preserves explicitness), and test `normalize_value(...) -> get_model(...)` for descriptor-only exclusions.

OutisLi (commented):
Adding the approved DPA4 exclusion-routing correctness finding.

deepmd/jax/descriptor/dpa4.py
```diff
@@ -0,0 +1,285 @@
...
+        if list_names is not None:
+            _promote_trainable_lists(submodule, list_names)
+    return module
+
+
+@flax_module
+class SO2Linear(SO2LinearDP):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.weight_m = _as_param_list(self.weight_m)
```
OutisLi: [P1] Honor `trainable=False` for the whole DPA4 descriptor

`SO2Linear` promotes every `weight_m` entry unconditionally here (and again in `deserialize`), while `_promote_trainable` also treats modules without a local `trainable` attribute as trainable; `RadialBasis` falls into that case. Consequently, even a minimal `DescrptDPA4(..., trainable=False)` still exposes the four `so2_linears.*.weight_m.0` arrays and `radial_basis.adam_freqs` as `nnx.Param`.

This is not only state metadata: after loading values from a model that has moved beyond its zero-initialized gates into an otherwise frozen descriptor, a real descriptor forward loss gives `radial_basis.adam_freqs` a nonzero gradient and the trainer's `wrt=nnx.Param` optimizer updates it. The reference PT implementation applies the root descriptor's freeze flag to every descendant parameter. Please avoid or demote these promotions when the descriptor is frozen, and add a regression asserting that a `trainable=False` DPA4 descriptor exposes no optimizer-visible `nnx.Param` leaves.

OutisLi (commented):
Adding the approved DPA4 descriptor-freeze correctness finding.

deepmd/jax/model/ener_model.py
```diff
@@ -11,6 +11,8 @@
 )
 
 
+@BaseModel.register("sezm_ener")
```
OutisLi: [P2] Handle the regular PT SeZM schema during backend conversion

These fitting-style aliases do not make the public PT-to-JAX conversion path work. The PyTorch backend explicitly advertises `.pt` checkpoints as IO inputs, and serializing a valid DPA4 training checkpoint produces the specialized top-level schema `{"type": "SeZM", "atomic_model": ...}`. Consequently, `dp convert-backend checkpoint.pt model.hlo` reaches JAX `BaseModel.deserialize()` and fails with `RuntimeError: Unknown model type: SeZM`.

DPA4 has no `.pth` alternative; its regular PT training artifact is the supported `.pt` checkpoint. Please canonicalize the PT SeZM whole-model schema before JAX deserialization and add an end-to-end `.pt` checkpoint conversion test.

OutisLi (commented):
Additional interoperability finding.

OutisLi (commented):
Adding the approved charge/spin-conditioning training finding.

deepmd/jax/model/model.py
```diff
@@ -109,6 +109,53 @@ def get_zbl_model(data: dict) -> DPZBLModel:
     )
 
 
+def get_sezm_model(data: dict) -> BaseModel:
+    """Build a DPA4/SeZM energy model from the pt-style model config."""
+    data = deepcopy(data)
+    if "spin" in data:
```
OutisLi: [P2] Reject unsupported charge/spin conditioning before training

`get_sezm_model()` accepts `descriptor.add_chg_spin_ebd=True`, but the JAX trainer never registers, loads, or forwards the corresponding `charge_spin` input: its prepared batch and `_evaluate_model_dict()` path only carry `fparam` and `aparam`. With a normalized DPA4 config and no `default_chg_spin`, the model therefore constructs successfully but fails on its first training forward with `ValueError: charge_spin is required for this SeZM descriptor`. When a default is configured, per-frame `charge_spin` data are instead silently ignored and the fixed default is always used.

Since this descriptor option is currently documented as PT-only, please reject it here alongside the other unsupported DPA4 features. Alternatively, if JAX support is intended, add `charge_spin` to the data requirements and every train/validation call path.

deepmd/jax/fitting/dpa4_ener.py
njzjz: Fixed in commit 7304a6773. GLUFittingNet now serializes and restores the effective per-layer trainable flags instead of rebuilding every NativeLayer as trainable. The JAX regression verifies that an all-false fitting round trip exposes no optimizer-visible nnx.Param leaves.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/jax/train/trainer.py
njzjz: Fixed in commit 7304a6773. Zero-size filtering now lives in the JAX serialization layer and is applied by both the trainer checkpoint writer and deserialize_to_file. The new end-to-end test converts a real DPA4 .pt checkpoint to Orbax .jax and restores it successfully.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/jax/model/ener_model.py
njzjz: Fixed in commit 7304a6773. JAX BaseModel now validates and unwraps the PT SeZM model wrapper, normalizes the sezm_atomic schema, and rejects unsupported dens, bridging, and LoRA state. A real current-format .pt checkpoint is converted to .jax and restored in the regression.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/jax/model/model.py
njzjz: Fixed in commit 7304a6773. An empty normalized pair_exclude_types default is now treated as absent when descriptor.exclude_types is populated. The regression runs model_args normalization before the JAX factory and preserves the descriptor exclusions.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/jax/model/model.py
njzjz: Fixed in commit 7304a6773. The JAX DPA4 factory now rejects descriptor.add_chg_spin_ebd before model construction, with focused factory coverage, so unsupported charge and spin conditioning cannot fail later in training or silently use only a default.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/jax/descriptor/dpa4.py
njzjz: Fixed in commit 7304a6773. Descriptor-level trainable=False now demotes every mapped descendant parameter, including SO2Linear and RadialBasis leaves, to tracked non-parameter NNX variables. The regression asserts that the frozen descriptor has no nnx.Param state.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

wanghan-iapcm (approved):
Both threads I raised are addressed and verified against the current HEAD (fix in 68cb12e0d):

- Silently-frozen params (`dpa4.py`): the promotion maps now include `FrameContract.weight`, `FrameExpand.weight`, and `SeZMInteractionBlock.adam_ffn_layer_scales`, and the new `test_optional_dpa4_weights_are_jax_parameters` enables `layer_scale` + `message_node_so3` and asserts these leaves are `ArrayAPIParam` — a test that fails on the pre-fix maps, so it genuinely locks the fix (with a companion frozen-descriptor test for the other side).
- Test coverage (`serialization.py` / trainer glue): `source/tests/jax/` now covers the zero-size checkpoint drop/restore round-trip, the `get_sezm_model` guard/mismatch paths, and `_match_label_shapes` / `_evaluate_model_dict` branches.

Approving on that basis.

deepmd/pt/model/task/sezm_ener.py
```diff
@@ -243,11 +243,18 @@ def __init__(
         super().__init__()
         if neuron is None:
             neuron = []
-        if isinstance(trainable, list):
-            trainable = all(trainable)
         self.in_dim = int(in_dim)
         self.out_dim = int(out_dim)
         self.neuron = [int(nn_dim) for nn_dim in neuron]
+        if isinstance(trainable, bool):
```
OutisLi: [P1] Reapply the saved freeze policy to deserialized PT layers

Recording `self.trainable` here does not make the newly created PyTorch parameters frozen. `GLULayer`/`MLPLayer` retain the flag as metadata but leave their `Parameter.requires_grad` at its default, and `SeZMEnergyFittingNet.deserialize()` first freezes a temporary collection and then replaces it with `SeZMNetworkCollection.deserialize()` without reapplying the policy. On this head, `SeZMEnergyFittingNet(trainable=False)` has every parameter frozen before `serialize()`, but after `SeZMEnergyFittingNet.deserialize(fitting.serialize())` every hidden/output parameter has `requires_grad=True` even though both the outer and nested `trainable` metadata remain false. This silently makes a frozen fitting optimizer-visible after a PT serialization round trip. Please apply each saved flag to the corresponding hidden/output parameters inside `GLUFittingNet` itself (and to the optional FiLM parameters), and add a PT `serialize -> deserialize` regression that asserts the actual `requires_grad` state.

OutisLi (changes requested):
The PT DPA4 fitting freeze policy is still not preserved across its serialization round trip.

deepmd/jax/model/model.py
```diff
@@ -109,6 +109,57 @@ def get_zbl_model(data: dict) -> DPZBLModel:
...
+        if descriptor_exclude_types and descriptor_exclude_types != pair_exclude_types:
+            raise ValueError(
+                "DPA4/SeZM pair_exclude_types and descriptor.exclude_types must "
+                "match when both are provided."
+            )
+    else:
+        pair_exclude_types = descriptor_exclude_types
+    data["pair_exclude_types"] = pair_exclude_types
+    data["descriptor"]["exclude_types"] = deepcopy(pair_exclude_types)
+    return get_standard_model(data)
```
OutisLi: [P1] Keep every in-cutoff edge for JAX DPA4 training

Routing DPA4 through `get_standard_model` makes the trainer use the standard dense `prepare_input` path. That builder sets `nsel = sum(sel)` and slices the sorted neighbor axis to `nsel`, whereas DPA4's conservative energy contract says `sel` is only the initial capacity and every neighbor within `rcut` must be retained. A four-atom nonperiodic frame with all pair distances below `rcut` produces only 4 directed edges for `sel=1`, instead of the 12 edges obtained for `sel=3`. This creates a training/deployment mismatch because PT DPA4 uses a sparse builder whose edge set is independent of `sel`. The same route also performs an O(N^2) pairwise search and evaluates padded O(N*sel) RBF/Wigner slots, wasting substantial work when `sel` is made large enough to avoid truncation. Please give JAX DPA4 a conservative dynamic/sparse edge path that retains all in-cutoff edges (`sel` only as initial capacity), and add a regression with `n_neighbors > sel` that checks edge-count and gradient parity.

OutisLi (commented):
JAX DPA4 currently routes through a fixed-width standard neighbor-list path, violating DPA4's conservative full-cutoff edge semantics.

deepmd/jax/descriptor/dpa4.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+@BaseDescriptor.register("SeZM")
+@BaseDescriptor.register("sezm")
+@BaseDescriptor.register("DPA4")
+@BaseDescriptor.register("dpa4")
+@flax_module
+class DescrptDPA4(DescrptDPA4DP):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
```
OutisLi, line 303: [P2] Honor or explicitly reject `random_gamma` during JAX training

This wrapper inherits `DescrptDPA4DP.call()`, whose dense and sparse-edge cache paths both hard-code `random_gamma=False`. Instrumenting a JAX descriptor configured with `random_gamma=True` shows that `build_edge_cache` still receives `False`, so both the default and an explicit `True` silently omit the documented per-edge, per-forward local-Z roll; PT correctly uses `self.random_gamma and self.training`. The new JAX tests all set this option to `False`, so they do not cover the default behavior.

Please either thread a JAX PRNG key through training, split it for each forward, and pass explicitly generated per-edge gamma into the cache, or fail fast for `random_gamma=True` until this is supported. Merely replacing the hard-coded value with `self.random_gamma` is insufficient because the current NumPy fallback RNG would execute during `nnx.jit` tracing and freeze one sample for all subsequent steps.

deepmd/jax/descriptor/dpa4.py
```diff
@@ -0,0 +1,296 @@
...
+    lambda v: SO2Linear.deserialize(v.serialize()),
+)
+
+
+@BaseDescriptor.register("SeZM")
+@BaseDescriptor.register("sezm")
+@BaseDescriptor.register("DPA4")
+@BaseDescriptor.register("dpa4")
+@flax_module
+class DescrptDPA4(DescrptDPA4DP):
```
OutisLi, line 288: [P2] Do not silently accept an inert `use_amp=True`

The DPA4 schema defaults `use_amp` to `True` and documents BF16 automatic mixed precision during GPU training, but the dpmodel/JAX implementation only stores `self.use_amp`; neither this wrapper nor the JAX trainer reads it or installs a mixed-precision policy. With otherwise identical descriptors, `use_amp=True` and `False` produce identical results, and the `True` JAXPR contains no BF16 casts. PT, by contrast, applies `torch.autocast` in its compute context.

Please implement a JAX mixed-precision policy, or fail fast for `use_amp=True` until it is available. At minimum, emit an explicit warning and document that the option is currently a no-op; silently accepting it misrepresents both the benchmark configuration and expected speed/memory behavior.

OutisLi (changes requested):
JAX DPA4 currently accepts two default training options, `random_gamma=True` and `use_amp=True`, while silently executing neither behavior. The random augmentation needs a JAX-safe PRNG path (or an explicit unsupported-option guard), and AMP must be implemented or surfaced as unsupported instead of being a no-op.

deepmd/jax/model/ener_model.py
OutisLi: [P1] Route the model-level DPA4 aliases through `update_sel`

The normal `dp --jax train` path calls `BaseModel.update_sel()` before `get_model()`. That dispatcher looks up the top-level `model.type`, but the JAX registry only contains the fitting/model aliases `dpa4_ener` and `sezm_ener`; `dpa4`, `DPA4`, `sezm`, and `SeZM` all fail here with `RuntimeError: Unknown model type` before any model is constructed. The PR's smoke test and benchmark both pass `--skip-neighbor-stat`, which masks this default-path failure. Moreover, `sel: auto` requires neighbor statistics, so it has no workaround.

Please route all four model-level aliases through the correct DPA4 `update_sel` preprocessing path (without collapsing the specialized DPA4 model identity into a generic energy model), and add an end-to-end JAX DPA4 training regression that does not pass `--skip-neighbor-stat`, ideally including `sel: auto`.

OutisLi (changes requested):
The default JAX DPA4 training entrypoint fails before model construction because neighbor-stat preprocessing cannot dispatch the model-level DPA4 aliases.

source/tests/jax/test_dpa4_conversion.py
```diff
@@ -0,0 +1,72 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+"""End-to-end PT checkpoint conversion coverage for JAX DPA4."""
+
+from copy import (
+    deepcopy,
+)
+
+import torch
```
OutisLi: [P3] Keep the JAX-only test suite collectable without PyTorch

This module imports `torch` and `deepmd.pt.*` unconditionally at collection time, but the `jax` extra does not depend on PyTorch. The current unified CI installs `[test,jax,torch]`, so it masks the issue; in a JAX-only development environment, collecting `source/tests/jax` now fails before any tests can run. Since this is intentionally a PT→JAX interoperability test, please gate it with `pytest.importorskip("torch")` before importing `deepmd.pt` modules, or move it to a cross-backend test suite.

OutisLi (commented):
One additional non-blocking test-isolation issue.

njzjz:

- @iProzd — 17 reviews on exact changed files (`deepmd/dpmodel/fitting/dpa4_ener.py`, `deepmd/jax/train/trainer.py`, `deepmd/jax/train/validation.py`, `deepmd/jax/utils/serialization.py`, `source/tests/jax/test_training.py`).

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

OutisLi (approved):
Re-reviewed the current descriptor-only head. The focused parameter-state tests and JAX consistency cases pass; JIT/gradient smoke coverage succeeds, and PT/JAX trainable parameter counts match across the optional DPA4 branches I checked. I found no remaining blocking issue in this scope.

PR #5749, opened 2026-07-07
feat(tf2): support DPA4 descriptor

njzjz:
## Summary

- Add and register the TF2 DPA4/SeZM descriptor adapter and its TensorFlow-backed submodule mappings.
- Add the TensorFlow array operations, dynamic-shape handling, and trackable parameter support required by the DPA4 descriptor.
- Enable TF2 coverage in the existing DPA4 descriptor consistency tests and add focused descriptor state, checkpoint, graph-shape, and array-operation tests.
- Keep this pull request descriptor-only; fitting, model factory, trainer, and model-conversion changes are excluded.

## Validation

- ruff format .
- ruff check .
- python -m py_compile deepmd/dpmodel/array_api.py deepmd/dpmodel/descriptor/dpa4.py deepmd/dpmodel/descriptor/dpa4_nn/so2.py deepmd/tf2/common.py deepmd/tf2/descriptor/dpa4.py source/tests/consistent/descriptor/test_dpa4.py source/tests/consistent/test_array_api.py source/tests/tf2/test_dpa4.py
- DP_TEST_TF2_ONLY=1 pytest source/tests/tf2/test_dpa4.py -v — 6 passed
- DP_TEST_TF2_ONLY=1 pytest source/tests/consistent/test_array_api.py::TestXpMaximumAtConsistent::test_tf_preserves_all_negative_infinity_segment -v — 1 passed

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh


njzjz:
Pushed follow-ups `789aa79ca` and `8136fb11` for the TF2 CI/review feedback.\n\nChanges kept:\n- Preserve empty-segment `-inf` values in the TensorFlow branch of `xp_maximum_at`.\n- Treat unknown-rank `tf.TensorShape(None)` as non-static by catching `ValueError`.\n- Avoid per-access set allocation in `TF2Module.__getattribute__`.\n- Route TF2 `se_atten_v2` deserialization through the `se_atten_v2` serializer path instead of DPA1.\n\nI then reverted the extra `source/tests/consistent` changes in `8136fb11`, per request.\n\nLocal validation:\n- TF2 smoke for `se_atten_v2` serialize/deserialize passed.\n- TensorFlow `xp_maximum_at` 1D and 2D empty-segment checks passed.\n- `DPTrainer._static_shape(tf.TensorShape(None))` returns `None`.\n- `ruff format . && ruff check .` passed.\n\nCI is rerunning.

deepmd/tf2/descriptor/dpa4.py
```diff
@@ -0,0 +1,267 @@
...
+    pass
+
+
+register_dpmodel_mapping(
+    WignerDCalculatorDP,
+    lambda v: WignerDCalculator(v.lmax, eps=v.eps, precision=v.precision),
+)
+
+
+_TRAINABLE_ATTRS: dict[str, tuple[str, ...]] = {
```
wanghan-iapcm, line 106: **Non-blocking (correctness — silently frozen params):** promotion here is strictly allowlist-gated — `_promote_trainable_tree` promotes to `tf.Variable` only the attributes named in `_TRAINABLE_ATTRS` / `_TRAINABLE_LIST_ATTRS` (keyed by `type(submodule).__name__`); anything else stays a plain array / non-trainable leaf, so the optimizer never updates it — it is silently frozen during training. A few trainable dpmodel params appear to be missing from the maps:

- `FrameContract.weight` / `FrameExpand.weight` — trainable in `deepmd/dpmodel/descriptor/dpa4_nn/grid_net.py` (pt registers them as `nn.Parameter`), reachable when `node_wise_so3` / `message_node_so3` instantiate a `mode=="cross"` grid net. Neither class is a key in the maps.
- `SeZMInteractionBlock.adam_ffn_layer_scales` — a trainable LayerScale list when `layer_scale=True` (`dpa4_nn/block.py`), also listed in `dpa4_nn/lora.py`'s unfreeze registry. The maps already cover the analogous `SO2Convolution.adam_so2_layer_scales`, but not this one.

These are opt-in (non-default) features, so a stock DPA4 config is unaffected — but when a user enables them, training silently skips those weights, and the forward-only consistency tests cannot catch it (a leaf gives identical output whether it is a trainable `tf.Variable` or a frozen constant). Consider driving promotion from the dpmodel `trainable` flag / mirroring pt's `nn.Parameter` set rather than a hand-maintained allowlist (which will keep silently dropping future params), and adding a test that asserts the set of trained variables equals the trainable dpmodel param set.

deepmd/tf2/common.py
```diff
@@ -266,11 +310,113 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
                     )
                     if converted is not value:
                         setattr(self, name, converted)
+            self._refresh_tf2_trackable_lists()
+
+        def _refresh_tf2_trackable_lists(self) -> None:
```
wanghan-iapcm, line 340: **Non-blocking (test coverage):** the new stateful glue in this PR ships without dedicated tests — only the two forward-consistency tests were added (`test_dpa4.py` / `test_dpa4_ener.py`) and `source/tests/tf2/` was not touched. Per CLAUDE.md ("UTs should cover all code paths, including both branches of boolean conditions"), worth covering:

- the deserialize -> trackable-refresh round-trip: `_refresh_tf2_trackable_lists` + the wrapped `deserialize` rebuild the `tf.Module` trackable containers on load, so a mismatch would silently corrupt checkpoint save/restore — trivially testable as a serialize -> deserialize -> re-serialize round-trip assertion.
- `get_sezm_model`'s guard branches (spin / bridging / lora / use_compile / preset_out_bias, plus the descriptor / fitting / exclude_types mismatches in `deepmd/tf2/model/model.py`) — none of these reachable branches are exercised.
- `_match_label_shapes` and the force-reshape branch in `_translate_model_ret_to_loss_dict` (`deepmd/tf2/train/trainer.py`) — both the reshape and the no-op branches.

njzjz:
Pushed follow-up commit 4328329eb for the remaining review feedback.

Changes:
- Promote optional cross-grid FrameContract/FrameExpand weights and FFN LayerScale leaves to tracked, trainable TensorFlow variables.
- Remove the redundant imports reported by CodeQL.
- Remove the duplicate DescrptSeAttenV2 refresh call; the tf2_module deserialize wrapper already performs it.
- Normalize explicit null descriptor/fitting blocks and add focused factory, trackable round-trip, and force-shape tests.

Validation:
- Focused TF2 tests: 14 passed.
- DPA4 fitting consistency: 28 passed, 36 skipped.
- ruff format . and ruff check . passed.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/tf2/train/trainer.py
```diff
@@ -1434,7 +1434,36 @@ def _translate_model_ret_to_loss_dict(
...
+        if label_dict is None:
+            return model_dict
+        force_hat = model_dict.get("force")
+        force = label_dict.get("force")
+        if force_hat is None or force is None:
+            return model_dict
+        force_hat_shape = cls._static_shape(force_hat)
+        force_shape = cls._static_shape(force)
+        if (
+            force_hat_shape is None
```
OutisLi: [P1] Reshape force outputs with dynamic shapes under reduce_retracing

This static-shape guard stops applying the normalization as soon as `tf.function(reduce_retracing=True)` generalizes the atom dimension. A task that samples systems with different atom counts traces the first batch with concrete shapes, then obtains shapes such as `(1, None, 3)` and `(1, None)`; `_static_shape()` returns `None`, so the second trace leaves prediction and label at different ranks. For the DPA4 force loss, TensorFlow then directly subtracts `(B, N, 3)` and `(B, 3N)` and fails at runtime (`Incompatible shapes: [1,3,3] vs. [1,9]` in a two-/three-atom reproduction).

Please reshape using dynamic `tf.shape(force)`/`tf.size`—with a runtime size assertion if needed—and add a `reduce_retracing=True` regression that invokes the same function with two different atom counts.

OutisLi (changes requested):
Requesting changes for the dynamic-shape DPA4 force-loss correctness issue described inline.

deepmd/tf2/descriptor/dpa4.py
```diff
@@ -0,0 +1,268 @@
...
+    lambda v: SO2Linear.deserialize(v.serialize()),
+)
+
+
+@BaseDescriptor.register("SeZM")
+@BaseDescriptor.register("sezm")
+@BaseDescriptor.register("DPA4")
+@BaseDescriptor.register("dpa4")
+@tf2_module
+class DescrptDPA4(DescrptDPA4DP):
```
OutisLi, line 308: [P1] Make the default DPA4 forward graph-safe for generalized atom counts

A real compiled DPA4 training step succeeds for the first fixed-size batch but cannot trace the same task for a second atom count. The default `radial_so2_mode="degree_channel"` instantiates `DynamicRadialDegreeMixer`; once `tf.function(reduce_retracing=True)` generalizes its edge dimension, `ndtensorflow.Array.shape` represents that dimension with a scalar tensor. `DynamicRadialDegreeMixer.call()` then evaluates `if x_local.shape != radial_feat.shape` in `dpa4_nn/so2.py`, so tuple comparison feeds a symbolic tensor to a Python `if` and raises `OperatorNotAllowedInGraphError`.

Please replace these Python shape predicates with graph-safe static checks or TensorFlow runtime assertions, and add an end-to-end compiled training test that invokes one task with two atom counts.

OutisLi (commented):
Adding the approved graph-safety correctness finding.

OutisLi (commented):
Adding the approved fitting-trainability correctness finding.

deepmd/tf2/fitting/dpa4_ener.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+@tf2_module
+class GLUFittingNet(GLUFittingNetDP):
+    pass
+
+
+register_dpmodel_mapping(
+    GLUFittingNetDP,
+    lambda v: GLUFittingNet.deserialize(v.serialize()),
```
OutisLi: [P1] Preserve the fitting trainability policy during TF2 conversion

The conversion through `GLUFittingNet.deserialize(v.serialize())` loses the requested trainability because `GLUFittingNet.serialize()` does not include `trainable`, so deserialization rebuilds every native layer with the default `trainable=True`. On this head, `SeZMEnergyFittingNet(..., trainable=False)` still exposes all three hidden/output TensorFlow variables through `trainable_variables`; their gradients are nonzero and an SGD step updates all of them. Since the trainer optimizes `model.trainable_variables`, an explicitly frozen fitting silently drifts.

Please preserve the effective trainable flags through conversion and add optimizer-step regressions for both `False` and a per-layer mask.

deepmd/tf2/descriptor/dpa4.py
```diff
@@ -0,0 +1,268 @@
...
+    if name not in attrs:
+        tf.Module.__setattr__(
+            module,
+            "_tf2_array_variable_list_attrs",
+            attrs | {name},
+        )
+
+
+def _promote_trainable(module: Any, names: tuple[str, ...]) -> None:
+    if not getattr(module, "trainable", True):
```
OutisLi: [P1] Track frozen DPA4 parameters as non-trainable variables

Returning here for `trainable=False` leaves the mapped DPA4 parameter arrays as untracked constants instead of `tf.Variable(trainable=False)`, so checkpoints cannot save or restore the frozen state. In a direct round trip, two frozen descriptors were initialized with different seeds; checkpoint restoration reported fully consumed and matched, but the target `type_embedding.adam_type_embedding` remained completely unchanged and still differed from the source. The frozen descriptor exposes only 8 tracked variables versus 46 for the same trainable configuration.

At the same time, `RadialBasis` has no local `trainable` attribute, so its allowlisted `adam_freqs` is still created as trainable even when the root descriptor is frozen. Please propagate the descriptor-level flag while promoting every parameter to a tracked variable, using `trainable=False` for frozen leaves, and test both optimizer membership and checkpoint restoration.

OutisLi (commented):
Adding the approved frozen-descriptor tracking correctness finding.

deepmd/tf2/model/model.py
```diff
@@ -107,6 +107,53 @@ def get_zbl_model(data: dict) -> DPZBLModel:
...
+    if data["fitting_net"]["type"] not in ("dpa4_ener", "sezm_ener"):
+        raise ValueError(
+            "Model type 'dpa4' requires the DPA4/SeZM energy fitting net, but got "
+            f"fitting_net type '{data['fitting_net']['type']}'."
+        )
+
+    descriptor_exclude_types = [
+        list(pair) for pair in (data["descriptor"].get("exclude_types") or [])
+    ]
+    if "pair_exclude_types" in data:
```
OutisLi: [P1] Preserve descriptor exclusions after config normalization

The standard `model_args().normalize_value(...)` path always inserts the default `pair_exclude_types=[]`, even when the user only specifies `descriptor.exclude_types`. A valid configuration with `descriptor.exclude_types=[[0, 1]]` therefore reaches this branch with an empty model-level value, which is treated as explicitly provided and rejected as a mismatch. I reproduced the normalized config as `pair_exclude_types=[]` and `descriptor.exclude_types=[[0, 1]]`; `get_model()` then raises before constructing the model.

Please preserve whether the model-level field was actually specified, or treat its empty default as absent when the descriptor supplies exclusions, and add a regression that normalizes the configuration before calling the TF2 factory. The current raw-dict test bypasses the failing production path.

OutisLi (commented):
Adding the approved normalized-exclusion routing correctness finding.

deepmd/tf2/descriptor/dpa4.py
OutisLi, line 308: [P2] Honor `random_gamma` during TF2 training

This wrapper inherits the DPModel forward implementation, where both the dense and sparse-edge cache builders hard-code `random_gamma=False`. I instrumented the TF2 descriptor with `random_gamma=True` and observed that the forwarded value was still `False`. Since the option defaults to `True`, every default TF2 DPA4 training run silently omits the configured local-Z roll augmentation, while the PT implementation correctly uses `self.random_gamma and self.training`. The new TF2 tests all explicitly set the option to `False`, so they do not exercise the default behavior.

Please propagate an explicit training state and generate gamma with a graph-safe TensorFlow RNG, or fail fast for `random_gamma=True` until it is supported. Calling the existing NumPy RNG directly inside `tf.function` would freeze randomness at trace time.

OutisLi (commented):
Adding the approved default-random-gamma training semantics finding.

deepmd/tf2/model/model.py
```diff
@@ -107,6 +107,53 @@ def get_zbl_model(data: dict) -> DPZBLModel:
...
+    data["descriptor"] = data.get("descriptor") or {}
+    data["fitting_net"] = data.get("fitting_net") or {}
+    data["descriptor"].setdefault("type", "dpa4")
+    data["fitting_net"].setdefault("type", "dpa4_ener")
+    if data["descriptor"]["type"] not in ("dpa4", "DPA4", "sezm", "SeZM"):
+        raise ValueError(
+            "Model type 'dpa4' requires a DPA4/SeZM descriptor, but got "
+            f"descriptor type '{data['descriptor']['type']}'."
+        )
+    if data["fitting_net"]["type"] not in ("dpa4_ener", "sezm_ener"):
```
OutisLi: [P2] Allow schema-supported property fitting through the DPA4 route

The shared DPA4 model schema explicitly includes invariant `property` fitting, and strict normalization accepts a `dpa4` / `dpa4` / `property` configuration. This new whitelist then rejects it as energy-only. The lower TF2 stack already supports the combination: passing the same DPA4 descriptor and property fitting configuration to `get_standard_model()` successfully constructs a `PropertyModel` with a `PropertyFittingNet`.

Please allow `property` through this route and add a normalized factory regression. If TF2 DPA4 is intentionally energy-only, the backend-facing schema and documentation need to reject that configuration consistently rather than accepting it during validation and failing at model construction.

OutisLi (commented):
Adding the approved DPA4 property-fitting factory compatibility finding.

deepmd/tf2/common.py
```diff
@@ -253,6 +256,47 @@ def tf2_module(module: type[T]) -> type[T]:
...
+            if value is None:
+                tf.Module.__setattr__(self, storage_name, None)
+                return
+            tensor = to_tf_tensor(value)
+            variable = tf.Variable(
+                tensor,
+                trainable=bool(getattr(self, "trainable", True)),
+                name=name,
+            )
+            tf.Module.__setattr__(self, storage_name, variable)
```
OutisLi, line 293: [P2] Release the stale tensors after parameter promotion

When a DPModel object is initialized, each promoted parameter already exists in the public attribute as an `xp.Array` backed by an EagerTensor. These setters create the new `_tf2_*_variable` storage but leave the original value in `module.__dict__`. `__getattribute__()` hides that value afterwards, yet the stale tensor remains strongly referenced for the lifetime of the model.

For a DPA4 descriptor with `channels=64`, `lmax=3`, and three blocks, all 97 promoted leaves retained such shadows: the stale tensors occupied 10.023 MiB while the tracked variables occupied 10.408 MiB, making resident parameter storage 1.96x the required size. Please remove the migrated public storage after creating the variable, or enable variable-backed attributes before their first assignment, and add a regression that promoted names no longer retain tensor values in the raw `__dict__`.

OutisLi (commented):
Adding the approved DPA4 parameter-promotion resident-memory finding.

deepmd/tf2/model/ener_model.py
```diff
@@ -11,6 +11,8 @@
 )
 
 
+@BaseModel.register("sezm_ener")
```
OutisLi: [P1] Support PT DPA4 `.pt` conversion into the TF2 backend

The public `dp convert-backend model.pt model.savedmodeltf` path cannot deserialize a current-head PT DPA4 checkpoint. The PT input hook succeeds, but its model payload has the custom outer type `SeZM` with a nested `sezm_atomic` model. The TF2 output hook then calls `BaseModel.deserialize()` and fails with `RuntimeError: Unknown model type: SeZM`.

Adding `SeZM` as another alias here is not sufficient: forcing the outer type to `sezm_ener` next fails on the incompatible nested atomic-model schema/version. Please add an adapter from the PT SeZM model payload to the generic TF2 energy/atomic-model representation, or emit a backend-neutral schema from serialization, and add an end-to-end regression using a current-format DPA4 `.pt` checkpoint.

OutisLi (commented):
Adding the approved PT DPA4 .pt to TF2 conversion integration finding.

deepmd/tf2/model/model.py
njzjz: Fixed in commit aa1ed8926. The TF2 factory now treats the normalized empty pair_exclude_types default as absent when descriptor exclusions are present. The regression normalizes the configuration before constructing the factory input.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/tf2/train/trainer.py
njzjz: Fixed in commit aa1ed8926. Force layout normalization now uses dynamic tf.shape and tf.size with a runtime size assertion, while preserving known incompatible layouts. The reduce_retracing regression calls the same function with two and three atom force tensors.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/tf2/fitting/dpa4_ener.py
njzjz: Fixed in commit aa1ed8926. GLUFittingNet now preserves the full per-layer trainability mask through serialization and TF2 conversion. Optimizer-step regressions cover both the all-false policy and a hidden-frozen/output-trainable mask, verifying that only enabled variables change.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/tf2/model/ener_model.py
njzjz: Fixed in commit aa1ed8926. TF2 BaseModel now adapts the regular PT SeZM wrapper and sezm_atomic payload to the generic energy model schema, with validation of unsupported modes. An isolated end-to-end regression creates a real .pt checkpoint and exports a loadable .savedmodeltf artifact.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/tf2/common.py
njzjz: Fixed in commit aa1ed8926. Variable promotion now removes the migrated public tensor slot after creating the tracked tf.Variable storage, including list-valued parameters. The regression walks all promoted attributes and verifies that no stale tensor shadows remain in raw module dictionaries.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/tf2/descriptor/dpa4.py
njzjz: Fixed in commit aa1ed8926. Native TF2 DPA4 construction now fails fast when random_gamma=True instead of silently disabling the augmentation. PT checkpoint conversion remains valid because SavedModel export is inference-only and explicitly normalizes the serialized training augmentation to random_gamma=False.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/tf2/descriptor/dpa4.py
njzjz: Fixed in commit aa1ed8926. Every mapped DPA4 parameter is now promoted to a tracked tf.Variable, with the descriptor-level freeze flag propagated to descendants that lack a local flag. The regression verifies zero trainable variables and full checkpoint restoration between differently seeded frozen descriptors.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/tf2/descriptor/dpa4.py
njzjz: Fixed in commit aa1ed8926. DynamicRadialDegreeMixer no longer compares symbolic shape tuples in Python, the TF2 wrapper adds a runtime shape assertion, and the descriptor no longer branches on a symbolic edge count. A compiled optimizer step now succeeds for two and three atom batches through the default degree-channel path.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/tf2/model/model.py
njzjz: Fixed in commit aa1ed8926. The DPA4 factory whitelist now accepts the schema-supported property fitting type and routes it through the existing standard TF2 stack. A normalized DPA4 property configuration now constructs a PropertyModel in the regression.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

wanghan-iapcm (dismissed):
Both threads I raised are addressed and verified against the current HEAD (fix in 4328329eb):

- Silently-frozen params (`tf2/dpa4.py`): the promotion maps now include `FrameContract.weight`, `FrameExpand.weight`, and `SeZMInteractionBlock.adam_ffn_layer_scales`, and `test_optional_dpa4_weights_are_tf2_trainable_variables` enables `layer_scale` + `message_node_so3` and asserts these leaves are tracked in `descriptor.trainable_variables` — a test that fails on the pre-fix maps. There is also a real optimizer-step gradient test and a deserialize trackable-refresh test.
- Test coverage (`tf2/common.py` / trainer glue): `source/tests/tf2/` now covers the deserialize -> trackable-refresh round-trip, the `get_sezm_model` guard/mismatch paths, and `_translate_model_ret_to_loss_dict`.

Approving on that basis.

wanghan-iapcm (dismissed):
Both threads I raised are addressed and verified against the current HEAD (fix in 4328329eb): the tf2 promotion maps now include FrameContract.weight, FrameExpand.weight, and SeZMInteractionBlock.adam_ffn_layer_scales, and test_optional_dpa4_weights_are_tf2_trainable_variables enables layer_scale + message_node_so3 and asserts these leaves are tracked in descriptor.trainable_variables (a test that fails on the pre-fix maps), with a real optimizer-step gradient test and a deserialize trackable-refresh test; source/tests/tf2/ also now covers the deserialize -> trackable-refresh round-trip, the get_sezm_model guard/mismatch paths, and _translate_model_ret_to_loss_dict. Approving on that basis.

wanghan-iapcm (approved):
Both threads I raised are addressed and verified against the current HEAD (fix in 4328329eb): the tf2 promotion maps now include FrameContract.weight, FrameExpand.weight, and SeZMInteractionBlock.adam_ffn_layer_scales, and test_optional_dpa4_weights_are_tf2_trainable_variables enables layer_scale + message_node_so3 and asserts these leaves are tracked in descriptor.trainable_variables (a test that fails on the pre-fix maps), with a real optimizer-step gradient test and a deserialize trackable-refresh test; source/tests/tf2/ also now covers the deserialize -> trackable-refresh round-trip, the get_sezm_model guard/mismatch paths, and _translate_model_ret_to_loss_dict. Approving on that basis.

deepmd/dpmodel/array_api.py
```diff
@@ -270,6 +282,32 @@ def xp_maximum_at(x: Array, indices: Array, values: Array) -> Array:
         return torch.scatter_reduce(
             x, 0, index, values, reduce="amax", include_self=True
         )
+    elif getattr(xp, "__name__", "") == "deepmd._vendors.ndtensorflow":
+        import tensorflow as tf
+
+        x_tensor = x.unwrap()
+        indices_tensor = tf.reshape(tf.cast(indices.unwrap(), tf.int64), (-1,))
+        values_tensor = values.unwrap()
+        reduced = tf.math.unsorted_segment_max(
```
OutisLi, line 295: [P1] Preserve the neutral -inf maximum for all-masked segments

This TensorFlow reduction does not implement the required xp_maximum_at semantics when every update for a touched segment is -inf. tf.math.unsorted_segment_max returns the dtype's lowest finite value (for float64, approximately -1.797693e308), rather than -inf. DPA4 attention deliberately supplies -inf for every masked/padded slot; because the returned sentinel is finite, attention.py treats it as a real group maximum. The subsequent exp(logit - group_max) overflows to inf, and multiplying by the zero edge weight produces 0 * inf = NaN.

On the current HEAD, an actual TF2 DescrptDPA4 with one atom, sel=4, n_blocks=1, and nlist shaped (1, 1, 4) filled with -1 returns an all-NaN descriptor in both eager mode and tf.function; the corresponding model energy, force, and gradients are NaN. The same input is finite with n_blocks=0 or attention disabled, which localizes the failure to this all-masked attention path. Any isolated atom, or any center whose neighbors are all removed by exclusions/cutoff, can therefore poison the total energy.

Please preserve the true neutral maximum or explicitly identify segments with no positive-weight edge before exponentiation, and add isolated-center regressions for eager, tf.function, and XLA energy/force paths.

OutisLi (changes requested):
TF2 DPA4 produces NaN energy and forces for a valid center with no effective neighbors. See the inline P1 finding.

deepmd/pt/model/task/sezm_ener.py
```diff
@@ -270,7 +277,7 @@ def __init__(
                     activation_function=self.activation_function,
                     precision=self.precision,
                     seed=child_seed(seed, layer_idx),
-                    trainable=trainable,
+                    trainable=self.trainable[layer_idx],
```
OutisLi: [P1] Apply the restored trainability policy to PyTorch parameters

Passing `trainable` here does not set `Parameter.requires_grad`: `GLULayer`/`MLPLayer` only stores the flag while constructing `nn.Parameter`s with the default `requires_grad=True`. This commit also removes the former final loop that explicitly applied `requires_grad`. As a result, `GLUFittingNet.deserialize` rebuilds every parameter as trainable, and `SeZMEnergyFittingNet.deserialize` replaces the correctly frozen collection with that rebuilt one.

On the current HEAD, serializing and deserializing a `SeZMEnergyFittingNet(..., trainable=False, neuron=[4])` changes the three fitting flags from `[False, False, False]` to `[True, True, True]`; an SGD step then updates all three parameters. This silently unfreezes a frozen fitting when a PT model/checkpoint is restored.

Please apply each layer's policy to `requires_grad` after construction/deserialization and add PT serialize-to-deserialize plus optimizer-step regressions for `trainable=False` (and the per-layer mask).

OutisLi (changes requested):
The PyTorch fitting trainability policy is not preserved across serialization/deserialization: a fitting saved with `trainable=False` is silently restored as trainable. This can change resumed-training behavior, so it needs to be fixed before merge.

njzjz:

- @iProzd — 10 reviews on exact changed files (`deepmd/dpmodel/array_api.py`, `deepmd/dpmodel/descriptor/dpa4.py`, `deepmd/dpmodel/fitting/dpa4_ener.py`, `deepmd/tf2/train/trainer.py`, `source/tests/tf2/test_training.py`).

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

deepmd/dpmodel/array_api.py
```diff
@@ -201,6 +201,18 @@ def xp_add_at(x: Array, indices: Array, values: Array) -> Array:
         import torch
 
         return torch.index_add(x, 0, indices, values)
+    elif getattr(xp, "__name__", "") == "deepmd._vendors.ndtensorflow":
+        import tensorflow as tf
+
+        x_tensor = x.unwrap()
+        indices_tensor = tf.reshape(tf.cast(indices.unwrap(), tf.int64), (-1, 1))
+        values_tensor = values.unwrap()
+        updates = tf.scatter_nd(
```
OutisLi: [P1] Handle zero-sized destinations in the TF2 `xp_add_at` path

`tf.scatter_nd` rejects an output shape with a zero-sized leading axis even when both `indices` and `updates` are empty. This is reachable from a valid DPA4 descriptor call with no edges. On this head, a one-atom descriptor with `n_blocks=1`, `edge_index.shape == (2, 0)`, `edge_vec.shape == (0, 3)`, and `edge_mask.shape == (0,)` fails here in eager mode with `InvalidArgumentError: Indices and updates specified for empty output shape`; the same DPModel call returns a finite descriptor. The dense path also fails for `nlist.shape == (1, 1, 0)` now that the block stage no longer skips an empty edge axis.

Please make this reduction graph-safely preserve `x` when its destination axis is empty (or use an equivalent TensorFlow primitive that supports the empty case), and add real `DescrptDPA4.call_with_edges(E=0)` eager and `tf.function` regressions, ideally plus the zero-width dense-neighbor case. The current direct `xp_maximum_at` regression does not exercise this failure.

OutisLi (changes requested):
TF2 DPA4 still crashes on the valid `E=0` case because the new `xp_add_at` implementation cannot scatter into a zero-sized destination. See the inline P1 finding.

deepmd/dpmodel/array_api.py
njzjz: Fixed in `48c5f1358`, and rebased onto master.

Reproduced exactly as described: a one-atom `DescrptDPA4` with `n_blocks=1` and `edge_index.shape == (2, 0)` fails in `xp_add_at` at `deepmd/dpmodel/array_api.py`, from `_scatter_rank_kernel` in `dpa4_nn/so2.py` — the destination there is `(n_ranks, E, channels)`, so `E == 0` makes it zero-sized while the rank indices stay non-empty. `tf.scatter_nd` then raises `InvalidArgumentError: Indices and updates specified for empty output shape` in eager mode and the equivalent `ValueError` at graph-construction time. The zero-width dense case (`nlist.shape == (1, 1, 0)`) hits the same line.

Rather than special-casing the empty destination, the TensorFlow branch now reduces with `tf.math.unsorted_segment_sum` over axis 0. It accumulates repeated indices exactly like `scatter_nd` (and like `np.add.at`), and accepts a zero-sized destination in both eager and graph mode.

Regressions added to `source/tests/tf2/test_dpa4.py`: a real `DescrptDPA4.call_with_edges` with `E=0` and the dense `(1, 1, 0)` neighbor list, each run eagerly and again under `tf.function(reduce_retracing=True)`, asserting finite output and eager/traced agreement. Both pass along with the rest of the focused TF2 suite (8 passed) and `source/tests/consistent/test_array_api.py` (31 passed).

deepmd/tf2/descriptor/dpa4.py
```diff
@@ -0,0 +1,321 @@
...
+
+@BaseDescriptor.register("SeZM")
+@BaseDescriptor.register("sezm")
+@BaseDescriptor.register("DPA4")
+@BaseDescriptor.register("dpa4")
+@tf2_module
+class DescrptDPA4(DescrptDPA4DP):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        if self.random_gamma:
```
OutisLi: [P1] Support `random_gamma=True` in TF2

DPA4 models use `random_gamma=True`, which is also the descriptor default. Rejecting it here makes standard existing DPA4 configurations and checkpoints impossible to construct or deserialize with TF2. This PR must support `random_gamma=True`; fail-fast is not an acceptable backend limitation.

OutisLi (changes requested):
TF2 DPA4 must support random_gamma=True; rejecting the standard DPA4 configuration is blocking.

deepmd/tf2/descriptor/dpa4.py
njzjz: Fixed in 7e8b2bc46. random_gamma=True is now supported with graph-safe TensorFlow randomness, and training/evaluation use separate traced paths so the randomized behavior does not leak into evaluation. All 10 targeted TF2 DPA4 tests pass.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

OutisLi (approved):

PR #5753, opened 2026-07-08
Using Git-diff to reduce the time cost of CI/CD

Gongxl972:
Modify the test_python.ymlworkflow and add several Python helper scripts under .github/scripts/.
The new scripts discover all test files, analyze test-to-source dependencies, and generate a .global_test_mapping.jsoncache file.
Subsequent test runs will only execute tests mapped to changed source files, reducing CI runtime.

## Summary by CodeRabbit

* **New Features**
  * Added smarter test selection in CI so changed files can trigger only the relevant test groups instead of always running everything.
  * Introduced a faster test pipeline with split test groups, cached timing data, and improved handling for incremental runs.

* **Bug Fixes**
  * Improved reliability when test-change data is missing or incomplete by falling back to a full test run.
  * Added better handling for changed source files and workflow-related updates to ensure the right tests still run.


.github/scripts/detect_affected_tests.py
```diff
@@ -0,0 +1,190 @@
...
+        # Case B: Changed file is a workflow/config file - trigger full test
+        if cf.startswith(".github/workflows/") or cf.startswith(".github/scripts/"):
+            debug(f"Changed workflow/script file {cf}, triggering full test")
+            print("selected_paths=source/tests")
+            print("skip_all=false")
+            print("need_full_test=true")
+            sys.exit(0)
+        
+        # Case C: Changed file is a source file, look for dependent tests
+        if cf.startswith("source/deepmd/"):
```
iProzd, line 155: **[blocking] Wrong path prefix — real source changes never match, so normal PRs skip all tests.**

The package lives at repo root `deepmd/` (there is no `source/deepmd/`; the workflow itself runs `pytest --cov=deepmd`). But `git diff --name-only` emits source changes as `deepmd/...`, while this check tests `source/deepmd/`. A PR that only edits `deepmd/**` therefore matches none of Case A/B/C, both `affected_tests` and `new_source_files_without_tests` stay empty, and we hit the `else` → `skip_all=true` → the entire Python suite is skipped and `pass` goes green. The same wrong prefix is in `build-global-mapping.py` (`dep = "source/" + module_path`) and in the workflow's `new_source_files` / syntax-check greps (`^source/deepmd/`). Note the asymmetry: tests are correctly `source/tests/`, only the package prefix is wrong. Fix: use `deepmd/` everywhere for the package.

.github/scripts/detect_affected_tests.py
```diff
@@ -0,0 +1,190 @@
...
+        print("skip_all=false")
+        print("need_full_test=true")
+    elif affected_tests:
+        # Found affected tests -> run only those
+        selected = " ".join(sorted(affected_tests))
+        debug(f"Affected tests found: {selected}")
+        print(f"selected_paths={selected}")
+        print("skip_all=false")
+        print("need_full_test=false")
+    else:
```
iProzd, line 182: **[blocking] Safe-default is inverted (fail-open).** When no rule matches, the selector currently *skips* tests. A merge gate must fail-closed: anything not provably irrelevant → FULL. Please remove the free `skip_all=true` path; only allow skipping for an explicit docs-only allowlist (e.g. all changed files match `*.md`). Every other bucket — unmatched paths, non-Python, data/fixtures, build/CI config, new files, deletes/renames, missing or unparsable mapping — must set `need_full_test=true`.

.github/scripts/detect_affected_tests.py
```diff
@@ -0,0 +1,190 @@
...
+        debug(f"Processing changed file: {cf}")
+        
+        # Case A: Changed file is a test file itself
+        if cf.startswith("source/tests/"):
+            debug(f"Changed file is a test file, adding to affected tests: {cf}")
+            affected_tests.add(cf)
+            continue
+        
+        # Case B: Changed file is a workflow/config file - trigger full test
+        if cf.startswith(".github/workflows/") or cf.startswith(".github/scripts/"):
```
iProzd, line 147: **[blocking] Non-Python / C++ / build changes are not covered.** Besides the workflow/scripts handled here, changes under `source/{lib,op,api_cc,api_c,lmp}/**`, `pyproject.toml`, `CMakeLists.txt`, `source/install/**`, and test data/fixtures (`*.pth`, `*.pbtxt`, `*.npy`) match no case and fall through to `skip_all=true` — yet many Python tests validate behavior *through* the compiled ops. These need an analogous force-FULL branch. (Covers CodeRabbit's `pyproject.toml`/`CMakeLists.txt` point too.)

.github/workflows/test_python.yml
```diff
@@ -7,147 +9,535 @@ on:
...
+          
+          export PYTHONPATH=$GITHUB_WORKSPACE:$PYTHONPATH
+          echo "PYTHONPATH configured: $PYTHONPATH"
+          
+          pytest -vv --cov=deepmd $SELECTED_PATHS \
+            --splits 12 --group ${{ matrix.group }} \
+            --store-durations --clean-durations \
+            --durations-path=.test_durations \
+            --splitting-algorithm least_duration \
+            --disable-warnings || true
```
iProzd, line 455: **[blocking] Two issues here:**
1. `|| true` makes the step always exit 0, so failing tests never fail the job and `alls-green` reports green — CI stops gating. The follow-up steps already use `if: always()`, so drop `|| true`.
2. Selection should be gated to PRs only. Wrap the detect/incremental path in `if: github.event_name == 'pull_request'`, and force `selected_paths=source/tests` (FULL) for `merge_group` and `push`. That keeps the merge-queue gate authoritative while still giving fast PR feedback.

iProzd (changes requested):
Thanks for tackling CI time — the three-stage structure (early-detection → mapping → sharded run) is a solid exploration. Before this can land, though, I think the design needs to be re-anchored on one principle:

**Test selection must never gate a merge; it may only speed up PR-iteration feedback.**

This repo already uses a GitHub merge queue (`merge_group` trigger). That gives us the right place to be strict: keep the FULL suite as the required check at `merge_group` (and on `push` to master), and only apply incremental selection on `pull_request`. With that split, an imperfect selector can at most cost a contributor one extra iteration — it can never let a regression reach master. As written, the selection logic also runs for non-PR events, which weakens the final gate itself.

Two more principles that must hold regardless of the selector's cleverness:
1. **Fail-closed:** anything the selector cannot *prove* is irrelevant → run FULL. Today the default is inverted (unmatched changes fall through to `skip_all=true`).
2. **The dependency signal must be sound.** Static top-level `import deepmd.*` scanning is an *under-approximation*: it misses transitive deps, registry/string-key dispatch (descriptors/models/fittings resolved from JSON keys), cross-backend `consistent/`+`universal/` parity tests, and all non-Python (C++/CUDA/LAMMPS/data) changes. For a numerically-sensitive multi-backend library this will silently skip real regressions. Prefer coverage-trace selection (`pytest-testmon`, which this fork is even named after) or a conservative path→backend-directory mapping over per-file import scanning.

Blocking items are in the inline comments (wrong `source/deepmd/` path prefix, `skip_all` safe-default inversion, `|| true` swallowing failures, non-Python changes not covered). Also worth doing first: the biggest, zero-risk win is caching/reusing the compiled build (build one wheel per Python version instead of recompiling the C++ extension in all 24 shards) — that likely saves more wall-clock than selection, with no correctness trade-off. Happy to pair on the merge-queue split.

PR #5758, opened 2026-07-09
perf(dpa1): optimize graph CUDA inference and deployment

OutisLi:
DPA1 graph inference previously decomposed descriptor, fitting, and
force/virial work into many generic tensor operations, retained large
edge-scale autograd state, and lacked a shared sparse topology contract across
Python export, the C++ runtime, and LAMMPS. Build an inference-oriented graph
pipeline that keeps eligible attention-free DPA1 models and their neighbor data
on the GPU while retaining explicit fallbacks for unsupported configurations.

- Add inference-only CUDA operators for attention-free concat and strip DPA1.
  The uncompressed path fuses environment construction, the three-layer
  embedding network, type handling, moment reduction, Gram contraction, and
  analytical edge-gradient backward. The geometrically compressed strip path
  fuses quintic table interpolation, type-pair gating, moments, descriptors,
  and analytical gradients. It supports masked edges, canonical or permutation
  CSR, int32 or int64 addressing, width buckets from 8 through 256,
  non-bucket padding and slicing, and axis widths up to 16.

- Add descriptor-agnostic fused energy fitting and force/virial operators.
  Run eligible fitting networks through pedantic FP32 cuBLAS GEMMs with TF32
  disabled and fused bias, activation, timestep, residual, and derivative
  epilogues, while retaining atomic and per-frame energies in FP64. Reduce both
  edge incidences with one warp per node and no global floating-point atomics,
  then accumulate frame virials through FP64 partial reductions without
  materializing an (E, 9) outer-product tensor.

- Define cumulative DP_CUDA_INFER levels. Level 1 uses separately registered
  descriptor, fitting, and CSR force operators with first-order autograd.
  Level 2 returns force and virial as values, using one opaque end-to-end
  operator for eligible uncompressed DPA1 and an explicit custom-operator chain
  for compressed DPA1. This removes the inference autograd tape, suppresses the
  unused rotation output, avoids retaining the descriptor only for shape
  metadata, and shortens saved-state lifetimes for memory reuse.

- Add make_fx- and export-composable Triton kernels for the PyTorch
  se-family environment matrix and the dense and graph DPA1 environment
  convolutions. Their closed-form first backward preserves the force path while
  avoiding decomposed gather, switch, normalization, gating, and segment
  reduction chains. DP_TRITON_INFER=2 resolves per-GPU launch tables and
  freeze-time tuning fills uncovered model shapes; level 3 optionally enables
  the compensated FP16x3 final embedding GEMM. The default CUDA graph path
  remains FP32.

- Select compressed-kernel resource policy at first uncaptured use from
  balanced or occupancy-oriented 128/256-thread launches, cache the result by
  device, direction, descriptor shape, topology, index width, and workload
  class, and retain architecture-specific fallbacks for small inputs, capture,
  or tuning failures. Build the native operators only with a CUDA-enabled
  PyTorch and include a lowest-supported portable PTX target alongside native
  code.

- Extend NeighborGraph with keyword-only destination/source CSR orders,
  int64 row pointers, and an explicit destination_sorted property without
  changing its original positional constructor. Generic builders preserve
  payload order and omit CSR by default; consumers opt into CSR or stable
  destination-major canonicalization, which moves masked guards to the suffix
  and makes destination order the identity. Edge masks remain authoritative
  inside every row, and export validates permutations, row membership, bounds,
  and canonical identity before tracing.

- Make graph-form AOTInductor deployment select the optimized path
  automatically for eligible .pt2 conversion and export compressed graph
  models directly. Keep frame, node, and edge axes dynamic; preserve custom
  operators through export; record the graph edge-vector dtype in metadata;
  and make per-atom virial part of the graph artifact contract required by the
  Kokkos consumer. Eligible compressed artifacts accept FP32 edge geometry
  directly, while generic and uncompressed graph artifacts retain the FP64
  geometry ABI. Use int64 Inductor indexing and bounded Triton launch tiling for
  large dynamic graphs.

- Extend DeepEval and the C++ API to consume the same canonical graph ABI.
  Add device-edge capability and dtype queries, FP32 and FP64 edge-vector
  overloads, runtime frame and atomic parameters, total-versus-owned node
  counts, and optional communication metadata. Host ingestion canonicalizes
  arbitrary payloads, while the device path constructs destination identity
  CSR and source order with histogram, prefix-sum, and counting scatter.

- Add pair_style deepmd/kk for device-resident edge and graph .pt2
  inference. Build compact model-cutoff edges from the Kokkos full neighbor
  list with count, scan, and fill passes; compact NULL-mapped atom types; emit
  FP32 or FP64 vectors according to artifact metadata; and scatter model-node
  outputs back on device. Single-rank execution folds periodic images onto
  local owners, whereas domain decomposition retains local-plus-halo nodes and
  explicitly reverse-communicates force and centroid per-atom virial through
  either device-aware or host-staged communication. Message-passing edge
  artifacts use their with-comm forward, including empty-rank phantom inputs;
  message-passing graph artifacts fail fast because that multi-rank ABI is not
  supported.

- Preserve graph correctness at ownership and masking boundaries. Exclude halo
  fitting outputs from energy, zero halo atomic parameters, retain virtual-atom
  and pair-exclusion masks in fused level 2, handle zero-node graphs, and reduce
  frame virial from node virials instead of a contended edge scatter. Validate
  and broadcast multi-frame fparam/aparam layouts, fix wide atomic-parameter
  allocation when filtering virtual atoms, and avoid copying incompletely
  constructed LAMMPS model members. Reduce graph-builder overhead with bounded
  GPU neighbor-capacity estimates and direct length-three periodic-shift
  reductions.

- Require regeneration of graph-form .pt2 artifacts because their positional
  ABI now includes n_local and both CSR views. Compressed graph export is
  limited to FP32, geometrically compressed, attention-free strip models
  without excluded type pairs, with output width at most 256 and
  axis_neuron <= 16. The uncompressed CUDA descriptor requires three FP32
  embedding layers whose widths stay equal or double within its compiled
  bounds. Other configurations continue through Triton, reference graph, or
  nlist lowering. deepmd/kk additionally requires a GPU edge/graph artifact,
  one model, and a valid atom map.

- Add regression coverage for CUDA and Triton forward/gradient parity,
  compressed and uncompressed end-to-end energy, force, global virial, atomic
  energy, and atom virial; smooth and non-smooth one- and two-sided type gates;
  residual and activation variants; non-power-of-two widths; int32/int64
  addressing; masked cached edges; canonical and permuted CSR; ownership;
  zero-node and many-frame reductions; make_fx, torch.compile, and dynamic
  graph export. Extend DeepEval and C++ tests for dynamic edge counts,
  multi-rank graph ingestion, parameter validation and broadcasting, atomic
  outputs, CSR construction, and virtual-atom parameter preservation.


source/api_cc/src/DeepPotPTExpt.cc
```diff
@@ -1866,35 +1948,116 @@ void DeepPotPTExpt::compute_edges_gpu(double* d_atom_energy,
...
+      using BuildGraphCSR =
+          std::tuple<torch::Tensor, torch::Tensor, torch::Tensor,
+                     torch::Tensor>(torch::Tensor, c10::SymInt, c10::SymInt);
+      static const auto build_graph_csr =
+          c10::Dispatcher::singleton()
+              .findSchemaOrThrow("deepmd::build_graph_csr", "")
+              .typed<BuildGraphCSR>();
+      std::tie(graph_pack.destination_order, graph_pack.destination_row_ptr,
+               graph_pack.source_row_ptr, graph_pack.source_order) =
+          build_graph_csr.call(edge_index, c10::SymInt(nnode),
```
wanghan-iapcm: **Non-blocking (robustness, not a current bug):** this GPU-edges graph branch builds the CSR from the raw `edge_index` via `build_graph_csr` and feeds the resulting `destination_order` into `run_model_graph`. That is correct today, but it relies on an *unenforced* invariant. For a geo-compressed model the graph kernel is exported with `destination_sorted=True`, so `edge_at_csr_position<Canonical=true>` (in `source/op/pt/dpa1_graph_compress.cu`) ignores `destination_order` and indexes the edge payload by CSR position directly — i.e. it assumes the incoming `edge_index`/`edge_vec` are already destination-major.

The host `compute` path guarantees that by calling `canonicalizeGraphPayload(...)` first (see L875); this path does not, and instead depends on the LAMMPS-Kokkos `build_edges_device` happening to emit edges in destination-major order (which it currently does, by construction of `edge_offset` / `loc2model`). So the correctness of the geo-compressed GPU path is a silent coupling between the Kokkos edge layout and the baked `Canonical=true` kernel, with nothing asserting it: a future change to the Kokkos neighbor-edge layout — or reuse of this path by another, non-sorted edge producer — would silently compute the descriptor from mismatched edges with no error (the in-tree C++ test only exercises the canonicalizing host path).

Suggest either calling `canonicalizeGraphPayload(...)` here as the host path does, or adding a debug assertion / comment documenting that `edge_index` must be destination-major on entry to this branch.

deepmd/kernels/autotune.py
```diff
@@ -0,0 +1,87 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
```
wanghan-iapcm, line 1: **Placement of `deepmd/kernels/` — it is torch-only; suggest nesting it under `pt_expt`.**

This package is not backend-agnostic: every module binds to torch — the Triton kernels are registered via `torch.library.triton_op`/`wrap_triton`, the CUDA kernels are exposed as `torch.ops.deepmd.*` — so it can never be consumed by `dpmodel`/`tf`/`jax`/`pd`. Today it is imported by both torch backends: `pt` (`pt/model/descriptor/{env_mat,se_atten}.py`, Triton path) and `pt_expt` (`pt_expt/descriptor/dpa1.py`, CUDA path). But `pt` (torch.jit) is slated for removal once jit leaves torch, which leaves `pt_expt` as the durable sole owner — and the kernels are already designed for its regime (`triton_op`/`torch.ops` that survive `torch.export`/`make_fx`).

As a bare top-level package, sibling to the backend-agnostic `deepmd/dpmodel/`, it reads as shared/neutral infrastructure, which it isn't. Suggest moving it under **`deepmd/pt_expt/kernels/`**:
- `pt_expt` is the long-term owner and the design target;
- during `pt`'s deprecation window, `pt` imports the kernels *from* `pt_expt` (dying → living direction, self-cleaning when `pt` is removed) — no cycle, since these modules only pull torch/triton + `dpmodel.utils.safe_gradient`;
- the C++ half can stay in `source/op/pt/` ("PyTorch ops").

Invariant to keep either way: `deepmd/dpmodel/**` must never import `deepmd.kernels` (currently clean — good).

OutisLi: But pt can also use this, so I believe a separate folder is a good choice. And there are also DPA4 kernels in this folder, so perhaps a relocation could be done through another separate PR?

deepmd/dpmodel/utils/neighbor_graph/graph.py
```diff
@@ -50,6 +54,16 @@ class NeighborGraph:
     """(2, A) int  [edge_a, edge_b] sharing a center; into [0, E). None if no angles."""
     angle_mask: Array | None = None
     """(A,) bool  real vs padding on the angle axis. None if no angles."""
+    destination_order: Array | None = field(default=None, kw_only=True)
+    """(E,) edge permutation grouped by destination; same dtype as edge_index."""
+    destination_row_ptr: Array | None = field(default=None, kw_only=True)
+    """(N + 1,) int64 offsets into ``destination_order``."""
+    source_row_ptr: Array | None = field(default=None, kw_only=True)
+    """(N + 1,) int64 CSR offsets into ``source_order``."""
+    source_order: Array | None = field(default=None, kw_only=True)
```
wanghan-iapcm, line 64: **Non-blocking (readability/consistency):** the field order here is a palindrome — `destination_order, destination_row_ptr, source_row_ptr, source_order`, i.e. `order, ptr, ptr, order`. The destination pair reads `(order, ptr)` but the source pair reads `(ptr, order)` — mirrored rather than parallel, which is easy to misread.

Since each `row_ptr` is documented as "offsets into `<...>_order`", the natural, parallel layout is *value array then its offsets* for both:

```
destination_order     # (E,)
destination_row_ptr   # (N+1,) offsets into destination_order
source_order          # (E,)
source_row_ptr        # (N+1,) offsets into source_order
```

The mirror currently propagates to every downstream signature/return/kwargs list — `csr.py::build_edge_csr` (Returns docstring + the return tuple) and `canonicalize_neighbor_graph` (the unpack and the re-construct), plus the torch twin `pt_expt/utils/graph_csr.py`. Worth applying the parallel order uniformly across the dataclass fields and all those call sites in one sweep. (If you'd rather follow CSR convention with `indptr` first, `row_ptr, order` for *both* pairs is equally fine — the point is just to stop mirroring.)

OutisLi: Keeping the current order: these dataclass fields are keyword-only, while the existing CSR tuple, exported operator schemas, and C++ deployment signatures consistently use `destination_order, destination_row_ptr, source_row_ptr, source_order`. Reordering every layer would add broad interface churn without changing behavior or making misuse harder; the named fields remain unambiguous.

OutisLi: Keeping the current order: these dataclass fields are keyword-only, while the existing CSR tuple, exported operator schemas, and C++ deployment signatures consistently use `destination_order, destination_row_ptr, source_row_ptr, source_order`. Reordering every layer would add broad interface churn without changing behavior or making misuse harder; the named fields remain unambiguous.

deepmd/entrypoints/convert_backend.py
```diff
@@ -36,19 +36,23 @@ def convert_backend(
...
+    # signature so scripts stay backend-agnostic. ``lower_kind="auto"`` lets the
+    # pt_expt hook pick the deploy-optimal lowering (the graph lower with its
+    # fused operator for a graph-eligible .pt2, the dense nlist lower otherwise);
+    # ``do_atomic_virial`` forwards ``--atomic-virial``.
     import inspect
 
     sig = inspect.signature(out_hook)
+    hook_kwargs: dict[str, Any] = {}
+    if "lower_kind" in sig.parameters:
+        hook_kwargs["lower_kind"] = "auto"
```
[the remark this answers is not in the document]
OutisLi: Restored automatic selection intentionally after clarifying the deployment contract for this PR: `convert-backend` must keep the existing command-free DPA1 workflow and select graph/canonical lowering for eligible `.pt2` artifacts. The conversion regression test now asserts `lower_kind="auto"`, and S/M/L production checkpoints were exported with the original command and verified as graph artifacts.

deepmd/pt_expt/model/edge_transform_output.py
wanghan-iapcm: better to put standard force and virial computation in an else case, in contrast to the `if` conditions.

wanghan-iapcm (changes requested):

deepmd/dpmodel/utils/neighbor_graph/graph.py
wanghan-iapcm: Reopening — I don't think the rationale holds up, on three points:

**(a) keyword-only addresses safety, not readability.** Making the fields keyword-only removes the risk of *positional misuse* (you can't accidentally pass `order` where `ptr` is expected). But the concern here isn't misuse — it's that the mirrored layout `order, ptr, ptr, order` is anti-intuitive every time a human reads or extends the structure. `keyword-only` does nothing for that; it answers a different objection.

**(b) a full sweep preserves consistency *and* fixes intuitiveness.** "The order is consistent across the CSR tuple, the exported op schema, and the C++ signatures" is an argument against a *partial* reorder that would leave layers mismatched — not against reordering. The proposal is a full sweep to the parallel layout `destination_order, destination_row_ptr, source_order, source_row_ptr` (value array then its offsets, for both pairs) across every site at once. After the sweep, consistency is preserved and the layout is intuitive. (If you'd rather follow CSR convention, `row_ptr, order` for *both* pairs is equally fine — the point is just to stop mirroring.)

**(c) the churn is cheap here.** The sites are bounded and mechanical (dataclass, `csr.py::build_edge_csr`, `canonicalize_neighbor_graph`, `pt_expt/utils/graph_csr.py`, the op schema, and the C++ `std::tie` unpack). pt_expt carries no serialized back-compat requirement, so reordering even the exported operator schema is free — there are no existing artifacts to keep compatible. And a desync can't ship silently: a mismatched order produces a wrong descriptor, so any consistency/inference test fails loudly. The only non-mechanical part is keeping the torch op schema and the C++ unpack in sync, which is ~2 sites and test-guarded.

Given that, I'd prefer the parallel order. Not blocking if you'd still rather keep it, but the readability objection isn't answered by the current reasoning.

deepmd/kernels/autotune.py
wanghan-iapcm: Deferring the relocation to a separate PR is fine by me — bundling it with DPA4 kernels here shouldn't block this perf PR. But to keep "another separate PR" from quietly becoming "never", please open a tracking issue for it and link it here.

Two things to capture in that issue so the eventual move is unambiguous:
- **Target:** `deepmd/kernels/` moves under `deepmd/pt_expt/kernels/` (pt_expt is the long-term owner and the design target). It reads as backend-neutral infrastructure today by sitting as a top-level sibling of `deepmd/dpmodel/`, but every module hard-binds torch (`torch.library.triton_op`/`wrap_triton`, `torch.ops.deepmd.*`), so it can never serve `dpmodel`/`tf`/`jax`/`pd`. "Both torch backends use it" is exactly the point — `pt` (torch.jit) and `pt_expt` (torch.export) are both torch.
- **Direction:** during `pt`'s deprecation window, `pt` imports the kernels *from* `pt_expt` (dying → living, self-cleaning when `pt` is removed), not the reverse — no cycle, since these modules only pull torch/triton + `dpmodel.utils.safe_gradient`.

Invariant to preserve either way: `deepmd/dpmodel/**` must never import `deepmd.kernels` (currently clean).

Could you file that issue and drop the link here so this thread can be closed against it?

OutisLi: Tracking issue opened with the requested target, dependency direction, and dpmodel invariant: https://github.com/deepmodeling/deepmd-kit/issues/5782

deepmd/dpmodel/utils/neighbor_graph/graph.py
OutisLi: Implemented the full sweep in 254015486: the generic graph ABI now consistently uses `destination_order, destination_row_ptr, source_order, source_row_ptr` across the dataclass, CSR builders, export/training/DeepEval inputs, CUDA operator schemas, C++ AOTI bridge, and tests. The compact canonical ABI is unchanged because it has no destination permutation and therefore no mirrored pair.

deepmd/pt_expt/model/edge_transform_output.py
OutisLi: Updated in 254015486: the standard array-API force/virial computation is now the explicit `else` branch paired with the fused path, followed by one shared return.

source/api_cc/src/DeepPot.cc
```diff
@@ -582,12 +582,116 @@ void DeepPotBackend::compute_edges_gpu(double* d_atom_energy,
...
+                                       double* d_force,
+                                       double* d_atom_virial,
+                                       const double* d_coord,
+                                       const int* d_atype,
+                                       const int* d_edge_index,
+                                       const double* d_edge_vec,
+                                       const int nloc,
+                                       const int nedge,
+                                       const std::vector<double>& fparam,
+                                       const std::vector<double>& aparam,
```
njzjz, line 600: I am wondering why these two arguments are vectors, but others are pointers. What is the design here?

doc/install/install-lammps.md
```diff
@@ -80,6 +80,16 @@ make -j4
...
+To build the GPU-resident `pair_style deepmd/kk`, enable both the LAMMPS
+Kokkos package and the DeePMD C++-API integration:
+
+```bash
+cmake -D PKG_KOKKOS=ON -D DEEPMD_LAMMPS_KOKKOS=ON \
+    -D LAMMPS_INSTALL_RPATH=ON -D BUILD_SHARED_LIBS=yes \
+    -D CMAKE_INSTALL_PREFIX=${deepmd_root} \
+    -D CMAKE_PREFIX_PATH=${deepmd_root} ../cmake
+```
+
```
njzjz: It seems to me that only the built-in mode is supported, which we don't use for distribution.

source/lmp/builtin.cmake
njzjz, line 1: It seems to me that only the built-in mode is supported, which we don't use for distribution.

source/lmp/builtin.cmake
```diff
@@ -68,7 +69,29 @@ target_sources(
...
+if(DEEPMD_LAMMPS_KOKKOS)
+  if(NOT PKG_KOKKOS)
+    message(FATAL_ERROR "DEEPMD_LAMMPS_KOKKOS requires PKG_KOKKOS")
+  endif()
+  if(NOT TARGET DeePMD::deepmd_cc)
+    message(
+      FATAL_ERROR
+        "DEEPMD_LAMMPS_KOKKOS requires the DeePMD C++ interface target")
+  endif()
+  target_link_libraries(lammps PUBLIC DeePMD::deepmd_cc)
```
njzjz: I am wondering why not to expose a C API. It's very to do it using the agent. The C++ API was designed to be deprecated due to a lack of forward compatibility. `DP_USE_CXX_API` is just kept for compatibility.

source/op/pt/CMakeLists.txt
```diff
@@ -1,8 +1,64 @@
...
+# the Python dispatch falls back to the reference path (see
+# deepmd.kernels.cuda.*.op_available). The CUDA language is scoped per
+# directory, so it must be enabled here: the sibling GPU library turns it on
+# only within its own subtree, which does not cover this target.
+if(USE_CUDA_TOOLKIT AND DEEPMD_TORCH_HAS_CUDA)
+  find_package(CUDAToolkit REQUIRED)
+  if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
+    # CUDA 12.9 and 13.x CCCL fail to compile CUB/Thrust with -arch=all.
+    if(CUDAToolkit_VERSION VERSION_GREATER_EQUAL "12.9" AND CUDAToolkit_VERSION
+                                                            VERSION_LESS "14.0")
```
njzjz: Are you sure it is still true for 13.x? 13.x is the latest version

source/api_cc/src/DeepPot.cc
OutisLi: The distinction is memory ownership: coordinates, topology, and outputs are device-resident buffers, while `fparam`/`aparam` are host control data assembled by LAMMPS computes/fixes/settings. Commit 45bf817b9 now exposes the supported C API as pointer+length pairs; the deprecated C++ facade keeps vectors only as a host-side convenience adapter, and its documentation now states that residency explicitly.

doc/install/install-lammps.md
OutisLi: Updated in 45bf817b9. The documentation now covers both built-in and distribution/plugin modes. A Kokkos-enabled plugin is configured with `DEEPMD_LAMMPS_KOKKOS=ON` plus `Kokkos_DIR`, and uses the same C API device-graph interface.

source/lmp/builtin.cmake
OutisLi: Fixed in 45bf817b9. Built-in mode has returned to the C API, and `source/lmp/plugin/CMakeLists.txt` now supports compiling `deepmd/kk` against an exported Kokkos package. Both C API Kokkos builds were compiled locally; the built-in path also completed a 129k-atom device inference run.

source/lmp/builtin.cmake
OutisLi: Implemented in 45bf817b9 as C API version 28. Device-edge/canonical execution and capability queries are exposed as C functions, `deepmd.hpp` provides the compatibility wrapper, and `deepmd/kk` no longer requires `DP_USE_CXX_API`.

source/op/pt/CMakeLists.txt
OutisLi: You are right: CUDA 13 fixed the relevant 12.9 issue and its NVCC documentation defines `-arch=all` over the currently supported architectures (Turing and newer). e5d0dbefc narrows the `all-major` workaround to CUDA 12.9 and restores `all` for 13.x.

wanghan-iapcm (approved):
All the review threads I raised are addressed and verified against the current HEAD:

- `neighbor_graph/graph.py` field order: the full sweep to the parallel layout `destination_order, destination_row_ptr, source_order, source_row_ptr` landed in 254015486 and is consistent across the dataclass, the dpmodel CSR builder, and the pt_expt torch twin.
- `edge_transform_output.py`: the standard array-API force/virial path is now the explicit `else` branch paired with the fused path, with a single shared return.
- `DeepPotPTExpt.cc` GPU-edges graph branch: the destination-major physical-edge contract is now documented at the call site, and the pair-exclusion case canonicalizes.
- `deepmd/kernels/` placement: deferred to tracking issue #5782 with the target (move under `pt_expt/kernels/`), the pt->pt_expt dependency direction, and the `dpmodel` must-not-import invariant captured, as agreed.

Approving on that basis. To be clear about scope: this reflects that my review points are resolved — it is not a full independent audit of every new CUDA/Triton kernel and the Kokkos deployment path.

njzjz (approved):

PR #5772, opened 2026-07-12
feat(dpa4): preserve cutoff smoothness in edge feature branches

OutisLi:
## What

- Add an `edge_norm` descriptor option to control normalization on feature branches that vanish at the cutoff, with matching PyTorch and DPModel implementations.
- When disabled, remove hidden radial-MLP RMSNorm, bypass environment-seed FiLM and cross-focus competition norms, and use a unit variance floor for post-SO(2) residual scaling.
- Serialize the new option and its nested module settings, expose it through input validation, and enable the cutoff-smooth mode in the DPA4 water example.
- Add cutoff-smoothness, serialization, configuration-propagation, and PyTorch/DPModel parity coverage.

## Why

DPA4 radial features carry a C3 cutoff envelope and should decay to zero at `rcut`. Per-edge RMSNorm divides out that envelope until the feature variance reaches the norm epsilon floor, which can introduce a narrow, non-physical curvature and force feature just inside the cutoff. The same small-signal amplification applies to other normalization sites driven by cutoff-vanishing inputs.

The new mode preserves the cutoff envelope: FiLM modulation returns smoothly to identity, cross-focus weights approach uniform, and post-SO(2) residual messages remain linear near zero.

## Impact

- `edge_norm=True` remains the default, so existing configurations retain their current behavior.
- Setting `edge_norm=False` opts into the recommended cutoff-smooth behavior.
- PyTorch and DPModel serialization and numerical behavior remain aligned in both modes.
- Normalization on persistent node states and FFN branches is unchanged.

## Checks

- Added a near-cutoff radial sweep regression that verifies the curvature spike is suppressed when radial RMSNorm is disabled.
- Added structure and serialization round-trip coverage for both normalization modes.
- Added block- and descriptor-level PyTorch/DPModel parity coverage for standard and unit-floor post-SO(2) scaling.
- Tests were run separately by the author; they were not rerun in this checkout per request.

## Summary by CodeRabbit

* **New Features**
  * Added `edge_norm` for DPA4 and SeZM descriptors to control edge-dependent normalization behavior.
  * Added `radial_norm` for radial embeddings and `focus_norm` for SO(2) focus competition normalization.
  * Added `so2_post_norm_eps` to tune SO(2 post-normalization stability.
  * Updated model save/restore so these normalization settings are preserved.
* **Documentation / Examples**
  * Updated the water DPA4 example input to set `edge_norm` explicitly.
* **Tests**
  * Added tests covering the new switches (including serialization) and parity/consistency across implementations.

### DIFF
--- a/deepmd/dpmodel/descriptor/dpa4.py
+++ b/deepmd/dpmodel/descriptor/dpa4.py
@@ -176,6 +176,15 @@ class DescrptDPA4(NativeOP, BaseDescriptor):
     radial_mlp
         Hidden layer sizes for radial networks. An output layer of size
         `(l_schedule[0]+extra_node_l+1)*channels` will be automatically appended.
+    edge_norm
+        Whether to apply channel RMSNorm on the descriptor's cutoff-vanishing
+        branches: the radial network hidden layers, the environment-seed FiLM
+        scale/shift logits, the cross-focus competition scalars, and the
+        post-SO(2) residual messages. ``False`` replaces the first three norms
+        with identity and changes only the post-SO(2) norm to unit-floor residual
+        scaling. The unit floor uses ``sqrt(1 + variance)`` so small messages
+        retain their cutoff envelope instead of receiving the standard
+        ``1/sqrt(eps)`` small-signal gain.
     use_env_seed
         If True, seed the initial node state with local-environment information:
         apply environment matrix FiLM conditioning on l=0 features using 4D
@@ -457,6 +466,7 @@ class DescrptDPA4(NativeOP, BaseDescriptor):
         basis_type: str = "bessel",
         n_radial: int = 16,
         radial_mlp: list[int] | None = None,
+        edge_norm: bool = True,
         use_env_seed: bool = True,
         random_gamma: bool = True,
         edge_cartesian: bool = False,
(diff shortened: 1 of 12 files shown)


wanghan-iapcm (approved):
Reviewed the new cutoff-smoothness normalization switches. The gating is correct and consistent across both backends: the dpmodel (`if self.edge_norm`/`None`) and pt (`nn.Identity()`) idioms are behaviorally equivalent, defaults (`edge_norm=True`, `so2_post_norm_eps=1e-5`) reproduce the pre-PR computation exactly so existing configs are unchanged, and both branches are exercised in the parity and structure tests. No correctness or dpmodel<->pt parity issues found.

Two non-blocking notes:

1. The new serialized options (`edge_norm`, `radial_norm`, `focus_norm`, `so2_post_norm_eps`) are added to `serialize()` without bumping `@version`/`LATEST_VERSION` or adding explicit `deserialize` back-compat. It is harmless today because `deserialize` uses `cls(**config)` and the omitted keys fall back to behavior-preserving defaults, and DPA4/SeZM is unreleased — this is the same standing point from #5599. Still worth bumping while it is free, so the audit trail reflects the schema change before the first release.

2. The regression coverage validates smoothness via structural checks and an isolated RadialMLP curvature sweep rather than the full-descriptor energy-curve smoothness harness. Since the stated goal is PES continuity across the cutoff, an end-to-end energy/force continuity check with edge_norm on/off would more directly lock the claim.

Approving.

PR #5788, opened 2026-07-14
fix(dpmodel): preserve model dictionaries during serialization

### LINKED ISSUE #5639: [Code scan] Avoid mutating model dictionaries during save_dp_model

## Problem

`save_dp_model()` shallow-copies the model dictionary and then mutates nested dictionaries/lists while traversing variables.

Evidence:

- `traverse_model_dict()` writes transformed child values back into the dict/list it is visiting: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/utils/serialization.py#L30-L63
- `save_dp_model()` only performs a top-level `model_dict.copy()`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/utils/serialization.py#L86-L97
- The `.dp`/`.hlo` path replaces nested arrays with HDF5 dataset paths: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/utils/serialization.py#L104-L117
- The YAML path replaces nested arrays with serialized variable dictionaries: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/utils/serialization.py#L118-L140

## Impact

Calling `save_dp_model()` can corrupt the caller's in-memory model dictionary by replacing arrays under nested `@variables` structures with strings or YAML dictionaries. A caller that saves twice, saves in two formats, or continues using the dictionary after saving can observe unexpected non-array values.

## Suggested Fix

Deep-copy `model_dict` before traversal or make `traverse_model_dict()` functional. Add a regression test that saves a nested model dict and asserts the original nested arrays remain arrays afterwards.

njzjz:
## Summary

- preserve the caller-owned model dictionary while save_dp_model() rewrites variables into YAML or HDF5 references;
- copy only the nested dict and list containers that traversal mutates, so large arrays are not duplicated;
- retain support for non-copyable variable objects such as h5py.Dataset;
- add regressions for nested YAML/native serialization and an HDF5-backed variable.

The existing save/load tests pass deepcopy(self.model_dict) into save_dp_model() and then validate only the serialized round trip. That disposable copy hides mutations of the object supplied by a real caller, so the shallow-copy bug was never observed.

The new tests pass a caller-owned nested structure directly, assert that its containers and variables remain unchanged, and cover both serialization implementations. The HDF5 dataset case also prevents a memory-heavy full deepcopy from becoming the fix.

## Validation

- targeted serialization tests and existing save/load tests: 6 passed;
- ruff format .;
- ruff check .;
- git diff --check.

Closes #5639.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Bug Fixes**
  * Saving models no longer alters the original nested data structures.
  * Preserves shared and non-copyable objects, including HDF5-backed datasets, during serialization.
  * Ensures saved models retain expected data when loaded across supported formats.

### DIFF
--- a/deepmd/dpmodel/utils/serialization.py
+++ b/deepmd/dpmodel/utils/serialization.py
@@ -63,6 +63,22 @@ def traverse_model_dict(
     return model_obj
 
 
+def _copy_model_containers(model_obj: Any) -> Any:
+    """Copy only containers that :func:`traverse_model_dict` mutates.
+
+    Arrays and other variable objects can be large or may not support ``deepcopy``.
+    They remain shared because traversal replaces their entries in parent containers
+    without modifying the variable objects themselves.
+    """
+    if isinstance(model_obj, dict):
+        if model_obj.get("@is_variable", False):
+            return model_obj
+        return {key: _copy_model_containers(value) for key, value in model_obj.items()}
+    if isinstance(model_obj, list):
+        return [_copy_model_containers(value) for value in model_obj]
+    return model_obj
+
+
 class Counter:
     """A callable counter.
 
@@ -91,9 +107,12 @@ def save_dp_model(filename: str, model_dict: dict) -> None:
     filename : str
         The filename to save to.
     model_dict : dict
-        The model dict to save.
+        The model dict to save. The caller retains ownership, and this function does
+        not modify it.
     """
-    model_dict = model_dict.copy()
+    # Give traversal an independent container tree while retaining references to
+    # potentially large or non-copyable variable objects.
+    model_dict = _copy_model_containers(model_dict)
     filename_extension = Path(filename).suffix
     extra_dict = {
         "software": "deepmd-kit",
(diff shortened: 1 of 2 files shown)


wanghan-iapcm (dismissed):
Correct and minimal. save_dp_model did a shallow model_dict.copy(), but traverse_model_dict reassigns nested dict/list entries in place, so serialization mutated the caller's dict (arrays replaced by YAML/HDF5 references). _copy_model_containers rebuilds only the container tree and shares the variable objects, which is safe because the save callbacks read variables (data=x, x.dtype, x.tolist()) and create new @is_variable dicts rather than mutating inputs. The regression tests genuinely fail pre-fix -- passing a caller-owned dict directly and asserting equality/identity -- and the h5py.Dataset test additionally prevents a naive full deepcopy from being the fix, since Datasets aren't deep-copyable. Good catch on why the existing tests missed it (they fed in a deepcopy). LGTM.

wanghan-iapcm (commented):
LGTM

wanghan-iapcm (approved):

PR #5789, opened 2026-07-14
fix(dpmodel): keep output reductions on active backend

### LINKED ISSUE #5640: [Code scan] Use array-API casts and reductions in fit_output_to_model_output

## Problem

`fit_output_to_model_output()` mixes NumPy-style array methods with array-API namespace operations during dense output reduction.

Evidence:

- The function selects an array namespace from `coord_ext`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/model/transform_output.py#L25-L37
- Reducible outputs are cast with `vv.astype(...)` instead of `xp.astype(vv, ...)`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/model/transform_output.py#L42-L57
- The intensive masked path divides by `np.sum(mask, ...)` instead of an array-namespace sum: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/model/transform_output.py#L45-L50

An external torch check showed `torch.Tensor` has no `.astype` method. Using `np.sum` on backend arrays can also copy to host or detach from the active namespace.

## Impact

Torch/JAX/array-API execution can fail or silently leave the active backend when reducing model outputs, especially for intensive outputs with masks.

## Suggested Fix

Use `xp.astype(vv, ...)` and `xp.sum(xp.astype(mask, ...), ...)` consistently. Add a torch or array-api-strict test for reducible extensive and intensive outputs.

njzjz:
## Summary

- translate the configured NumPy energy precision into the dtype object of the active array namespace;
- use namespace-native casts and reductions in the dense dpmodel output transform;
- apply the same dtype correction to the ragged graph transform, which otherwise passes a NumPy dtype class to Torch;
- add focused Torch regressions for extensive, masked-intensive, and graph reductions.

There were no direct tests exercising the backend-agnostic dpmodel transform helpers with Torch arrays. Normal PyTorch and Paddle model tests call their backend-specific transform implementations instead, while the existing dpmodel graph tests use NumPy. As a result, neither Tensor.astype nor the NumPy dtype passed into the Torch namespace was reached.

The new tests invoke the generic dpmodel helpers directly and verify that results remain Torch tensors in the configured energy precision.

## Validation

- new transform regressions plus existing ragged graph tests: 10 passed;
- additional reviewer probes passed for NumPy, Torch, JAX, and array-api-strict;
- ruff format .;
- ruff check .;
- git diff --check.

Closes #5640.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved energy reduction accuracy by using backend-specific precision for accumulation and per-frame/per-node division, instead of a fixed global dtype.
  * Ensured intensive/extensive reductions consistently cast reducible values—and any provided masks or computed divisors—to the correct backend dtype before summing/averaging.

* **Tests**
  * Added Torch backend-preservation tests for both dense and graph reduction paths, covering intensive vs extensive behavior with and without masks and validating dtype promotion and correctness.


wanghan-iapcm (dismissed):
Correct fix, well aligned with the array-api discipline. The generic dpmodel reduction helpers were hard-coding NumPy ops (Tensor.astype, np.sum) and passing a NumPy dtype class into the active namespace, so they crashed/mis-typed on Torch -- exactly the pt_expt path via make_model. Resolving the energy precision through get_xp_precision and using xp.astype/xp.sum fixes it, and NumPy numerics are preserved (the int-vs-float64 mask divisor is value-identical). The three Torch regressions genuinely fail pre-fix (raw torch.Tensor has no .astype; the graph path rejects the NumPy dtype) and assert a float64 torch.Tensor result. Minor, non-blocking: only Torch is committed-tested here, but it's the representative namespace that exposes both bugs and the NumPy path stays covered by existing tests. LGTM.

deepmd/dpmodel/model/transform_output.py
```diff
@@ -45,16 +52,14 @@ def fit_output_to_model_output(
             if vdef.intensive:
                 if mask is not None:
                     model_ret[kk_redu] = xp.sum(
-                        vv.astype(GLOBAL_ENER_FLOAT_PRECISION), axis=atom_axis
-                    ) / np.sum(mask, axis=-1, keepdims=True)
+                        xp.astype(vv, energy_dtype), axis=atom_axis
+                    ) / xp.sum(xp.astype(mask, energy_dtype), axis=-1, keepdims=True)
                 else:
                     model_ret[kk_redu] = xp.mean(
```
wanghan-iapcm, line 58: This intensive-without-mask branch (`xp.mean(xp.astype(vv, energy_dtype), ...)`) is not exercised by any test -- the new tests cover extensive and intensive-WITH-mask, but not `mask is None`. Mutation testing (reverting just this line to the pre-fix `vv.astype(...)` while keeping the rest of the fix) still passes all three tests. Please add a dense intensive, no-mask Torch case.

deepmd/dpmodel/model/edge_transform_output.py
```diff
@@ -81,15 +88,13 @@ def fit_output_to_model_output_graph(
...
         redu = segment_sum(vv_e, frame_id, nf)  # (nf, *shape)
         if vdef.intensive:
             if mask is not None:
-                cnt = segment_sum(
-                    xp.astype(mask, GLOBAL_ENER_FLOAT_PRECISION), frame_id, nf
-                )
+                cnt = segment_sum(xp.astype(mask, energy_dtype), frame_id, nf)
             else:
-                cnt = xp.astype(n_node, GLOBAL_ENER_FLOAT_PRECISION)
+                cnt = xp.astype(n_node, energy_dtype)
```
wanghan-iapcm, line 157: Same gap on the graph path: this `cnt = xp.astype(n_node, energy_dtype)` fallback (intensive, no mask) is untested, and the graph extensive (`vdef.intensive == False`) branch is never taken by a test either. A graph intensive-no-mask case (and ideally a graph extensive case) would close the `mask is None` boolean and the extensive branch.

wanghan-iapcm (changes requested):
The fix itself is correct and complete (verified: get_xp_precision resolves the namespace-native dtype, NumPy numerics are preserved bit-identically, and the three new tests genuinely fail pre-fix). One blocking gap before merge: the `mask is None` reduction branches are untested in both files. Mutation testing confirms it -- reverting the fix on those exact lines while keeping the rest still passes all three tests, so a regression there would go undetected. Per the repo convention of covering every reachable branch (including both sides of a boolean), please add coverage for the no-mask intensive path (and ideally the graph extensive path). Requesting changes only for the test coverage; the code is good.

njzjz:

- @iProzd — 2 reviews on exact changed files (`deepmd/dpmodel/model/transform_output.py`).

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

wanghan-iapcm (approved):
Both review comments are addressed with genuine regression tests, verified against HEAD. The dense intensive/no-mask branch (transform_output.py) and the graph n_node-fallback + extensive branches (edge_transform_output.py) are each now exercised by a Torch test that asserts the result stays a torch tensor in energy precision. Confirmed empirically that all three pre-fix idioms raise on a Torch tensor (Tensor.astype -> AttributeError, xp.astype with a NumPy dtype class -> TypeError, np.sum on a tensor -> TypeError), so every new test fails pre-fix and passes after -- closing the mutation-survival gap the original comments flagged. No residual numpy idioms remain in either transform; the fix is complete.

njzjz:
Addressed the CodeRabbit nitpick in `bb988aea`: the Torch-dependent test module now imports `pytest` and uses module-level `pytest.importorskip("torch")`, so the entire module skips cleanly when Torch is unavailable while preserving all six test cases when it is installed.

Revalidated the current PR head: `source/tests/common/dpmodel/test_transform_output.py` passes (6 tests), and Ruff reports no issues for the file.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

PR #5790, opened 2026-07-14
fix(api_cc): size remapped atomic parameters by stride

### LINKED ISSUE #5618: [Code scan] Size remapped atomic-parameter buffers by daparam

## Problem

`select_real_atoms_coord` under-allocates the remapped atomic-parameter buffer when `daparam > 1`.

Evidence:

- The buffer is resized to `nframes * nreal` or `nframes * nloc_real`, without multiplying by `daparam`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/common.cc#L188-L193
- The subsequent `select_map(..., stride=daparam, ...)` writes `nframes * nreal * daparam` entries: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/common.cc#L985-L1008

## Impact

For models with `dim_aparam > 1`, shared neighbor-list wrapper paths can write past the end of `aparam`. This is memory corruption and may also feed incorrect atomic parameters to inference.

## Suggested Fix

Resize `aparam` with `* daparam`, matching the `select_map` stride. Add a nlist inference regression test using `dim_aparam > 1`, ideally with ASAN or debug STL enabled.

njzjz:
## Summary

- allocate remapped atomic-parameter buffers for every frame, retained atom, and daparam component;
- keep the allocation count aligned with the stride passed to select_map;
- add typed C++ regressions for float and double, two frames, local and ghost virtual atoms, and both local-only and all-atom aparam layouts.

## Impact

The previous allocation omitted the daparam factor. For models with dim_aparam greater than one, select_map wrote more scalar components than the vector contained, causing undefined behavior and potential heap memory corruption in shared neighbor-list API paths.

Existing multiframe and neighbor-list API tests use models whose daparam is one. The aparam_nall=true fixtures also use one component per atom, so the missing multiplier was numerically hidden.

The new direct helper tests exercise the exact missing combination: daparam=2 with virtual-atom remapping, multiple frames, local parameters, and local-plus-ghost parameters.

## Validation

- backend-neutral C++ unit-test target compiled and linked successfully;
- TestSelectMap typed suite: 12 passed;
- the four new cases fail with the pre-fix allocation and pass with the corrected size;
- clang-format applied to changed C++ files;
- ruff format .;
- ruff check .;
- git diff --check.

Closes #5618.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of multi-component atom parameters when filtering virtual atoms.
  * Ensured remapped parameter data maintains the correct size and values for local and ghost atoms.

* **Tests**
  * Added coverage for multi-frame data and virtual-atom filtering scenarios.

### DIFF
--- a/source/api_cc/src/common.cc
+++ b/source/api_cc/src/common.cc
@@ -189,8 +189,11 @@ void deepmd::select_real_atoms_coord(std::vector<VALUETYPE>& dcoord,
   select_map<int>(datype, datype_, fwd_map, 1);
   // aparam
   if (daparam > 0) {
+    // Atomic parameters store ``daparam`` consecutive components per atom.
+    // Keep the allocation consistent with the stride passed to ``select_map``
+    // below so every remapped component has a valid destination element.
     aparam.resize(static_cast<size_t>(nframes) *
-                  (aparam_nall ? nall_real : nloc_real));
+                  (aparam_nall ? nall_real : nloc_real) * daparam);
     select_map<VALUETYPE>(aparam, aparam_, fwd_map, daparam, nframes,
                           (aparam_nall ? nall_real : nloc_real),
                           (aparam_nall ? nall : (nall - nghost)));
--- a/source/api_cc/tests/test_select_map.cc
+++ b/source/api_cc/tests/test_select_map.cc
@@ -94,3 +94,85 @@ TYPED_TEST(TestSelectMap, selectmap_type1) {
     EXPECT_EQ(this->expected_atype_out_1[ii], this->atype_out_1[ii]);
   }
 }
+
+TYPED_TEST(TestSelectMap, select_real_atoms_coord_aparam_local) {
+  constexpr int nframes = 2;
+  constexpr int daparam = 2;
+  constexpr int nall = 5;
+  constexpr int nghost = 2;
+  constexpr int ntypes = 2;
+
+  // Type 2 represents a virtual atom.  The input contains virtual atoms in
+  // both the local [0, 3) and ghost [3, 5) portions of the neighbor list.
+  const std::vector<int> atype = {0, 2, 1, 0, 2};
+  std::vector<TypeParam> coord(nframes * nall * 3);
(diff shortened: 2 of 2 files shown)


njzjz (commented):
This piece of codes was created by @Yi-FanLi in 69e712c, so I request @Yi-FanLi to review.

wanghan-iapcm (dismissed):
Correct fix. select_map writes out[kk*nall1*stride + to_ii*stride + dd] with stride=daparam and does not resize out, so the destination needs nframes*nall1*daparam elements; the old resize omitted the daparam factor, overflowing the buffer for dim_aparam>1. The new resize matches select_map's indexing exactly. The regressions genuinely fail pre-fix on the size assertion (4!=8, 6!=12) before even hitting the out-of-bounds writes, and they cover both the aparam_nall=false (local) and aparam_nall=true (all) branches, plus float/double and local+ghost virtual atoms. Good explanation of why daparam=1 fixtures hid it. LGTM.

wanghan-iapcm (approved):
Re-reviewed via the code-review skill. Correct and complete: the new nframes*nall1*daparam allocation exactly matches select_map's max write index+1 (select_map doesn't resize its output, so the caller's resize is load-bearing), and the sibling dcoord/atype buffers were already correctly stride-sized, so no analogous bug remains. Day-one issue from #2436 where the aparam line never followed the stride-multiply pattern its dcoord neighbor established in the same commit. The two regressions fail pre-fix deterministically (size 4!=8 and 6!=12 via ASSERT_EQ) and cover both aparam_nall branches with local+ghost virtual atoms; pre-sizing aparam_out keeps the failure a clean size-assertion rather than an ASAN crash. The CodeRabbit TYPED_TEST-won't-compile note is a false positive (the fixture is a template). LGTM.

PR #5791, opened 2026-07-14
fix(tf): avoid empty AMix neighbor mapping pointers

### LINKED ISSUE #5653: [Code scan] Avoid empty idx_mapping pointers in AMix env-mat CPU paths

## Problem

The AMix env-mat CPU paths form `&idx_mapping[0]` even when `idx_mapping` is empty.

The non-NVNMD AMix CPU path declares an empty `std::vector<int> idx_mapping`, calls `_prepare_coord_nlist_cpu()`, and then unconditionally passes `&idx_mapping[0]`:

- https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/op/tf/prod_env_mat_multi_device.cc#L1292-L1314

The NVNMD AMix CPU path does the same:

- https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/op/tf/prod_env_mat_multi_device_nvnmd.cc#L781-L803

For no-PBC, LAMMPS-pointer, and tensor-neighbor-list modes, no coordinate-copy map is built, so `idx_mapping` remains empty. The callee only needs the map when `b_nlist_map` is true:

- https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/lib/src/neighbor_list.cc#L918-L955

But forming `&idx_mapping[0]` on an empty vector is still undefined behavior and can trip debug STL/UBSan builds.

## Impact

AMix env-mat calls in the no-copy modes can fail under debug or sanitizer builds, and the code relies on a pointer value that is not valid by the C++ standard.

## Suggested fix

Pass `idx_mapping.data()` only after ensuring the vector is nonempty, pass `nullptr` when `b_nlist_map == false`, or skip the map argument entirely in the no-map branch. Add sanitizer or `_GLIBCXX_DEBUG` coverage for AMix no-PBC and tensor-neighbor-list modes.

njzjz:
## Summary

- pass the neighbor-index mapping only when periodic coordinate expansion populated it;
- pass nullptr for non-mapping modes instead of forming an element pointer from an empty vector;
- apply the contract to both the standard TensorFlow AMix CPU fallback and the NVNMD quantized AMix CPU path;
- add direct CPU path coverage for external/LAMMPS neighbor lists and NVNMD no-PBC neighbor construction.

The ordinary AMix CPU no-PBC and PBC modes now use a batch fast path, so the affected legacy fallback is reached primarily by the external pointer-based neighbor-list mode. There was no direct test for that AMix path. The NVNMD AMix path also had no direct no-PBC test.

In release libstdc++, taking element zero from an empty vector can appear to work when the callee does not dereference the pointer, so higher-level numerical tests do not reliably expose the undefined behavior. The defect is diagnosed by static analysis or a libstdc++ debug/sanitizer build. The added tests ensure both affected call sites remain exercised and validate their neighbor indices, types, masks, valid entries, and padding.

## Validation

- TensorFlow-only deepmd_op target compiled and linked successfully against TensorFlow 2.21;
- the two focused tests loaded the newly built op: 2 passed;
- the full existing prod-env-mat test file plus the initial NVNMD coverage: 6 passed during development;
- libstdc++ debug-mode reproduction confirms the pre-fix empty-vector expression aborts;
- clang-format applied to changed C++ files;
- ruff format .;
- ruff check .;
- git diff --check.

Closes #5653.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed CPU neighbor/type/mask post-processing so neighbor-list index mapping is only provided when it’s valid, preventing unsafe pointer usage across AMix CPU modes (including NVNMD and external neighbor list scenarios).

- **Tests**
  - Added new AMix test coverage for NVNMD without periodic boundaries, PBC/NVNMD remapping when neighbors are mapped, and an external (legacy ABI) neighbor list path.
  - Extended CI to run additional TensorFlow-gated AMix neighbor mapping tests with hardened C++ standard library settings when leak checks are enabled.

### DIFF
--- a/.github/workflows/test_cc.yml
+++ b/.github/workflows/test_cc.yml
@@ -92,7 +92,8 @@ jobs:
           export LD_LIBRARY_PATH=${{ github.workspace }}/dp_test/lib:$LD_LIBRARY_PATH
           pytest -q \
             source/tests/tf/test_prod_env_mat.py::TestProdEnvMat::test_nopbc_nvnmd_amix \
-            source/tests/tf/test_prod_env_mat.py::TestProdEnvMat::test_pbc_nvnmd_amix_maps_copied_neighbors
+            source/tests/tf/test_prod_env_mat.py::TestProdEnvMat::test_pbc_nvnmd_amix_maps_copied_neighbors \
+            source/tests/tf/test_prod_env_mat.py::TestProdEnvMat::test_external_nlist_amix
         env:
           OMP_NUM_THREADS: 1
           TF_INTRA_OP_PARALLELISM_THREADS: 1
--- a/.github/workflows/test_cc.yml
+++ b/.github/workflows/test_cc.yml
@@ -63,7 +63,7 @@ jobs:
           TF_INTRA_OP_PARALLELISM_THREADS: 1
           TF_INTER_OP_PARALLELISM_THREADS: 1
           CMAKE_GENERATOR: Ninja
-          CXXFLAGS: ${{ matrix.check_memleak && '-fsanitize=leak' || '' }}
+          CXXFLAGS: ${{ matrix.check_memleak && '-fsanitize=leak -D_GLIBCXX_ASSERTIONS' || '' }}
           LSAN_OPTIONS: suppressions=${{ github.workspace }}/.github/workflows/suppr.txt
           ENABLE_TENSORFLOW: ${{ matrix.enable_tensorflow && 'TRUE' || 'FALSE' }}
           ENABLE_PYTORCH: ${{ matrix.enable_pytorch && 'TRUE' || 'FALSE' }}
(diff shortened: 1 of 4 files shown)


wanghan-iapcm (dismissed):
Correct fix. Forming &idx_mapping[0] on the empty mapping vector (non-PBC/external-nlist modes) is UB and aborts under a hardened libstdc++ (_GLIBCXX_ASSERTIONS), even though release builds tolerate it because the pointer is never dereferenced. Passing idx_mapping.data() only when b_nlist_map is set, else nullptr, is safe: I confirmed use_nei_info_cpu (neighbor_list.cc) dereferences nlist_map exclusively inside the if(b_nlist_map) branch, so nullptr in the false case is never touched. Both the standard AMix and NVNMD AMix call sites are fixed. One non-blocking note: the new tests exercise the two affected fallbacks (good coverage) and abort pre-fix under a debug/hardened libstdc++, but deepmd CI runs only -fsanitize=leak, which won't catch this UB -- so they won't guard against reintroduction in CI unless a _GLIBCXX_ASSERTIONS or ASAN job is added. LGTM.

iProzd (approved):

wanghan-iapcm (approved):
Re-reviewed via the code-review skill. The fix is correct and complete: forming &idx_mapping[0] on the empty mapping vector is UB (aborts under a hardened libstdc++), and passing idx_mapping.data() only when b_nlist_map is set else nullptr is safe -- use_nei_info_cpu (neighbor_list.cc) dereferences nlist_map exclusively inside if(b_nlist_map). Both AMix call sites are fixed and there are no missed siblings (the other &idx_mapping[0] sites are wrapped in if(b_nlist_map){}). Two non-blocking notes: (1) the NVNMD site's b_nlist_map==true branch is only covered indirectly by test_nvnmd_entrypoints; a direct PBC NVNMD AMix test would make that explicit. (2) The new tests are coverage-only -- pre-fix and post-fix produce identical output in a release build, and CI runs only -fsanitize=leak, so a reintroduction of this exact UB wouldn't fail in CI without a _GLIBCXX_ASSERTIONS/ASan job. Neither blocks; the fix itself is provably correct. LGTM.

PR #5792, opened 2026-07-14
fix(jax): return stored HLO neighbor count

### LINKED ISSUE #5669: [Code scan] Return the stored neighbor count from JAX HLO models

## Problem

The JAX HLO wrapper stores the neighbor selection as `self.sel`:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/jax/model/hlo.py#L60-L86

But `HLO.get_nnei()` returns `self.nsel`, which is never initialized:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/jax/model/hlo.py#L281-L290

The sibling TF SavedModel wrapper implements the same API by delegating to `get_nsel()`:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/jax/jax2tf/tfmodel.py#L285-L294

Serialization/export code expects `model.get_nnei()` to return an integer neighbor dimension:

https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/jax/utils/serialization.py#L229-L235

## Impact

Any code path that queries `get_nnei()` on an HLO-loaded JAX model raises `AttributeError: 'HLO' object has no attribute 'nsel'` instead of returning `sum(sel)`. This breaks the common model API for HLO models and can trip tooling that treats JAX model wrappers uniformly.

## Suggested fix

Change `HLO.get_nnei()` to return `self.get_nsel()` or `sum(self.sel)`, matching `TFModelWrapper`. Add a small HLO wrapper unit test that constructs an `HLO` object or a minimal fake and checks `get_nnei() == sum(sel)`.

njzjz:
## Summary

- make HLO.get_nnei() derive the total neighbor width from the stored sel metadata;
- delegate to get_nsel() so the metadata contract has one source of truth;
- add a focused regression for a serialized HLO model's selection metadata.

The constructor stores self.sel but never creates self.nsel. Existing consistency and model tests exercise neighbor counts on dpmodel, TensorFlow, and PyTorch wrappers, but no JAX test directly queried the HLO wrapper. Serialization and export helpers can therefore reach an AttributeError even though the broader model tests pass.

The new unit test isolates the metadata API without constructing real StableHLO byte artifacts, which are unrelated to this accessor.

## Validation

- targeted HLO metadata test: 1 passed;
- pre-fix implementation fails the same test with AttributeError;
- ruff format .;
- ruff check .;
- git diff --check.

Closes #5669.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected neighbor-count reporting for serialized JAX HLO model metadata.
  * Neighbor counts now accurately reflect the stored selection values.
* **Tests**
  * Added regression coverage to verify neighbor-count calculations.

### DIFF
--- a/deepmd/jax/model/hlo.py
+++ b/deepmd/jax/model/hlo.py
@@ -369,7 +369,7 @@ class HLO(BaseModel):
 
     def get_nnei(self) -> int:
         """Returns the total number of selected neighboring atoms in the cut-off radius."""
-        return self.nsel
+        return self.get_nsel()
 
     def get_sel(self) -> list[int]:
         return self.sel
--- a/source/tests/jax/test_deep_dos.py
+++ b/source/tests/jax/test_deep_dos.py
@@ -72,6 +72,10 @@ class TestDeepDOSJAX(unittest.TestCase):
     def test_numb_dos_survives_export(self) -> None:
         self.assertEqual(self.dp.get_numb_dos(), 2)
 
+    def test_neighbor_count_survives_export(self) -> None:
+        """Load the exported HLO and expose its serialized selection width."""
+        self.assertEqual(self.dp.deep_eval.dp.get_nnei(), 40)
+
     def test_global_dos_only(self) -> None:
         (dos,) = self.dp.eval(self.coords, self.cells, self.atypes, atomic=False)
         self.assertEqual(dos.shape, (1, 2))
--- a/source/tests/jax/test_hlo.py
+++ b/source/tests/jax/test_hlo.py
@@ -1,0 +1,17 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+"""Regression tests for metadata exposed by serialized JAX HLO models."""
+
+from deepmd.jax.model.hlo import HLO
+
+
+def test_hlo_get_nnei_uses_stored_selection() -> None:
+    """Return the neighbor width encoded in an HLO model's selection metadata.
+
+    Constructing a complete ``HLO`` instance requires valid serialized
+    StableHLO artifacts.  This API only depends on the stored ``sel`` metadata,
+    so bypass initialization to test that metadata contract directly.
+    """
+    model = HLO.__new__(HLO)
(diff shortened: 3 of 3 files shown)


wanghan-iapcm (dismissed):
Correct. HLO.__init__ stores self.sel but never sets self.nsel, so get_nnei()'s return self.nsel was a guaranteed AttributeError for any caller (serialization/export). Delegating to get_nsel() (sum(self.sel)) fixes it with a single source of truth. The regression genuinely fails pre-fix (AttributeError on self.nsel) and passes post-fix; using HLO.__new__ to bypass the StableHLO-artifact construction is fine since this accessor depends only on self.sel. LGTM.

wanghan-iapcm (approved):
Re-reviewed via the code-review skill. Correct and complete: HLO.__init__ stores self.sel but never sets self.nsel, so the old return self.nsel was a guaranteed AttributeError; delegating to get_nsel() (sum(self.sel)) matches the standard cross-backend get_nnei->get_nsel pattern. Verified self.nsel is the only orphaned attribute in the file (no other latent AttributeError), and the bug was present from day one (#4256), not a regression. The regression test genuinely fails pre-fix (AttributeError) and is non-tautological. One optional, non-blocking note: the test exercises the accessor in isolation via HLO.__new__; an end-to-end freeze/serialize test would also cover the real serialization.py call site, but that needs StableHLO artifacts. LGTM.

njzjz:

- @OutisLi — 4 reviews on exact changed files (`deepmd/jax/model/hlo.py`).

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

PR #5793, opened 2026-07-14
fix(ipi): reject unknown atom names

### LINKED ISSUE #5648: [Code scan] Reject unknown i-PI atom names instead of mapping them to type 0

## Problem

The i-PI driver silently maps any atom name missing from the JSON `atom_type` table to DeePMD type `0`.

`dp_ipi` reads the user-provided type table and the atom names from the XYZ file, then constructs `Convert`:

- https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/ipi/driver.cc#L74
- https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/ipi/driver.cc#L82-L85

Inside `Convert`, the lookup uses `std::map::operator[]`:

- https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/ipi/src/Convert.cc#L12-L14

For `std::map<std::string, int>`, `name_type_map[atomname[ii]]` inserts a missing key with the default integer value `0`. A misspelled XYZ element name or an omitted JSON entry therefore runs successfully, but the atom is evaluated as model type `0`.

## Impact

This can produce wrong i-PI energies and forces without any diagnostic. The failure mode is especially easy to miss because type `0` is a valid DeePMD type.

## Suggested fix

Use `find()`/`at()` instead of `operator[]`, report the unknown atom name, and fail before inference. If possible, also validate that configured type IDs are in the model's supported type range.

Add a `dp_ipi` regression where `coord_file` contains an atom name that is absent from `atom_type`; the driver should exit with a clear error instead of evaluating the atom as type `0`.

njzjz:
## Summary

- look up coordinate-file atom names without `std::map::operator[]`;
- reject names missing from `atom_type` instead of inserting them as DeePMD type 0;
- catch conversion/configuration errors locally before model loading, print a stable diagnostic, and return exit code 1;
- add a backend-independent CLI regression proving validation occurs before model loading.

## Impact

Type 0 is a valid model type. The previous `operator[]` lookup silently inserted every unknown name with value 0, so a typo or incomplete `atom_type` mapping could produce plausible-looking but scientifically incorrect energies and forces.

The existing i-PI tests use complete O/H mappings and exercise successful socket and inference workflows. None supplies an unknown coordinate-file name, so the default-insertion behavior was never observed.

The new test uses atom `Xx` with only `O` configured and deliberately points `graph_file` at a missing model. It requires the unknown atom diagnostic and exact exit code before any model-loading error can occur.

## Exception boundary and CodeQL

The first version wrapped the entire driver in a helper called from `main`. That preserved runtime behavior but caused CodeQL to treat the pre-existing command-line configuration-file access as a newly introduced uncontrolled path flow. The revised implementation keeps file parsing directly in `main` and limits the exception boundary to `Convert` construction, which is the operation that can reject an unknown atom name. This avoids suppressing the security scan while retaining the intended stable diagnostic.

## Validation

- full C++ build completed successfully and installed the updated `dp_ipi`;
- incremental `dp_ipi` target build succeeded after the localized exception-boundary revision;
- backend-independent unknown-name regression passed after the revision;
- complete i-PI test suite previously passed: 7 tests across TensorFlow, PyTorch, and the new validation case;
- `Convert` compiled separately under C++11 with `-Wall -Wextra -Werror`;
- both C API and C++ API driver conditional paths passed syntax checks;
- clang-format check passed;
- `ruff format .`;
- `ruff check .`;
- `git diff --check`.

Closes #5648.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid or missing atom-type configuration now produces clear error messages and a clean failure.
  * Unknown atom names are rejected before model loading, preventing misleading downstream errors.
  * Model initialization failures are handled safely with user-visible diagnostics.

* **Tests**
  * Added end-to-end coverage for unknown atom names and missing atom-type configuration.

### DIFF
--- a/source/ipi/driver.cc
+++ b/source/ipi/driver.cc
@@ -73,18 +73,25 @@ int main(int argc, char* argv[]) {
   const char* host = host_str.c_str();
   std::string graph_file = jdata["graph_file"];
   std::string coord_file = jdata["coord_file"];
-  std::map<std::string, int> name_type_map = jdata["atom_type"];
+  std::map<std::string, int> name_type_map;
+  try {
+    name_type_map = jdata.at("atom_type").get<std::map<std::string, int>>();
+  } catch (const std::exception& error) {
+    std::cerr << "dp_ipi: invalid atom_type configuration: " << error.what()
+              << std::endl;
+    return 1;
+  }
   bool b_verb = jdata["verbose"];
 
   std::vector<std::string> atom_name;
   {
-    std::vector<std::vector<double> > posi;
-    std::vector<std::vector<double> > velo;
-    std::vector<std::vector<double> > forc;
+    std::vector<std::vector<double>> posi;
+    std::vector<std::vector<double>> velo;
+    std::vector<std::vector<double>> forc;
     XyzFileManager::read(coord_file, atom_name, posi, velo, forc);
   }
 
-  std::unique_ptr<Convert<double> > cvt;
+  std::unique_ptr<Convert<double>> cvt;
   try {
     cvt.reset(new Convert<double>(atom_name, name_type_map));
   } catch (const std::exception& error) {
@@ -227,4 +234,5 @@ int main(int argc, char* argv[]) {
   if (msg_buff != NULL) {
     delete[] msg_buff;
   }
+  return 0;
 }
--- a/source/ipi/driver.cc
+++ b/source/ipi/driver.cc
@@ -1,8 +1,10 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 #include <cstdint>
+#include <exception>
 #include <fstream>
 #include <iomanip>
 #include <iostream>
+#include <memory>
 
 #include "Convert.h"
 #ifdef DP_USE_CXX_API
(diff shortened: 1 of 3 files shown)


wanghan-iapcm (dismissed):
Correct fix. Convert used name_type_map[atomname] (std::map::operator[]), which inserts a value-initialized 0 for a missing key -- and type 0 is a valid DeePMD type, so an unknown or mistyped atom name silently produced wrong-but-plausible energies/forces. Switching to .find() + std::invalid_argument, caught in main to return exit 1 with a stable diagnostic before model loading, is the right fix. The regression genuinely fails pre-fix: with a deliberately missing graph_file, the old code reaches model loading (a different error / non-1 exit) instead of the exact 'Unknown atom name' message, which nicely proves validation happens before model init. Keeping file parsing in main to avoid a spurious CodeQL uncontrolled-path finding is a thoughtful touch. LGTM.

njzjz:
Follow-up on the `run_driver` nit: I did not add an unreachable `return 0;`. The function’s `while (true)` loop has no `break`, and every explicit termination path returns from inside the loop, so control cannot flow off the end of the non-`main` function and the current build emits no `-Wreturn-type` warning. Adding a statement after that loop would not change behavior or make the existing cleanup reachable. Converting the long-lived buffers to RAII would be a separate lifecycle refactor rather than part of the unknown-atom validation fix.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

wanghan-iapcm (approved):
Re-reviewed via the code-review skill. Correct fix: Convert's operator[] silently inserted unknown atom names as the valid type 0; switching to .find()+std::invalid_argument, caught in main around only the Convert construction (file parsing kept in main to avoid the CodeQL uncontrolled-path false positive), returning exit 1 with a stable diagnostic before model loading. All cvt. call sites were converted to cvt->, the unique_ptr+raw-new is the correct C++11 pattern, and the try/catch scope doesn't swallow model-load errors. A reviewer executed the regression and confirmed it genuinely fails pre-fix (exit 134 / unrelated message) and passes post-fix (exit 1 / exact message); both accept and reject branches are covered. One non-blocking, pre-existing note: a *missing* atom_type key in the config still aborts via an uncaught nlohmann type_error at driver.cc:76 (before the try), so the PR body's broader 'configuration errors' phrasing slightly overclaims -- worth a follow-up, out of scope here. LGTM.

PR #5794, opened 2026-07-14
fix(c-api): contain model deviation frame errors

### LINKED ISSUE #5622: [Code scan] Do not throw uncaught C++ exceptions through model-deviation C APIs

## Problem

Several model-deviation C API helpers throw `std::runtime_error` before entering `DP_REQUIRES_OK`, and `DP_REQUIRES_OK` only catches `deepmd::deepmd_exception`.

Evidence:

- `DP_REQUIRES_OK` catches only `deepmd::deepmd_exception`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_c/include/c_api_internal.h#L11-L19
- `DP_DeepPotModelDeviCompute_variant` throws `std::runtime_error` for `nframes > 1`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_c/src/c_api.cc#L732-L750
- The same pattern exists in spin and nlist model-deviation variants: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_c/src/c_api.cc#L843-L861 and https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_c/src/c_api.cc#L955-L976 and https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_c/src/c_api.cc#L1080-L1101

## Impact

Throwing C++ exceptions through an `extern "C"` API can terminate C callers instead of storing an error retrievable with `DP_*CheckOK`.

## Suggested Fix

Set `dp->exception` and return for unsupported `nframes`, or throw/catch `deepmd::deepmd_exception` and `std::exception` consistently at the C boundary. Add a C API regression test passing `nframes = 2` and checking `DP_*CheckOK`.

njzjz:
## Summary

- validate model-deviation frame counts before any model, neighbor-list, or input-array access
- report unsupported counts through `DP_*CheckOK` instead of throwing a C++ exception across the C ABI
- cover all 12 public entry points that accept `nframes`, including float/double, DeepPot/DeepSpin, neighbor-list/non-neighbor-list, and v2/v3 variants

## Root cause and fix

The four shared model-deviation helpers threw `std::runtime_error` for multiple frames before reaching `DP_REQUIRES_OK`. Because that exception was outside the existing C API error boundary, a C caller could observe an unwinding C++ exception rather than a retrievable error string.

This change adds one shared validation helper. Any `nframes != 1` value now stores a precise error in the model-deviation object's shared exception state and returns immediately. Keeping the check at the top of each shared helper also makes zero and negative counts safe: no pointer range is constructed and no model or neighbor-list object is accessed.

The four deprecated v1 APIs are not in the 12-entry regression matrix because they do not accept `nframes` and always delegate with one frame.

## Regression coverage

The previous tests exercised C API exception formatting and model errors, but did not call the model-deviation compute entry points with an unsupported frame count. Consequently, they never crossed the exact boundary where `std::runtime_error` escaped.

The new regression uses a fresh error carrier for every public entry point, asserts that no exception escapes, and verifies the exact `CheckOK` message. On the previous implementation, the `nframes = 2` matrix fails immediately at `ASSERT_NO_THROW`; the additional `-1` and `0` cases protect the early-validation invariant.

Validation performed:

- focused regression: 4 tests passed
- expanded tests in `test_deepmd_exception.cc`: 11 tests passed
- complete no-backend C API target: 19 passed, 215 skipped
- clang-format 22.1.5 check
- `ruff format --check .`
- `ruff check .`
- `git diff --check`

Closes #5622.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Bug Fixes**
  * Model-deviation C API now rejects unsupported frame counts using the standard error-checking flow, avoiding exceptions reaching C callers.
  * “Exactly one frame” validation is enforced consistently across DeepPot and DeepSpin compute variants, including neighbor-list and precision-specific variants.
  * Validation is performed early to prevent unintended input processing for invalid frame counts.
* **Tests**
  * Added boundary and parameterized test coverage to ensure all relevant multi-frame model-deviation C API variants fail with the expected error for unsupported frame counts.

### DIFF
--- a/source/api_c/src/c_api.cc
+++ b/source/api_c/src/c_api.cc
@@ -733,7 +733,7 @@ inline void flatten_vector(std::vector<VALUETYPE>& onedv,
 namespace {
 
 constexpr char model_devi_nframes_error[] =
-    "Model-deviation C APIs support exactly one frame.";
+    "DeePMD-kit Error: Model-deviation C APIs support exactly one frame.";
 
 bool validate_model_devi_nframes(DP_DeepBaseModelDevi* dp, const int nframes) {
   if (nframes == 1) {
--- a/source/api_c/src/c_api.cc
+++ b/source/api_c/src/c_api.cc
@@ -730,6 +730,27 @@ inline void flatten_vector(std::vector<VALUETYPE>& onedv,
   }
 }
 
+namespace {
+
+constexpr char model_devi_nframes_error[] =
+    "Model-deviation C APIs support exactly one frame.";
+
+bool validate_model_devi_nframes(DP_DeepBaseModelDevi* dp, const int nframes) {
+  if (nframes == 1) {
+    return true;
+  }
+
+  // These helpers serve extern "C" entry points, so an unsupported frame
+  // count must be reported through DP_*CheckOK instead of unwinding a C++
+  // exception into a C caller. Keep this validation before constructing any
+  // pointer ranges or accessing the model/neighbor list: invalid input must
+  // remain safe even when those arguments are otherwise unusable.
+  dp->exception = model_devi_nframes_error;
+  return false;
+}
+
+}  // namespace
+
 template <typename VALUETYPE>
 void DP_DeepPotModelDeviCompute_variant(
     DP_DeepPotModelDevi* dp,
@@ -746,8 +767,8 @@ void DP_DeepPotModelDeviCompute_variant(
     VALUETYPE* atomic_energy,
     VALUETYPE* atomic_virial,
     const VALUETYPE* charge_spin = nullptr) {
-  if (nframes > 1) {
-    throw std::runtime_error("nframes > 1 not supported yet");
+  if (!validate_model_devi_nframes(dp, nframes)) {
+    return;
   }
   // init C++ vectors from C arrays
   std::vector<VALUETYPE> coord_(coord, coord + natoms * 3);
@@ -857,8 +878,8 @@ void DP_DeepSpinModelDeviCompute_variant(DP_DeepSpinModelDevi* dp,
                                          VALUETYPE* virial,
                                          VALUETYPE* atomic_energy,
                                          VALUETYPE* atomic_virial) {
-  if (nframes > 1) {
-    throw std::runtime_error("nframes > 1 not supported yet");
+  if (!validate_model_devi_nframes(dp, nframes)) {
+    return;
   }
   // init C++ vectors from C arrays
   std::vector<VALUETYPE> coord_(coord, coord + natoms * 3);
@@ -972,8 +993,8 @@ void DP_DeepPotModelDeviComputeNList_variant(
     VALUETYPE* atomic_energy,
     VALUETYPE* atomic_virial,
     const VALUETYPE* charge_spin = nullptr) {
-  if (nframes > 1) {
-    throw std::runtime_error("nframes > 1 not supported yet");
+  if (!validate_model_devi_nframes(dp, nframes)) {
+    return;
   }
   // init C++ vectors from C arrays
   std::vector<VALUETYPE> coord_(coord, coord + natoms * 3);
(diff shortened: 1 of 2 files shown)


wanghan-iapcm (dismissed):
Correct and thorough. The four model-devi helpers threw std::runtime_error from extern C entry points, so a C caller saw an exception unwind across the ABI instead of a retrievable error. Routing through validate_model_devi_nframes, which sets dp->exception and returns, matches the existing C-API error mechanism -- I confirmed DP_DeepPot/SpinModelDevi inherit DP_DeepBaseModelDevi and DP_DeepBaseModelDeviCheckOK returns string_to_char(dpbase->exception). Tightening nframes>1 to nframes!=1 is a real improvement: it rejects 0/negative before the std::vector(coord, coord+natoms*3) construction, which the -1/0 parametrized cases exercise against null pointers. The regression genuinely fails pre-fix (nframes=2 escapes the throw so ASSERT_NO_THROW fails; -1/0 would crash on null-pointer range construction) and covers all 12 nframes-accepting entry points across float/double, Pot/Spin, nlist/non-nlist, v2/v3. LGTM.

source/api_c/src/c_api.cc
```diff
@@ -730,6 +730,27 @@ inline void flatten_vector(std::vector<VALUETYPE>& onedv,
...
+  if (nframes == 1) {
+    return true;
+  }
+
+  // These helpers serve extern "C" entry points, so an unsupported frame
+  // count must be reported through DP_*CheckOK instead of unwinding a C++
+  // exception into a C caller. Keep this validation before constructing any
+  // pointer ranges or accessing the model/neighbor list: invalid input must
+  // remain safe even when those arguments are otherwise unusable.
+  dp->exception = model_devi_nframes_error;
```
wanghan-iapcm, line 748: Non-blocking (consistency): this is the only place in c_api.cc that writes dp->exception directly. Every other error goes through DP_REQUIRES_OK/DP_NEW_OK, which store a deepmd::deepmd_exception whose constructor prepends "DeePMD-kit Error: " -- so all other C-API error strings carry that prefix, but this one doesn't. A caller inspecting error messages will see a differently-formatted string. Consider either prefixing model_devi_nframes_error with "DeePMD-kit Error: " or throwing a deepmd::deepmd_exception and letting the existing macro store it, for uniform formatting. Purely cosmetic; the fix is correct as-is.

wanghan-iapcm (approved):
Re-reviewed via the code-review skill. Correct and complete: the four model-devi helpers threw std::runtime_error from extern "C" entry points (UB across the C ABI, unretrievable); routing through validate_model_devi_nframes (sets dp->exception, returns) with nframes!=1 fixes it and also rejects 0/negative before any pointer-range construction. Verified DP_DeepPot/SpinModelDevi inherit DP_DeepBaseModelDevi and DP_DeepBaseModelDeviCheckOK returns string_to_char(exception). A reviewer built both versions and confirmed the tests fail pre-fix (12-variant ASSERT_NO_THROW failures; nframes=-1/0 SIGSEGV) and pass post-fix; all 12 nframes-accepting entry points are covered (exhaustive), and the accept path stays covered by the existing hpp tests. One non-blocking consistency note inline. LGTM.

njzjz:

- @OutisLi — 24 reviews on exact changed files (`source/api_c/src/c_api.cc`, `source/api_c/tests/test_deepmd_exception.cc`).

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

PR #4321, opened 2024-11-07
feat(tf/pt): add/refact lammps & C++ support for spin model

iProzd:
The conversations in #4216 are not readable. I opened this clean one for reviewers.

## Release Notes

- **New Features**
  - Introduced support for spin configurations within the DeepMD framework.
  - Added `PairDeepSpin` class for integrating spin models into LAMMPS simulations.
  - Enhanced `DeepSpin` and `DeepSpinModelDevi` classes with new methods for energy and force calculations.
  - Updated `DeepPotTF` class to compute atomic energies and virials.
  - Added a new function for writing LMP data that includes spin information.

- **Bug Fixes**
  - Improved error handling for unsupported configurations in spin calculations.

- **Tests**
  - Added comprehensive unit tests for `DeepSpin` and `DeepSpinModelDevi` classes to validate functionality and ensure accuracy in computations.
  - Introduced tests for verifying spin interactions without periodic boundary conditions.
  - Developed tests for MPI-enabled versions of spin interactions.

- **Documentation**
  - Updated scripts and tests to reflect new functionalities and ensure clarity in usage.

PR #5795, opened 2026-07-14
fix(pt): remap spins after null atom filtering

### LINKED ISSUE #5619: [Code scan] Remap spin values when PyTorch spin nlist filters NULL atoms

## Problem

`DeepSpinPT` filters coordinates/types through `select_real_atoms_coord`, but it does not apply the same `bkw_map` remapping to the spin tensor.

Evidence:

- `select_real_atoms_coord` returns filtered `dcoord`, `datype`, `fwd_map`, and `bkw_map`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/DeepSpinPT.cc#L146-L155
- The spin tensor is built from the original `spin` buffer and reshaped to `{1, nall_real, 3}`, effectively taking the first `nall_real` original rows: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/DeepSpinPT.cc#L159-L166
- `DeepSpinPTExpt` already uses `bkw_map` to build the filtered spin tensor correctly: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/DeepSpinPTExpt.cc#L411-L424

## Impact

If NULL-type atoms are not all at the end, spins are attached to the wrong atoms during PyTorch spin nlist inference. This can silently produce incorrect forces/energies.

## Suggested Fix

Build `spin_wrapped` from `spin[bkw_map[ii]]`, matching `DeepSpinPTExpt`. Add a PyTorch spin nlist test with `atype` containing `-1` in the middle and distinct spin vectors.

njzjz:
## Summary

- compact DeepSpinPT spin vectors with the same `bkw_map` used for coordinates and atom types
- add a TorchScript `.pth` neighbor-list regression with a NULL atom in the middle of the input
- use a strict representation-equivalence tolerance so both float and double instances detect the old mapping error

## Root cause and fix

`select_real_atoms_coord` removes NULL-type atoms and returns compacted coordinates/types plus forward and backward maps. The DeepSpinPT neighbor-list path used the compacted coordinates and types, but reshaped a prefix of the original spin buffer to `nall_real` rows. That only preserves atom-to-spin association when all filtered NULL atoms happen to be at the end.

The fix rebuilds the spin buffer in `bkw_map` order before creating the tensor. Spin is a per-atom side channel, so it now follows exactly the same compacted ordering as coordinates and atom types, including local and ghost partitions.

## Regression coverage

The existing tests covered ordinary neighbor-list inference, but did not place `atype = -1` between real atoms. Therefore the incorrect prefix and the correct mapped spin buffers were indistinguishable in the tested layouts.

The new test evaluates the same six real atoms in two representations:

1. a compact input; and
2. a padded input with a distinctive NULL coordinate/spin inserted in the middle.

It compares energy, virial, real-atom force and magnetic force, and verifies that the NULL output slots are zero. The test explicitly checks the `.pth` backend so it exercises DeepSpinPT rather than PTExpt.

The file's historical float reference tolerance is `0.1`. The old mapping error is smaller than that (maximum observed differences: energy `0.01493`, force `0.02159`, magnetic force `0.08961`, virial `0.03228`), so this equivalence regression uses `1e-10` for double and `1e-5` for float. With the old production line restored, both typed cases fail; with the fix, both pass.

Validation performed:

- built `runUnitTests_cc`, `deepmd_backend_pt`, and `deepmd_op_pt` with CPU PyTorch
- old implementation check: 0 passed, 2 failed as expected
- focused fixed regression: 2 passed
- broader NoPBC DeepSpinPT neighbor-list group: 6 passed
- clang-format 22.1.5 check
- `ruff format --check .`
- `ruff check .`
- `git diff --check`

Closes #5619.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected spin handling when NULL atoms are filtered from an input system.
  * Ensured spin data remains correctly mapped to real atoms, including when NULL atoms appear in the middle of the system.

* **Tests**
  * Added coverage verifying consistent energy, forces, and virial results for systems containing intermediate NULL atoms.
  * Confirmed NULL atoms receive zero forces.

### DIFF
--- a/source/api_cc/src/DeepSpinPT.cc
+++ b/source/api_cc/src/DeepSpinPT.cc
@@ -160,7 +160,17 @@ void DeepSpinPT::compute(ENERGYVTYPE& ener,
   at::Tensor coord_wrapped_Tensor =
       torch::from_blob(coord_wrapped.data(), {1, nall_real, 3}, options)
           .to(device);
-  std::vector<VALUETYPE> spin_wrapped = spin;
+  // ``select_real_atoms_coord`` compacts coordinates and atom types in
+  // ``bkw_map`` order.  Spin is another per-atom side channel, so it must use
+  // the same mapping; taking a prefix of the original spin buffer is only
+  // correct when every filtered NULL atom happens to be at the end.
+  std::vector<VALUETYPE> spin_wrapped(static_cast<size_t>(nall_real) * 3);
+  for (int ii = 0; ii < nall_real; ++ii) {
+    for (int dd = 0; dd < 3; ++dd) {
+      spin_wrapped[static_cast<size_t>(ii) * 3 + dd] =
+          spin[static_cast<size_t>(bkw_map[ii]) * 3 + dd];
+    }
+  }
   at::Tensor spin_wrapped_Tensor =
       torch::from_blob(spin_wrapped.data(), {1, nall_real, 3}, options)
           .to(device);
--- a/source/api_cc/tests/test_deeppot_dpa_pt_spin.cc
+++ b/source/api_cc/tests/test_deeppot_dpa_pt_spin.cc
@@ -314,6 +314,103 @@ TYPED_TEST(TestInferDeepSpinDpaPtNopbc, cpu_lmp_nlist) {
   }
 }
 
+TYPED_TEST(TestInferDeepSpinDpaPtNopbc, cpu_lmp_nlist_null_atom_in_middle) {
+  using VALUETYPE = TypeParam;
+  const std::vector<VALUETYPE>& coord = this->coord;
+  const std::vector<VALUETYPE>& spin = this->spin;
+  const std::vector<int>& atype = this->atype;
+  const std::vector<VALUETYPE>& box = this->box;
+  deepmd::DeepSpin& dp = this->dp;
+
+  // This fixture deliberately uses the TorchScript ``.pth`` artifact, which
+  // dispatches through DeepSpinPT rather than the separate PTExpt ``.pt2``
+  // implementation.
+  ASSERT_EQ(deepmd::get_backend(kModelPath), deepmd::DPBackend::PyTorch);
+
+  auto compute_with_nlist = [&dp, &box](
+                                double& energy, std::vector<VALUETYPE>& force,
+                                std::vector<VALUETYPE>& force_mag,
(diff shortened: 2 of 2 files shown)


iProzd (approved):

wanghan-iapcm (dismissed):
Correct fix. coord_wrapped = dcoord is compacted by select_real_atoms_coord with dcoord[ii] = coord[bkw_map[ii]], so rebuilding spin_wrapped[ii] = spin[bkw_map[ii]] gives spin exactly the same compacted ordering as coordinates and types; the old prefix of the original spin buffer was only correct when NULL atoms happened to be at the end. The regression is well designed: it uses the .pth backend explicitly (DeepSpinPT, not PTExpt), inserts a NULL atom with a distinctive spin in the middle so misalignment is observable, and -- importantly -- tightens the tolerance to 1e-10/1e-5, below the 0.015-0.09 error magnitude, since the file's default 0.1 tolerance would have masked it. It covers both float and double and checks that NULL output slots are zero. LGTM.

wanghan-iapcm (approved):
Re-reviewed via the code-review skill. Correct and complete: coord_wrapped=dcoord is compacted so dcoord[ii]=coord[bkw_map[ii]], and rebuilding spin_wrapped[ii]=spin[bkw_map[ii]] gives spin the identical compacted ordering (same bkw_map gather/scatter idiom the output side already uses); the old prefix was only correct when NULLs were at the end. The incomplete-fix sweep is clean: DeepSpinPTExpt already uses bkw_map (correct since #5370) and DeepSpinTF folds spin into the coord array before filtering, so neither is affected; the no-nlist compute() overload correctly keeps the plain copy. The regression fails pre-fix (author's run: 0 passed, 2 failed) and, importantly, tightens the tolerance to 1e-10/1e-5 -- below the 0.01-0.09 error -- since the file default 0.1 would have masked it; both float and double, .pth backend, NULL-slot forces asserted zero. Bug dates to #4321. LGTM.

PR #5156, opened 2026-01-15
feat(pt): support spin virial (Rebased and updated version of #4545 by @iProzd.)

OutisLi:
## Release Notes

* **New Features**
  * Added virial (stress tensor) calculation support for spin-enabled models
  * Introduced virial loss metrics (RMSE_v, MAE_v) for spin model training

* **Bug Fixes**
  * Enabled virial data propagation through PyTorch and TensorFlow backends
  * Fixed virial output handling in C API for spin models

* **Documentation**
  * Updated training documentation to include virial data format specifications for spin models

PR #5796, opened 2026-07-14
fix(c-api): return DeepSpin atomic virials

### LINKED ISSUE #5621: [Code scan] Return DeepSpin model-deviation atomic virials from C API

## Problem

The non-nlist `DP_DeepSpinModelDeviCompute*_2` path computes atomic virials when requested, but never copies them back to the C output buffer.

Evidence:

- The documented C API exposes an `atomic_virial` output buffer for `DP_DeepSpinModelDeviCompute2`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_c/include/c_api.h#L1200-L1233
- The implementation calls the overload that fills `av` when `atomic_virial` is non-null: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_c/src/c_api.cc#L879-L889
- The copy-out block for `atomic_virial` is commented out: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_c/src/c_api.cc#L909-L918
- The nlist variant does copy atomic virials back, which suggests the non-nlist omission is accidental: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_c/src/c_api.cc#L1152-L1160

## Impact

C API callers that request `atomic_virial` for DeepSpin model deviation get an unchanged output buffer even though the backend computed the data.

## Suggested Fix

Restore the flatten/copy block for `atomic_virial`. Add C API and `deepmd.hpp` tests that request `atomic_virial` from `DeepSpinModelDevi`.

njzjz:
## Summary

- restore non-neighbor-list DeepSpin model-deviation atomic-virial copy-out in the raw C API
- restore flat-to-nested atomic-virial reshaping in the `deepmd.hpp` wrapper
- correct the model-major output-size documentation for all four DeepSpin model-deviation v2 C APIs
- add independent float/double regressions for both API layers

## Root cause and fix

The non-neighbor-list DeepSpin model-deviation helper requested atomic virials from the backend, but its flatten/copy block was commented out. The header-only wrapper had a second commented-out loop, so even a repaired raw C buffer would not be transferred into the nested `atom_virial` output.

This change restores both operations using the same model-major layout already used by the working neighbor-list path:

`model * natoms * 9 + atomic_virial_element`

The C API documentation for the double/float, neighbor-list/non-neighbor-list DeepSpin model-deviation v2 functions now consistently states that all output arrays include the `nmodels` dimension.

## Regression coverage

This escaped existing coverage for two reasons:

- the API-CC model-deviation tests call the C++ implementation directly and therefore bypass both C API layers; and
- the relevant atomic-virial size/value assertions in the hpp C API test were commented out.

The new raw C typed test evaluates one model to obtain a reference, pre-fills a two-model output buffer with NaNs, and verifies every element of both model-major blocks. The hpp test independently pre-fills its nested output buffers with NaNs and checks size, population, and values, so it also detects a missing wrapper reshape after the raw layer is fixed.

With the previous copy/reshape behavior restored, the four focused typed cases produce 0 passes and 4 failures because the NaN sentinels remain unchanged. With this fix, the expanded DeepSpin model-deviation group passes all 8 tests.

Validation performed:

- built `runUnitTests_c`, `deepmd_backend_pt`, and `deepmd_op_pt` with CPU PyTorch
- old implementation check: 0 passed, 4 failed as expected
- focused fixed raw C and hpp atomic tests: 4 passed
- expanded DeepSpin model-deviation group: 8 passed
- independent review and C++14 syntax checks found no issues
- clang-format 22.1.5 check
- `ruff format --check .`
- `ruff check .`
- `git diff --check`

Closes #5621.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed model-deviation atomic virial outputs so `atomic_virial` is correctly computed and returned when requested.
  * Ensured atomic virial values are consistent across supported computation variants and for all models.
* **Documentation**
  * Updated output parameter documentation to correctly describe per-model shapes for energy, force, virial, atomic energy, and atomic virial.
* **Tests**
  * Added and strengthened tests to verify atomic virial is populated (no missing values) and matches expected results across models.

### DIFF
--- a/source/api_c/include/c_api.h
+++ b/source/api_c/include/c_api.h
@@ -1207,15 +1207,18 @@ void DP_DeepPotModelDeviCompute2(DP_DeepPotModelDevi* dp,
  *dim_fparam.
  * @param[in] aparam The atom parameters. The array can be of size nframes x
  *natoms x dim_aparam.
- * @param[out] energy Output energy.
- * @param[out] force Output force. The array should be of size natoms x 3.
- * @param[out] force_mag Output magnetic force on each atom. The array should be
- * of size natoms x 3.
- * @param[out] virial Output virial. The array should be of size 9.
- * @param[out] atomic_energy Output atomic energy. The array should be of size
- *natoms.
- * @param[out] atomic_virial Output atomic virial. The array should be of size
- *natoms x 9.
+ * @param[out] energy Output energies of all models. The array should be of size
+ *nmodels.
+ * @param[out] force Output forces of all models. The array should be of size
+ *nmodels x natoms x 3.
+ * @param[out] force_mag Output magnetic forces of all models. The array should
+ *be of size nmodels x natoms x 3.
+ * @param[out] virial Output virials of all models. The array should be of size
+ *nmodels x 9.
+ * @param[out] atomic_energy Output atomic energies of all models. The array
+ *should be of size nmodels x natoms.
+ * @param[out] atomic_virial Output atomic virials of all models. The array
+ *should be of size nmodels x natoms x 9.
  * @warning The output arrays should be allocated before calling this function.
  *Pass NULL if not required.
  * @since API version 24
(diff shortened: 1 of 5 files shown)


iProzd (approved):

wanghan-iapcm (approved):
Reviewed via the code-review skill. Correct and complete: the non-nlist DeepSpin model-devi path had its atomic_virial copy-out (c_api.cc) and flat->nested reshape (deepmd.hpp) commented out -- a #5156 oversight (that PR uncommented the sibling virial copy-out after spin-virial support landed, but missed atomic_virial). The backend already produces av, the compute call populates it (same 6-output overload as the working atomic_energy block), and the model-major nmodels*natoms*9 layout matches the already-correct neighbor-list path. The c_api.h doc corrections (adding the nmodels dimension) are accurate. Both regressions genuinely fail pre-fix -- the raw-C buffer and the hpp avmd stay NaN (resize-to-same-size is a no-op preserving the sentinel), and NaN is the right sentinel since it distinguishes never-written from written-zero; both layers, float+double. One non-blocking cleanup: the sibling virial (vdir/vmd) assertions in the same hpp test functions remain commented even though #5156 fixed that path -- worth uncommenting alongside. LGTM.

PR #5798, opened 2026-07-14
fix(dpmodel): guard empty magnetic loss masks

### LINKED ISSUE #5637: [Code scan] Guard EnergySpinLoss when batches contain no magnetic atoms

## Problem

`EnergySpinLoss` divides by the number of selected magnetic atoms without guarding the all-false mask case.

Evidence:

- The magnetic-force path computes `n_valid = xp.sum(mask_float)`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/loss/ener_spin.py#L189-L200
- The MSE and MAE metric paths divide by `n_valid * 3`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/loss/ener_spin.py#L201-L208
- The MAE loss path also divides per frame by `per_frame_count_fm`, which can be zero for frames without magnetic atoms: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/loss/ener_spin.py#L209-L221

## Impact

A batch with no magnetic atoms can produce `NaN` magnetic-force loss values and poison the total loss, even if the magnetic force prefactor or `find_force_mag` later makes the contribution logically absent.

## Suggested Fix

Return a zero magnetic-force contribution when `n_valid == 0`, and guard per-frame denominators before division. Add a dpmodel loss regression test with `mask_mag` all false.

njzjz:
## Summary

- replace zero magnetic-mask denominators with Array API-safe nonzero denominators
- make all-empty magnetic-force masks contribute finite zero instead of NaN for both MSE and MAE losses
- preserve the existing display semantics when the magnetic-force label itself is absent
- add backend-neutral NumPy and Torch Array API regression coverage

## Root cause and fix

`EnergySpinLoss` masks non-magnetic atoms out of the force-magnitude residual, then divides its reductions by `n_valid * 3`. When a batch contains no magnetic atoms, both numerator and denominator are zero, producing NaN. Multiplying that value by a zero prefactor or a missing-label flag does not recover it because `0 * NaN` is still NaN.

The implementation now constructs:

```python
safe_n_valid = xp.where(n_valid > 0, n_valid, xp.ones_like(n_valid))
```

and uses it for every global magnetic-force MSE/MAE reduction. The masked numerator is already zero for an empty set, so the contribution becomes exactly zero. Guarding the denominator itself is important: array backends may evaluate both branches of a later `where`, and an unselected divide-by-zero can still produce invalid values or gradients. The expression is compatible with Array API namespaces and JAX tracing; non-empty batches retain the exact previous formula.

This PR intentionally does not change the native PyTorch loss or redefine its missing-label display metrics. When `find_force_mag == 0`, the total loss is finite zero while `display_if_exist` continues to report NaN for the unavailable magnetic metric.

Existing spin-loss mask tests always selected at least one magnetic atom in each batch. They covered partial masks and all-magnetic batches, but never the global zero denominator.

The new backend-neutral regression parameterizes:

- `mse` and `mae`;
- float32 and float64; and
- `find_force_mag` present and absent.

It verifies a finite zero total loss in all eight combinations, zero magnetic metrics for a present-but-empty label, and the existing NaN display behavior for an absent label. A separate pt_expt test exercises the same shared dpmodel implementation through the Torch Array API namespace.

On the previous implementation, all eight backend-neutral cases fail with NaN. With the fix, the combined new and related targeted tests pass.

## Validation

- old backend-neutral regression: 8 failed with NaN as expected
- fixed NumPy and pt_expt targeted tests: 20 passed
- `array_api_strict` MSE/MAE empty-mask scenarios passed
- JAX was not installed locally; the denominator expression was kept free of Python data-dependent branching for JIT compatibility
- a broader consistent-loss run produced 35 passed / 45 skipped plus 2 unrelated pre-existing MAE parity failures because the shared editable native-PT install came from another worktree; neither failure exercises the empty-mask path
- `ruff format --check .`
- `ruff check .`
- `git diff --check`

Closes #5637.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed magnetic-force loss calculations when no magnetic atoms are present.
  - Prevented NaN or invalid loss values in both MSE and MAE modes.
  - Ensured magnetic-force metrics and overall loss report finite zero values for empty magnetic masks.
  - Aligned results across supported computational backends.

- **Tests**
  - Added regression coverage across supported precisions and loss configurations.

### DIFF
--- a/deepmd/dpmodel/loss/ener_spin.py
+++ b/deepmd/dpmodel/loss/ener_spin.py
@@ -268,14 +268,18 @@ class EnergySpinLoss(Loss):
             # zero out non-magnetic atoms
             diff_fm = (force_mag_label - force_mag_pred) * mask_float
             n_valid = xp.sum(mask_float)
+            # Guard the denominator itself because array backends may evaluate
+            # both branches of ``where``. This is safe under JAX tracing and
+            # makes an all-empty magnetic mask contribute exactly zero.
+            safe_n_valid = xp.where(n_valid > 0, n_valid, xp.ones_like(n_valid))
             if self.loss_func == "mse":
-                l2_force_mag_loss = xp.sum(xp.square(diff_fm)) / (n_valid * 3)
+                l2_force_mag_loss = xp.sum(xp.square(diff_fm)) / (safe_n_valid * 3)
                 loss += pref_fm * l2_force_mag_loss
                 more_loss["rmse_fm"] = self.display_if_exist(
                     xp.sqrt(l2_force_mag_loss), find_force_mag
                 )
                 if mae:
-                    mae_fm = xp.sum(xp.abs(diff_fm)) / (n_valid * 3)
+                    mae_fm = xp.sum(xp.abs(diff_fm)) / (safe_n_valid * 3)
                     more_loss["mae_fm"] = self.display_if_exist(mae_fm, find_force_mag)
             elif self.loss_func == "mae":
                 abs_diff_fm = xp.abs(diff_fm)  # [nf, na, 3], zeros for non-magnetic
@@ -283,7 +287,7 @@ class EnergySpinLoss(Loss):
                 # force_mag MSE, force_real MAE and the displayed mae_fm) so the
                 # loss is batch-size independent: a 2-frame batch equals the mean
                 # of the two single-frame losses.
-                l1_force_mag_loss = xp.sum(abs_diff_fm) / (n_valid * 3)
+                l1_force_mag_loss = xp.sum(abs_diff_fm) / (safe_n_valid * 3)
                 loss += pref_fm * l1_force_mag_loss
                 more_loss["mae_fm"] = self.display_if_exist(
                     l1_force_mag_loss, find_force_mag
--- a/deepmd/pt/loss/ener_spin.py
+++ b/deepmd/pt/loss/ener_spin.py
@@ -370,15 +370,14 @@ class EnergySpinLoss(TaskLoss):
                 # over an empty tensor (NaN), and ``0 * NaN`` is still NaN, so
                 # ``nan_to_num`` keeps such a batch's force_mag term at zero
                 # loss and zero gradient instead of poisoning the whole step.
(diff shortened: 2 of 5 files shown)


iProzd (approved):

wanghan-iapcm (approved):
Reviewed via the code-review skill. Correct and behavior-preserving: for a non-empty magnetic mask safe_n_valid==n_valid (loss unchanged), and for an empty mask the numerator is already 0 so the contribution is a finite 0; guarding the denominator (not wrapping the result in where) is the right approach for JAX, which evaluates both where branches. All three magnetic-force reductions are covered. The regressions genuinely fail pre-fix (verified: all 8 numpy and 4 torch cases fail on isfinite, since 0*NaN=NaN regardless of find_force_mag), across mse/mae, float32/64, numpy+torch, with correct display_if_exist semantics (0 for a present-but-empty label, NaN for an absent label).

One non-blocking cross-backend note: the separate torch.jit backend deepmd/pt/loss/ener_spin.py already guards the accumulated loss via torch.nan_to_num, but its displayed more_loss['rmse_fm']/['mae_fm'] are NOT guarded and still evaluate to NaN for an empty magnetic mask. After this merges, dpmodel/pt_expt will report 0.0 for those display metrics while pt reports NaN for the same logical case, and source/tests/consistent/loss/test_ener_spin.py never exercises n_magnetic=0, so the divergence is untested. Worth aligning pt's display metrics (and/or adding an n_magnetic=0 consistency case) in a follow-up. Separately -- and clearly out of scope here -- the same NaN class (a mask-count of 0 used as a divisor) is still open in the energy/virial 1/real_natoms reciprocals across ener.py/ener_spin.py/pt equivalents for all-ghost frames; that belongs to the broader loss-padding-mask work. LGTM for the magnetic-mask fix.

PR #5800, opened 2026-07-14
fix(tf): close C++ API sessions

### LINKED ISSUE #5657: [Code scan] Close TensorFlow sessions in C++ API destructors

## Problem

Several TensorFlow C++ API wrappers create a `tensorflow::Session` with `NewSession`, but their destructors only delete the `GraphDef`.

- `DeepPotTF` creates the session here: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/DeepPotTF.cc#L458
- Its destructor only deletes `graph_def`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/DeepPotTF.cc#L419
- The raw session pointer is stored in the class: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/include/DeepPotTF.h#L296

The same lifetime pattern appears in the other TensorFlow API wrappers:

- `DeepSpinTF` creates a session at https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/DeepSpinTF.cc#L458, but its destructor only deletes `graph_def` at https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/DeepSpinTF.cc#L419.
- `DeepTensorTF` creates a session at https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/DeepTensorTF.cc#L57, but its destructor only deletes `graph_def` at https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/DeepTensorTF.cc#L23.
- `DipoleChargeModifierTF` creates a session at https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/DataModifierTF.cc#L60, but its destructor only deletes `graph_def` at https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/api_cc/src/DataModifierTF.cc#L26.

The constructor catch blocks also only delete `graph_def` if `init()` throws, so failures after `NewSession` can leak the partially constructed session before the destructor can run.

## Impact

Repeated construction/destruction of TensorFlow C++ API objects can leak TensorFlow sessions, including associated CPU/GPU resources and runtime threads. This is especially relevant for long-running host applications, plugins, or test processes that load many models.

## Suggested fix

Close and delete the TensorFlow session in each destructor, and clean it up on constructor failure after `NewSession` succeeds. Prefer RAII ownership for both `session` and `graph_def` so partial initialization and exceptions follow the same cleanup path.

njzjz:
Closes #5657.

## Summary

- initialize the TensorFlow `Session*` member to `nullptr` in `DeepPotTF`, `DeepSpinTF`, `DeepTensorTF`, and `DipoleChargeModifierTF`;
- centralize exclusive session teardown as `Close()` followed by `delete` and pointer reset;
- use an initialization rollback guard so public `init()` calls are failure-atomic and can be retried without overwriting a session left by an earlier failed attempt;
- clean sessions from both initializing-constructor exception paths and normal destructors;
- make `DeepPotJAX` initialization failure-atomic by unconditionally releasing all partially created TensorFlow C API handles (`TF_Session`, `TF_Graph`, `TF_SessionOptions`, `TF_Status`, `TFE_Context`, `TFE_ContextOptions`, and `TF_Function` references);
- add direct lifecycle regressions for failed initialization/retry and successful initialization/normal destruction.

## Root cause

The TensorFlow wrappers owned raw sessions created by `NewSession`, but their destructors and initializing-constructor catch blocks only deleted the `GraphDef`. Default construction followed by a failing public `init()` could also retain a partially initialized session, allowing a retry to overwrite the pointer.

`DeepPotJAX` had the same failure mode with TensorFlow C API handles: its destructor only released resources when `inited == true`, so an exception before the final initialization commit leaked every handle created up to that point.

The TensorFlow rollback guard now owns cleanup until initialization has completely committed. `DeepPotJAX::clear_tf_resources()` is null-safe for partial initialization and is called both from the exception path and the destructor. Successful initialization commits only after all metadata has been loaded.

## Test coverage

The lifecycle test intentionally white-box-links `deepmd_backend_tf` and `deepmd_backend_jax` into `runUnitTests_cc`, rather than going through the `dlopen` plugin facade. This makes the concrete backend destructors directly visible to the LSan test process.

The tests cover:

- repeated failed initialization followed by safe destruction for all four TensorFlow C++ wrappers;
- successful initialization followed by normal destruction for all four TensorFlow C++ wrappers;
- repeated failed `DeepPotJAX` initialization;
- successful `DeepPotJAX` initialization followed by normal destruction when the generated SavedModel test artifact is available.

Ordinary builds verify retry safety and destructor behavior; the CI leak-sanitizer leg provides the direct leak regression signal.

## Validation

- built `deepmd_backend_tf`, `deepmd_backend_jax`, `runUnitTests_cc`, and `deepmd_op` in a TensorFlow/JAX C++ test configuration with LAMMPS plugin mode disabled;
- `TestTensorFlowSessionLifecycle.*`: 3 passed, 1 skipped locally because the generated JAX SavedModel artifact was unavailable;
- `ruff check .`;
- `ruff format .`;
- clang-format check;
- CMake formatting hook;
- `git diff --check`.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

### DIFF
--- a/source/api_cc/include/DeepPotJAX.h
+++ b/source/api_cc/include/DeepPotJAX.h
@@ -242,6 +242,13 @@ class DeepPotJAX : public DeepPotBackend {
                            const bool atomic);
 
  private:
+  /** Release every TensorFlow C API handle owned by this backend.
+   *
+   * This helper is intentionally safe for partially initialized objects so an
+   * exception from init() cannot leak handles or make a later retry unsafe.
+   */
+  void clear_tf_resources() noexcept;
+
   bool inited;
   // device
   std::string device;
@@ -284,12 +291,12 @@ class DeepPotJAX : public DeepPotBackend {
   /**  TF C API objects.
    * @{
    */
-  TF_Graph* graph;
-  TF_Status* status;
-  TF_Session* session;
-  TF_SessionOptions* sessionopts;
-  TFE_ContextOptions* ctx_opts;
-  TFE_Context* ctx;
+  TF_Graph* graph = nullptr;
+  TF_Status* status = nullptr;
+  TF_Session* session = nullptr;
+  TF_SessionOptions* sessionopts = nullptr;
+  TFE_ContextOptions* ctx_opts = nullptr;
+  TFE_Context* ctx = nullptr;
   std::vector<TF_Function*> func_vector;
   /**
    * @}
--- a/source/api_cc/include/tf_private.h
+++ b/source/api_cc/include/tf_private.h
@@ -18,4 +18,49 @@ typedef tensorflow::tstring STRINGTYPE;
 #else
 typedef std::string STRINGTYPE;
 #endif
+
+/**
+ * @brief Close and release an exclusively owned TensorFlow session.
+ *
+ * TensorFlow permits concurrent Run calls, but requires them all to finish
+ * before the sole Close call.  Owners must therefore stop concurrent use
+ * before destruction; this helper only centralizes the non-throwing cleanup
+ * needed by destructors and partially constructed API objects.
+ *
+ * @param[in,out] session Exclusively owned session, reset to nullptr.
+ */
+inline void close_and_delete_session(tensorflow::Session*& session) noexcept {
+  if (session == nullptr) {
+    return;
+  }
+  session->Close().IgnoreError();
+  delete session;
+  session = nullptr;
+}
+
+/**
+ * @brief Roll back a session assigned during a potentially throwing init call.
+ *
+ * Call release() only after initialization has completed successfully.
+ */
+class SessionCleanupGuard {
+ public:
+  explicit SessionCleanupGuard(tensorflow::Session*& session) noexcept
+      : session_(session), active_(true) {}
+
+  ~SessionCleanupGuard() {
+    if (active_) {
+      close_and_delete_session(session_);
+    }
+  }
+
+  SessionCleanupGuard(const SessionCleanupGuard&) = delete;
+  SessionCleanupGuard& operator=(const SessionCleanupGuard&) = delete;
+
+  void release() noexcept { active_ = false; }
+
+ private:
+  tensorflow::Session*& session_;
+  bool active_;
+};
 }  // namespace deepmd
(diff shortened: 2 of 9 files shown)


wanghan-iapcm (approved):
Reviewed via the code-review skill. Correct and complete: all four tensorflow::Session* owners now close+delete the session; the SessionCleanupGuard binds session by reference (so it cleans the pointer NewSession later assigns), releases only after inited=true, and its member-declaration order matches the reordered init lists (no -Wreorder). No double-free on the throwing-constructor path (the in-init guard nulls session before the catch block runs), and Close().IgnoreError() is the right cleanup-path call. Nicely, unlike a pure-UB fix, this genuinely fails pre-fix under CI's -fsanitize=leak leg, which does run runUnitTests_cc, and all four wrappers create a session before the forced failure so the leak is LSan-visible. Long-overdue fix (PR #3062 fixed the analogous graph_def leak in 2023 but never the session).

Two non-blocking notes:
1. Same leak family, still open elsewhere: DeepPotJAX (source/api_cc/src/DeepPotJAX.cc) has the identical leak-on-throwing-init pattern with TF C-API handles (TF_Session/TF_Graph/TFE_Context/TF_Function) -- its multi-arg constructor has no try/catch and ~DeepPotJAX only frees resources if inited==true, so a throw after TF_LoadSessionFromSavedModel but before inited=true leaks everything. Out of #5657's scope (different handle type), but worth a follow-up issue.
2. The new lifecycle test covers only the failed-init/retry path; the successful-init-then-normal-destroy path (the headline #5657 scenario, which exercises the destructor's Close() on a LIVE session) is only covered implicitly by the pre-existing functional tests under the same LSan leg -- a direct case would make that explicit. Also worth calling out in the PR description that this is the first test to white-box-link deepmd_backend_tf into runUnitTests_cc rather than going through the dlopen plugin facade. LGTM.

PR #5803, opened 2026-07-14
fix(lammps): free model deviation buffers

### LINKED ISSUE #5644: [Code scan] Free LAMMPS model-deviation atomic communication buffers

## Problem

The LAMMPS pair style allocates model-deviation `out_each` communication buffers but never frees them.

Evidence:

- The buffer pointers are stored on `PairDeepBaseModel`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/lmp/pair_base.h#L103-L108
- The constructor initializes `stdf_comm_buff_size` but does not initialize these pointers to null: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/lmp/pair_base.cpp#L373-L399
- `init_style()` allocates `counts`, `displacements`, `stdfsend`, `stdfrecv`, `tagsend`, and `tagrecv` when `out_each == 1`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/lmp/pair_base.cpp#L479-L490
- The destructor only destroys `setflag`, `cutsq`, and `scale`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/lmp/pair_base.cpp#L428-L434
- The compute path may also reallocate the four atom-sized buffers when `atom->natoms` grows: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/source/lmp/pair_deepmd.cpp#L562-L591

## Impact

Repeatedly creating/destroying `pair_style deepmd` or changing simulations that use model-deviation `atomic` output leaks LAMMPS-managed memory. Uninitialized pointer members also make it harder to safely add cleanup later.

## Suggested Fix

Initialize all buffer pointers to `nullptr` in the constructor and destroy them in `PairDeepBaseModel::~PairDeepBaseModel()`. If `init_style()` can run more than once for the same pair object, guard or destroy existing buffers before recreating them. Add a LAMMPS lifecycle test or ASAN run that creates and clears a `deepmd` pair style with `atomic` model-deviation output.

njzjz:
Closes #5644.

## Summary

- initialize all six model-deviation gather buffer pointers to `nullptr`;
- centralize allocation and growth in a shared `ensure_model_deviation_buffers()` helper;
- reuse process-count buffers across repeated `init_style()` calls instead of overwriting live allocations;
- grow the four atom-sized buffers through LAMMPS `Memory::grow()` when the global atom count increases;
- release all six buffers from the shared base-class destructor;
- use the same lifecycle path for both DeepMD and DeepSpin pair styles;
- add a repeated `run(0)` / `clear()` smoke test for reinitialization and pair destruction.

## Root cause

`init_style()` created the communication arrays whenever atomic model-deviation output was enabled, but repeated LAMMPS runs can call `init_style()` again on the same pair object. Those creates overwrote the stored pointers. The compute paths separately destroyed and recreated four arrays on atom-count growth, while the pair destructor released none of the six arrays.

The pair object is the sole owner of these buffers. Allocation, reuse, growth, and destruction are now handled in one base-class implementation, with `nullptr` initialization making every cleanup path safe.

The existing model-deviation tests validate numerical output. Leaking a LAMMPS-managed buffer does not change those values or necessarily crash, so the old implementation passes ordinary functional tests.

The new repeated-init/clear test is intentionally only a lifecycle smoke: it exercises repeated `init_style()` and pair destruction and can catch obvious double-free or use-after-free regressions. It is not presented as a leak detector, and the old implementation also passes it while leaking.

A direct stable old-fail/new-pass demonstration requires a dedicated LeakSanitizer build with appropriate handling of unrelated TensorFlow/Python/LAMMPS allocations. No RSS or process-memory threshold was added because it would be allocator- and timing-dependent.

## Validation

- built the DeePMD LAMMPS plugin, TensorFlow backend, and TensorFlow op against `stable_22Jul2025_update2`;
- built a matching Update2 LAMMPS shared library and Python wrapper;
- existing `test_pair_deepmd_model_devi`: passed;
- new `test_pair_deepmd_model_devi_atomic_reinit_clear_smoke`: passed;
- combined targeted result: 2 passed;
- clang-format 22.1.5 check;
- `ruff format .`;
- `ruff check .`;
- `git diff --check`;
- independent final review: no findings.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved allocation, cleanup, and reuse of model-deviation output buffers during repeated pair initialization and simulation teardown.
  * Prevented stale pointers and potential memory leaks by centralizing model-deviation buffer management.
  * Ensures model-deviation gather buffers correctly grow with system size changes.

* **Tests**
  * Added a smoke test that runs the DeepMD model-deviation workflow twice and then clears the simulation to verify cleanup.
  * Extended CI with a LeakSanitizer step for model-deviation teardown coverage.

### DIFF
--- a/.github/workflows/test_cc.yml
+++ b/.github/workflows/test_cc.yml
@@ -87,24 +87,6 @@ jobs:
           ENABLE_JAX: ${{ matrix.enable_tensorflow && '1' || '0' }}
           ENABLE_PADDLE: ${{ matrix.enable_paddle && '1' || '0' }}
         if: ${{ !matrix.check_memleak }}
-      - name: Test LAMMPS model-deviation teardown with LSan
-        run: |
-          export LD_LIBRARY_PATH=${{ github.workspace }}/dp_test/lib:$LD_LIBRARY_PATH
-          pytest -q source/lmp/tests/test_lammps.py::test_pair_deepmd_model_devi_atomic_reinit_clear_smoke
-        env:
-          OMP_NUM_THREADS: 1
-          TF_INTRA_OP_PARALLELISM_THREADS: 1
-          TF_INTER_OP_PARALLELISM_THREADS: 1
-          LAMMPS_PLUGIN_PATH: ${{ github.workspace }}/dp_test/lib/deepmd_lmp
-          LSAN_OPTIONS: suppressions=${{ github.workspace }}/.github/workflows/suppr.txt
-          ENABLE_TENSORFLOW: "1"
-          ENABLE_PYTORCH: "1"
-          ENABLE_JAX: "1"
-          ENABLE_PADDLE: "0"
-        # The focused TensorFlow/PyTorch LSan leg makes the buffer lifecycle
-        # smoke test fail on a reintroduced leak without running the full
-        # LAMMPS suite under sanitizer-incompatible Paddle dependencies.
-        if: ${{ matrix.check_memleak && matrix.enable_tensorflow }}
       # test ipi
       - run: |
           export PATH=${{ github.workspace }}/dp_test/bin:$PATH
--- a/.github/workflows/test_cc.yml
+++ b/.github/workflows/test_cc.yml
@@ -87,6 +87,24 @@ jobs:
           ENABLE_JAX: ${{ matrix.enable_tensorflow && '1' || '0' }}
           ENABLE_PADDLE: ${{ matrix.enable_paddle && '1' || '0' }}
         if: ${{ !matrix.check_memleak }}
+      - name: Test LAMMPS model-deviation teardown with LSan
+        run: |
+          export LD_LIBRARY_PATH=${{ github.workspace }}/dp_test/lib:$LD_LIBRARY_PATH
+          pytest -q source/lmp/tests/test_lammps.py::test_pair_deepmd_model_devi_atomic_reinit_clear_smoke
+        env:
+          OMP_NUM_THREADS: 1
+          TF_INTRA_OP_PARALLELISM_THREADS: 1
+          TF_INTER_OP_PARALLELISM_THREADS: 1
+          LAMMPS_PLUGIN_PATH: ${{ github.workspace }}/dp_test/lib/deepmd_lmp
+          LSAN_OPTIONS: suppressions=${{ github.workspace }}/.github/workflows/suppr.txt
+          ENABLE_TENSORFLOW: "1"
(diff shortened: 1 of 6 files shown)


wanghan-iapcm (approved):
Reviewed via the code-review skill. Correct and complete: all six model-devi buffers are nullptr-initialized, ensure_model_deviation_buffers() reuses counts/displacements (nprocs is fixed for the run) and memory->grow()s the four atom buffers (grow-on-nullptr == create in LAMMPS), and destroy_model_deviation_buffers() frees all six from the base destructor (destroy-on-nullptr is a safe no-op, so no double-free), covering both PairDeepMD and PairDeepSpin. The sweep found no other unfreed model-devi buffers, and the destructor-never-freed gap has existed since #628. Both the atomic and non-atomic branches are exercised (the latter via existing teardown).

One non-blocking note: the new reinit/clear smoke test genuinely exercises the fixed paths, but it cannot fail pre-fix in CI -- the leak is silent (not a crash/UAF), and the LAMMPS pytest step is gated with 'if: not matrix.check_memleak', so the -fsanitize=leak leg never runs it while every leg that runs it has no sanitizer. As written it is a crash-safety/coverage guard, not a leak-regression guard (which the PR body candidly acknowledges). Wiring source/lmp/tests into the check_memleak leg (or a dedicated LSan job) would turn it into a real regression guard for this leak class. LGTM.

njzjz:

- @OutisLi — 49 reviews on exact changed files (`.github/workflows/test_cc.yml`, `source/lmp/pair_base.cpp`, `source/lmp/pair_base.h`, `source/lmp/pair_deepmd.cpp`, `source/lmp/pair_deepspin.cpp`).
- @iProzd — 7 commits on changed files (`source/lmp/pair_base.cpp`, `source/lmp/pair_base.h`, `source/lmp/pair_deepmd.cpp`, `source/lmp/pair_deepspin.cpp`, `source/lmp/tests/test_lammps.py`).

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

PR #5807, opened 2026-07-16
fix(dpa1): support nonperiodic input statistics

OutisLi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved support for nonperiodic data by allowing input samples to omit the box information.
  * Statistical calculations now produce consistent results whether box is missing or explicitly set to `None`.
* **Tests**
  * Added regression coverage to ensure mean and standard deviation outputs remain bit-identical when `box` is absent or `box=None`.

### DIFF
--- a/deepmd/dpmodel/utils/env_mat_stat.py
+++ b/deepmd/dpmodel/utils/env_mat_stat.py
@@ -274,11 +274,9 @@ class EnvMatStatSe(EnvMatStat):
             device=array_api_compat.device(data[0]["coord"]),
         )
         for system in data:
-            coord, atype, box = (
-                system["coord"],
-                system["atype"],
-                system["box"],
-            )
+            coord = system["coord"]
+            atype = system["atype"]
+            box = system.get("box")
             nframes, nloc = atype.shape[:2]
             pair_excl = None
             if "pair_exclude_types" in system:
--- a/deepmd/pd/utils/env_mat_stat.py
+++ b/deepmd/pd/utils/env_mat_stat.py
@@ -107,11 +107,9 @@ class EnvMatStatSe(EnvMatStat):
                 "last_dim should be 1 for raial-only or 4 for full descriptor."
             )
         for system in data:
-            coord, atype, box = (
-                system["coord"],
-                system["atype"],
-                system["box"],
-            )
+            coord = system["coord"]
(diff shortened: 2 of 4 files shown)


wanghan-iapcm (changes requested):
Thanks — the fix itself is correct: `system.get("box")` collapses "box key absent" and explicit `box=None` to the same nonperiodic path, which `extend_input_and_build_neighbor_list` already supports, and the new test genuinely fails pre-fix (KeyError on the box-less sample) and passes after.

One change needed before merge: the fix is incomplete across backends. `deepmd/pt` re-exports the dpmodel class so it is fixed automatically, but the Paddle backend keeps its own independent copy of `EnvMatStatSe.iter` that still does unconditional `system["box"]` indexing, so a nonperiodic (box-less) sample — e.g. an LMDB frame, where `lmdb_data.py` only sets `box` when `"box" in frame` — will still raise the same `KeyError` on pd:

https://github.com/deepmodeling/deepmd-kit/blob/d155800a053f73b98e0915bd9510bca069059c5a/deepmd/pd/utils/env_mat_stat.py#L110-L114

pd's own neighbor-list builder already handles `box=None`, so the dict access is the only blocker. Please mirror the `.get("box")` change into `deepmd/pd/utils/env_mat_stat.py` (and ideally add a matching pd or cross-backend test), consistent with the dp/pt/pd parity established for this same file in #5267.

wanghan-iapcm (approved):
Thanks for the follow-up. Commit 72c211e4 mirrors the .get("box") fix into the Paddle backend (deepmd/pd/utils/env_mat_stat.py) and adds a pd regression test (test_nonperiodic_input_stats_accept_missing_box) that omits the box key entirely, so it fails pre-fix with KeyError and passes after. dpmodel/pt (re-export), jax (no own copy), and pd are now all consistent, each covered by a test that fails pre-fix. Request satisfied. LGTM.

njzjz:

- @iProzd — 7 reviews on exact changed files (`deepmd/dpmodel/utils/env_mat_stat.py`, `source/tests/common/dpmodel/test_env_mat_stat_graph.py`).

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

PR #5808, opened 2026-07-16
fix(dpa4): stabilize envelope-gated attention normalization

OutisLi:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved gated attention numerical stability for tiny edge/source weights, including smoother behavior at cutoff boundaries.
  - Refined attention normalization using explicit “physical null mass” and improved mixed-precision handling.
  - Updated cutoff envelope computation to a cancellation-free, factorized form for more reliable non-negative envelopes.

- **Tests**
  - Added regression coverage for cutoff-boundary behavior, vanishing-weight continuity, and second-derivative (Hessian) correctness.

- **Documentation**
  - Updated DPA4 example READMEs and input configs to use DPA4-Mini-style parameter sets.
  - Added an informational message when freezing SeZM models about setting the Triton inference level.


wanghan-iapcm (approved):
The numerics check out: the log-space softmax is algebraically equivalent with a correct log-sum-exp shift (the max-shift cancels through the null term, so autograd stays correct), the all-vanish case gives a well-defined alpha=0, the removed isfinite/eps guards were the original stabilization rather than a targeted patch and are provably unnecessary under the new null-mass formulation, and the C3 envelope factorization is exact (verified for p=1..7) and reaches exactly 0 at the cutoff. dpmodel and pt are faithful mirrors. LGTM.

One non-blocking follow-up on test coverage: the new extreme-regime tests (TestCutoffNumerics — nextafter-cutoff, src_weight=1e-30, high-logit vanishing, tiny-source-weight Hessian) are written only against the pt implementation. The dpmodel implementation is a separate rewrite and is the source of truth for the pt_expt and jax backends (pt_expt's descriptor subclasses this C3CutoffEnvelope), yet the existing cross-backend parity test (test_dpa4_dpmodel_parity.py) only feeds moderate rng.normal() values and never reaches these underflow/near-cutoff extremes. So nothing currently proves the dpmodel branch is equally stable in the exact regime this PR targets. Consider mirroring these edge cases into the parity file or adding a dpmodel-side equivalent, so the stability guarantee holds for all backends that consume dpmodel.

Minor: the Step 1 comment in deepmd/dpmodel/descriptor/dpa4_nn/attention.py lists edge_mask in the "edge_env**2 * src_weight * edge_mask underflows" product, but edge_mask is a binary 0/1 gate handled purely via the active predicate (never folded into log_weight), as the adjacent comment and the pt version correctly state. Worth aligning the wording.

PR #5809, opened 2026-07-16
fix(disp): div 0 error when using disp_avg under multitask

OutisLi:
## Summary by CodeRabbit

* **New Features**
  * Added per-task averaged training metrics for display intervals, with support for single-task and multi-task training.
  * Validation metrics are now reported per task alongside averaged training metrics.
  * Tasks without sampled training steps display `NaN` values rather than incomplete results.
* **Bug Fixes**
  * Improved consistency between metric columns, displayed steps, and reported values in multi-task training logs.
* **Documentation**
  * Clarified the behavior and backend support of the `disp_avg` training option.
* **Tests**
  * Added coverage for averaged metrics, restarts, validation, unsampled tasks, and metric reporting.


njzjz (commented):
Thanks — the second half of this is a clear bug fix, but the accumulator seeding worries me.

## The logging hoist is right

On master the `for _key in self.model_keys:` loop that calls `log_loss_valid` and emits the `_trn`/`_val` lines sits **inside the `else:` (non-`disp_avg`) branch**, so a multi-task run with `disp_avg: true` produced no per-task log lines at all — `train_results` was filled and then dropped on the floor. Hoisting it to its own loop over `model_keys` fixes that, and threading `check_total_rmse_nan=False` for a task with `step_count_per_task == 0` is the right way to keep the deliberate NaN from tripping the NaN guard. `dict.fromkeys(task_losses, float("nan"))` for the zero-step case is a clean replacement for the silent empty dict.

## `initialize_task_loss_accumulator` runs a real training step to learn column names

```python
self.optimizer.zero_grad(set_to_none=True)
task_input, task_label, _ = self.get_data(is_train=True, task_key=_task_key)
if not task_input:
    return
_, _, task_more_loss = self.wrapper(**task_input, cur_lr=pref_lr, label=task_label, task_key=_task_key)
```

I understand *why*: `lcurve.out` writes its header once, so every task's column set has to be known at the first display step, and `more_loss` keys only exist after a forward. But this pays for that with four side effects on a path that is supposed to be pure reporting:

1. **It consumes a training batch.** `get_data(is_train=True, task_key=...)` advances that task's training iterator. So the task the display is about to report as "not sampled this interval" *is* sampled — the batch is drawn, fed forward, and thrown away. That shifts the data stream and epoch bookkeeping for that task by one batch per display step until it is first sampled naturally. Your own test has to mock `dp_random.choice` with an exact `[0, 1]` sequence, which is a hint at how sensitive this is.
2. **No `torch.no_grad()`.** The enclosing block runs after `self.wrapper.eval()`, so this forward builds a full autograd graph in eval mode and discards it — wasted memory and time on every display step until the task is seeded, and for a large task that is not cheap.
3. **`self.optimizer.zero_grad(set_to_none=True)` mutates optimizer state from inside the display path.** The non-`disp_avg` branch does the same thing on master, so there is precedent, but that branch is at least computing numbers it then reports; here the loss is discarded.
4. **DDP.** The `if not task_input: return` early exit is evaluated per rank. If it can ever be true on some ranks and not others, the ranks disagree on whether to run a forward and the collectives desynchronize. Worth confirming `get_data` cannot return a falsy `task_input` on a subset of ranks.

Two directions that avoid all four:

- **Keep the keys, not the counters.** The reset at the end of the display block already zeroes *values* and preserves keys, so a task only needs seeding if it has never been sampled since training started. If the header could be deferred until every task has been seen once, or written with a per-task placeholder set derived from the loss configuration (which terms are enabled: energy / force / virial / …) rather than from an executed `more_loss`, no forward is needed at all.
- **If a forward really is unavoidable**, at minimum wrap it in `torch.no_grad()`, and draw from the validation loader rather than the training one so the training stream is untouched.

## Smaller points

- `initialize_task_loss_accumulator` is redefined on every display step; it does not close over anything that changes except `pref_lr`, so it could be a method or moved above the loop.
- The test asserts `self.assertIn("nan", data_lines[1])` — that checks the literal token appears somewhere in the row. Asserting that the NaN falls in *that task's* columns (via `header_columns.index(...)`) would pin the actual contract; as written it would still pass if the NaN landed in the wrong task's column.
- Worth a line in the test docstring or a comment saying which task is unsampled and why, so the `[0, 1]` mock sequence is not load-bearing but unexplained.

njzjz (approved):

PR #5828, opened 2026-07-16
fix(pt-expt): group LMDB statistics by atom count

OutisLi:
## Summary
- expose fixed-`nloc` statistical partitions for PT-experimental LMDB data while preserving one logical training dataset
- let the common statistics collector consume shape-compatible partitions and cap sampling to each partition's available batches
- add regression coverage for mixed-`nloc` statistics packing and end-to-end PT-experimental training

## Test plan
- [x] `python -m pytest source/tests/pt_expt/test_lmdb_training.py source/tests/consistent/test_make_stat_input.py -v`
- [x] repository pre-commit hooks, including Ruff and pylint
- [ ] full OMat24 initialization (intentionally omitted because of dataset size; covered by synthetic mixed-`nloc` tests)

## Summary by CodeRabbit

* **New Features**
  * Added statistical data collection and batching across statistical systems, including support for mixed fixed-atom-count frames in LMDB datasets.
  * Exposed per-statistical-system access to batch retrieval and system/batch counts for trainer workflows.
* **Documentation**
  * Clarified how statistical views are computed independently from the underlying training dataset.
  * Expanded statistical input semantics for variable atom counts and per-system packing behavior.
* **Tests**
  * Added coverage for mixed-nloc LMDB statistics and end-to-end training with statistical batching enabled.

### DIFF
--- a/deepmd/pt_expt/utils/lmdb_dataset.py
+++ b/deepmd/pt_expt/utils/lmdb_dataset.py
@@ -32,7 +32,10 @@ class LmdbDataSystem:
     Exposes the small surface that pt_expt's trainer touches:
     ``get_batch(sys_idx=None)``, ``add_data_requirements(list)``, and
     ``get_nsystems()``. Internally uses :class:`LmdbDataReader` for I/O and
-    :class:`SameNlocBatchSampler` to draw same-nloc batches.
+    :class:`SameNlocBatchSampler` to draw same-nloc batches. Statistics use a
+    separate logical-system view in which every ``nloc`` group is sampled
+    independently, matching the PyTorch DataLoader adapter without changing
+    the identity of the LMDB as one training dataset.
 
     Parameters
     ----------
@@ -76,6 +79,8 @@ class LmdbDataSystem:
             block_targets=block_targets,
         )
         self._iter = iter(self._sampler)
+        self._stat_nlocs = tuple(sorted(self._reader.nloc_groups))
+        self._stat_offsets = [0] * len(self._stat_nlocs)
 
     # ------------------------------------------------------------------
     # pt_expt trainer surface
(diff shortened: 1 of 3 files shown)


wanghan-iapcm (approved):
The core fix is correct and well-tested: grouping frames by atom count into fixed-nloc partitions is sound, the min(nbatches, ceil(nframes/bs)) cap and sequential offset-with-wraparound visit every frame once per pass, the _collate_indices extraction is behavior-preserving, and the getattr-based shims fall back byte-identically to the legacy get_nsystems()/get_batch(sys_idx=) path for non-LMDB datasets (TF/jax/pd). test_stat_input_partitions_mixed_nloc_batches genuinely fails pre-fix (a single-system LMDB would concatenate ragged mixed-nloc shapes). LGTM.

Two non-blocking follow-ups:

1. Batch-count test coverage. The new test uses 4 frames with batch_size=2 — evenly divisible, so it cannot distinguish the correct ceil-division in get_stat_numb_batches from a floor-division regression, and it never exercises the offset-wraparound reset or the undersampling case (nbatches smaller than available). This is exactly the class of bug that shipped in #5488 (floor-division under-reporting batch counts in the sibling pt LMDB dataloader), caught only when reviewers added non-divisible-boundary tests. Please add a case with a non-divisible frame count (e.g. 5 frames at batch_size=2) and an undersampling case to lock the counting contract.

2. Stale Parameters docstring. In deepmd/utils/model_stat.py the make_stat_input summary now mentions that data sources may expose dedicated statistical-system methods, but the Parameters blocks of both make_stat_input and collect_batches still document the contract as only "must support get_nsystems() and get_batch(sys_idx=)". Since the implementation now probes get_stat_nsystems/get_stat_numb_batches/get_stat_batch and they change the effective per-shape system count, please name those three optional hooks in the Parameters section so the documented contract matches the code.

PR #5863, opened 2026-07-19
CI: run Test C++ workflow on self-hosted runner

njzjz:
## Summary by CodeRabbit

* **Chores**
  * Updated the automated testing workflow to select the runner conditionally for the `testcc` job.
  * Added an explicit “Install GCC” step (including build tools) to improve build reliability.

### DIFF
--- a/.github/workflows/test_cc.yml
+++ b/.github/workflows/test_cc.yml
@@ -14,39 +14,35 @@ name: Test C++
 jobs:
   testcc:
     name: Test C++
-    runs-on: ${{ matrix.runner }}
+    runs-on: ${{ matrix.check_memleak == false && github.repository_owner == 'deepmodeling' && 'cpu' || 'ubuntu-22.04' }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


wanghan-iapcm (approved):

OutisLi:
The intended CI routing is a two-stage process:

1. When a PR is created or updated, the pull_request workflow should run the unit tests on a GitHub-hosted runner.
2. After maintainers approve the PR and add it to the merge queue, the merge_group workflow should run the unit tests on the self-hosted cpu runner. The PR is merged only after this merge-group run succeeds.

The current github.repository_owner == deepmodeling condition does not implement this distinction, because a fork PR targeting deepmodeling/deepmd-kit also runs in the base repository context and therefore has deepmodeling as the repository owner. Please route the self-hosted jobs based on github.event_name == merge_group instead. All pull_request jobs should remain on ubuntu-22.04. The Install GCC condition should use the same routing predicate as runs-on.

PR #5882, opened 2026-07-20
ci(cuda): overlap AOTI .pt2 compiles with GPU unit tests

wanghan-iapcm:
## What

The CUDA python job runs `source/tests` as one serial pytest process. The tests that freeze a `.pt2` (torch.export + AOTInductor) are CPU-bound: inductor / g++ / ptxas generate code for minutes while the GPU sits idle. On the CI GPU the device measured ~98% idle (peak <250 MiB) for the entire duration of that test group, so those minutes are pure serialization in front of the GPU-bound tests.

Since #5896 moved the C++ build + ctest + LAMMPS to a separate `test_cc` job, this PR's scope is the python side plus caching:

- **Two pytest lanes in `test_python`, concurrently on the one GPU:** lane 1 = the CPU-bound `.pt2`-freezing tests (`-m aoti_compile`), `nice`d and capped to half the cores so the GPU lane keeps CPU priority; lane 2 = the GPU-bound remainder (`-m "not aoti_compile"`). Wall-clock becomes `max(lane 1, lane 2)` instead of their sum.
- **Persistent AOTInductor compile cache** (`actions/cache`) restored in BOTH jobs: the `test_cc` fixture builds (`gen_*.py`) drive the same inductor backend on the same architectures, so they reuse kernels the python lanes (or previous runs) already compiled, and unchanged models never recompile across runs.
- **Per-lane JUnit reports** uploaded as artifacts for lane-balance monitoring.

## How the partition stays correct

The `aoti_compile` marker is auto-applied in `source/tests/conftest.py` to any test whose module references a `.pt2` freeze entry point (`deserialize_to_file` / `_trace_and_export` / `aoti_compile_and_package`). No hand-maintained file list — new compiling tests are partitioned automatically. The two selections are complementary and exhaustive (verified locally: A + B = total, disjoint). A fail-fast guard aborts if an AOT compile happens in an untagged test.

The custom-op `.so` install in `test_cc_local.sh` is **atomic** (`copy` to a temp file + `os.replace`) rather than in place: overwriting the destination inode while another process has it `mmap`'d corrupts its live code pages and SIGSEGVs it. `os.replace` swaps the directory entry to a new inode, so live mappings stay intact.

## Measured on CUDA CI

Lane wall-clocks from the green run at sha `86128ef7` (measured under the pre-#5896 single-job layout; the pytest lanes are unchanged by the rebase):

- Lane 1 (`aoti_compile`): **4063 s (~68 min)**
- Lane 2 (`-m "not aoti_compile"`): **8158 s (~2h16m)**

Serial python pytest is by construction the sum, 12221 s (~3h24m); the overlap runs them in `max(4063, 8158) =` **8158 s (~2h16m)** — the `test_python` job's pytest step drops by **~68 min (~33%)**, with the entire compile stream hidden behind the GPU lane. The inductor cache additionally removes recompilation of unchanged models from both jobs on subsequent runs.

## Notes / limitations

- Overlap quality scales with the runner's core count (`nice` is a no-op unless the CPU saturates); the numbers above are for the CI runner.
- Whole-file granularity: a compiling module's few non-compiling tests also run in the compile lane (harmless).
- The compile lane tolerates pytest exit 5 ("no tests collected") so the split is safe on branches without `.pt2`-freezing tests.
- The two jobs share one cache key per sha; if both try to save it, the second save is skipped with a warning (caches are immutable), which is harmless.
- Generic CI orchestration, independent of any feature branch.

## Summary by CodeRabbit

* **Tests / CI**
  * Improved CUDA CI by running AOT-related and non-AOT GPU test groups concurrently, with separate JUnit reports and “no tests collected” treated as success.
  * Persisted the compilation cache across workflow runs to speed up repeated runs.
  * Added formal AOT compilation test tagging via a new pytest marker and a fail-fast guard if AOT compilation occurs without the expected tag.
* **Build & Test Scripts**
  * Updated local C++ test runner to support skipping build vs ctest phases, improve runtime library setup, handle leak-sanitizer generator behavior, and generate fixtures more reliably (clean stale outputs and run generation in parallel).

### DIFF
--- a/.github/workflows/test_cuda.yml
+++ b/.github/workflows/test_cuda.yml
@@ -94,17 +94,20 @@ jobs:
           # Cap the compile lane's inductor compile threads to ~half the cores
           # so the GPU lane keeps CPU for its host-side dispatch; nice it too.
           nc=$(( $(nproc) / 2 )); [ "$nc" -lt 1 ] && nc=1
-          ( t0=$(date +%s)
+          # rc=0 + `|| rc=$?` so a lane's failure does not let the step's
+          # default `set -e` skip the wall-clock echo (the timing we want most
+          # when a lane fails).
+          ( t0=$(date +%s); rc=0
             TORCHINDUCTOR_COMPILE_THREADS=$nc nice -n 15 \
               python -m pytest source/tests -m "aoti_compile" \
-                -p no:cacheprovider --junit-xml=junit-aoti.xml
-            rc=$?; echo "[lane aoti_compile] wall=$(( $(date +%s) - t0 ))s exit=$rc"
+                -p no:cacheprovider --junit-xml=junit-aoti.xml || rc=$?
+            echo "[lane aoti_compile] wall=$(( $(date +%s) - t0 ))s exit=$rc"
             exit $rc ) &
           pid_a=$!
-          ( t0=$(date +%s)
+          ( t0=$(date +%s); rc=0
             python -m pytest source/tests -m "not aoti_compile" \
-              -p no:cacheprovider --junit-xml=junit-gpu.xml
-            rc=$?; echo "[lane gpu] wall=$(( $(date +%s) - t0 ))s exit=$rc"
+              -p no:cacheprovider --junit-xml=junit-gpu.xml || rc=$?
+            echo "[lane gpu] wall=$(( $(date +%s) - t0 ))s exit=$rc"
             exit $rc ) &
           pid_b=$!
           rc_a=0; wait $pid_a || rc_a=$?
--- a/.github/workflows/test_cuda.yml
+++ b/.github/workflows/test_cuda.yml
@@ -63,7 +63,29 @@ jobs:
           DP_ENABLE_NATIVE_OPTIMIZATION: 1
           DP_ENABLE_PYTORCH: 1
       - run: dp --version
-      - run: python -m pytest source/tests
+      - name: Run Python tests (overlap AOTI compiles with GPU unit tests)
+        # The tests that freeze a .pt2 (-m aoti_compile) are CPU-bound: the GPU
+        # sits ~idle (measured ~98% idle, <250 MiB) while inductor/g++/ptxas
+        # generate code. Run them concurrently with the GPU-bound tests that do
+        # NOT compile a .pt2 so the compile CPU-time overlaps the GPU unit tests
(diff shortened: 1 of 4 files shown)


wanghan-iapcm:
Thanks — you're right that a source grep is the wrong basis for the scheduling boundary, on both counts. Replaced it (f660993, b4a7996).

The `aoti_compile` set is now an explicit curated list of the modules that actually freeze a `.pt2`, derived by running the suite under a detector that wraps the real compile entry points (`aoti_compile_and_package`, `_trace_and_export`) instead of scanning text:

- **Missed freeze(...) cases** — `test_dp_freeze`, `test_finetune`, `test_multitask`, `test_dp_test`, `test_change_bias`, `infer/test_models` are now in the compile lane (none name the low-level entry points, so the grep skipped them all).
- **Over-tagging** — the detector confirmed 8 grep-candidate files (`test_dpa1`, `test_dpa4`, `test_dpa1_triton`, `test_get_model_dpa4`, `test_dpa2_graph_lower`, `test_sezm_model`, `test_entrypoint`, `test_descriptor_dpa1_triton`) never actually compile (mock / `.pth` / `.pte`), so they stay in the GPU lane.

An always-on guard wraps the same entry points every run and **fails the session** (with the offending nodeids) if any test freezes a `.pt2` without the marker, so the list can't silently drift; it runs in the CPU `test_python` CI too, so a missed producer is caught cheaply rather than only in the CUDA job.

Other points: compile-thread constraint added (`TORCHINDUCTOR_COMPILE_THREADS=nproc/2` + `nice`); each lane now echoes its wall-clock and uploads its own JUnit report to compare T_A/T_B against the ~3h46m serial baseline. The current labelled CUDA run is on the pre-fix commit (grep split, so a lower bound); I'll re-trigger on the fixed markers and post the lane timings + GPU util/VRAM before relying on the concurrency beyond what those numbers support. Agreed the lane isn't categorically CPU-only — the ~98% idle / <250 MiB is empirical and the benchmark will confirm on the fixed split.

Also made the inductor compile cache explicit (shared `TORCHINDUCTOR_CACHE_DIR` + `actions/cache`) so the C++ fixture builds reuse the Python lane's kernels. (The `UnicodeDecodeError` note is moot now — the conftest no longer reads test source.)

njzjz (approved):
LGTM

PR #5779, opened 2026-07-13
feat(dpmodel,pt_expt,cc): graph-native DPA-2 — first message-passing model on the NeighborGraph lower (PR-G-dpa2)

wanghan-iapcm:
DPA-2 becomes graph-native full-stack — the first **message-passing** model on the NeighborGraph lower (PR-G-dpa2 of the [NeighborGraph series](https://github.com/wanghan-iapcm/deepmd-kit/discussions/4); follows dpa1's route from #5581/#5583/#5604/#5714/#5715/#5717/#5733). The dense path coexists untouched; the graph route is opt-in for inference (`--lower-kind graph`) and the default for pt_expt training/eager on graph-eligible configs, exactly like dpa1.

## What's new

**dpmodel (backend-agnostic math)**
- `DescrptBlockRepformers.call_graph`: every repformer op ported to the flat edge list — conv/grrg/drrd/g1g1/`LocalAtten` via `segment_*` over `dst`; the dense `nnei×nnei` gated attention (`Atten2Map`/`Atten2MultiHeadApply`/`Atten2EquiVarApply`) via `center_edge_pairs(ordered, include_self)` + per-head `segment_softmax` (extended to trailing feature dims).
- `DescrptDPA2.call_graph` + `uses_graph_lower` + dense-call adapter. The multi-level nlist (`build_multiple_neighbor_list`) becomes per-block edge masks; in the shape-static adapter layout the per-block **slice** replicates the dense `nlist[:, :, :ns]` truncation, making the adapter **bit-exact vs the dense path at ANY sel** (verified to 5e-16 at deliberately binding sel with attention enabled) — this is what keeps the cross-backend consistency suites green after the routing flip. Ineligible configs (`use_three_body`, compression, spin) keep the legacy dense route.
- Owned-node energy mask: `fit_output_to_model_output_graph` consumes `n_local`, excluding halo rows from the differentiated energy (prevents cross-rank double counting; force stays full-N, halo partials reverse-commed by LAMMPS).

**pt_expt / export**
- Routing, autograd force/virial, compiled training, and freeze eligibility are inherited generically from the dpa1 machinery.
- Per-layer MPI halo refresh on the graph path: `_exchange_ghosts_graph` — identity on ghost-free single-rank graphs (`src` IS the owner), `deepmd_export::border_op` overwrite of halo rows on extended-region multi-rank graphs.
[...]

PR #5884, opened 2026-07-20
feat(dpa4): DPA4/SeZM on the NeighborGraph lower — graph-native port, multi-rank, native spin, charge-spin/bridging

wanghan-iapcm:
## Summary

Brings the DPA4/SeZM descriptor fully onto the NeighborGraph ("graph") lower — the sel-free, edge-native route that dpa1 (#5583) and dpa2 (#5779) already use — and adds the conditioning inputs DPA4 needs on that route. Stacked on the merged dpa2 graph work (#5779); this PR is DPA4's turn on the same infrastructure.

Four layers, built and reviewed incrementally:

1. **DPA4 graph-native port** — dense `call` becomes a thin adapter over one graph-native math owner (`_call_graph_impl`); the model-level graph seam drives DPA4 through `call_graph`. Parity vs the dense route at fp64 1e-12; parity vs the pt reference at ~6e-15.
2. **Multi-rank (with-comm) graph `.pt2`** — per-block `border_op` ghost-feature exchange on the graph lower; graph-kind `.pt2` embeds the nested with-comm artifact. Dense (nlist) lower stays comm-less and fails fast on multi-rank; bridging models also fail fast (a rank cannot observe a ghost owner's full outgoing-edge set).
3. **Native spin (magnetic moments)** — per-local-atom spin `(nf, nloc, 3)` conditions the descriptor and is a second autograd leaf giving `force_mag = -dE/dspin`. Rides the graph route only; frozen to a graph-kind `.pt2` (spin at positional ABI index 10), evaluated in Python (`DeepEval`) and C++ (`DeepSpinPTExpt` graph route), run in LAMMPS (`pair_style deepspin`). Single-rank. `DPA4NativeSpinModel` in dpmodel + pt_expt.
4. **Charge-spin FiLM + SFPG bridging** — the last dense-only conditioning inputs, now wired onto the graph route (the edge-native trunk already implemented both; this threads `charge_spin` and flips the capability gates).

## Validation

- dpmodel ↔ pt reference DPA4 parity: 387 tests.
- native-spin autograd `force_mag` vs finite difference: 5.5e-10 (atol 1e-6); pt weight-copied parity 1e-12.
- charge-spin graph-vs-dense parity 1e-12; SFPG bridging frozen-sphere invariance <1e-10 (ablation reopens the leak >1e-6).
- C++ (CUDA, Tesla T4): `deeppot_universal` DPA4 dense + graph rows 38 passed / 8 skipped (NoPbc profile); native-spin `DeepSpinPTExpt` graph gtest 16 passed (double/float × PBC/NoPbc).
- LAMMPS (CUDA): DPA4 graph energy/force vs live DeepEval ~1e-13; native-spin `deepspin` energy/force/force_mag; 2-rank multi-rank fail-fast aborts cleanly.
- Full Python native-spin + charge-spin + bridging battery green on CUDA (device-conditional tolerances where the graph-route `index_add` atomics introduce 1-2 fp64 ULP run-to-run nondeterminism).

## Known limitations

- **Single-rank only for spin / charge-spin FiLM / bridging.** Multi-rank spin-graph (ghost-spin exchange) is a follow-up; the graph-kind spin `.pt2` carries `has_comm_artifact=false` and C++ fails fast on multi-rank. Bridging is intrinsically single-rank (per-node freeze fold over the full outgoing-edge set).
- **Charge-spin FiLM + native spin combined** is rejected at build (`add_chg_spin_ebd` + native spin) — a small follow-up.
- **dpmodel is energy-only** for the native-spin wrapper; force/force_mag come from pt_expt autograd.
- **Whole-model conversion of pt-serialized `sezm_native_spin` checkpoints is not claimed** — the deserialize alias covers the type string with dpmodel structure only.
- **Sel-free carry-all divergence when sel binds** — the graph route uses the full physical edge set, so it diverges from a sel-truncated dense route only when sel actually binds (the same deliberate divergence as dpa1/dpa2).
- **The legacy pt-side `edge_vec` `.pt2` lower schema** (DPA4's original inference rail) is quarantined with `DeprecationWarning`, not removed — its removal is a follow-up gated on repointing the pt-SeZM freeze (and, for its spin part, on graph-spin support, now landed here).

## Summary by CodeRabbit

* **New Features**
  * Added native per-atom spin conditioning for DPA4/SeZM on the NeighborGraph route, including magnetic outputs (e.g., `force_mag`) and spin masking.
  * Introduced native-spin `.pt2` export/inference support on the graph route (with appropriate single-rank/MPI behavior).
  * Extended graph-route input/output handling for spin and charge+spin routing.
* **Bug Fixes**
  * Improved energy+spin loss for flat per-atom label shapes via consistent reshaping.
* **Documentation**
  * Expanded guidance on graph vs dense capabilities, native-spin limitations, and legacy `edge_vec` deprecation.
* **Tests**
  * Added/extended coverage for graph routing, spin outputs, export parity, and multi-rank fail-fast/empty-rank scenarios.


wanghan-iapcm:
Validation note: the full CUDA battery was run on a Tesla T4 (C++ `deeppot_universal` DPA4 dense+graph 38 passed/8 skipped; native-spin `DeepSpinPTExpt` graph gtest 16 passed; Python native-spin+charge-spin+bridging battery green; DPA4-graph LAMMPS + native-spin MPI fail-fast passed). The single-rank native-spin `test_lammps_dpa4_spin_graph_pt2.py::test_pair_deepspin`/`_virial` were refactored from stale hardcoded references (a post-merge `dpa4_nn` numerics shift) to a live `DeepPot.eval` reference in the last two commits; that specific re-run is pending a remote-box re-provision, but CI regenerates the fixtures and exercises the same path.

deepmd/dpmodel/descriptor/dpa4.py
```diff
@@ -1562,8 +1481,6 @@ def call_with_edges(
             bridging_switch=self.bridging_switch,
             edge_envelope=self.edge_envelope,
             radial_basis=self.radial_basis,
-            has_exclude_types=bool(self.exclude_types),
-            edge_type_keep_mask=self._edge_type_keep_mask,
             # Random local-Z roll is a training-only augmentation;
             # the model is roll-equivariant, so inference fixes gamma.
             random_gamma=False,
```
OutisLi: [P1] Preserve PT training-time random_gamma semantics

The PT reference passes `random_gamma=self.random_gamma and self.training` in both its dense and edge paths. This shared core always passes `False`, and the pt_expt descriptor inherits it without an override, so the default `random_gamma=True` is silently ignored for all pt_expt DPA4 training through both the graph route and dense adapter. I verified at this HEAD that a train-mode PT descriptor forwards `True` to its edge-cache builder while the corresponding pt_expt descriptor forwards `False`. The new graph/native-spin parity tests explicitly set `random_gamma=False`, so they do not cover the default behavior. Please introduce a pt_expt runtime hook or argument equivalent to `self.random_gamma and self.training`, while keeping dpmodel/eval inference deterministic, and add train/eval coverage.

OutisLi (changes requested):
The PT-equivalence contract is currently broken for the default training-time random-gamma behavior. Please restore that behavior and cover it with train/eval tests; details are in the inline comment.

deepmd/pt_expt/model/get_model.py
```diff
@@ -205,6 +212,46 @@ def get_sezm_model(data: dict) -> EnergyModel:
...
+    if data["descriptor"].get("type", "dpa4") not in ("dpa4", "DPA4", "sezm", "SeZM"):
+        raise NotImplementedError(
+            "spin scheme 'native' requires the DPA4/SeZM descriptor"
+        )
+    if data["descriptor"].get("add_chg_spin_ebd", False):
+        raise NotImplementedError(
+            "charge-spin FiLM combined with native spin on the graph route "
+            "is a follow-up; use one or the other"
+        )
+    use_spin = [bool(flag) for flag in spin_cfg["use_spin"]]
```
OutisLi: [P1] Normalize symbolic/indexed `use_spin` before boolean conversion

The public schema accepts `list[bool]`, `list[int]`, and `list[str]`, and the PT reference uses `_normalize_spin_use_spin` to expand type indices or element symbols against `type_map`. Direct boolean conversion changes that contract: with `type_map=["Ni", "O"]`, `use_spin=["Ni"]` becomes `[True]` and fails with ``use_spin` length must equal `ntypes``, while `use_spin=[0, 1]` becomes `[False, True]` instead of selecting both types. Valid `input.json` files can therefore either fail to construct or silently gate the spin embedding and `mask_mag` for the wrong species. Please apply the same normalization as PT before constructing `Spin` and the descriptor, and add pt_expt coverage for both symbol and index forms.

OutisLi (changes requested):
PT-equivalence review: the public `spin.use_spin` forms are not normalized consistently in pt_expt.

deepmd/pt_expt/model/get_model.py
```diff
@@ -205,6 +212,46 @@ def get_sezm_model(data: dict) -> EnergyModel:
...
+    spin = Spin(
```
OutisLi, line 281: [P1] Preserve `spin.allow_missing_label` in the constructed model

The PT reference passes `allow_missing_label=model_params["spin"].get("allow_missing_label", False)` into `Spin`, but this builder omits it. Consequently a normalized config with `allow_missing_label=True` produces `model.spin.allow_missing_label == False`; `get_additional_data_requirement()` then marks the spin input as `must=True`, so datasets intentionally lacking `spin.npy` are rejected instead of receiving the documented zero default. Please forward this field to `Spin(...)` and add a regression asserting that the resulting spin data requirement has `must=False` and `default=0.0`.

OutisLi (changes requested):
PT-equivalence review: `spin.allow_missing_label` is accepted by the public config but dropped by the pt_expt native-spin builder.

deepmd/dpmodel/descriptor/dpa4.py
wanghan-iapcm: Fixed in [3c108c4b5](https://github.com/deepmodeling/deepmd-kit/commit/3c108c4b5073995e174bc077f529c8fdb67a1984). The shared trunk now passes `random_gamma=self.random_gamma and self._in_training_mode()`; the `_in_training_mode()` runtime hook is `False` on dpmodel (inference/reference backend, deterministic) and overridden in the pt_expt wrapper with the torch module's `training` flag, restoring pt's `self.random_gamma and self.training` semantics on both the graph route and the dense adapter. Export stays fixed-gamma (the freeze path calls `model.eval()` before tracing).

Coverage added: a pt_expt spy test at the edge-cache seam pins train→`True` / eval→`False` / dpmodel→`False` plus eval-mode bit-determinism (spy-based because the model is roll-equivariant, so an output-difference check has no deterministic teeth), and a dpmodel test pins the hook `False` with bit-identical repeated calls at `random_gamma=True`.

One known residual: under the *optional* inductor training compile, the numpy gamma draw would be baked at trace time (lost augmentation, not wrong physics — roll equivariance). Eager training (the default) draws fresh per step like pt. A torch-native draw behind the array-API seam is a follow-up.

deepmd/pt_expt/model/get_model.py
wanghan-iapcm: Fixed in [73a851f09](https://github.com/deepmodeling/deepmd-kit/commit/73a851f094b0b012de9b5988b9bbb11761a44821). Added a pure shared helper `deepmd.utils.spin.normalize_spin_use_spin(use_spin, type_map) -> list[bool]` implementing the same three-form contract as pt's `_normalize_spin_use_spin` (boolean passthrough / index scatter / symbol lookup with `ValueError` on unknown symbols), and both native-spin builder twins (dpmodel `get_dpa4_native_spin_model` and the pt_expt counterpart) now normalize through it before constructing `Spin` and the descriptor.

Coverage added: direct unit tests for the helper (all three forms, empty list, unknown-symbol raise, input purity) plus config-form parametrizations on both builders asserting the normalized mask reaches `model.spin.use_spin` AND the descriptor (`['Ni']` → `[True, False]`, `[0, 1]` → `[True, True]`, etc.).

deepmd/pt_expt/model/get_model.py
wanghan-iapcm: Fixed in [73a851f09](https://github.com/deepmodeling/deepmd-kit/commit/73a851f094b0b012de9b5988b9bbb11761a44821). Both native-spin builder twins now forward `allow_missing_label=spin_cfg.get('allow_missing_label', False)` into `Spin(...)`.

Regression added as requested: with `allow_missing_label=True` the constructed model reports `model.spin.allow_missing_label is True` and `get_additional_data_requirement` yields the spin item with `must=False` and `default=0.0`; the default-config branch is pinned too (`must=True`).

deepmd/pt_expt/train/training.py
```diff
@@ -261,6 +261,22 @@ def get_additional_data_requirement(_model: Any) -> list[DataRequirementItem]:
                 "aparam", _model.get_dim_aparam(), atomic=True, must=True
             )
         )
+    if _model.has_spin():
```
OutisLi, line 264: [P1] Preserve PT inference requirements for ener_spin change-bias

`pt_expt change-bias` first constructs `get_loss({"inference": True}, ...)` and then calls this helper. The PT `EnergySpinLoss` explicitly enables every `has_*` flag in inference mode, but the dpmodel loss used by pt_expt has no `inference` parameter and silently absorbs it through `**kwargs`; with the default zero prefactors its `label_requirement` is therefore empty. This branch loads `spin`, but `energy` is never loaded. I reproduced this on the NiO spin data at this HEAD: the command path completes with a zero bias delta, whereas adding the missing energy requirement changes both type biases by `-5.15069006`. Please preserve the PT inference semantics in `EnergySpinLoss` and add an end-to-end native-spin `change-bias` regression so this cannot silently no-op.

OutisLi (changes requested):
The native-spin change-bias path still diverges from the PT reference and can silently leave the output bias unchanged; see the inline comment.

deepmd/pt_expt/model/get_model.py
```diff
@@ -205,6 +213,49 @@ def get_sezm_model(data: dict) -> EnergyModel:
...
+        key with ``scheme == "native"``.
+    """
+    data = copy.deepcopy(data)
+    spin_cfg = data.pop("spin")
+    data.setdefault("descriptor", {})
+    if data["descriptor"].get("type", "dpa4") not in ("dpa4", "DPA4", "sezm", "SeZM"):
+        raise NotImplementedError(
+            "spin scheme 'native' requires the DPA4/SeZM descriptor"
+        )
+    if data["descriptor"].get("add_chg_spin_ebd", False):
```
OutisLi: [P1] Preserve PT native spin combined with charge-spin FiLM

The PT reference does not reject this configuration: `SeZMNativeSpinModel.forward` and its exportable lower both accept `charge_spin` alongside the native `spin` input. I built the same `add_chg_spin_ebd=True` native-spin model at this HEAD; PT returns `energy`, `force`, and `force_mag`, and changing only `charge_spin` changes the energy (`5.97e-05` in the probe), whereas pt_expt stops here with `NotImplementedError`. This is not a limitation of the migrated graph math or eager wrapper: bypassing this builder guard and constructing the current `DPA4NativeSpinModel` over the same descriptor produces both a nonzero charge-spin response and a nonzero magnetic force. Please port the combined public configuration rather than dropping PT behavior, including a `charge_spin` slot in the native-spin graph export/metadata and the corresponding serialization, DeepEval/C++ inference, training, and freeze regressions.

OutisLi (changes requested):
The pt_expt native-spin migration still rejects a public conditioning combination supported by the PT reference; see the inline comment.

source/tests/common/dpmodel/test_dpa4_call_graph.py
```diff
@@ -0,0 +1,977 @@
...
+    coord[:, 1, :] = directions  # B rotates around A, radius fixed
+    coord[:, 2, :] = np.array([2.4, 0.0, 0.0])  # C: ordinary neighbor, static
+    atype = np.tile(np.array([[0, 1, 0]], dtype=np.int64), (n_points, 1))
+    nlist = build_neighbor_list_np(coord, dd.get_rcut(), nnei=4)
+    graph = make_graph_from_nlist(coord, nlist)
+    out, _ = dd.call_graph(graph, atype.reshape(-1))
+    return np.asarray(out).reshape(n_points, 3, -1)
+
+
+def test_call_graph_bridging_frozen_sphere_invariance() -> None:
```
OutisLi, line 538: [P1] Implement and test the complete PT bridging model

This regression proves only the descriptor-side `InnerClamp`/SFPG invariance. In the PT reference, `bridging_method="ZBL"` also constructs a model-level `InterPotential` and adds its analytical ZBL atomic energy before the shared edge autograd, which is what supplies the short-range repulsive energy, force, and virial. That half is absent here: pt_expt still rejects every non-`none` `bridging_method`, its PT-checkpoint deserializer rejects bridging models, and there is no `InterPotential` equivalent in dpmodel/pt_expt. The current PT regressions confirm that ZBL creates the potential and adds positive energy, while the pt_expt regression confirms only `NotImplementedError`. Consequently users cannot construct the claimed single-rank bridging model, and manually enabling these descriptor switches would omit the physical ZBL term. The separate DP-ZBL model is not equivalent to SeZM's native SFPG plus additive ZBL path. Please port the model-level potential and add PT-parity training, freeze, DeepEval/C++ inference tests rather than treating the descriptor-only gate as complete bridging support.

OutisLi (changes requested):
The new graph tests cover only the SFPG descriptor gate, while the PT model-level analytical ZBL half of bridging is still absent and rejected in pt_expt; see the inline comment.

deepmd/dpmodel/model/base_model.py
```diff
@@ -135,6 +144,18 @@ def deserialize(cls, data: dict) -> "BaseBaseModel":
...
 
                     return SpinModel.deserialize(data)
+                if model_type in ("dpa4_native_spin", "sezm_native_spin"):
+                    # DPA4NativeSpinModel is not a BaseModel subclass either
+                    # (same reasoning as SpinModel above); dispatch directly.
+                    # NOTE: the "sezm_native_spin" alias only covers the wire
+                    # TYPE STRING with dpmodel structure -- it does not claim
+                    # whole-model conversion of pt-serialized checkpoints
+                    # (pt's SeZMAtomicModel dict layout differs).
+                    from deepmd.dpmodel.model.dpa4_native_spin_model import (
```
OutisLi: [P1] Keep native-spin deserialization backend-aware and model-agnostic

Even without claiming whole-model conversion of PT-serialized native-spin checkpoints, the pt_expt backend cannot round-trip its own native-spin serialization through its public BaseModel entry point. At this HEAD, deepmd.pt_expt.model.model.BaseModel.deserialize(model.serialize()) enters this hard-coded dpmodel branch and returns deepmd.dpmodel.model.dpa4_native_spin_model.DPA4NativeSpinModel, not a torch.nn.Module; its backbone consequently lacks forward_common_lower_graph_exportable. This completely bypasses the pt_expt subclass's dedicated deserialize() override. One concrete production failure is finetuning from a native-spin .pte: training.py calls BaseModel.deserialize(data["model"]).to(DEVICE), but the returned NumPy model has no .to().

Please make this dispatch backend-aware and add a public-BaseModel serialize/deserialize regression asserting the pt_expt class, torch.nn.Module, .to(), graph-export method, and forward parity. More generally, native spin is a model input/output/autograd contract rather than a DPA4-specific base-model type; please factor a shared native-spin wrapper/registry/abstraction so future native-spin descriptors do not require additional hard-coded dpa4_native_spin branches.

OutisLi (changes requested):
Requesting changes for the backend-breaking native-spin BaseModel deserialization dispatch described inline.

deepmd/pt_expt/utils/serialization.py
```diff
@@ -1292,15 +1402,33 @@ def _trace_and_export(
...
+    # Detect spin model. Two flavors share the ``is_spin`` gate below (both
+    # need the spin-only metadata fields — ``ntypes_spin``/``use_spin`` —
+    # and the nlist-lower spin ABI probes), but only the NATIVE flavor
+    # (``dpa4_native_spin``/``sezm_native_spin``, ``DPA4NativeSpinModel``)
+    # rides the graph lower: the virtual-atom flavor (``spin_ener``,
+    # ``SpinModel``) doubles the atom count and has no graph-lower
+    # implementation. ``is_native_spin`` distinguishes them at every seam
+    # below (model rebuild, graph rejection, graph sample-input/dynamic-shape
+    # ABI, trace call site).
+    is_native_spin = data["model"].get("type") in (
```
OutisLi: [P1] Route native-spin freeze to graph before choosing the export ABI

The public freeze path still defaults --lower-kind to nlist and chooses a .pte suffix from that value, while this native-spin wrapper has no dense/nlist lower. I reproduced the ordinary default path at this HEAD with a freshly built native-spin model: deserialize_to_file(..., lower_kind="nlist") reaches the dense-spin branch and fails with TypeError: make_model.<locals>.CM.forward_common_lower_exportable() got multiple values for argument 'fparam'. Existing native-spin export tests all pass lower_kind="graph" explicitly, so they do not cover the user-facing default. This also contradicts the new documentation saying native-spin models are always graph-frozen regardless of --lower-kind.

Please resolve native-spin to graph at the public freeze layer before selecting both the export ABI and default output suffix (or reject nlist explicitly and make the default auto/graph), and add a regression that freezes a native-spin checkpoint through the default public CLI/function path without supplying --lower-kind, asserting a graph-kind .pt2 artifact.

OutisLi (changes requested):
Requesting changes for the broken default native-spin freeze route described inline.

deepmd/pt_expt/train/training.py
wanghan-iapcm: same issue in energy model. not in the scope of the current pr. will be fixed by a future pr.

deepmd/dpmodel/model/base_model.py
wanghan-iapcm: Fixed in [24abf3ce3](https://github.com/deepmodeling/deepmd-kit/commit/24abf3ce35850843e50cca24d24b1943a02c5ec8), [71239f945](https://github.com/deepmodeling/deepmd-kit/commit/71239f945a3ceef23eb04836b7c74af1074b39a4), [b03715416](https://github.com/deepmodeling/deepmd-kit/commit/b03715416bb52c73ca5182f01af8737dfa17a89f) — redesigned per your second point rather than patching the dispatch.

Native spin is now a `make_native_spin_model(T_Model)` **class factory** (the make_model pattern, mirroring pt's is-a `SeZMNativeSpinModel(SeZMModel)`): each backend instantiates it on its OWN standard `EnergyModel` and registers the concrete `NativeSpinEnergyModel` in its OWN plugin registry under the descriptor-agnostic wire type `native_spin`. The hard-coded `dpa4_native_spin` branch in `BaseBaseModel.deserialize` is deleted entirely — dispatch is backend-aware by construction, and `deepmd.pt_expt.model.model.BaseModel.deserialize(model.serialize())` returns the pt_expt `torch.nn.Module` with `.to()` and the graph-export machinery. Serialization is the make_model flat shape + a `spin` field (the old nested wrapper shape and its dpa4-specific wire strings are gone; nothing was released).

A future native-spin descriptor needs zero model/dispatch/builder code: builders gate on the `descriptor.supports_native_spin()` capability instead of a descriptor-type list, and a standard-typed config with `spin.scheme='native'` now dispatches generically.

Regression added with your exact asserts: public-BaseModel round trip → pt_expt class, `torch.nn.Module`, `.to(DEVICE)` (the finetune failure), `forward_common_lower_graph_exportable`, and forward parity at 1e-12 (`TestPublicBaseModelRoundTrip`), plus dpmodel wire-type/registry pins.

njzjz (approved):

deepmd/pt_expt/model/get_model.py
wanghan-iapcm: Ported in [c4ddea343](https://github.com/deepmodeling/deepmd-kit/commit/c4ddea343ab5b6a25049883bd2fda63039efe80f), [57a008e30](https://github.com/deepmodeling/deepmd-kit/commit/57a008e30cf30622571b2f52c1cebac70e0eb52d), [9cf9f36fc](https://github.com/deepmodeling/deepmd-kit/commit/9cf9f36fc4f95863f778f087b6f77837a17bd28e), [d1dac261e](https://github.com/deepmodeling/deepmd-kit/commit/d1dac261e5061c17f2949e35a58ff42fedca66f8) — the combined public configuration is now supported end to end rather than rejected:

- **Builders**: both guards lifted; `add_chg_spin_ebd` + `spin.scheme='native'` constructs the combined model in dpmodel and pt_expt.
- **Eager**: weight-copied fp64 parity vs pt's `SeZMNativeSpinModel` with the same `charge_spin` at rtol/atol 1e-12 (energy/force/force_mag, two conditioning values), plus nonzero charge-spin AND spin responses in the same model (your probe, integer-valued: the `ChargeSpinEmbedding` lookup is categorical).
- **Graph export ABI**: `charge_spin` occupies a conditional slot-13 tail (spin stays pinned at index 10); symbolic-trace parity at 1e-12 with a LIVE slot (an integer-valued change moves the traced energy — pins that it is not baked), dynamic-shape spec + synthetic sample + freeze metadata (`has_chg_spin_ebd`/`dim_chg_spin`) included.
- **Serialization/freeze/DeepEval**: a combined model freezes to a graph `.pt2` and DeepEval evaluates it end to end through the compiled artifact — slot live, parity vs eager at 1e-10; `DeepPot.eval(..., spin=..., charge_spin=...)`.
- **Training**: 2-step trainer smoke on the NiO spin data through the `default_chg_spin` metadata path (this also exposed and fixed a latent dpmodel bug: the default-branch converted the numpy default with the numpy namespace against a torch device).
- **C++**: `DeepSpinPTExpt::run_model_graph` appends the conditional charge_spin tail from `default_chg_spin` metadata, mirroring its nlist/edge siblings (metadata-default is the C++ production contract, as elsewhere; compiles clean).

Known residual: no dedicated C++ gtest row for a combined fixture yet (the Python DeepEval e2e covers the identical artifact); multi-rank stays out of scope (native spin has no with-comm ABI by contract).

source/tests/common/dpmodel/test_dpa4_call_graph.py
wanghan-iapcm: Ported in [0c388bec8](https://github.com/deepmodeling/deepmd-kit/commit/0c388bec89188787c3a666a8cac7b99a3cc09780), [32514115e](https://github.com/deepmodeling/deepmd-kit/commit/32514115ee68a82470b4a3f8aaa1ba406f38f9ee), [412cc2129](https://github.com/deepmodeling/deepmd-kit/commit/412cc2129cf4401e016d7989c33b9433a501ea20), [69a2697f0](https://github.com/deepmodeling/deepmd-kit/commit/69a2697f059120496d92ace42059107813c858a6) — the complete model-level ZBL term is now implemented, not just the descriptor-side SFPG/InnerClamp gate:

- **`InterPotential` (backend-agnostic)**: exact array-API port of pt's universal-ZBL math (constants/screening verbatim; per-edge half energies scattered per atom; virtual-type wrap+mask; fp64). Unit tests port pt's known-value OO/OH cases, masking, and error paths, plus a torch-namespace autograd smoke.
- **Injection**: owned by `DPAtomicModel` — added to the per-atom energy after fitting+masking, before the model layer's edge autograd (pt's sezm Step 5). `graph.edge_vec` there is pt_expt's autograd leaf, so the ZBL force/virial flow through the same `edge_energy_deriv` backward as the learned energy. The dense (nlist) route raises for bridging models (pt has no dense conservative path; fail-closed beats silently dropping the term).
- **Builders/serialization/interop**: `bridging_method`/`bridging_r_inner`/`bridging_r_outer` build working models in dpmodel and pt_expt (radii feed the descriptor's InnerClamp/BridgingSwitch, as in pt); `bridging_method` rides the atomic dict (emitted only when active); the pt-checkpoint deserializer now maps pt's wrapper key instead of rejecting — a pt-serialized ZBL checkpoint loads into pt_expt with 1e-12 energy parity.
- **Regressions**: weight-copied fp64 parity vs pt's `SeZMModel` with ZBL at 1e-12 (energy/force/virial); ZBL-vs-plain positive repulsion on identical weights; central-FD force; graph `.pt2` freeze + DeepEval through the compiled artifact at 1e-10; 2-step training smoke; native-spin × bridging composition (pt supports it — ours composes via the atomic layer); `bridging_r_*` defaults; with-comm gate off (pt's `supports_edge_parallel()==False` contract → `has_comm_artifact=False`, single-rank only). argcheck's `doc_only_pt_supported` flags dropped from the three keys.

Known residuals: no dedicated C++ gtest/LAMMPS fixture row (the term is baked into the AOTI artifact — C++ needs no code change; the Python DeepEval e2e exercises the identical compiled artifact), and multi-rank stays off by the pt contract.

deepmd/pt_expt/model/get_model.py
OutisLi: The implementation itself now passes my targeted PT-parity, graph-export, training, and compiled-artifact checks, but the current HEAD still leaves `source/tests/pt_expt/model/test_get_model_dpa4.py::TestGetModelDPA4::test_native_spin_add_chg_spin_ebd_raises` asserting that this configuration must raise `NotImplementedError`. I reproduced that failure at `f51ca5e3`, and the current Python CI shard fails on the same stale negative regression. Please remove or replace it with a positive construction assertion (the new end-to-end coverage can remain the main behavioral regression). I am leaving this thread unresolved until the branch is internally consistent and CI is green.

source/tests/common/dpmodel/test_dpa4_call_graph.py
OutisLi: The ZBL implementation now passes my formula-level, PT-parity, finite-difference, multi-frame, pair-exclusion, training, and compiled-artifact checks. However, the current HEAD still contains two stale negative regressions that make CI fail: `TestGetModelDPA4.test_unsupported_keys_raise` still lists `bridging_method="ZBL"` as unsupported, and `TestDPA4Interop.test_guard_branches_raise` still requires a PT ZBL wrapper to raise `NotImplementedError`. I reproduced both failures at `f51ca5e3`, matching the current Python CI shards. Please remove the bridging entry from the unsupported-key table and replace/remove the obsolete interop guard (the new positive ZBL checkpoint-interoperability test is the appropriate behavior). I am leaving this thread unresolved until those expectations and CI are updated.

deepmd/pt_expt/utils/serialization.py
OutisLi: Rechecked at f51ca5e3ae9180c08ea4941fcc717cff353afa38: this is still unresolved. The public freeze() and deserialize_to_file() defaults remain lower_kind="nlist", and _resolve_lower_kind() only reroutes when the caller explicitly passes "auto". The native-spin export tests still pass lower_kind="graph" explicitly, so the ordinary default path is not covered and still selects the dense/nlist ABI (and .pte default suffix) for a graph-only native-spin model. Please resolve native-spin to graph before choosing both the export ABI and default suffix, and add a regression through the default public freeze path with no lower-kind argument.

deepmd/dpmodel/descriptor/dpa4.py
OutisLi: Rechecked at f51ca5e3ae9180c08ea4941fcc717cff353afa38. The train/eval boolean gate is fixed and its targeted tests pass, but the runtime draw is still not PT-equivalent: the shared `_build_edge_wigner` creates a fresh `np.random.default_rng()` draw on the host and converts it to the caller device, whereas the PT reference uses `torch.rand(..., device=edge_quat.device)`. I verified that after resetting both `torch.manual_seed(123)` and `np.random.seed(123)`, two pt_expt/shared-core calls produce different edge quaternions while two reference-PT calls are identical. Thus the default training path is not controlled by Torch RNG state, performs an E-sized host draw plus host-to-device transfer on every CUDA forward, and has the compile-time freezing issue already noted above. The new spy test only checks the boolean and cannot detect any of these differences. Please provide a backend-native RNG seam so pt_expt draws with device-local `torch.rand`, and add a seed/replay regression against the PT semantics.

deepmd/pt_expt/train/training.py
OutisLi, line 264: [P1] Preserve the capability-safe non-spin fallback

This unconditional call regresses the existing contract of `get_additional_data_requirement` and diverges from the PT reference, which uses `getattr(_model, "has_spin", False)` and only reads `spin.allow_missing_label` after the capability is present. At this HEAD, both existing `TestAdditionalDataRequirement::test_has_default_fparam` and `test_no_default_fparam` fail here with `AttributeError: _M has no attribute has_spin`; these are also current CI failures. Forwarding `allow_missing_label` does not require narrowing the helper to models that implement the new spin API. Please retain the previous capability-safe `has_spin` check and the nested `spin.allow_missing_label` fallback, while still producing `must=False, default=0.0` for the native-spin model.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -90,6 +91,31 @@
...
+    category as the physical one (``energy_derv_r``) -- only ``.magnetic``
+    tells them apart -- so this checks ``magnetic`` before falling back to
+    the category table.  Returns ``None`` for outputs the graph-spin ABI
+    does not produce (``mask``/``mask_mag`` -- the graph route has no owner
+    site for the boolean spin-active mask, unlike the dense/nlist spin path
+    which derives it via ``communicate_extended_output``); callers fall
+    back to a NaN-filled placeholder for those, same as any other
+    unavailable output.
+    """
+    if odef.name in ("mask", "mask_mag"):
```
OutisLi: [P1] Return the real magnetic mask for graph-native-spin artifacts

This branch guarantees an invalid public result: `DeepPot.eval()` always returns `results["mask_mag"]` for a spin model, but mapping `mask_mag` to `None` makes `_eval_model_graph_spin` fill the entire requested output with NaNs. For `use_spin=[True, False]` and atom types `[0,0,0,1,1,1]`, the documented result is `[True,True,True,False,False,False]`, not an unavailable-output placeholder. The new end-to-end test currently receives `_mm` and deliberately does not assert it, so it codifies the gap instead of catching it. The adapter already has `atom_types` and the artifact metadata contains `use_spin`, so it can construct the Boolean mask without adding neural-network work (alternatively, export `mask_mag` from the graph forward). Please restore the PT/public DeepPot semantics and assert the exact mask through the graph `.pt2` path.

source/tests/pt_expt/model/test_get_model_dpa4.py
```diff
@@ -209,6 +209,22 @@ def test_unsupported_keys_raise(self) -> None:
             with self.assertRaisesRegex(NotImplementedError, msg_regex):
                 get_model(raw)
 
+    def test_native_spin_non_dpa4_descriptor_raises(self) -> None:
+        """Native-scheme spin requires a DPA4/SeZM descriptor."""
+        raw = _make_raw_model_config()
```
OutisLi, line 244: [P1] Exercise the descriptor capability gate with a standard model config

`_make_raw_model_config()` fixes `model.type` to `dpa4`, so replacing its descriptor with `se_e2_a` violates the DPA4 model-type contract before the new descriptor-agnostic native-spin capability check can run. The resulting `ValueError` (`Model type dpa4 requires a DPA4/SeZM descriptor`) is therefore correct, while this stale expectation currently fails CI. After the native-spin abstraction refactor, this regression should use a complete valid `type="standard"` + `se_e2_a` energy-model config and then assert the `supports_native_spin()` rejection; alternatively, if it intends to test the DPA4 alias contract, it should expect the `ValueError`. Please prefer the former so the generalized capability gate remains covered rather than reintroducing a DPA4-specific restriction.

deepmd/pt_expt/model/get_model.py
```diff
@@ -202,7 +214,65 @@ def get_sezm_model(data: dict) -> EnergyModel:
...
+        allow_missing_label=spin_cfg.get("allow_missing_label", False),
+    )
+    data["descriptor"]["use_spin"] = use_spin
+    model_type = str(data.get("type", "standard")).lower()
+    backbone_builder = (
+        get_sezm_model if model_type in ("dpa4", "sezm") else get_standard_model
     )
+    try:
+        backbone_model = backbone_builder(data)
+    except TypeError as err:
```
OutisLi, line 293: [P2] Do not translate every backbone construction `TypeError` into a native-spin capability error

This `try` covers the entire descriptor + fitting/model construction, so unrelated constructor and internal programming errors are reported as “descriptor ... does not accept `use_spin`”. I reproduced this at `f51ca5e3` by adding an invalid `bogus_option` to an otherwise valid native-spin DPA4 fitting config: both pt_expt here and the twin dpmodel catch report `NotImplementedError: descriptor type "dpa4" does not accept use_spin`, while the chained cause is actually `InvarFitting.__init__() got an unexpected keyword argument "bogus_option"`. The same masking would apply to any `TypeError` raised inside descriptor/fitting construction, making real regressions substantially harder to diagnose. Please narrow the translation to the specific unexpected-`use_spin` failure (or establish descriptor capability before the full backbone build) and let unrelated `TypeError`s propagate with their real context; apply the same fix to `deepmd/dpmodel/model/model.py` and add a regression for an unrelated fitting-construction error.

deepmd/pt_expt/utils/serialization.py
wanghan-iapcm: Fixed in [`35dd2289d`](https://github.com/wanghan-iapcm/deepmd-kit/commit/35dd2289d). `freeze()` now resolves native-spin to the graph lower right after the checkpoint model is built — before the export ABI is chosen and before the default output suffix is derived (suffix defaulting moved from the CLI block into `freeze()`, since the correct suffix depends on the resolved kind). `deserialize_to_file` additionally fail-fasts with a clear `ValueError` on `native_spin` + non-graph lower kind for direct programmatic callers. Regressions: a fast test pinning the `ValueError` guard, and `test_native_spin_default_freeze_routes_to_graph` freezing a native-spin checkpoint through the default public `freeze()` path (suffixless output, no `lower_kind`) and asserting a graph-kind `.pt2` (`lower_input_kind == "graph"`, `is_spin`) — run locally, passed (AOTI compile, skipif-CI per file convention).

deepmd/pt_expt/train/training.py
wanghan-iapcm: Fixed in [`e23f5f79d`](https://github.com/wanghan-iapcm/deepmd-kit/commit/e23f5f79d), taking the other branch of the contract: instead of restoring the `getattr` fallback, the `_M` fakes in `TestAdditionalDataRequirement` now implement `has_spin()` (same pattern as their existing `has_chg_spin_ebd`). Rationale: `has_spin` is a concrete base-declared capability on every pt_expt model, so the helper relies on the pinned interface rather than a defensive `getattr` — a `getattr(default=False)` would silently mask a model that forgot to inherit the capability, whereas the direct call fails loudly. The PT reference keeps `getattr` because its model zoo predates the pinned base method; pt_expt does not have that constraint. The helper still produces `must=False, default=0.0` for the native-spin model, and `spin.allow_missing_label` is only read when `has_spin()` is True. Both previously failing tests pass locally.

deepmd/pt_expt/infer/deep_eval.py
wanghan-iapcm: Fixed in [`2ff9f252d`](https://github.com/wanghan-iapcm/deepmd-kit/commit/2ff9f252d). `_eval_model_graph_spin` now synthesizes `mask_mag` from the artifact metadata and the input atom types (`use_spin[atype][..., None]` — the same math as the eager model's `spin_mask[atype] > 0`), restoring the public DeepPot semantics without widening the exported ABI or adding network work. The e2e graph-`.pt2` parity test now receives `mm` and asserts the exact boolean mask (`[T,T,T,F,F,F]` for `use_spin=[True, False]`, atypes `[0,0,0,1,1,1]`) through the public `DeepPot.eval` — run locally, passed.

source/tests/pt_expt/model/test_get_model_dpa4.py
wanghan-iapcm: Fixed in [`62a2bfffb`](https://github.com/wanghan-iapcm/deepmd-kit/commit/62a2bfffb), taking the preferred option: the regression now uses a complete valid `type="standard"` + `se_e2_a` energy-model config and asserts the generic `supports_native_spin()` capability rejection (`NotImplementedError` mentioning native spin), so the descriptor-agnostic gate stays covered. The dpa4-typed variant is retained as a separate test asserting the family contract (`ValueError: requires a DPA4/SeZM descriptor`). Both pass locally.

deepmd/pt_expt/model/get_model.py
wanghan-iapcm: Fixed in [`27678e617`](https://github.com/wanghan-iapcm/deepmd-kit/commit/27678e617). Both builders (pt_expt `get_model.py` and the dpmodel twin in `deepmd/dpmodel/model/model.py`) now translate the `TypeError` into the native-spin capability error only when the message names `use_spin`; any unrelated construction `TypeError` re-raises with its real context. Regressions added in both suites: a bogus `fitting_net` option now surfaces as the original `TypeError` naming the option (`bogus_option`) instead of the masked capability error. Both pass locally.

source/tests/common/dpmodel/test_dpa4_call_graph.py
wanghan-iapcm: Follow-up: the bridging implementation has been redesigned as an atomic-model COMPOSITION in [`6a2e9de69`](https://github.com/wanghan-iapcm/deepmd-kit/commit/6a2e9de69) (with refinements in [`eccc7c3de`](https://github.com/wanghan-iapcm/deepmd-kit/commit/eccc7c3de) and [`e8ecfd414`](https://github.com/wanghan-iapcm/deepmd-kit/commit/e8ecfd414)). Instead of a flag+hook on the DP atomic model, a bridged model is now `LinearEnergyModel(LinearEnergyAtomicModel([dpa4_atomic, InterPotentialAtomicModel], weights="sum"))`: the analytical ZBL term is a first-class `BaseAtomicModel` (`inter_potential`), the sum lives in the existing linear combination layer, and the DPA4 atomic model carries no bridging state. Force/virial parity with pt's dpa4+ZBL is exact (jittered-weight parity at 1e-12 plus an exact analytical reference check), because both children differentiate through the shared `graph.edge_vec` leaf. Serialization uses the energy-specific wire type `linear_ener`; pt checkpoints with `bridging_method` set are rejected with a rebuild-from-config message (no conversion is claimed). Tests: `source/tests/pt_expt/model/test_zbl_bridging.py` and `source/tests/common/dpmodel/test_zbl_bridging.py` (freeze+DeepEval e2e, FD force, training smoke, serialize roundtrip, with-comm gate).

source/api_cc/src/DeepSpinPTExpt.cc
```diff
@@ -780,6 +881,38 @@ void DeepSpinPTExpt::compute(ENERGYVTYPE& ener,
...
+    graph_pack.atype = node_atype;
+    graph_pack.n_node = n_node_tensor;
+    graph_pack.n_local = n_node_tensor;
+    graph_pack.edge_index = edge_tensors.edge_index;
+    graph_pack.edge_vec = graph_edge_fp32_
+                              ? edge_tensors.edge_vec.to(torch::kFloat32)
+                              : edge_tensors.edge_vec;
+    // No pair-type-exclusion table on the spin route yet (known limitation;
+    // see task report): edge_mask is used as-is.
+    graph_pack.edge_mask = edge_tensors.edge_mask;
```
OutisLi: [P1] Apply model-level pair exclusions on the C++ graph-spin ingestion seam

Rechecked at `acd68f8c4`: both graph-spin branches still pass the raw builder `edge_mask` into the exported lower, while `metadata.json` carries `pair_exclude_types` and the graph lower contract requires every external feeder to pre-apply it. Unlike `DeepPotPTExpt`, `DeepSpinPTExpt::init` never builds `pair_exclude_table_`, so C++ and Python `DeepEval` diverge for a valid generalized native-spin model. I reproduced this with `type="standard"`, a native-spin DPA4 descriptor, model-level `pair_exclude_types=[[0,1]]`, and an empty descriptor exclusion: metadata correctly contains `[[0,1]]`, but the raw C++-equivalent graph keeps 12 edges versus 4 after the required exclusion; on the same jittered weights this changes energy from `1.01471518` to `1.20616363` and gives max force delta `0.94182123`. The `type="dpa4"` alias may hide this by redundantly copying the pairs into `descriptor.exclude_types`, but the generic native-spin abstraction and metadata contract cannot rely on that DPA4-specific duplication. Please mirror `DeepPotPTExpt`: parse/cache the exclusion table once, call `applyPairExclusion` before canonicalization in both cached-nlist and standalone graph branches, and add an anti-vacuous C++ regression with model-level exclusions only.

deepmd/pt_expt/model/get_model.py
```diff
@@ -196,13 +211,108 @@ def get_sezm_model(data: dict) -> EnergyModel:
...
+    Parameters
+    ----------
+    data : dict
+        The data to construct the model. Must carry a top-level ``"spin"``
+        key with ``scheme == "native"``.
+    """
+    data = copy.deepcopy(data)
+    spin_cfg = data.pop("spin")
+    data.setdefault("descriptor", {})
+    if str(data.get("bridging_method", "none")).lower() not in ("none", ""):
```
OutisLi: [P1] Preserve PT native-spin + ZBL bridging support

This guard rejects a valid normalized configuration that the PT reference already supports: `spin.scheme="native"` together with `bridging_method="ZBL"`. I verified the same config builds a `SeZMNativeSpinModel` with an `InterPotential` in PT and runs forward with finite energy, force, and magnetic force, while both the dpmodel and pt_expt builders fail here before construction.

This is not a limitation of the new composition design. With the current code, a `LinearEnergyAtomicModel` containing the spin-aware learned atomic model plus `InterPotentialAtomicModel` can already be wrapped by `NativeSpinEnergyModel`; that composed model runs and produces finite energy/force/force_mag and the expected mask_mag. The analytical child accepts and ignores `spin`, while the learned child consumes it.

Please remove this early rejection in both backends, inject `use_spin` and the bridging InnerClamp into the learned descriptor, compose the learned and analytical atomic models, and then wrap the resulting composition in `NativeSpinEnergyModel`. A regression should cover normalized config construction, eager energy/force/force_mag, serialization, and graph freeze/DeepEval so this remains equivalent to the PT implementation.

deepmd/dpmodel/atomic_model/inter_potential.py
```diff
@@ -0,0 +1,409 @@
...
+    def __init__(
+        self,
+        type_map: list[str],
+        mode: str = "zbl",
+        rcut: float = 0.0,
+        sel: "list[int] | int" = 0,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(type_map, **kwargs)
+        self.potential = InterPotential(type_map=list(type_map), mode=mode)
```
OutisLi, line 305: [P1] Refresh the ZBL element lookup when the type map changes

`LinearEnergyAtomicModel.change_type_map()` recursively updates every child, but this child inherits the generic `BaseAtomicModel.change_type_map()`, which only changes `self.type_map` and the generic stats/exclusion state. The `InterPotential` created here keeps its original `atomic_numbers`, so the public type IDs and the ZBL lookup diverge.

I reproduced this at the current HEAD. Starting from `type_map=["H", "O"]`, a type-0/type-0 pair at 1 A has energy 1.02558505 eV. After `change_type_map(["O", "H"])`, type 0 should be O, but `potential.atomic_numbers` remains `[1, 8]` and the energy stays 1.02558505 eV instead of the fresh O-O result 23.29469704 eV. Adding a new element produces `IndexError` when that type is evaluated because the lookup table retains its old length.

This also breaks checkpoint continuity: serialization records the new public type map, so deserialization rebuilds `[8, 1]` and the same pair jumps from 1.02558505 to 23.29469704 eV after restart.

Please override the type-map update for `InterPotentialAtomicModel` (or make `InterPotential` own a backend-safe update) so its element lookup is rebuilt on the active backend/device together with the public map. Regression tests should cover reordering, adding an element, and serialize/deserialize prediction parity in both dpmodel and pt_expt.

source/tests/common/dpmodel/test_dpa4_call_graph.py
OutisLi: The composition redesign does not require dropping PT ZBL checkpoint compatibility. I tested the conversion directly at current HEAD `fd6fa77ee` using only fields already present in the serialized PT wrapper:

1. normalize/deserialize the nested `sezm_atomic` with the existing `_normalize_pt_sezm_atomic()` path;
2. construct `InterPotentialAtomicModel` from the wrapper `bridging_method` and the learned model type map/rcut/sel (the serialized descriptor already carries the matching InnerClamp/BridgingSwitch radii);
3. combine both children in `LinearEnergyAtomicModel(weights="sum")` and wrap it in `LinearEnergyModel`.

Against the original PT `SeZMModel` with the identical serialized weights and frame, this produced max absolute differences of `8.88e-16` for energy, `3.33e-16` for atomic energy, `1.22e-15` for force, `2.66e-15` for virial, and `1.11e-15` for atomic virial. Thus the wrapper contains everything needed for an exact conversion; the different output dict shape is only a dispatch/construction detail.

Please replace the current `bridging_method != none` rejection in `_unwrap_pt_sezm_model()` with this conversion and replace the negative regression with PT-checkpoint numerical parity. Rebuilding from a config is not an equivalent migration path because it neither guarantees that the original config is available nor restores the trained checkpoint weights. I am keeping this P1 thread unresolved.

deepmd/pt_expt/model/native_spin_model.py
```diff
@@ -0,0 +1,280 @@
...
+)
+from deepmd.pt_expt.model.ener_model import (
+    EnergyModel,
+)
+from deepmd.pt_expt.model.model import (
+    BaseModel,
+)
+
+
+@BaseModel.register("native_spin")
```
OutisLi, line 24: [P1] Accept regular-PT `sezm_native_spin` checkpoints

The backend-local `native_spin` round trip is fixed, but migration from the regular PT backend is still broken. At current HEAD `e48b6e281`, serializing `deepmd.pt`ʼs `SeZMNativeSpinModel` produces a wrapper with `type="sezm_native_spin"`, nested `atomic_model.type="sezm_atomic"`, and `spin`; passing that exact dict to `deepmd.pt_expt.model.model.BaseModel.deserialize()` fails with `RuntimeError: Unknown model type: sezm_native_spin. Did you mean: native_spin?`.

The current PT parity regression does not exercise this path: it constructs a fresh pt_expt model from config and copies the descriptor and fitting separately, thereby bypassing top-level wire dispatch and checkpoint migration. This is not an architectural incompatibility. I converted the same serialized checkpoint by feeding its nested atomic dict through the existing `sezm_atomic` normalizer, deserializing its `spin`, and wrapping the result in `NativeSpinEnergyModel`; on a nontrivial magnetic-force frame, PT vs pt_expt max differences were `4.44e-16` energy, `5.55e-16` atom energy, `1.42e-14` force, `2.89e-15` force_mag, and `7.99e-15` virial, with exact mask_mag. The converted pt_expt model then passed its own public BaseModel serialize/deserialize round trip exactly.

Please keep `native_spin` as the canonical pt_expt wire type, but add a compatibility read path for regular PTʼs current `sezm_native_spin` wrapper (including the nested DPA4 atomic conversion), and add a regression that directly evaluates `BaseModel.deserialize(pt_model.serialize())` rather than copying components manually.

deepmd/pt_expt/utils/serialization.py
```diff
@@ -1337,8 +1499,38 @@ def _trace_and_export(
...
+                    "export with the dense lower"
+                )
+        # Defense-in-depth: every production caller (freeze entrypoint,
+        # compress, _resolve_lower_kind auto) gates on model_uses_graph_lower
+        # upstream, but a direct programmatic call with lower_kind="graph"
+        # on a graph-INELIGIBLE model (e.g. set_davg_zero=False dpa2,
+        # use_three_body, or disable_graph_lower()) would otherwise trace a
+        # silently divergent artifact.  Assert the gate at the innermost
+        # layer too.
+        from deepmd.pt_expt.model.graph_lower import (
```
OutisLi, line 1535: [P2] Remove the duplicated graph-eligibility guard

This entire import/check/error block is repeated verbatim immediately below (current lines 1535-1554). On the success path it calls `model_uses_graph_lower(model)` twice; on failure the first copy already raises, so the second copy can never add protection. The two copies were introduced by separate commits, which is exactly patch-stacking residue and creates an unnecessary risk that future eligibility conditions or diagnostics drift between them. Please keep one complete guard and remove the duplicate rather than carrying both copies.

deepmd/pt_expt/utils/serialization.py
```diff
@@ -136,15 +138,80 @@ def _needs_with_comm_artifact(model: torch.nn.Module) -> bool:
...
+    # Analytical bridging models are single-rank only (pt's
+    # ``supports_edge_parallel() == False`` contract: ZBL + SFPG fold each
+    # node's full outgoing-edge set, which a single rank cannot observe for
+    # ghost owners) -- never compile a with-comm artifact for them.
+    from deepmd.dpmodel.atomic_model.linear_atomic_model import (
+        LinearEnergyAtomicModel,
+    )
+
+    atomic_model = getattr(model, "atomic_model", None)
+    if isinstance(atomic_model, LinearEnergyAtomicModel):
```
OutisLi, line 193: [P1] Do not classify every linear composition as single-rank

This concrete-type guard conflates one special analytical bridging composition with every `LinearEnergyAtomicModel`. The PR now makes the generic linear graph route eligible when every child supports it and forwards the same `comm_dict` to every child, so ordinary message-passing compositions are valid here. I reproduced this at `2f4e63dd8` with a public `linear_ener` config containing two DPA2 children: `model_uses_graph_lower(model)` is `True`, `atomic_model.has_message_passing()` is `True`, and both children report `has_message_passing_across_ranks() == True`, yet `_needs_with_comm_artifact(model, "graph")` returns `False` solely because of this check. The exported model therefore omits the with-comm artifact required for the children&apos;s cross-rank `border_op` exchanges.

Please represent this as capabilities rather than a wrapper-type special case: aggregate whether any child needs cross-rank message passing, separately aggregate whether every child supports edge-parallel execution, and let only the unsupported analytical child make a composition single-rank. Add a regression for a graph-eligible two-DPA2 `linear_ener` model asserting that the with-comm artifact is requested.

source/api_cc/src/DeepSpinPTExpt.cc
wanghan-iapcm: Fixed in [ecfb7a897](https://github.com/deepmodeling/deepmd-kit/commit/ecfb7a897b17b9f6ba7f4a7aa63b7c7d5a3cb0d0).

`DeepSpinPTExpt::init` now parses and uploads the keep table once (twin of `DeepPotPTExpt::init`), and **both** graph branches — cached-nlist and standalone — call `applyPairExclusion` before canonicalization. The two "known limitation" comments that stood exactly where those calls belong are gone.

Anti-vacuous C++ regression added (`source/api_cc/tests/test_deepspin_dpa4_pairexcl_ptexpt.cc`), built as you asked with model-level exclusions only: `deeppot_dpa4_spin_pairexcl.pt2` carries `pair_exclude_types=[[0,1]]` at the model level with `descriptor.exclude_types` left **empty**, and is *derived* from the no-exclusion baseline by patching the two JSON blobs (`gen_common.derive_pair_exclude_pt2`) — no second inductor compile, and a byte-identical AOTI artifact, so the two archives can differ only through the ingestion seam. Excluding every Ni-O pair moves the energy by `3.83e-01`. Validated on a Tesla T4: 6/6 pass (both branches vs the Python `DeepEval` reference at 1e-10, plus `excluded != baseline`).

On why this was not caught earlier, since it matters beyond this fix: `gen_dpa4_spin.py` was never in `source/install/test_cc_local.sh`'s generator list, while that script starts by deleting every `*.pt2`. The fixture therefore never exists in CI and the **entire** native-spin graph C++ suite has been `GTEST_SKIP`-ing — indistinguishable from passing in the summary. It is now generated, so those 16 pre-existing tests run in CI for the first time (they pass on the T4).

deepmd/dpmodel/atomic_model/inter_potential.py
wanghan-iapcm: Fixed in [e48b6e281](https://github.com/deepmodeling/deepmd-kit/commit/e48b6e281).

You are right on every point, including the restart-time divergence. The element lookup belongs to `InterPotential`, so it now owns the update rather than the atomic model patching it: a new `InterPotential.change_type_map` replaces the symbols, the count and the nuclear-charge table together, rebuilding the table **from the symbols** instead of permuting indices — so reordering, adding and dropping are all covered with no index bookkeeping to drift. The rebuilt array keeps the current one's namespace/dtype/device, so on a wrapped backend (pt_expt buffer, possibly CUDA) it is updated in place, as you suggested with "make `InterPotential` own a backend-safe update". `InterPotentialAtomicModel.change_type_map` calls `super()` and delegates.

Regressions in both backends, per your list:

- **dpmodel** (`source/tests/common/dpmodel/test_zbl_bridging.py::TestInterPotentialChangeTypeMap`): reorder matches a freshly built model (with an anti-vacuity assertion that the two element pairs are far apart), added element extends the lookup and is addressable, dropped element shrinks it, and serialize/deserialize predicts identically after the remap.
- **pt_expt** (`source/tests/pt_expt/model/test_zbl_bridging.py::TestInterPotentialChangeTypeMapPtExpt`): the same, plus an explicit precondition test that the lookup really is a torch buffer inside the wrapper, and assertions that after the rebuild it is still a tensor on the model device.

One scoping note: the pt_expt tests apply `change_type_map` to the ZBL child rather than the whole composition, because the DPA4/SeZM learned child raises `change_type_map is not supported for SeZM` — a separate pre-existing limitation, so a bridged DPA4 model still cannot be remapped as a whole.

deepmd/pt_expt/model/get_model.py
wanghan-iapcm: Fixed in [e48b6e281](https://github.com/deepmodeling/deepmd-kit/commit/e48b6e281), [2f4e63dd8](https://github.com/deepmodeling/deepmd-kit/commit/2f4e63dd8) and [cadd73772](https://github.com/deepmodeling/deepmd-kit/commit/cadd73772).

You are right that this is not a limitation of the composition design. The rejection is removed in both backends, and no special case replaced it: `get_standard_model` already OWNS assembling the atomic model — descriptor/fitting, exclusions and the bridging composition — so `get_native_spin_model` now injects `use_spin`, delegates to it, and re-classes whatever atomic model comes back, be it a single learned model or a `LinearEnergyAtomicModel` over `[learned, InterPotential]`. The learned child consumes `spin`; the analytical child accepts and ignores it, exactly as you described.

Two supporting changes were required:

- **Eligibility gate.** It used to read `backbone.atomic_model.descriptor`, which does not exist on a composition. It is now the atomic model's own `supports_native_spin()` capability — concrete `False` on `BaseAtomicModel`, delegated to the descriptor by `DPAtomicModel`, and `any(child)` on `LinearEnergyAtomicModel` (deliberately `any`, unlike `uses_graph_lower`'s `all`: every child must run on the shared graph, but spin only has to reach one consumer; with none, the magnetic force would be identically zero). No caller assumes a descriptor+fitting architecture any more.
- **Serialization.** `NSM.serialize` hard-coded the backbone wire type to `"standard"`, which fails on a composed backbone whose dict has a different shape and `@version`. It now records `backbone_type`; archives written before this field still read as `"standard"`.

Coverage, per your list: normalized-config construction (both backends, with the bridging radii still reaching the learned child's `InnerClamp`), eager energy/force/force_mag with the expected `mask_mag`, the analytical term checked **exactly** against an independent in-test pair loop at 1e-10 rather than a positivity smoke, serialize/deserialize parity at 1e-12, and graph freeze + `DeepPot` parity at 1e-10 (`test_native_spin_with_bridging_graph_freeze_and_deep_eval`, asserting `is_spin`, `use_spin` and the single-rank `has_comm_artifact=false` contract). The stale `test_native_spin_with_bridging_fails_fast` is replaced by these.

deepmd/pt_expt/model/native_spin_model.py
wanghan-iapcm: would not support this compatibility as the native spin is redesigned in the pt-expt backend.

deepmd/dpmodel/atomic_model/linear_atomic_model.py
```diff
@@ -233,6 +250,116 @@ def enable_compression(
...
+    def supports_native_spin(self) -> bool:
+        """Spin-capable when ANY child consumes the spin input.
+
+        Unlike :meth:`uses_graph_lower` (every child must run on the shared
+        graph), spin only has to reach ONE consumer: analytical terms accept
+        and ignore it, so a composition of a spin-aware learned model with a
+        ZBL term is a valid native-spin model.  With no consumer at all the
+        magnetic force would be identically zero, which is not a spin model.
+        """
+        return any(m.supports_native_spin() for m in self.models)
```
OutisLi, line 301: [P1] Keep `charge_spin` capability coherent through a native-spin bridging composition

The non-bridged native-spin + charge-spin path is now supported, but wrapping the same learned atomic model with the ZBL bridging composition silently erases that capability. I reproduced this at `cadd73772` with the public three-way config (`spin.scheme="native"`, `add_chg_spin_ebd=True`, `bridging_method="ZBL"`). The learned `DPAtomicModel` reports `has_chg_spin_ebd() == True` and `get_dim_chg_spin() == 2`, while the resulting `NativeSpinEnergyModel(LinearEnergyAtomicModel(...))` reports `False` and `0` because this composition inherits the base defaults. Eager evaluation still consumes an explicitly supplied `charge_spin`, so the model is accepted and functional in one path, but training omits the `charge_spin` data requirement, export metadata says that the input does not exist, and the native-spin graph trace builds slot 13 as `None` and fails with `ValueError: charge_spin is required for this SeZM descriptor`.

Semantically, `charge_spin` is a frame-level `[charge, global spin]` condition, whereas native spin is a continuous per-atom vector used for `-dE/ds_i`. The global-spin entry may be redundant or ambiguous next to local spins, but the charge entry is not; the ordinary PT backend also explicitly supports both inputs. My preference is therefore to preserve PT parity now and aggregate the charge-spin capability through the composition, with a future API refactor splitting global charge and global spin if the combined meaning proves unclear. Concretely, active consumers should agree on the input dimension, and a composition should advertise a default only when all consumers have a compatible common default.

If the intended dpmodel/pt_expt policy is instead that native spin and charge-spin conditioning are incompatible, please reject the combination explicitly and early in both builders and document that backend divergence. Either policy is defensible, but the current partial state -- accepted eager model, missing training input, false metadata, and broken freeze -- is not. Please add a three-way regression covering construction/eager behavior plus data requirements, graph freeze, metadata, and DeepEval (or the explicit fail-fast contract).

deepmd/pt_expt/utils/serialization.py
wanghan-iapcm: will not be addressed in this pr, tracked in https://github.com/deepmodeling/deepmd-kit/issues/5906

deepmd/pt_expt/utils/serialization.py
wanghan-iapcm: Fixed in https://github.com/deepmodeling/deepmd-kit/commit/18b7459816a065d4ef8e88f63242252147d658f1 — one copy kept.

Pinned by a source-level regression rather than a behavioral one: behavior cannot distinguish one guard from two, since the first already raises. The test reads the module file rather than calling `inspect.getsource`, because conftest wraps `_trace_and_export` and `getsource` returns the wrapper.

deepmd/dpmodel/descriptor/dpa4.py
wanghan-iapcm: Fixed in https://github.com/deepmodeling/deepmd-kit/commit/1e80e0b33394c8ea0480a9dfbc44f92561454bf3 — thanks for pushing on this; measuring it turned up a third problem beyond the two you named.

Besides being uncontrolled by torch RNG state and paying a host-to-device copy per CUDA forward, a host-side draw inside a traced region is captured as a graph **constant**. Traced with `make_fx`, the numpy version produces no random op at all (`lift_fresh_copy` of a `get_attr`) and re-running the traced graph returns identical values, whereas `torch.rand` traces as a live `rand.default` and keeps varying. So under compiled training the augmentation was silently replaying one fixed roll.

The seam is `xp_uniform` in `deepmd/dpmodel/array_api.py`: torch draws with `torch.rand` on the edge device, other backends use `deepmd.utils.random`, which `setup_seed` seeds alongside torch. Worth flagging one deliberate asymmetry: the nine existing `is_torch_array` branches in that file all compute the *same* values by a faster route, whereas this one changes which numbers come out. The docstring says so explicitly and restricts the helper to a per-forward random stream, never anything a cross-backend parity test inspects.

Regression added in `test_dpa4.py::test_random_gamma_draw_obeys_torch_rng_state` (replay under `torch.manual_seed`, plus device locality and an anti-vacuity check); it fails against the previous numpy draw. `xp_uniform` itself is covered in `test_array_api.py::TestXpUniform` for both branches.

deepmd/dpmodel/atomic_model/linear_atomic_model.py
wanghan-iapcm: Addressed in https://github.com/deepmodeling/deepmd-kit/commit/5f837d5ee0ab88d43a7665c64bea2b42b038d860, which landed independently from an audit of the same area — `LinearEnergyAtomicModel` now forwards `has_chg_spin_ebd` / `get_dim_chg_spin` / `has_default_chg_spin` / `get_default_chg_spin`, plus the default-fparam pair, instead of letting them fall through to the base defaults.

Your exact three-way config now reports `has_chg_spin_ebd=True` and `get_dim_chg_spin=2` on `NativeSpinEnergyModel(LinearEnergyAtomicModel(...))`. Since the graph freeze gates the `charge_spin` slot on `dim_chg_spin > 0`, the `ValueError: charge_spin is required` you hit should be gone with it. Your diagnosis of the mechanism was exactly right: the eager path consumed `charge_spin` while the freeze read the accessors, so the artifact disagreed with its own eager model.

Coverage is partial against what you asked for: `TestCompositionForwardsConditioningCapabilities` (dpmodel) covers bridging + charge-spin capability forwarding and `test_bridged_metadata_carries_charge_spin_dim` covers metadata, but there is no end-to-end three-way regression running native spin *on top* through freeze and DeepEval. Resolving on the capability fix; happy to reopen if you would rather the full three-way test land in this PR.

deepmd/pt_expt/model/get_model.py
wanghan-iapcm: The stale negative test is gone — `test_native_spin_add_chg_spin_ebd_raises` no longer exists anywhere under `source/tests/`, so the branch is internally consistent on this point and the end-to-end coverage is the behavioral regression as you suggested.

OutisLi (changes requested):
[P1] Preserve explicit `lower_kind="nlist"` instead of treating graph capability as a requirement

`model_uses_graph_lower(m)` establishes that a model can/defaults to use the graph lower; it does not establish that the dense lower is unavailable. This branch therefore rewrites every non-graph choice, including an explicit `lower_kind="nlist"`, for every graph-eligible DPA1/DPA2/DPA4 model.

I reproduced this at `0053e2799` with a plain non-spin, non-ZBL DPA4 checkpoint: `freeze(..., lower_kind="nlist")` passes `lower_kind="graph"` to `deserialize_to_file` and changes a suffixless output to `.pt2`. That conflicts with the CLI contract (`--lower-kind nlist` is the documented default), `_resolve_lower_kind` (only `auto` selects automatically; explicit kinds are preserved), and this PRs own DPA4 export test, which states that the dense kind stays reachable for back-compat. This is numerically observable for models such as DPA1, where the documented graph/dense attention semantics can differ by about 1e-4.

Please keep an explicit `nlist` request intact. If graph should become the preferred default, expose/use `auto` for that policy, and represent the distinct graph-required/dense-supported capability explicitly so only models that truly lack a dense lower are forced to graph. This also avoids replacing the earlier concrete-type special case with an overly broad capability test.

OutisLi (changes requested):
[P1] Do not collapse per-child `fparam` defaults to the first child

The new `any(...)` plus first-default aggregation in `LinearEnergyAtomicModel.has_default_fparam()` / `get_default_fparam()` is not a valid model-level default for a general `linear_ener`. The composition exposes one external `fparam` tensor shared by all children, but when it is omitted each fitting uses its own configured default.

I reproduced this at `0053e2799` with two valid learned `se_atten` energy children whose defaults are `[0.0]` and `[1.0]`. The parent reports `has_default_fparam=True` and `get_default_fparam=[0.0]`; consequently `get_additional_data_requirement` marks the input optional and injects `[0.0]` into both children. The same model gives `2.1515824682051807` for `fparam=None` (each child uses its own default) and `2.1507697800584133` after the parent fallback, a silent `8.13e-4` change. The compiled training wrappers and frozen metadata use this parent accessor too, so training/exported inference can disagree with eager omission semantics. The PT reference reports no model-level default for this composition and therefore requires an explicit shared `fparam` in training.

Please consider only active consumers (`get_dim_fparam() > 0`) and advertise a parent default only when every active child has a compatible dimension and the same default value. Dimension-zero analytical children should be ignored so the intended learned-model plus ZBL composition still inherits the learned default. If active children disagree or one lacks a default, expose no parent default (or reject the incompatible composition explicitly). Please cover differing defaults, matching defaults, and a dimension-zero analytical child.

wanghan-iapcm:
Re: [this review](https://github.com/deepmodeling/deepmd-kit/pull/5884#pullrequestreview-4781085983) — fixed in https://github.com/deepmodeling/deepmd-kit/commit/f937b896ef9bdb69c82929836ff00b765a2dd2a4. Your diagnosis was right and the bug was mine: the accessors were forwarded in 5f837d5ee but aggregated with `any()` plus the first child's value, which is not a valid model-level default for a shared input.

A parent default is now advertised only when every **active** consumer (`get_dim_*() > 0`) has one **and** they all agree in shape and value; otherwise none is exposed, so omission stays omission and each child applies its own default. Dimension-zero children are excluded as non-consumers, so the intended learned + ZBL composition still inherits the learned default, as you asked.

I applied the same aggregation to `charge_spin`, which had the identical defect and which your report did not mention.

Regressions cover the four cases you listed — differing defaults, matching defaults, a dimension-zero analytical child, and an active child with no default:

| case | `has_default_fparam` | `get_default_fparam` |
|---|---|---|
| `[0.0]` / `[1.0]` | False | None |
| `[1.0]` / `[1.0]` | True | `[1.0]` |
| learned + ZBL (dim-0 child) | True | `[0.5]` |
| active child, no default | False | None |

Two things I did not do, so the record is accurate: the regressions use a stub child and pin the accessor contract, not your end-to-end 8.13e-4 divergence with two `se_atten` children — that number in the commit message is your measurement, not mine. And the downstream readers (`get_additional_data_requirement`, the compiled training wrappers, the frozen metadata) follow from the accessors but are not separately pinned.

Your other P1 ([4781068808](https://github.com/deepmodeling/deepmd-kit/pull/5884#pullrequestreview-4781068808), explicit `lower_kind="nlist"`) is a policy question I have referred to the author, since the requested graph-required-vs-dense-supported capability was implemented earlier in this branch and removed deliberately. I will follow up on that thread separately rather than silently pick a side.

OutisLi (approved):
Reviewed the latest HEAD. The previously requested changes have been addressed.

OutisLi (changes requested):
[P2] Preserve the PT-aligned forward-method ordering in `dpa4.py`. The PT `SeZM` implementation deliberately presents this code top-down: construction, the peer forward entrypoints, their shared implementation, the core block computation, and then lower-level helpers. The dpmodel port currently orders the section as `call -> _run_graph -> call_graph -> _forward_blocks`, so the private `_run_graph` separates the two peer entrypoints and makes readers traverse roughly 260 lines of internals before reaching `call_graph`. Please make this a move-only organization change and use `__init__ -> call -> call_graph -> _run_graph -> _forward_blocks -> remaining helpers`. Do not reorganize it around DPA2 conventions; the goal is to keep the port structurally aligned with the intentionally arranged PT implementation. This should not change runtime, checkpoint, or serialization behavior.

OutisLi (changes requested):
[P2] Correct the PT DPA4 freeze and multi-rank capability matrix in `dpa4.md`. The important block currently says that the default `dp --pt freeze` output is a dense-kind `.pt2` and is single-rank only. The implementation does the opposite: `deepmd.pt.entrypoints.main.freeze` detects DPA4/SeZM and calls `freeze_sezm_to_pt2`; `SeZMModel.export_lower_input_kind()` returns `edge_vec` (a compact edge-list ABI, not a padded dense nlist), and a plain model reports `supports_edge_parallel()==True`, so freeze embeds `model/extra/forward_lower_with_comm.pt2` and sets `has_comm_artifact=true`. The PT export tests explicitly assert that artifact. Consequently the standard PT-frozen plain DPA4 archive supports multi-rank LAMMPS. The text also conflates the fixed PT `edge_vec` export route with pt_expt `--lower-kind graph`; these are different graph/edge ABIs, but neither should be described as the default PT dense route. Please rewrite this block to distinguish PT legacy `edge_vec` from the pt_expt NeighborGraph ABI and state the correct multi-rank support for both plain-energy archives.

OutisLi (approved):

PR #5885, opened 2026-07-21
fix(pt_expt): preserve graph routing for raw checkpoints

### LINKED ISSUE #5862: [BUG] pt_expt DeepPot(.pt) uses dense lower for graph-native DPA1 and returns incorrect outputs
## Bug summary

`dp --pt-expt test -m model.ckpt.pt` can return catastrophically incorrect energy, force, and virial predictions for a graph-eligible PT-experimental DPA1 checkpoint, although the checkpoint's direct public `model.forward(...)` predictions and in-training validation are correct.

The raw `.pt` loader in `deepmd/pt_expt/infer/deep_eval.py::_load_pt` reconstructs the model but installs an eager runner that calls `model.forward_common_lower(...)`, forcing the legacy dense neighbor-list path. A graph-eligible DPA1 model's public `forward(...)` instead reaches the graph-native `call_common` path by default. These paths are not numerically equivalent when descriptor statistics are nonzero: the dense DPA1 implementation retains a padding-neighbor `-davg/dstd` residual, as already documented in `deepmd/dpmodel/descriptor/dpa1.py`. Large `sel` values amplify the discrepancy because most dense slots are padding.

This violates the stated `.pt` inference contract and the intent of `source/tests/pt_expt/infer/test_deep_eval_pt_checkpoint.py`: `DeepPot(.pt)` should reproduce the source model's public forward outputs. That test currently uses only `se_e2_a`, so it does not exercise graph-eligible DPA1 routing.

On a trained OMat24 checkpoint (`se_atten_v2`, `attn_layer=0`, `sel=416`), the same five validation frames gave:

```text
Direct public model.forward (graph): Energy MAE 0.03955 eV, Force MAE 0.02755 eV/Å
DeepPot/checkpoint dp test (dense):   Energy MAE 2600.81 eV, Force MAE 12.97899 eV/Å
```

Type maps matched. Single-frame and batched inference agreed with each other, and native and Vesin neighbor-list builders also agreed, isolating the discrepancy to graph public forward versus the dense lower selected by raw-checkpoint DeepEval.

## DeePMD-kit Version

- Clean Python source checkout: `1be0082cceff7934a88874af6daa0fb8fa60b9f3`
- Installed package metadata: `v3.2.0b1.dev110+g8399520c5.d20260704`

The minimal reproducer below sets `DP_CUDA_INFER=0` and exercises only eager Python/PyTorch paths; it does not invoke compiled graph custom operators.

## Backend and its version

- Backend: PyTorch Experimental (`pt_expt`)
- PyTorch: `2.11.0+cu128`
- CUDA: `12.8`
- Python: `3.13.13`

## How did you download the software?

Built from source.

## Input Files, Running Commands, Error Log, etc.

The production command was:

```bash
dp --pt-expt test \
  -m models/model.ckpt.pt \
  -s /path/to/val.lmdb
```

The following standalone reproducer requires no external model or dataset. It constructs a small graph-eligible DPA1 model, sets deterministic nonzero descriptor statistics, saves a normal PT-experimental training checkpoint, and compares its public forward with `DeepPot(.pt)`:

```python
import copy
import os
import tempfile

import numpy as np
import torch
from deepmd.infer import DeepPot
from deepmd.pt_expt.model import get_model
from deepmd.pt_expt.train.wrapper import ModelWrapper
from deepmd.pt_expt.utils.env import DEVICE

params = {
    "type_map": ["H", "O"],
    "descriptor": {
        "type": "dpa1",
        "sel": 20,
        "rcut_smth": 0.5,
        "rcut": 4.0,
        "neuron": [3, 6],
        "axis_neuron": 2,
        "attn": 4,
        "attn_layer": 0,
        "smooth_type_embedding": True,
        "set_davg_zero": False,
        "type_one_side": True,
        "precision": "float64",
        "seed": 1,
    },
    "fitting_net": {
        "type": "ener",
        "neuron": [8, 8],
        "precision": "float64",
        "seed": 1,
    },
}

model = get_model(copy.deepcopy(params)).to(torch.float64).to(DEVICE).eval()
with torch.no_grad():
    model.atomic_model.descriptor.se_atten.mean.fill_(0.01)
    model.atomic_model.descriptor.se_atten.stddev.fill_(0.1)

coords = np.array(
    [[[1.0, 1.0, 1.0], [2.0, 1.0, 1.0], [1.0, 2.0, 1.0]]]
)
cells = np.eye(3).reshape(1, 9) * 10.0
atom_types = np.array([0, 1, 0], dtype=np.int32)

coord_t = torch.tensor(
    coords, dtype=torch.float64, device=DEVICE
).requires_grad_(True)
atype_t = torch.tensor(
    atom_types.reshape(1, -1), dtype=torch.int64, device=DEVICE
)
cell_t = torch.tensor(cells, dtype=torch.float64, device=DEVICE)
reference = model.forward(coord_t, atype_t, cell_t)

fd, checkpoint = tempfile.mkstemp(suffix=".pt")
os.close(fd)
try:
    wrapper = ModelWrapper(model, model_params=params)
    torch.save({"model": wrapper.state_dict()}, checkpoint)
    energy, force, _ = DeepPot(checkpoint, auto_batch_size=False).eval(
        coords, cells, atom_types
    )
finally:
    os.unlink(checkpoint)

reference_energy = reference["energy"].detach().cpu().numpy()
reference_force = reference["force"].detach().cpu().numpy()
print("public graph energy:", reference_energy.reshape(-1))
print("DeepPot .pt energy:", energy.reshape(-1))
print("energy max abs delta:", np.max(np.abs(energy - reference_energy)))
print("force max abs delta:", np.max(np.abs(force - reference_force)))
```

Run with:

```bash
DP_CUDA_INFER=0 python reproduce.py
```

Observed output:

```text
public graph energy: [2.42767467]
DeepPot .pt energy: [2.33223403]
energy max abs delta: 0.09544063670727354
force max abs delta: 0.18991794401885587
```

Expected behavior: all outputs from `DeepPot(checkpoint)` match the source model's public `forward(...)` within numerical tolerance.

## Steps to Reproduce

1. Use a current clean deepmd-kit checkout with the PT-experimental backend.
2. Save the script above as `reproduce.py`.
3. Run `DP_CUDA_INFER=0 python reproduce.py`.
4. Observe that raw-checkpoint `DeepPot` does not match the source model's public graph forward.

## Further Information, Files, and Links

The relevant control flow is:

- `deepmd/pt_expt/infer/deep_eval.py::_load_pt` defines `_eager_runner` using `model.forward_common_lower(...)`, which is the dense lower.
- `deepmd/pt_expt/model/ener_model.py::EnergyModel.forward` calls `call_common(...)`; PT-experimental graph-eligible DPA1 resolves this to the graph-native route.
- `deepmd/dpmodel/descriptor/dpa1.py` documents that the dense body leaks a phantom padding-neighbor `-davg/dstd` residual when `davg != 0`, whereas the graph path omits it.

A structural fix should make raw `.pt` DeepEval preserve the source model's public-forward semantics, selecting the graph-native route for graph-eligible DPA1 checkpoints instead of unconditionally adapting every `.pt` checkpoint to the dense lower ABI. A regression test can extend `test_deep_eval_pt_checkpoint.py` with a graph-eligible DPA1 fixture whose descriptor mean is nonzero and assert energy/force/virial parity with `model.forward(...)`.

OutisLi:
## Summary

- select the eager lower ABI for raw PT-experimental checkpoints with the restored model's `model_uses_graph_lower` capability
- route graph-eligible energy models through the existing graph DeepEval contract while preserving dense and spin checkpoint behavior
- automatically choose the scalable graph builder (`nv` on CUDA, then Vesin, then dense fallback) so `dp test` does not stall in the single-core NumPy O(N²) builder
- add a deterministic nonzero-statistics DPA1 regression covering both plain and compiled checkpoint layouts

Closes #5862

## Test plan

- [x] `pytest source/tests/pt_expt/infer/test_deep_eval_pt_checkpoint.py source/tests/pt_expt/infer/test_graph_deepeval.py -v`
- [x] CUDA auto route resolves to `nv`; CPU auto route resolves to Vesin; both pass the raw checkpoint parity regression
- [x] `pytest source/tests/pt_expt/model/test_dpa1_graph_lower.py source/tests/pt_expt/model/test_graph_builder_dispatch.py -v`
- [x] repository pre-commit hooks, including Ruff, formatting, Velin, and pylint
- [x] representative OMat24 benchmark (8 frames × 128 atoms): dense median 0.1769 s, auto/NV median 0.0092 s
- [x] cross-backend validation with the supplied OMat24 weights: graph checkpoint outputs match public forward within float32 numerical tolerance

## Summary by CodeRabbit

* **New Features**
  * Added graph-routed (“graph-lower”) eager `.pt` checkpoint loading and evaluation support.
  * `neighbor_graph_method` now defaults to `auto` and applies graph-lower routing automatically.
  * Graph-routed models expose lower-input kind and graph edge dtype metadata.

* **Bug Fixes**
  * Enforced correct neighbor-graph behavior when inappropriate neighbor-list settings are provided (with clearer warnings).
  * Ensured graph-routed inference outputs match the model’s public `forward(..., do_atomic_virial=True)`.

* **Tests**
  * Expanded coverage for plain and compiled graph-routed DPA1 `.pt` checkpoints.

### DIFF
--- a/deepmd/pt_expt/infer/deep_eval.py
+++ b/deepmd/pt_expt/infer/deep_eval.py
@@ -142,12 +142,14 @@ class DeepEval(DeepEvalBackend):
         nlist-form ``.pt2``, and dense-routed ``.pt`` checkpoints):
         ``"auto"`` / ``"vesin"`` / ``"native"``. Explicit non-default values
         are ignored with a warning for graph-routed artifacts.
-    neighbor_graph_method : str, default: "dense"
+    neighbor_graph_method : str, default: "auto"
         Carry-all graph builder for graph-form ``.pt2`` artifacts and
         graph-routed ``.pt`` checkpoints
-        (``metadata["lower_input_kind"] == "graph"``): ``"dense"`` / ``"ase"``
-        (backend-agnostic) or ``"vesin"`` / ``"nv"`` (on-device O(N)). A
-        non-default value on any other artifact raises at construction because
+        (``metadata["lower_input_kind"] == "graph"``): ``"auto"`` selects
+        ``"nv"`` on CUDA when nvalchemiops is available, then ``"vesin"`` when
+        available, and finally falls back to ``"dense"``. Explicit
+        ``"dense"`` / ``"ase"`` / ``"vesin"`` / ``"nv"`` choices are preserved.
+        A non-default value on any other artifact raises at construction because
         the knob would silently do nothing there; use ``nlist_backend`` for the
         nlist path instead. All builders emit the same neighbor set, so the
         choice is performance-only. Consolidating the two knobs into a single
@@ -164,14 +166,14 @@ class DeepEval(DeepEvalBackend):
         auto_batch_size: bool | int | AutoBatchSize = True,
         neighbor_list: Optional["ase.neighborlist.NewPrimitiveNeighborList"] = None,
         nlist_backend: str = "auto",
-        neighbor_graph_method: str = "dense",
+        neighbor_graph_method: str = "auto",
         **kwargs: Any,
     ) -> None:
         self.output_def = output_def
         self.model_path = model_file
         self.neighbor_list = neighbor_list
-        # Graph-lower builder selection: "dense"/"ase" (backend-agnostic) or
-        # "vesin"/"nv" (on-device O(N)).
+        # Graph-lower builder selection is resolved once after model metadata
+        # identifies the lower ABI.
         self._neighbor_graph_method = neighbor_graph_method
         self._is_pt2 = model_file.endswith(".pt2")
 
@@ -191,9 +193,9 @@ class DeepEval(DeepEvalBackend):
         # ``neighbor_graph_method`` is consumed only by graph-lower evaluation.
         # Fail fast instead of silently ignoring it on nlist-form artifacts,
         # where the corresponding builder knob is ``nlist_backend``.
-        if neighbor_graph_method != "dense" and getattr(self, "metadata", {}).get(
-            "lower_input_kind"
-        ) not in ("graph", "dpa1_canonical"):
+        if neighbor_graph_method not in ("auto", "dense") and getattr(
+            self, "metadata", {}
+        ).get("lower_input_kind") not in ("graph", "dpa1_canonical"):
             raise ValueError(
                 f"neighbor_graph_method={neighbor_graph_method!r} only applies to "
                 "graph-routed artifacts (lower_input_kind == 'graph'); this "
@@ -201,7 +203,7 @@ class DeepEval(DeepEvalBackend):
                 "neighbor-list builder for the nlist path."
             )
 
-        self._setup_nlist_backend(nlist_backend)
+        self._setup_neighbor_backend(nlist_backend)
 
         if isinstance(auto_batch_size, bool):
             if auto_batch_size:
@@ -215,8 +217,32 @@ class DeepEval(DeepEvalBackend):
         else:
             raise TypeError("auto_batch_size should be bool, int, or AutoBatchSize")
 
-    def _setup_nlist_backend(self, nlist_backend: str) -> None:
-        """Resolve the neighbor-list construction strategy from a user choice.
+    @staticmethod
+    def _resolve_neighbor_graph_method(method: str) -> str:
+        """Resolve the graph builder once for the active device."""
+        if method not in ("auto", "dense", "ase", "vesin", "nv"):
+            raise ValueError(
+                f"Unknown neighbor_graph_method {method!r}; "
+                "expected 'auto', 'dense', 'ase', 'vesin', or 'nv'."
+            )
+        if method != "auto":
+            return method
+
+        from deepmd.pt.utils.nv_nlist import (
+            is_nv_available,
+        )
+        from deepmd.pt_expt.utils.env import (
+            DEVICE,
+        )
+
+        if DEVICE.type == "cuda" and is_nv_available():
+            return "nv"
+        if is_vesin_torch_available():
+            return "vesin"
+        return "dense"
+
+    def _setup_neighbor_backend(self, nlist_backend: str) -> None:
+        """Resolve the graph or neighbor-list construction strategy.
 
         ``"native"`` uses the dense all-pairs builder; ``"vesin"`` forces the
         O(N) ``vesin.torch`` cell list (raising if it is unavailable or the
@@ -245,6 +271,9 @@ class DeepEval(DeepEvalBackend):
                     UserWarning,
                     stacklevel=2,
                 )
+            self._neighbor_graph_method = self._resolve_neighbor_graph_method(
+                self._neighbor_graph_method
+            )
             self._use_vesin = False
             self._nlist_builder = None
             return
--- a/deepmd/pt_expt/infer/deep_eval.py
+++ b/deepmd/pt_expt/infer/deep_eval.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 import json
+import warnings
 from collections.abc import (
     Callable,
 )
@@ -76,11 +77,11 @@ if TYPE_CHECKING:
     )
 
 
-# Public output keys emitted by the graph-form AOTI forward
-# (``forward_lower_graph_exportable``) keyed by the output-variable category that
-# ``request_defs`` carries.  The graph path is LOCAL-only (``N == sum(n_node)``
-# nodes, no ghosts), so its outputs are already at local-atom resolution -- no
-# ``communicate_extended_output`` fold-back is needed.
+# Public output keys emitted by graph-lower forwards, keyed by the
+# output-variable category that ``request_defs`` carries. The graph path is
+# local-only (``N == sum(n_node)`` nodes, no ghosts), so its outputs are already
+# at local-atom resolution and require no ``communicate_extended_output``
+# fold-back.
 _GRAPH_CATEGORY_TO_KEY = {
     OutputVariableCategory.OUT: "atom_energy",
     OutputVariableCategory.REDU: "energy",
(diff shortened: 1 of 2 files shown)


deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -211,20 +217,66 @@ def __init__(
...
+        artifacts use ``neighbor_graph_method`` instead, reject an explicit ASE
+        neighbor list, and warn when an explicit nlist backend is ignored.
+        Results are unchanged either way; only the neighbor-search cost differs.
         """
         if nlist_backend not in ("auto", "vesin", "native"):
             raise ValueError(
                 f"Unknown nlist_backend '{nlist_backend}'; "
                 "expected 'auto', 'vesin', or 'native'."
             )
+        if self.metadata.get("lower_input_kind") in ("graph", "dpa1_canonical"):
```
wanghan-iapcm, line 260: This new graph-routed branch adds two reachable, user-facing guardrails that no test exercises: (1) the `raise ValueError` when an explicit `neighbor_list` is passed for a graph artifact, and (2) the `warnings.warn` when `nlist_backend != "auto"` for a graph artifact. Both are cheap to cover — construct `DeepEval`/`DeepPot` on the new graph DPA1 checkpoint with `neighbor_list=<something>` (assertRaises) and with `nlist_backend="vesin"` (assertWarns). Worth adding so every reachable branch here is tested.

wanghan-iapcm (commented):
The routing fix is correct by construction (it keys on the same `model_uses_graph_lower` flag that drives the model's own `forward()`, so `.pt` eval goes graph iff `forward()` does), and the DPA1 nonzero-stats regression is a genuine test for #5862. One test-coverage note inline.

njzjz (commented):
Nice piece of work — routing raw `.pt` checkpoints through the graph lower and auto-selecting the O(N) builder is a real usability win, and the parity test against `model.forward(..., do_atomic_virial=True)` at 1e-10 is the right assertion to anchor it on. A few things I would want resolved before this lands.

## 1. `neighbor_list=` now raises for graph-eligible `.pt` checkpoints (backward-incompatible)

```python
if self.metadata.get("lower_input_kind") in ("graph", "dpa1_canonical"):
    if self.neighbor_list is not None:
        raise ValueError("neighbor_list only applies to nlist-routed artifacts; ...")
```

Before this PR `_load_pt` always set `lower_input_kind = "nlist"`, so `neighbor_list=` was accepted for every `.pt`. After it, any checkpoint where `model_uses_graph_lower(model)` is true takes the new branch and the same call raises. That reaches user code that never opted into anything: `DP` in `deepmd/calculator.py` forwards `neighbor_list` straight through (`calculator.py:100`), so an existing ASE script pairing a DPA1 `.pt` with a custom `NewPrimitiveNeighborList` goes from working to `ValueError` on upgrade.

Since the PR's own premise is that all builders emit the same neighbor set, the ASE list is not *wrong* here — it is just redundant. I would either honor it (route to `neighbor_graph_method="ase"`, which already exists) or downgrade to a warning, and in either case call the change out in the PR description. Raising is the one option that breaks callers silently at import-of-model time.

## 2. The fail-fast guard stops catching an explicit `"dense"`

```python
if neighbor_graph_method not in ("auto", "dense") and ... != "graph":
    raise ValueError(...)
```

The guard exists because "the knob would silently do nothing there". With the default moved to `"auto"`, `"dense"` is no longer the default but is still exempt, so `DeepEval(nlist_model, neighbor_graph_method="dense")` now silently does nothing — exactly the case the check was written to prevent. Comparing against a *sentinel* rather than a value set fixes this cleanly: default the parameter to `None`, treat `None` as `"auto"` after routing is known, and raise for any non-`None` value on a non-graph artifact.

## 3. The two knobs disagree on policy in mirrored situations

`neighbor_graph_method` on an nlist artifact → `ValueError`. `nlist_backend` on a graph artifact → `UserWarning`. Same mistake, same consequence (the knob does nothing), opposite outcome. Worth picking one. If the warning stays, `stacklevel=2` points at `DeepEval.__init__` rather than the user's `DeepPot(...)` call — the warning as emitted names `deep_eval.py`, so it is not actionable. It needs to account for `__init__` **and** the `DeepEval` wrapper frame.

## 4. `auto` resolution is untested for the branches CI does not hit

```python
cls.expected_graph_method = "nv" if (cuda and is_nv_available()) else "vesin" if ... else "dense"
```

The test computes the expected method with the same predicate the implementation uses, then asserts they match — so it pins "the resolver agrees with itself", not "the resolver picks a builder that is correct". The parity assertion against `model.forward` does check correctness, but only for whichever builder happens to be available on that runner; on a CPU-only CI box `nv` is never exercised at all, and the `"performance-only" ` claim in the docstring is never verified.

Since that claim is what makes the default change safe, I would add a loop over the builders that *are* available (skipping the rest) asserting they produce identical outputs, rather than only checking the one `auto` chose. That also gives `ase` — which is still reachable explicitly and currently has no coverage on this path — a home.

## 5. Smaller points

- `_resolve_neighbor_graph_method` imports `deepmd.pt.utils.nv_nlist` from inside `pt_expt`. Presumably deliberate reuse, but a `pt_expt` → `pt` dependency is worth a one-line comment saying so, otherwise it reads like a typo for `pt_expt.utils`.
- The device is sampled once at construction (`DEVICE.type == "cuda"`), which is fine, but the resolved value is stored in the same attribute that held the user's request. Keeping the request and the resolution in separate attributes would make the `_neighbor_graph_method` assertion in the test unambiguous about which one it is checking.
- The parity fixture is 3 atoms in a 10 Å cell with `rcut=4.0`, so every atom sees every other and no periodic image matters. An atom with an empty neighbor row, and a cell small enough to need images, would exercise the parts of the graph builders most likely to differ between backends.
- `docs/readthedocs.org:deepmd` is red on this PR. That is the shared 40-minute RTD timeout affecting many open PRs, not something in this diff — #5888 is the fix.

PR #5903, opened 2026-07-24
feat(pt_expt): call-time DeepEval auto ladder with nf==1 vesin gate

Shaurya2k06:
## Summary

- Add shared `resolve_auto_graph_builder(device, nf)` for **inference / DeepEval only** (training keeps `resolve_neighbor_graph_method` from #5913).
- Resolve `neighbor_graph_method="auto"` **at eval call time** with the batch frame count: CUDA prefers `nv`; `vesin` only when `nf == 1` and importable; otherwise `dense`. Matches `_select_neighbor_builder`.
- Multi-frame `auto_batch_size` / `dp test` batches therefore stay off vesin's per-frame Python loop.
- Parametrize vesin (and nv) vs dense energy/force parity over `nf in {1, 4}`.

This is **not** a model-level / training default flip — those already shipped in #5912 / #5913. The remaining change is the inference auto ladder: re-introducing vesin only under the `nf == 1` gate that review asked for on #5912.

## Validation

- `ruff check` / `ruff format` on touched files
- `pytest` resolver ladder + DeepEval resolution + vesin parity for `nf=1` and `nf=4`

## Summary by CodeRabbit

- **New Features**
  - Improved automatic neighbor-graph builder selection during inference.
  - CUDA prioritizes optimized GPU processing, with Vesin used for eligible single-frame cases and dense processing as a fallback.
  - CPU inference uses Vesin for single-frame cases when available; otherwise, it uses dense processing.
  - Training continues to use dense processing for automatic selection.

- **Bug Fixes**
  - Ensured automatic and unspecified builder settings produce consistent energy and force results.

- **Tests**
  - Expanded coverage for single- and multi-frame backend selection and CPU/CUDA fallback behavior.

### DIFF
--- a/deepmd/pt_expt/utils/graph_builder.py
+++ b/deepmd/pt_expt/utils/graph_builder.py
@@ -18,6 +18,93 @@ if TYPE_CHECKING:
 
 log = logging.getLogger(__name__)
 
+# Warn once per process: resolve_auto_graph_builder runs per batch after
+# call-time resolution, but "install nvalchemi-toolkit-ops" is a one-shot
+# action for the user.
+_warned_auto_no_nv = False
+
+
+def resolve_auto_graph_builder(
+    device: torch.device | str,
+    nf: int = 1,
+) -> str:
+    """Resolve ``neighbor_graph_method="auto"`` to a concrete inference builder.
+
+    Single owner of the inference / DeepEval auto ladder. Training uses
+    :func:`resolve_neighbor_graph_method`, which never selects ``vesin``.
+
+    Mirrors :func:`deepmd.pt.model.model.sezm_model._select_neighbor_builder`:
+    ``vesin`` is eligible only for a single-frame batch (``nf == 1``), because
+    its API loops frames in Python (~1 ms/frame). Multi-frame batches stay on
+    ``nv`` (CUDA) or ``dense`` so ``auto_batch_size`` / ``dp test`` do not
+    regress to the per-frame loop.
+
+    Policy
+    ------
+    * CUDA + ``nvalchemiops``: ``nv`` (any ``nf``).
+    * ``nf == 1`` + ``vesin.torch``: ``vesin``.
+    * otherwise: ``dense``.
+
+    ``ase`` is never chosen automatically. All builders emit the same carry-all
+    neighbor set; the choice is performance-only. Builders run eagerly outside
+    traced / compiled regions, so this does not change ``.pt2`` artifacts.
+
+    Parameters
+    ----------
+    device : torch.device or str
+        Device the coordinates live on (or will be moved to). Controls whether
+        the CUDA-only ``nv`` builder is eligible.
+    nf : int, default: 1
+        Number of frames in the batch. ``vesin`` is selected only when
+        ``nf == 1`` and ``vesin.torch`` is importable.
+
+    Returns
+    -------
+    str
+        One of ``"nv"``, ``"vesin"``, or ``"dense"``.
+
+    Raises
+    ------
+    ValueError
+        If ``nf`` is not a positive ``int`` (``bool`` is rejected).
+    """
+    global _warned_auto_no_nv
+
+    from deepmd.pt.utils.nv_nlist import (
+        is_nv_available,
+    )
+    from deepmd.pt_expt.utils.vesin_neighbor_list import (
+        is_vesin_torch_available,
+    )
+
+    # ``bool`` is a subclass of ``int``; reject it explicitly.
+    if type(nf) is not int:
+        raise ValueError(f"nf must be a positive int, got {nf!r}")
+    if nf < 1:
+        raise ValueError(f"nf must be >= 1, got {nf}")
+
+    dev = torch.device(device)
+    nv_available = is_nv_available()
+    if dev.type == "cuda" and nv_available:
+        return "nv"
+    if nf == 1 and is_vesin_torch_available():
+        return "vesin"
+    if dev.type == "cuda" and not nv_available:
+        if not _warned_auto_no_nv:
+            _warned_auto_no_nv = True
+            log.warning(
+                "nvalchemi-toolkit-ops is unavailable; falling back from "
+                "neighbor_graph_method='auto' to the dense graph builder"
+                + (
+                    ""
+                    if nf == 1
+                    else " (vesin is not used for nf>1; its API loops frames in Python)"
+                )
+                + ". Install it with `pip install nvalchemi-toolkit-ops` to enable "
+                "the NV graph builder."
+            )
+    return "dense"
+
 
 def resolve_neighbor_graph_method(
     requested: str,
@@ -36,6 +123,8 @@ def resolve_neighbor_graph_method(
     -------
     str
         The concrete builder name, either ``"dense"`` or ``"nv"``.
+        Training auto never selects ``vesin`` (per-frame Python loop); use
+        :func:`resolve_auto_graph_builder` for inference auto selection.
 
     Raises
     ------
--- a/deepmd/pt_expt/utils/graph_builder.py
+++ b/deepmd/pt_expt/utils/graph_builder.py
@@ -19,6 +19,83 @@ if TYPE_CHECKING:
 log = logging.getLogger(__name__)
 
 
+def resolve_auto_graph_builder(
+    device: torch.device | str,
+    nf: int = 1,
+) -> str:
+    """Resolve ``neighbor_graph_method="auto"`` to a concrete inference builder.
+
+    Single owner of the inference / DeepEval auto ladder. Training uses
+    :func:`resolve_neighbor_graph_method`, which never selects ``vesin``.
+
+    Mirrors :func:`deepmd.pt.model.model.sezm_model._select_neighbor_builder`:
+    ``vesin`` is eligible only for a single-frame batch (``nf == 1``), because
+    its API loops frames in Python (~1 ms/frame). Multi-frame batches stay on
+    ``nv`` (CUDA) or ``dense`` so ``auto_batch_size`` / ``dp test`` do not
+    regress to the per-frame loop.
+
+    Policy
+    ------
+    * CUDA + ``nvalchemiops``: ``nv`` (any ``nf``).
+    * ``nf == 1`` + ``vesin.torch``: ``vesin``.
+    * otherwise: ``dense``.
+
+    ``ase`` is never chosen automatically. All builders emit the same carry-all
+    neighbor set; the choice is performance-only. Builders run eagerly outside
+    traced / compiled regions, so this does not change ``.pt2`` artifacts.
+
+    Parameters
+    ----------
+    device : torch.device or str
+        Device the coordinates live on (or will be moved to). Controls whether
+        the CUDA-only ``nv`` builder is eligible.
+    nf : int, default: 1
+        Number of frames in the batch. ``vesin`` is selected only when
+        ``nf == 1`` and ``vesin.torch`` is importable.
+
+    Returns
+    -------
+    str
+        One of ``"nv"``, ``"vesin"``, or ``"dense"``.
+
+    Raises
+    ------
+    ValueError
+        If ``nf`` is not a positive ``int`` (``bool`` is rejected).
+    """
+    from deepmd.pt.utils.nv_nlist import (
+        is_nv_available,
+    )
+    from deepmd.pt_expt.utils.vesin_neighbor_list import (
+        is_vesin_torch_available,
+    )
+
+    # ``bool`` is a subclass of ``int``; reject it explicitly.
+    if type(nf) is not int:
+        raise ValueError(f"nf must be a positive int, got {nf!r}")
+    if nf < 1:
+        raise ValueError(f"nf must be >= 1, got {nf}")
+
+    dev = torch.device(device)
+    if dev.type == "cuda" and is_nv_available():
+        return "nv"
+    if nf == 1 and is_vesin_torch_available():
+        return "vesin"
+    if dev.type == "cuda" and not is_nv_available():
+        log.warning(
+            "nvalchemi-toolkit-ops is unavailable; falling back from "
+            "neighbor_graph_method='auto' to the dense graph builder"
+            + (
+                ""
+                if nf == 1
+                else " (vesin is not used for nf>1; its API loops frames in Python)"
+            )
+            + ". Install it with `pip install nvalchemi-toolkit-ops` to enable "
+            "the NV graph builder."
+        )
+    return "dense"
+
+
 def resolve_neighbor_graph_method(
     requested: str,
     device: torch.device,
@@ -36,6 +113,8 @@ def resolve_neighbor_graph_method(
     -------
     str
         The concrete builder name, either ``"dense"`` or ``"nv"``.
+        Training auto never selects ``vesin`` (per-frame Python loop); use
+        :func:`resolve_auto_graph_builder` for inference auto selection.
 
     Raises
     ------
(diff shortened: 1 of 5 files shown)


Shaurya2k06:
Rebased onto current master after #5912 / #5913 landed the training/eager auto path.

**What changed in this update**
- Dropped the superseded duplicate model/training default flip (that already lives in `graph_builder.resolve_neighbor_graph_method` + training config).
- Kept the unique follow-up: inference DeepEval `auto` now uses shared `resolve_auto_graph_builder` with the original ladder (CUDA: nv→vesin→dense; CPU: vesin→dense).
- Training auto still keeps CPU on `dense` so multi-frame training does not hit vesin's per-frame Python loop (that was the CI failure mode on the previous tip).
- Vesin docstring now states the importability condition for auto selection.

CodeQL empty-`except` alert and the prior Test Python reshape failures were on the old tip that selected vesin during training; they should clear on this tip.

OutisLi (approved):

deepmd/pt_expt/utils/graph_builder.py
```diff
@@ -19,6 +19,47 @@
...
+        if is_nv_available():
+            return "nv"
+        if is_vesin_torch_available():
+            return "vesin"
+        log.warning(
+            "nvalchemi-toolkit-ops and vesin[torch] are unavailable; falling "
+            "back from neighbor_graph_method='auto' to the dense graph builder."
+        )
+        return "dense"
+    if is_vesin_torch_available():
```
wanghan-iapcm: This is the change the PR really makes, and it re-adds what #5912 removed.

On #5912 I raised this against an identical ladder:

> This branch puts vesin on the default path, which contradicts the builder's own documented contract. [...] After this change, any environment with `vesin.torch` installed and without CUDA + nvalchemiops gets that per-frame Python loop by default. That includes runs under `auto_batch_size`, which deliberately batches many frames into a single `_eval_model_graph` call - exactly where the vectorised dense builder handles all frames at once and the vesin loop is worst. [...] Either prefer `dense` when `nframes > 1`, or update the contract to say vesin is now a default and re-examine whether the per-frame loop is acceptable there.

That rung was dropped and #5912 merged without it. This commit restores it and takes the second half of that either/or -- the module docstring is rewritten -- but not the first: there is no `nframes` gate, and no measurement re-examining the loop.

The reason I do not think the docstring rewrite settles it is that the repository already has an answer to this exact question, and it goes the other way. [`_select_neighbor_builder`](https://github.com/deepmodeling/deepmd-kit/blob/master/deepmd/pt/model/model/sezm_model.py#L528-L562) picks between the same two builders:

```python
if device.type == "cpu" and nf == 1 and is_vesin_torch_available():
    return VesinNeighborList()
```

with the rationale stated directly above it: *"Every other case -- any CUDA input or any multi-frame batch -- uses `nvalchemiops`, whose batched kernel amortizes the launch cost across frames."* It takes `nf`, and it is called per forward. This resolver takes only `device` and is called once from `_setup_neighbor_backend` inside `DeepEval.__init__`, before any frame count exists -- so it cannot express that policy even if we wanted it to. That is a structural difference, not a parameter we forgot to thread through.

Three things make the exposure wider than it looks. `vesin[torch]` is an unconditional dependency of the `torch` extra, so this is the CPU default for essentially every pt user rather than an opt-in for people who installed something extra. `auto_batch_size` defaults to `True`, and `_eval_model_graph` receives whole batches from `execute_all`, so multi-frame is the normal case for `dp test` and `dp model-devi`, not the exception. And the builder's own scope note puts the cost at *"~1 ms/frame call overhead"*. For a small system over many frames the arithmetic points at a slowdown, which inverts the PR's stated goal.

What would resolve it, in order of preference: move the resolution to call time and gate on `nf == 1` for vesin, matching `_select_neighbor_builder`; or keep construction-time resolution and drop the vesin rung, leaving it explicit opt-in as #5912 concluded; or keep it and post a benchmark over a realistic `dp test` batch showing dense is not faster. Any of the three is fine by me -- what I do not want is the decision being reversed silently.

deepmd/pt_expt/utils/graph_builder.py
```diff
@@ -19,6 +19,47 @@
 log = logging.getLogger(__name__)
 
 
+def resolve_auto_graph_builder(
```
wanghan-iapcm, line 27: The summary describes two changes that are not in this diff.

It says the PR will *"Flip the pt_expt model-level default (`None` / `"auto"`) from hard-coded `"dense"` to that ladder"* and *"use the same helper in ... compiled training's eager `_forward_graph`"*. Neither `deepmd/pt_expt/model/make_model.py` nor `deepmd/pt_expt/train/training.py` appears in the changed files, and both still resolve the same way they did before:

- [`make_model.py`](https://github.com/deepmodeling/deepmd-kit/blob/master/deepmd/pt_expt/model/make_model.py) `_resolve_graph_method` still ends at `getattr(self, "neighbor_graph_method", "dense")`, so a model driven directly still defaults to `dense` on every device;
- `training.py` still imports only `resolve_neighbor_graph_method`, and `_forward_graph` still reads `getattr(_model, "neighbor_graph_method", "dense")`.

`build_neighbor_graph_for_method` also has no `"auto"` branch -- it raises `ValueError` on anything it does not recognise -- so `"auto"` could not reach it even if the model default did produce it.

I think this is stale text rather than a missing change: those two flips landed in #5912 and #5913, which are already ancestors of this branch. The "Why existing tests missed this" paragraph has the same problem, since it describes fixing compiled training's hardcoded dense. Worth rewriting the body to what the commit does -- add vesin to the inference auto ladder and extract the shared helper -- because as written a reviewer would look for a model-level behaviour change that is not here, and a bisect later would be misled about where the flip came from.

deepmd/pt_expt/utils/graph_builder.py
```diff
@@ -19,6 +19,47 @@
...
+    device: torch.device | str,
+) -> str:
+    """Resolve ``neighbor_graph_method="auto"`` to a concrete inference builder.
```
wanghan-iapcm, line 31: Missing the numpydoc sections the rest of this module uses.

This is a public function -- no leading underscore, imported by `deep_eval.py` -- and its sibling `resolve_neighbor_graph_method` a few lines below carries full `Parameters` / `Returns` / `Raises` blocks. This one documents the ladder in prose and a bullet list with no `Parameters` for `device` and no `Returns` for the `str`.

`_select_neighbor_builder` in `deepmd/pt/model/model/sezm_model.py`, which does the analogous job, also documents `nf` and `device` under `Parameters` and its return under `Returns`. Matching that is a small edit and keeps the API docs uniform.

source/tests/pt_expt/model/test_graph_builder_dispatch.py
```diff
@@ -138,6 +138,33 @@ def test_explicit_nv_rejects_cpu():
         resolve_neighbor_graph_method("nv", torch.device("cpu"))
 
 
+@pytest.mark.parametrize(
```
wanghan-iapcm, line 148: These five cases pin the ladder, and I want to say they are the right shape -- patching availability and device rather than recomputing the cascade in the assertion is exactly what was missing from the equivalent test on #5912.

The gap is that they are the *only* new coverage, and they are pure resolver assertions: they never build a graph. The cross-builder numerical checks that do build one, `test_vesin_matches_dense_energy_force` and `test_nv_matches_dense_energy_force` further up this file, share the `_eval` helper, and it is single-frame:

```python
coord = torch.tensor(rng.random((1, 6, 3)) * 4.0, ...)
```

So parity between vesin and dense is established only at `nf == 1` -- the regime the vesin builder's own docstring says it is *"intended for"* -- while this PR makes vesin the default precisely for the batched regime, where nothing compares it against dense. That matters slightly more than a generic coverage note because the graph lower accumulates with `segment_sum` over edges and edge ordering after `canonicalize` is an independent implementation per builder, so multi-frame agreement does not follow from single-frame agreement; it needs its own case.

A parametrization of `_eval` over `nf` in `{1, 4}` would cover it, and it is the test I would want in place before the default moves, whichever way the vesin question above is settled.

wanghan-iapcm (changes requested):
Thanks for picking this up. Before the objections: the part I asked for on #5912 and did not get there, you have delivered here. The five ladder branches are driven deterministically by patching `is_nv_available` / `is_vesin_torch_available` / the device rather than recomputing the cascade inside the assertion, so `test_resolve_auto_graph_builder_ladder` genuinely fails without the change instead of comparing the code to a copy of itself. Extracting the ladder into one function so DeepEval and training stop carrying private copies is also the right direction.

My problem is with what the remaining commit actually changes, and with the description of it. Four comments inline.

The short version: the default flip this PR is named for already shipped -- #5912 landed DeepEval's `auto` ladder and #5913 landed training's, and both are ancestors of this branch. What is left is one substantive commit, and its real content is that `vesin` becomes an automatically selected builder. That is precisely the rung that was removed from #5912 in response to review before it merged, and the reasoning that removed it has not changed.

To be clear about what does *not* conflict: #5913 owns `resolve_neighbor_graph_method` for training and this adds a separate inference-only function, so the two are complementary and the branch merges cleanly. I am not asking you to rebase or coordinate with it.

Shaurya2k06:
Addressed the review on the current tip ([`be6d9bb`](https://github.com/Shaurya2k06/deepmd-kit/commit/be6d9bb4)):

1. **Vesin on auto** — took the preferred option: resolve at call time in `_build_eval_graph` and gate vesin on `nf == 1`, matching `_select_neighbor_builder`. Multi-frame `auto_batch_size` batches stay on `nv`/`dense`.
2. **PR body** — rewritten to what the commit actually does (inference ladder + shared helper only; no model/training flip — those are #5912/#5913).
3. **numpydoc** — `resolve_auto_graph_builder` now has Parameters / Returns like its sibling.
4. **Coverage** — ladder cases include `nf in {1, 4}`; `test_vesin_matches_dense_energy_force` (and nv) parametrized over the same.

deepmd/pt_expt/utils/graph_builder.py
Shaurya2k06: fixed

deepmd/pt_expt/utils/graph_builder.py
Shaurya2k06: fixed

source/tests/pt_expt/model/test_graph_builder_dispatch.py
Shaurya2k06: fixed

deepmd/pt_expt/utils/graph_builder.py
Shaurya2k06: yep done too

deepmd/pt_expt/utils/graph_builder.py
```diff
@@ -19,6 +19,75 @@
...
+    """
+    from deepmd.pt.utils.nv_nlist import (
+        is_nv_available,
+    )
+    from deepmd.pt_expt.utils.vesin_neighbor_list import (
+        is_vesin_torch_available,
+    )
+
+    if nf < 1:
+        raise ValueError(f"nf must be >= 1, got {nf}")
```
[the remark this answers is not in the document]
Shaurya2k06: Done. nf now requires a real int (bool excluded) before the >= 1 check; tests cover None, 1.5, True/False, 0, and -1.

deepmd/pt_expt/utils/graph_builder.py
```diff
@@ -19,6 +19,83 @@
...
+
+    dev = torch.device(device)
+    if dev.type == "cuda" and is_nv_available():
+        return "nv"
+    if nf == 1 and is_vesin_torch_available():
+        return "vesin"
+    if dev.type == "cuda" and not is_nv_available():
+        log.warning(
```
wanghan-iapcm: This warning now fires once per batch instead of once per `DeepEval`.

Moving resolution to call time is right, but it also moved this log line onto the per-batch path. `_build_eval_graph` is reached from `_eval_model_graph`, which [`_eval_func`](https://github.com/deepmodeling/deepmd-kit/blob/abe2e033a20af33c4bee341af7d59e418b69be64/deepmd/pt_expt/infer/deep_eval.py#L1174-L1182) hands to `self.auto_batch_size.execute_all(...)`, so it runs once per batch rather than once at construction. On a CUDA host without `nvalchemiops` evaluating multi-frame batches -- exactly the configuration this message is written for -- a `dp test` or `dp model-devi` run will now repeat it for every batch. Previously `_setup_neighbor_backend` resolved once and the user saw it a single time.

The message is also the more useful of the two things happening here, so burying it in repetition is a real loss: the point is "install nvalchemi-toolkit-ops", which is a one-time action, not a per-batch one.

A module-level warn-once flag is the smallest fix:

```python
_warned_no_nv = False
...
if dev.type == "cuda" and not nv:
    global _warned_no_nv
    if not _warned_no_nv:
        _warned_no_nv = True
        log.warning(...)
```

Caching the whole resolver per `(device.type, nf, availability)` would work too and would also pick up the second point: `is_nv_available()` is called twice on this path, once at line 80 and again in this condition. Hoisting it to a local reads better and halves the import probes -- which matter slightly more than they look, because on CPU-only hosts `is_nv_available` wraps its import in `_suppress_native_stderr`, so it is doing `dup`/`dup2`/`close` syscalls on every call now rather than once per model.

None of this affects results -- all builders emit the same carry-all neighbor set -- so it is not a correctness concern, just something that belongs with this change rather than after it.

wanghan-iapcm (commented):
All four points are addressed, and this is a real rework rather than a wording change. Checked against `abe2e033a`.

The vesin rung is now what I was asking for. `_resolve_neighbor_graph_method` takes `nf` and returns `"auto"` unresolved when it is omitted, so construction defers and `_build_eval_graph` resolves from the batch frame count -- which is the structural change, not just a new parameter. I went looking for the ways that could go wrong and they are all covered: `coord_input` at both call sites is `coords.reshape(nframes, natoms, 3)`, so `np.asarray` is safe and `shape[0]` really is the frame count; `"auto"` leaks nowhere, since the only reader of `self._neighbor_graph_method` is the dispatch in `_build_eval_graph`; and the construction-time raise for a non-default value on a non-graph artifact is intact. Citing `_select_neighbor_builder` in the docstring is the right call, since that is the policy this now has to stay in step with.

The rewritten summary is accurate, including the explicit note that the model-level and training flips shipped in #5912 / #5913 -- that was the part most likely to mislead a bisect later. The numpydoc sections match the sibling function. And `_eval` parametrized over `nf in {1, 4}`, with `atype` and `box` expanded to match, closes the parity gap in the batched regime rather than only asserting the resolver's return value.

One thing the move to call-time resolution brings with it, inline.

Worth noting the tests had not run when I read this -- every substantive workflow was sitting at `action_required`. I have approved them, so `Test Python` / `Test C++` / `Test CUDA` are going now. Everything above is from reading the head, so the new ladder cases and the `nf=4` parity case still need a green run to count as verified.

deepmd/pt_expt/utils/graph_builder.py
Shaurya2k06: Fixed.
 
- _warned_auto_no_nv , warn once per process
- nv_available = is_nv_available() hoisted ,  one probe per resolve
- test that a second call does not warn again

wanghan-iapcm (approved):
All five points are resolved, checked against `2d386b63d`.

The warn-once fix is right: a module-level `_warned_auto_no_nv` with the `global` declaration, and the `is_nv_available()` probe hoisted so a resolve costs one call instead of two. `test_resolve_auto_graph_builder_warns_once` calls the resolver twice on `("cuda", nf=4)` and asserts exactly one message, which fails on the previous head where both calls warned -- so it is a real regression test, not a restatement of the code.

I also closed out the caveat from my last review. `vesin[torch]` is an unconditional entry in the torch extra, so the `skipif not is_vesin_torch_available()` guard does not trip in CI, which means `test_vesin_matches_dense_energy_force[4]` genuinely executed. That was the whole point of the parity gap, and the suite is green across all 54 checks this time rather than sitting at `action_required`, so the ladder and parity coverage is now verified by a run rather than by reading.

The end state is the one I was arguing for on #5912: vesin is reachable automatically, but only where its per-frame Python loop is not a liability, and the policy lives in one function that cites `_select_neighbor_builder` as the thing it has to stay in step with. Thanks for working through the iterations on this one.

PR #5904, opened 2026-07-25
fix(tabulate): correct sorted padding two-embed gradients

### LINKED ISSUE #5891: [Code scan] Sorted SE-Attention padding has inconsistent `two_embed` gradients and double backward
This issue comes from a focused review of SE-Attention tabulation at commit `21b73ab303581f467bba0cae101a880c35cc2d82`.

## Problem

For sorted padding, the forward kernel folds the full padding tail at the first sentinel. It reads only that sentinel's `two_embed` value and multiplies its contribution by the tail length:

- [read the first sentinel's `two_embed`](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/source/lib/src/gpu/tabulate.cu#L341-L345)
- [multiply the folded contribution by `nnei - breakpoint`](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/source/lib/src/gpu/tabulate.cu#L348-L357)

However, the backward kernel writes the unscaled `two_embed` gradient to every entry in the folded tail:

- [copy `oldres * res` across the tail](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/source/lib/src/gpu/tabulate.cu#L475-L481)

The CPU implementation has the same behavior. This is not the Jacobian of the implemented forward function. If the sentinel is at `p` and the tail length is `m`, forward depends on `t[p]` as `m * base * (1 + t[p])`; its derivative with respect to the tail is `[m * base, 0, ...]`, not `[base, base, ...]`.

The double-backward path is inconsistent with the implemented first backward as well. It reads only `dz_dy_dtwo[p]` and then multiplies the whole expression by the tail length:

- [read one `dz_dy_dtwo` entry](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/source/lib/src/gpu/tabulate.cu#L554-L562)
- [multiply it by `nnei - breakpoint`](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/source/lib/src/gpu/tabulate.cu#L587-L591)

Because first backward copies the gradient to all tail positions, differentiating that backward requires summing all corresponding `dz_dy_dtwo` entries, not repeating only the first one.

## Reproduction

A minimal sorted-tail case was run inside an `srun --partition=main --gres=gpu:5090:1` allocation.

First-order gradient versus central finite difference:

```text
custom [0.2, 0.2, 0.2]
finite_difference [0.5999999999617422, 0.0, 0.0]
```

Double backward using tail weights `[1, 2, 4]`:

```text
first_backward [0.2, 0.2, 0.2]
double_backward [0.6000000000000001, 0.0, 0.0, 0.0]
expected_first_component 1.4
```

A direct CUDA-library harness reproduced the same double-backward result on an NVIDIA GeForce RTX 5090:

```text
atten_gradgrad actual=0.6 expected=1.4
```

## Impact

The custom op fails finite-difference checks with respect to `two_embed`, and higher-order gradients can silently be wrong. This affects both CPU and GPU backends and both TensorFlow/PyTorch callers of the shared library implementation.

## Suggested fix

Choose one coherent contract and implement all three stages consistently:

1. If forward keeps the folded representation, make only the first sentinel depend on `two_embed`, scale that gradient by the tail length, and mirror it in grad-grad.
2. If the intended function depends independently on every tail `two_embed`, do not fold those values in forward, and make grad-grad sum all tail cotangents.

Add finite-difference tests that perturb tail `two_embed` entries independently and a double-backward test with nonuniform tail cotangents.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

njzjz:
## Summary

- preserve the folded sorted-padding forward contract and make only the first sentinel depend on `two_embed`
- scale that sentinel gradient by the padding-tail length while leaving later padding gradients zero on CPU and GPU
- add finite-difference and nonuniform-cotangent grad-grad regression tests, including a GPU tail longer than the four-warp tile

## Validation

- `source/build/lib/tests/runUnitTests_lib --gtest_filter='TestTabulateSeA.*:TestTabulateSeASortedPaddingTwoEmbed.*'`
- CUDA 12.4 build of `deepmd_op_cuda` and `runUnitTests_lib`
- RTX 5090 Slurm run of all 8 existing/new SE-A CPU and GPU tests
- `ruff format .`
- `ruff check .`
- `clang-format --dry-run --Werror` on changed C++/CUDA files

Closes #5891

Note: #5844 modifies the same GPU kernel for the shared-breakpoint fix, so the later-merging branch may need a small conflict resolution; the gradient contract fixed here is independent.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected sorted-padding two-embedding gradient behavior so only the first sentinel receives the full tail multiplicity; later padding sentinel gradients remain zero.
  * Updated CPU and GPU gradient/grad-grad accumulation to apply the padding-tail repeat factor consistently.
  * Ensured the GPU two-embedding gradient buffer is cleared only when applicable to avoid stale values.
* **Tests**
  * Strengthened sorted-padding assertions and expanded GPU coverage to confirm forward output consistency with CPU before numerical gradient checks.

### DIFF
--- a/source/lib/src/gpu/tabulate.cu
+++ b/source/lib/src/gpu/tabulate.cu
@@ -453,10 +453,26 @@ __global__ void tabulate_fusion_se_a_grad_fifth_order_polynomial(
       Csub += (nnei - breakpoint) * res_grad *
               (enable_se_atten ? res * t + res : res);
       if (enable_se_atten) {
-        // from ii to ii + (nnei - breakpoint)
-        for (int ii2 = ii; ii2 < ii + nnei - breakpoint; ii2++) {
-          dy_dtwo[block_idx * nnei * last_layer_size + ii2 * last_layer_size +
+        if (!unloop) {
+          dy_dtwo[block_idx * nnei * last_layer_size + ii * last_layer_size +
                   jj] = oldres * res;
+        } else {
+          // Several warps can encounter different entries in the same sorted
+          // padding tail. Only its first sentinel is used by forward, so only
+          // that entry receives the repeat-scaled two-embedding gradient.
+          bool first_padding_sentinel = ii == 0;
+          if (ii > 0) {
+            const int_64 previous_em_index =
+                block_idx * nnei * MTILE + (ii - 1) * MTILE;
+            first_padding_sentinel = em_x[block_idx * nnei + ii - 1] != ago ||
+                                     em[previous_em_index + 1] != 0. ||
+                                     em[previous_em_index + 2] != 0. ||
+                                     em[previous_em_index + 3] != 0.;
+          }
+          if (first_padding_sentinel) {
+            dy_dtwo[block_idx * nnei * last_layer_size + ii * last_layer_size +
+                    jj] = (nnei - ii) * oldres * res;
+          }
         }
       }
     }
@@ -534,8 +550,9 @@ __global__ void tabulate_fusion_se_a_grad_grad_fifth_order_polynomial(
     if (enable_se_atten) {
       FPTYPE t = two_embed[block_idx * nnei * last_layer_size +
                            ii * last_layer_size + thread_idx];
-      // dz_dy_dtwo * res * em
-      // res above should be used instead of res + res * t below
+      // For sorted padding, only the first sentinel has a nonzero
+      // two-embedding gradient. The repeat factor below applies its full tail
+      // multiplicity to this cotangent.
       two_grad = dz_dy_dtwo[block_idx * nnei * last_layer_size +
                             ii * last_layer_size + thread_idx] *
                  res;
@@ -1071,6 +1088,12 @@ void tabulate_fusion_se_a_grad_gpu(FPTYPE* dy_dem_x,
   DPErrcheck(gpuDeviceSynchronize());
   DPErrcheck(gpuMemset(dy_dem_x, 0, sizeof(FPTYPE) * nloc * nnei));
   DPErrcheck(gpuMemset(dy_dem, 0, sizeof(FPTYPE) * nloc * nnei * 4));
+  if (two_embed != nullptr) {
+    // The sorted-padding fast path writes only the first sentinel. Explicitly
+    // clear the unused tail because framework output buffers are uninitialized.
+    DPErrcheck(
+        gpuMemset(dy_dtwo, 0, sizeof(FPTYPE) * nloc * nnei * last_layer_size));
+  }
 
   tabulate_fusion_se_a_grad_fifth_order_polynomial<FPTYPE, MM, KK>
       <<<nloc, KK * WARP_SIZE, sizeof(FPTYPE) * MM * last_layer_size>>>(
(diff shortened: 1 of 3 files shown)


wanghan-iapcm (approved):
Verified this by building `tabulate.cc` at both the base and the head and running finite-difference drivers against the real `deepmd::` entry points, rather than by reading the diff.

The new contract is exactly the derivative of the folded forward. Since the fold multiplies the sentinel row's contribution by `(nnei - jj)` and uses only that row's two-embedding value before breaking, the output depends on `two_embed[jj]` with that multiplicity and does not depend on the later padding rows at all. Analytic and central-difference agree at the sentinel (1.0) and on the tail (0.0), where the old code produced 0.2 in every slot -- wrong in magnitude at the sentinel and wrong in support on four rows the forward never reads. The old tail sum happened to come out right, which is presumably why this survived so long. `dy_dem` and `dy_dem_x` are byte-identical to base.

The part I found most convincing is that `grad_grad` needed no functional change, because it was already the JVP of the new contract. That means `grad` and `grad_grad` were implementing mutually inconsistent contracts before this PR, and second derivatives of compressed se_atten were wrong accordingly. With a non-uniform cotangent, the finite difference of the backward gives 6.2 at base against the grad-grad kernel's 1.0; at head both are 1.0. So this is not only a gradient fix, it restores self-consistency between the two stages -- and the two comment-only hunks there correctly describe what the surrounding code already did.

The added `gpuMemset` is genuinely required rather than defensive: the tail used to be fully overwritten by the copy loop, so a partial write now needs it, and the TF op hands in `allocate_output` memory that is uninitialized. The CPU side already had the equivalent memset, so the "retains the zero initialized above" comment is accurate. `grad_grad` has a single output that is fully zeroed on both sides, so there is no missing counterpart there.

On the tests: these fail pre-fix, which I checked by running them rather than inferring it. Eleven assertions fail at the base and none at the head, with discrepancies of 0.2 to 5.2 against a 1e-9 tolerance. The oracles are independent of the implementation -- central difference of the forward, plus a grad-to-grad-grad consistency relation with a deliberately non-uniform cotangent so that any tail leakage shows up -- rather than restated implementation constants. Seeding the GPU output buffer with 7.0 to make the memset load-bearing, and choosing `nnei = 6` so the tail exceeds the four-warp tile and covers entries no warp visits, are both good touches.

Worth recording why nothing caught this earlier: `dy_dtwo` has been produced since 2023 and no test asserted its values at all until three days ago. The one that did asserted CPU/GPU parity, which cannot catch a bug both sides share. This PR is the first time the output has been checked against an oracle independent of the kernels.

PR #4227, opened 2024-10-17
Add 4 pt descriptor compression

cherryWangY:
se_a, se_atten(DPA1), se_t, se_r

## Release Notes

- **New Features**
  - Introduced a model compression feature across multiple descriptor classes, enhancing performance and efficiency.
  - Added `enable_compression` methods to various classes, allowing users to enable and configure compression settings.

- **Bug Fixes**
  - Improved error handling for unsupported compression scenarios and parameter validation.

- **Tests**
  - Added comprehensive unit tests for new compression functionalities across multiple descriptor classes to ensure accuracy and reliability.

- **Documentation**
  - Enhanced documentation for new methods and classes to clarify usage and parameters related to compression.

PR #5905, opened 2026-07-25
fix(pt): honor is_sorted in SE-Attention tabulation

### LINKED ISSUE #5890: [Code scan] PyTorch SE-Attention tabulation ignores `is_sorted` in forward and first backward
This issue comes from a focused review of SE-Attention tabulation at commit `21b73ab303581f467bba0cae101a880c35cc2d82`.

## Problem

The public PyTorch op accepts and stores `is_sorted`, but its forward helper does not accept that flag:

- [`TabulateFusionSeAForward` has no `is_sorted` parameter](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/source/op/pt/tabulate_multi_device.cc#L21-L28)
- [both CPU and GPU backend calls omit the final argument](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/source/op/pt/tabulate_multi_device.cc#L60-L71)
- [the omitted backend argument defaults to `true`](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/source/lib/include/tabulate.h#L6-L16)

The first-backward path has the same problem:

- [`TabulateFusionSeAGradForward` has no `is_sorted` parameter](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/source/op/pt/tabulate_multi_device.cc#L75-L85)
- [its CPU and GPU calls also omit the flag](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/source/op/pt/tabulate_multi_device.cc#L112-L125)
- [`TabulateFusionSeAttenGradOp` receives the flag but does not pass it into that helper](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/source/op/pt/tabulate_multi_device.cc#L857-L878)

Only grad-grad passes the saved flag. TensorFlow passes it in all three paths.

This matters for compressed PyTorch SE-Attention: [`exclude_types` sets `self.is_sorted = false`](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/deepmd/pt/model/descriptor/se_atten.py#L446-L452), and [that value is supplied to the custom op](https://github.com/deepmodeling/deepmd-kit/blob/21b73ab303581f467bba0cae101a880c35cc2d82/deepmd/pt/model/descriptor/se_atten.py#L755-L766), but the op behaves as if it were `true`.

## Reproduction

Inside an `srun --partition=main --gres=gpu:5090:1` allocation, a minimal PyTorch custom-op case with an unsorted sentinel-like row produced:

```text
is_sorted=false [0.6000000000000001, 0.0, 0.0, 0.0]
is_sorted=true  [0.6000000000000001, 0.0, 0.0, 0.0]
```

Calling the CUDA library entry point directly with the same logical case proves that the backend kernel itself distinguishes the flag:

```text
atten_unsorted false=1.9 true=0.6 expected_false=1.9 expected_true=0.6
```

The validation GPU was an NVIDIA GeForce RTX 5090 (driver 580.95.05, compute capability 12.0).

## Impact

Compressed PyTorch SE-Attention models with excluded type pairs can silently fold an unsorted neighbor row as sorted, producing incorrect descriptors and first-order gradients. Grad-grad then uses the requested flag, so the autograd stages can also disagree with one another.

## Suggested fix

Add `is_sorted` to both PyTorch helper signatures and pass it explicitly to the CPU/GPU library functions. Add forward, backward, and double-backward tests where `is_sorted=false` produces a different result from `true`.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

njzjz:
## Summary

- forward `is_sorted` through the PyTorch SE-Attention forward helper to the CPU and GPU library entry points
- forward the saved flag through the first-backward helper so all autograd stages use the same sorting contract
- require a sorted lower neighbor list while DPA1 geometric compression is active, matching the tabulation fold invariant
- add float32 and float64 regression coverage for unsorted forward, backward, and double-backward behavior, including a nonzero table derivative
- add compressed DPA1 `forward_lower` coverage for unsorted rcut+skin lists and an end-to-end plain `se_atten` case with `exclude_types` and `set_davg_zero: true`
- document why the non-attention `se_a` path intentionally keeps the sorted fold enabled

Fixes #5890

## Historical origin

- The PyTorch custom-op forwarding bug was introduced by #3877, which added `tabulate_fusion_se_atten` while omitting the backend `is_sorted` argument in forward and first backward.
- The compressed DPA1 lower-nlist exposure became reachable in #4227, which added PyTorch `se_atten` geometric compression while `need_sorted_nlist_for_lower()` still returned `False`. #3993 introduced that API/default earlier, but compression did not yet use the sorted fold.

## Performance note

With non-empty `exclude_types`, the corrected `is_sorted=False` path traverses the full neighbor list instead of applying the sorted-padding fold. This may be measurably slower, but avoids silently folding interleaved excluded rows.

## Testing

- CPU-only PyTorch `deepmd_op_pt` C++ target build
- `python -m pytest source/tests/pt/test_tabulate_fusion_se_atten.py -v` (12 passed)
- `python -m pytest source/tests/pt/model/test_compressed_se_atten_forward_lower.py -v` (1 passed)
- `python -m pytest source/tests/pt/test_model_compression_se_atten.py::TestDeepPotATPBCExcludeTypes::test_1frame -v` (1 passed)
- `ruff format .`
- `ruff check` on all modified Python files
- commit hooks, including Ruff, pylint, and clang-format
- `dp --version`
- `dp --pt -h`
- Python imports for `deepmd` and `deepmd.pt`

`ruff check .` was also run; it reports five pre-existing findings in `deepmd/jax/jax_md/__init__.py` and `deepmd/tf/entrypoints/__init__.py`, outside this PR.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

### DIFF
--- a/source/op/pt/tabulate_multi_device.cc
+++ b/source/op/pt/tabulate_multi_device.cc
@@ -25,6 +25,7 @@ void TabulateFusionSeAForward(const torch::Tensor& table_tensor,
                               const torch::Tensor& em_tensor,
                               const torch::Tensor& two_embed_tensor,
                               int64_t last_layer_size,
+                              bool is_sorted,
                               torch::Tensor& descriptor_tensor) {
   // check input shape
   if (table_tensor.dim() != 2) {
@@ -60,7 +61,8 @@ void TabulateFusionSeAForward(const torch::Tensor& table_tensor,
   if (device == "GPU") {
 #if GOOGLE_CUDA || TENSORFLOW_USE_ROCM
     deepmd::tabulate_fusion_se_a_gpu(descriptor, table, table_info, em_x, em,
-                                     two_embed, nloc, nnei, last_layer_size);
+                                     two_embed, nloc, nnei, last_layer_size,
+                                     is_sorted);
 #else
     throw std::runtime_error(
         "The input tensor is on the GPU, but the GPU support for the "
@@ -68,7 +70,8 @@ void TabulateFusionSeAForward(const torch::Tensor& table_tensor,
 #endif  // GOOGLE_CUDA || TENSORFLOW_USE_ROCM
   } else if (device == "CPU") {
     deepmd::tabulate_fusion_se_a_cpu(descriptor, table, table_info, em_x, em,
-                                     two_embed, nloc, nnei, last_layer_size);
+                                     two_embed, nloc, nnei, last_layer_size,
+                                     is_sorted);
   }
 }
 
@@ -80,6 +83,7 @@ void TabulateFusionSeAGradForward(const torch::Tensor& table_tensor,
                                   const torch::Tensor& two_embed_tensor,
                                   const torch::Tensor& dy_tensor,
                                   const torch::Tensor& descriptor_tensor,
+                                  bool is_sorted,
                                   torch::Tensor& dy_dem_x_tensor,
                                   torch::Tensor& dy_dem_tensor,
                                   torch::Tensor& dy_dtwo_tensor) {
@@ -111,18 +115,18 @@ void TabulateFusionSeAGradForward(const torch::Tensor& table_tensor,
   // compute
   if (device == "GPU") {
 #if GOOGLE_CUDA || TENSORFLOW_USE_ROCM
-    deepmd::tabulate_fusion_se_a_grad_gpu(dy_dem_x, dy_dem, dy_dtwo, table,
-                                          table_info, em_x, em, two_embed, dy,
-                                          nloc, nnei, last_layer_size);
+    deepmd::tabulate_fusion_se_a_grad_gpu(
+        dy_dem_x, dy_dem, dy_dtwo, table, table_info, em_x, em, two_embed, dy,
+        nloc, nnei, last_layer_size, is_sorted);
 #else
     throw std::runtime_error(
         "The input tensor is on the GPU, but the GPU support for the "
         "customized OP library is not enabled.");
 #endif  // GOOGLE_CUDA || TENSORFLOW_USE_ROCM
   } else if (device == "CPU") {
-    deepmd::tabulate_fusion_se_a_grad_cpu(dy_dem_x, dy_dem, dy_dtwo, table,
-                                          table_info, em_x, em, two_embed, dy,
-                                          nloc, nnei, last_layer_size);
+    deepmd::tabulate_fusion_se_a_grad_cpu(
+        dy_dem_x, dy_dem, dy_dtwo, table, table_info, em_x, em, two_embed, dy,
+        nloc, nnei, last_layer_size, is_sorted);
   }
 }
 
@@ -643,7 +647,7 @@ class TabulateFusionSeAGradOp
     // compute
     TabulateFusionSeAGradForward<FPTYPE>(
         table_tensor, table_info_tensor, em_x_tensor, em_tensor, at::Tensor(),
-        dy_tensor, descriptor_tensor, dy_dem_x_tensor, dy_dem_tensor,
+        dy_tensor, descriptor_tensor, true, dy_dem_x_tensor, dy_dem_tensor,
         dy_dtwo_tensor);
     // save data
     ctx->save_for_backward({table_tensor, table_info_tensor, em_x_tensor,
@@ -785,7 +789,7 @@ class TabulateFusionSeAOp
     // compute
     TabulateFusionSeAForward<FPTYPE>(table_tensor, table_info_tensor,
                                      em_x_tensor, em_tensor, at::Tensor(),
-                                     last_layer_size, descriptor_tensor);
+                                     last_layer_size, true, descriptor_tensor);
     // save data
     ctx->save_for_backward({table_tensor, table_info_tensor, em_x_tensor,
                             em_tensor, descriptor_tensor});
@@ -870,8 +874,8 @@ class TabulateFusionSeAttenGradOp
     torch::Tensor dy_dtwo_tensor = torch::zeros_like(two_embed_tensor);
     TabulateFusionSeAGradForward<FPTYPE>(
         table_tensor, table_info_tensor, em_x_tensor, em_tensor,
-        two_embed_tensor, dy_tensor, descriptor_tensor, dy_dem_x_tensor,
-        dy_dem_tensor, dy_dtwo_tensor);
+        two_embed_tensor, dy_tensor, descriptor_tensor, is_sorted,
+        dy_dem_x_tensor, dy_dem_tensor, dy_dtwo_tensor);
 
     ctx->save_for_backward({table_tensor, table_info_tensor, em_x_tensor,
                             em_tensor, two_embed_tensor, descriptor_tensor});
@@ -969,9 +973,9 @@ class TabulateFusionSeAttenOp
     torch::Tensor descriptor_tensor =
         torch::empty({em_tensor.size(0), 4, last_layer_size}, options);
     // compute
-    TabulateFusionSeAForward<FPTYPE>(table_tensor, table_info_tensor,
-                                     em_x_tensor, em_tensor, two_embed_tensor,
-                                     last_layer_size, descriptor_tensor);
+    TabulateFusionSeAForward<FPTYPE>(
+        table_tensor, table_info_tensor, em_x_tensor, em_tensor,
+        two_embed_tensor, last_layer_size, is_sorted, descriptor_tensor);
     // save data
     ctx->save_for_backward({table_tensor, table_info_tensor, em_x_tensor,
                             em_tensor, two_embed_tensor, descriptor_tensor});
--- a/source/op/pt/tabulate_multi_device.cc
+++ b/source/op/pt/tabulate_multi_device.cc
@@ -645,6 +645,10 @@ class TabulateFusionSeAGradOp
     torch::Tensor dy_dem_tensor = torch::zeros_like(em_tensor);
     torch::Tensor dy_dtwo_tensor = at::Tensor();
     // compute
+    // The non-attention se_a path invokes this op per type-pair block, so
+    // exclusions cannot interleave zero rows. Compressed forward_lower also
+    // requests a sorted nlist through the Python-side
+    // DescrptBlockSeA.need_sorted_nlist_for_lower() contract.
     TabulateFusionSeAGradForward<FPTYPE>(
         table_tensor, table_info_tensor, em_x_tensor, em_tensor, at::Tensor(),
         dy_tensor, descriptor_tensor, true, dy_dem_x_tensor, dy_dem_tensor,
@@ -787,6 +791,8 @@ class TabulateFusionSeAOp
     torch::Tensor descriptor_tensor =
         torch::empty({em_tensor.size(0), 4, last_layer_size}, options);
     // compute
+    // Keep the sorted fold enabled: exclusions are uniform within each se_a
+    // type-pair invocation, and compressed forward_lower sorts its nlist first.
     TabulateFusionSeAForward<FPTYPE>(table_tensor, table_info_tensor,
                                      em_x_tensor, em_tensor, at::Tensor(),
                                      last_layer_size, true, descriptor_tensor);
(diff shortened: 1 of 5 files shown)


source/tests/pt/test_tabulate_fusion_se_atten.py
```diff
@@ -1645,5 +1645,131 @@ def test_second_order_backward(self) -> None:
...
+            [[[36.0] * 4, [108.0] * 4, [27.0] * 4]],
+            dtype=self.dtype,
+            device=env.DEVICE,
+        )
+        expected_dtwo = torch.stack(
+            (0.2 * coefficients, 0.8 * coefficients, 0.4 * coefficients)
+        )
+
+        torch.testing.assert_close(
+            unsorted_grads[0], torch.zeros_like(unsorted_grads[0])
```
wanghan-iapcm: Small point on this assertion: it cannot fail either way.

The fixture sets only the `a0` coefficient (`table[0, :, 0]` and `table[1, :, 0]`), so `polynomial5_grad` has no non-zero term and the `em_x` gradient is identically zero regardless of `is_sorted`. This line therefore passes on the pre-fix code too, unlike the `dy_dem` and `dy_dtwo` assertions just below it, which are genuinely discriminating.

Giving the table a non-zero `a1` would make this output carry information, and would have a second benefit: the same degeneracy zeroes the `dz_xx * var_grad` term in the grad-grad kernel, so that term is currently unexercised by the second-order test as well. The rest of the expectations would need recomputing, so it is only worth doing if you think the extra coverage is worth the churn -- the test already fails pre-fix on the other assertions, so this is about depth rather than about whether the regression is caught.

source/op/pt/tabulate_multi_device.cc
```diff
@@ -643,7 +647,7 @@ class TabulateFusionSeAGradOp
     // compute
     TabulateFusionSeAGradForward<FPTYPE>(
         table_tensor, table_info_tensor, em_x_tensor, em_tensor, at::Tensor(),
-        dy_tensor, descriptor_tensor, dy_dem_x_tensor, dy_dem_tensor,
+        dy_tensor, descriptor_tensor, true, dy_dem_x_tensor, dy_dem_tensor,
```
wanghan-iapcm, line 654: Not a request, just a note for the next person in this file: these `true` literals are load-bearing, and nothing here says so.

They are correct today for two independent reasons -- se_a applies exclusion per type-pair block, so an excluded block is uniformly zero within a single op invocation, and `DescrptBlockSeA.need_sorted_nlist_for_lower()` returns `self.compress` so the compressed path never sees an unsorted list. The second of those is a Python-side guarantee documented only in `deepmd/pt/model/descriptor/se_a.py`, with no counterpart anywhere in this file.

So someone editing this file has no local signal that changing these constants, or that a change on the Python side, would break the invariant. A one-line comment pointing at `need_sorted_nlist_for_lower` would carry the coupling across the language boundary.

wanghan-iapcm (commented):
The plumbing is right. I walked all four autograd classes and every library entry point at head: `is_sorted` now reaches forward, first backward and grad-grad with the user's value in each; `ctx->saved_data["is_sorted"]` has exactly two stores and two matching reads, so no backward can read a key its forward never wrote; and since `TORCH_LIBRARY_FRAGMENT` registers a function pointer, the schema is inferred from the signature and cannot drift out of sync. The public signature is unchanged, so the TorchScript stub and the pt_expt fake kernel stay valid.

Worth stating explicitly what the pre-fix state was, because it is a bit worse than "a flag was ignored": the grad-grad helper already forwarded `is_sorted` correctly, while forward and first-backward did not. So the three autograd stages disagreed with each other -- the op was not consistently wrong, it was internally incoherent, which is why the second-order finite-difference check in the new test is such a sharp discriminator.

The hard-coded `true` on the non-attention path is correct rather than merely unchanged. `is_sorted` defaults to `true` in `source/lib/include/tabulate.h`, so behaviour is preserved exactly; and se_a applies exclusion per type-pair block with a separate op call per embedding index, so an excluded block is uniformly zero within one invocation and never interleaves. TF was already correct (it declares `is_sorted` as an op attr and reads it in every kernel), Paddle has no tabulate op, and pt_expt calls this same op so it is fixed transitively.

**On why no existing test caught this.** This is the part I found genuinely interesting, because there *is* already an end-to-end test that trains, freezes and compresses a se_atten model with `exclude_types` and compares against the uncompressed model at 1e-10 -- `TestDeepPotATPBCExcludeTypes` in `source/tests/pt/test_model_compression_se_atten.py`. It passes pre-fix, and not because of tolerance or because the op is bypassed. It passes because in that model the fold is never armed.

`forward` builds the env-mat from the unmasked nlist and applies the exclusion mask afterwards, so an excluded row is exactly all-zero including its `em_x`. Padding rows are not zero: `prod_env_mat` normalises without re-masking, so a padding row becomes `(-avg_r/std_r, 0, 0, 0)`. The kernel's guard needs `ago == xx`, where `ago` is the last neighbour's scalar -- and `se_atten_v2` defaults `set_davg_zero=False`, so that sentinel is non-zero while excluded rows are zero. The guard never matches an excluded row and fires only on the genuine trailing padding block, where the `(nnei - jj)` fold is exact. The nlist ordering is actually maximally adversarial there (the excluded O-H pairs are the nearest neighbours, sitting ahead of real ones), and it still cannot trigger.

So the untested cell is compressed se_atten, non-empty `exclude_types`, **and** an env-mat where the excluded row's `em_x` equals the padding sentinel -- which is what `set_davg_zero=True` gives you, and that is the default for plain `se_atten` rather than `se_atten_v2`. Nothing in the repo crosses it.

Your op-level test does construct exactly that configuration and is a real regression test: hand-derived closed-form expectations rather than the op compared against itself, all three autograd stages, both dtypes, and it runs in CI since the custom ops are built there. Pre-fix the forward returns `0.6*c` where you assert `1.4*c`, and `dy_dem` comes out `[108, 0, 0]` against your `[36, 108, 27]` -- not a tolerance question.

What would round it out is one end-to-end case crossing the cell above: a compressed se_atten (not v2) with `exclude_types` and `set_davg_zero: True`, asserted against the uncompressed model the way `TestDeepPotATPBCExcludeTypes` already does. That is the configuration a user would actually hit, and it is the one the current suite structurally cannot reach.

**The one thing I would like your view on before approving.** This PR makes the `is_sorted=True` fold genuinely live for se_atten for the first time, and the Python-side guarantee that keeps it safe was never added for that descriptor family:

https://github.com/deepmodeling/deepmd-kit/blob/dba50c5a43a26d9697b4e186ba353fa106d6817c/deepmd/pt/model/descriptor/se_atten.py#L813-L817

Compare the se_e2_a version, where the method returns `self.compress` specifically so that the sorted-nlist precondition holds, with a comment explaining the coupling:

https://github.com/deepmodeling/deepmd-kit/blob/dba50c5a43a26d9697b4e186ba353fa106d6817c/deepmd/pt/model/descriptor/se_a.py#L863-L879

The exposure is a compressed DPA1 with *empty* `exclude_types`, which gives `is_sorted = True`, so the fold applies -- but `need_sorted_nlist_for_lower()` returning `False` means the `forward_lower` neighbour list from C++/LAMMPS, built at rcut+skin and not pre-sorted, can place an out-of-rcut neighbour with `sw == 0` ahead of real ones. The guard then fires early and multiplies that row by `(nnei - jj)`, silently producing a wrong descriptor. That is the same failure mode fixed for se_e2_a. `se_t.py` and `se_t_tebd.py` have the identical unconditional `return False`.

This hazard predates your PR -- before it the flag was ignored and the kernel behaved as sorted regardless, so the LAMMPS exposure was already there. I am not claiming you introduced it. But it is the other half of the same invariant, and this PR is what makes the flag meaningful, so I would rather we decide deliberately whether it belongs here or in an immediate follow-up than leave it implicit. `return self.geo_compress` would mirror the se_a fix.

Two smaller notes inline. Also worth a line in the description: with non-empty `exclude_types` the fold is now genuinely disabled, so compressed DPA1 pays a full `nnei` loop where it previously short-circuited. Correct, but a measurable cost that users may notice.

njzjz:
Addressed in 8a3a02381:

- `DescrptBlockSeAtten.need_sorted_nlist_for_lower()` now returns `self.geo_compress`, so only the geometric tabulation path requests the extra lower-nlist filtering/sort. Type-embedding-only compression remains unchanged.
- Added a functional compressed-DPA1 `forward_lower` regression with an unsorted rcut+skin list containing both in-cutoff and out-of-cutoff neighbors.
- Changed the existing exclude-types end-to-end case to plain `se_atten` with `set_davg_zero: true` and `tebd_input_mode: strip`, so it crosses the previously untested cell.
- Made the operator fixture's `a1` coefficients nonzero, so the `em_x` gradient and the grad-grad `dz_xx * var_grad` term are exercised.
- Documented the two invariants supporting the non-attention `se_a` hard-coded `true` values.

I also audited `se_t.py` and `se_t_tebd.py`. Their compressed path uses `tabulate_fusion_se_t`, which does not have the `is_sorted` early-fold contract, so they do not need the analogous `geo_compress` override.

History-wise, there are two distinct origins:

- The bug fixed by #5890 was introduced by #3877, which added the PyTorch tabulation op but omitted `is_sorted` from the backend calls in forward and first backward.
- The lower-nlist exposure discussed here became reachable in #4227, which added PyTorch `se_atten` geometric compression while `need_sorted_nlist_for_lower()` still returned `False`. #3993 introduced that method/default earlier, but there was no compressed `se_atten` fold at that point.

Validated with the rebuilt PyTorch C++ op, all 12 SE-Attention tabulation tests, the new `forward_lower` regression, and the targeted end-to-end compression comparison.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

source/tests/pt/test_tabulate_fusion_se_atten.py
njzjz: Done in 8a3a02381. The fixture now gives both table segments nonzero `a1` coefficients and updates the closed-form forward/backward expectations. The `em_x` gradient is consequently nonzero, and the second-order finite-difference test now exercises the `dz_xx * var_grad` contribution as well. The full tabulation test file passes for both float64 and float32.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

source/op/pt/tabulate_multi_device.cc
njzjz: Done in 8a3a02381. I added local comments at both non-attention forward and first-backward `true` call sites, documenting the per-type-pair exclusion invariant and the Python-side `DescrptBlockSeA.need_sorted_nlist_for_lower()` coupling.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

wanghan-iapcm (approved):
All five points from the previous round check out against `8a3a02381`.

`DescrptBlockSeAtten.need_sorted_nlist_for_lower()` now returns `self.geo_compress`, which is the right predicate -- narrower than se_a's `self.compress`, since only the geometric path reaches the se_a tabulate op, and `geo_compress` is initialized in `__init__` so uncompressed models are unaffected. The new `forward_lower` regression is a genuine one: it fails pre-fix on the `need_sorted_nlist_for_lower()` assertion alone, and lines 114-115 verify the constructed list really does contain both in-cutoff and out-of-cutoff real neighbors rather than assuming it.

On the operator fixture, I re-derived the expectations after the `a1` change: every one of the nine previously-verified numbers scales by exactly 1.5, and the `em_x` assertion is now discriminating instead of vacuous. The reconfigured end-to-end case crosses the cell I described -- with `set_davg_zero: true` the padding rows normalize to all-zero and so match excluded rows exactly, which is what arms the guard; pre-fix that case folds an excluded row with real neighbors behind it and misses the 1e-10 comparison.

One correction on my side: my note about `se_t.py` and `se_t_tebd.py` was over-broad and your rebuttal is right. The `ago == xx && ... && is_sorted` fold and its `break` exist only in the three `se_a` kernels, on CPU (`tabulate.cc` 186/275/379) and GPU (`tabulate.cu` 329/414/518); `se_t`, `se_t_tebd` and `se_r` take no `is_sorted` parameter and have no fold, so an override there would have no effect. Nothing to change.

Thanks also for the performance note in the description.

PR #5908, opened 2026-07-26
fix(pair-tab): validate uniform distance grid to avoid silently wrong potentials

hanaol:
## Summary

`PairTab` (used for `use_srtab` and ZBL `tab_file` short-range tables) assumes the distance grid is uniformly spaced, but never validates it. A non-uniform table is silently accepted and produces incorrect potentials with no error or warning.

## Problem

`PairTab.reinit()` infers a single stride from the first two rows only:

```python
self.hh = self.vdata[1][0] - self.vdata[0][0]
```
`self.hh` is then baked into the spline coefficients and stored in tab_info, and the C++ inference kernel (source/lib/src/pair_tab.cc) maps distance to table index purely as `idx = floor((r - rmin) / hh)` . No per-row distances are kept, so a non-uniform grid is impossible to represent.

Also, the uniform-grid requirement is documented (doc/model/pairtab.md: the table is defined "on an evenly discretized grid"), but nothing enforced it.

## Fix
Validate in `reinit()` that every distance interval matches the inferred stride, and raise a `ValueError` otherwise. The check runs before any spline is built, so a bad table is rejected up front instead of mis-indexed later.

## Tests
Added `TestPairTabGridSpacing` in
source/tests/common/dpmodel/test_pairtab_preprocess.py:

test_non_uniform_grid — a table with a changing stride now raises ValueError.
test_uniform_grid — a uniformly spaced table is still accepted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pairwise tables now validate the loaded distance grid before updating internal state.
  * Non-uniform, duplicate, or descending grids now fail fast with a clear `ValueError` advising to regrid.
  * Evenly spaced uniform grids still initialize with the expected spacing.
* **Tests**
  * Added unit coverage for non-uniform, duplicate, descending, and uniform distance grids (including small-grid edge cases).
  * Added verification that a failed reinitialization preserves the previously serialized state.


hanaol:
Thanks for the review — both findings addressed in f87eeb78, with two intentional deviations from the suggested diffs:

1. **Strictly increasing grid.** Implemented as a separate check with its own error message rather than folding `np.all(dx > 0)` into the existing `allclose` condition. A descending grid with a constant negative step *is* evenly spaced, so reusing the "not evenly spaced" text would have misdescribed the failure. Duplicate rows and descending tables now raise "not strictly increasing"; genuinely uneven grids keep the original message.

2. **Atomic `reinit()`.** Validation now runs against locals and instance state is committed only once every check passes. I also moved the `ntypes` assert inside the atomic region — otherwise a table with a bad column count would still have left the object half-updated.

Regression tests added for the duplicate-distance and descending-grid cases, plus one asserting that a failed `reinit()` on a live `PairTab` leaves the serialized table byte-for-byte unchanged.

deepmd/utils/pair_tab.py
```diff
@@ -57,17 +57,39 @@ def reinit(self, filename: str, rcut: float | None = None) -> None:
...
+        hh = vdata[1][0] - vdata[0][0]
+        dx = np.diff(vdata[:, 0])
+        if not np.all(dx > 0):
+            raise ValueError(
+                f"The distance grid in the pairwise table {filename} is not "
+                "strictly increasing. The tabulated potential must be provided "
+                "on a uniform grid with distances sorted in ascending order and "
+                "without duplicated rows. Please regrid the table."
+            )
+        if not np.allclose(dx, hh, rtol=1e-5, atol=1e-8):
```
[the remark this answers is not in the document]
hanaol: confirmed and fixed in 040569ed.

deepmd/utils/pair_tab.py
```diff
@@ -57,17 +57,39 @@ def reinit(self, filename: str, rcut: float | None = None) -> None:
...
+        hh = vdata[1][0] - vdata[0][0]
+        dx = np.diff(vdata[:, 0])
+        if not np.all(dx > 0):
+            raise ValueError(
+                f"The distance grid in the pairwise table {filename} is not "
+                "strictly increasing. The tabulated potential must be provided "
+                "on a uniform grid with distances sorted in ascending order and "
+                "without duplicated rows. Please regrid the table."
+            )
+        if not np.allclose(dx, hh, rtol=1e-5, atol=0):
```
wanghan-iapcm: This is the one I would like changed before merge.

The check bounds each **interval** (`dx[i]` against `hh`), but every consumer indexes by **absolute node position**: `pair_tab.cc` computes `idx = floor((r - rmin)/hh)`, and `_make_data()` builds the spline assuming node `i` sits at `rmin + i*hh`. Those are different quantities, and the gap cuts both ways.

**It rejects tables that are correct today.** Text tables have finite printed precision, so each `r_i` is rounded. That leaves the node positions essentially exact while making each `dx_i` deviate by up to one ulp of the print format -- which is a large *relative* deviation when `hh` is small. Concretely, for `np.linspace(0.0, 6.0, 1000)` (true stride 0.006006006...) written at several precisions and read back with `np.loadtxt`:

| print fmt | inferred `hh` | max per-interval rel. dev. | this check | max node offset |
|---|---|---|---|---|
| `%.8f` | 0.00600601 | 1.7e-06 | passes | 0.0007 cells |
| `%.6f` | 0.006006 | 1.7e-04 | **rejects** | **0.0010 cells** |
| `%.5f` | 0.00601 | 1.7e-03 | rejects | 0.66 cells |
| `%.4f` | 0.006 | 1.7e-02 | rejects | 1.00 cells |

(node offset = `max |r_i - (rmin + i*hh)|`, in units of `hh`.)

The `%.6f` row is a table whose nodes are within one-thousandth of a cell of where the index arithmetic puts them -- evaluated correctly today by both kernels -- and it now raises `ValueError` telling the user to regrid. The check cannot tell it apart from the `%.4f` row, which is off by a full cell and genuinely broken. The affected band is wide rather than exotic: rejection needs `10^-p / hh > 1e-5` while harmlessness needs `0.5 * 10^-p << hh`, so at 6-decimal printing any stride from roughly 5e-7 to 0.1 is caught whenever the stride is not exactly representable at that precision -- which is the normal case for anything built as `linspace(0, rcut, N)`. No in-tree file trips it only because they all use round strides like 0.001.

**It also accepts tables it should not.** `N` intervals each within `rtol` may all lean the same way, so the last node can drift by up to `N * 1e-5 * hh`. For the shipped 6000-row table that is about 6% of a bin while still passing.

Both follow from measuring intervals instead of positions, so checking the position directly fixes both:

```python
n = vdata.shape[0]
expected = rmin + hh * np.arange(n)
if not np.allclose(vdata[:, 0], expected, rtol=0, atol=1e-2 * abs(hh)):
    raise ValueError(...)
```

An `atol` of a small fraction of `hh` still rejects the 0.66-cell and 1.0-cell cases above and every case in your test suite, while tolerating printed-precision rounding. I would avoid making it much tighter than ~1e-2 -- at 1e-3 the `%.6f` example lands almost exactly on the boundary.

Worth noting this is the second pass at this tolerance: njzjz-bot correctly pointed out that the earlier `atol=1e-8` accepted a non-uniform fine grid, and dropping to `atol=0` is what introduced the strictness above. That is a sign the constant is not the thing to keep tuning -- switching the measured quantity resolves njzjz-bot's case too, since `[0, 1e-10, 1.1e-9, 2.1e-9, 3.1e-9]` has node offsets of order a full cell and fails a position check comfortably.

Two related smaller points on the same lines, both cheap: `hh` comes from `vdata[1][0] - vdata[0][0]`, so if the *first* interval is the anomalous one, every other row gets reported as the offender against a reference stride that is itself wrong -- deriving the reference as `(rmax - rmin) / (n - 1)` avoids that. (Careful: if you also feed that value into `self.hh` it changes `tab_info` and therefore results for existing models, so I would keep that separate from the validation change unless you intend it.) And the message would be much more actionable if it named the first bad row, e.g. via `int(np.argmax(np.abs(vdata[:, 0] - expected) > tol))`.

Finally, on coverage: all seven new tests patch `numpy.loadtxt` with in-memory float64 arrays, so none of them goes through the text-parsing round trip where this actually bites. A test that writes a real file with a realistic format (`np.savetxt(f, tab, fmt="%.6f")`) and asserts it is *accepted* would pin this down.

wanghan-iapcm (commented):
The diagnosis here is exactly right and I want to say so first. `hh` is inferred from the first two rows, baked into the spline coefficients and `tab_info`, and then both evaluators index purely on it -- [`source/lib/src/pair_tab.cc`](https://github.com/deepmodeling/deepmd-kit/blob/b00aeb305f01714a9b614a57c187e705410f1bf4/source/lib/src/pair_tab.cc#L18-L32) does `uu = (rr - rmin) * hi; int idx = uu;` with no validation whatsoever -- so a non-uniform table cannot be represented and was being silently mis-indexed. Enforcing what [`doc/model/pairtab.md`](https://github.com/deepmodeling/deepmd-kit/blob/b00aeb305f01714a9b614a57c187e705410f1bf4/doc/model/pairtab.md) has always documented ("on an evenly discretized grid") is the right call, and `reinit()` is the right place: it is the only path that reads a file, and the checks run before `_check_table_upper_boundary()`, so they only ever inspect the user's raw table and never the padding rows the code appends itself. The "compute into locals, commit afterwards" refactor is a good instinct too.

I checked the regression coverage by running your new tests against unfixed code rather than reasoning about it: 5 of the 7 fail pre-fix, and the only two that pass are the intended accept-cases; all 11 pass after. `test_duplicate_distances` is a nice one -- pre-fix it does not raise at all, it just emits `RuntimeWarning: divide by zero encountered in scalar divide` from `rcut_idx = int(np.ceil(self.rcut / self.hh - ...))` with `hh == 0` and carries on, which is the silent corruption this PR is about. I also confirmed nothing in-tree breaks: `examples/water/zbl/H2O_tab_potential.txt` and the `zbl_tab_potential` fixtures sit ~1e-13 relative against a 1e-5 tolerance, `examples/water/d3/dftd3.txt` ~1e-14, and every mocked grid in the existing tests is uniform to ~1e-16. And since [`deepmd/tf/utils/pair_tab.py`](https://github.com/deepmodeling/deepmd-kit/blob/b00aeb305f01714a9b614a57c187e705410f1bf4/deepmd/tf/utils/pair_tab.py) is a pure alias re-export, the fix correctly lands in the one shared reader that tf, tf2, pt, pt_expt, dpmodel and jax all go through.

One thing I would like resolved before this merges, inline below: the uniformity check measures a different quantity than the one the consumers actually depend on, and that makes it both too strict and too lax in ways that are easy to demonstrate.

A few smaller notes I am recording rather than asking you to act on. The `Table file` section of `doc/model/pairtab.md` -- the part a user reads when authoring a table -- does not state the even-grid requirement anywhere (only the Theory footnote does), so someone who hits the new `ValueError` has nothing to look up; a sentence there would pay for itself. `deserialize()` assigns `vdata`/`hh`/`tab_info`/`tab_data` directly and never routes through `reinit()`, so a model serialized from a bad table before this fix still loads silently -- I think ingest-time is the right gate and the serialized `vdata` is post-extrapolation anyway, so this is a note, not a request. The atomicity guarantee is narrower than `test_failed_reinit_keeps_state` suggests: after `self.vdata = vdata` the method still runs `_check_table_upper_boundary()` and `_make_data()`, and a `CubicSpline` failure there would leave new `vdata` beside stale `tab_data`. And `reinit()`'s docstring gained two raise paths but no numpydoc `Raises` section (there is one to copy in `deepmd/utils/version.py`).

hanaol:
@wanghan-iapcm @njzjz-bot Follow-up commits address the node-position issue:

- `040569ed` fix(pair-tab): use a scale-aware tolerance for the grid spacing check
- `2aa67040` fix(pair-tab): validate grid by absolute node position, not interval
- `3527afe7` fix(pair-tab): add explicit dtype to np.arange in grid validation

Summary of what changed in `deepmd/utils/pair_tab.py`:

- Validation now checks each row's distance against `rmin + hh_ref * i` (with `hh_ref = (rmax - rmin) / (n - 1)`) rather than bounding each `dx[i]` against `hh`, so it measures the quantity the consumers (`pair_tab.cc`'s index arithmetic, `_make_data()`'s spline) actually rely on.
- `hh_ref` is used only for validation and kept separate from `self.hh` (still `vdata[1][0] - vdata[0][0]`), so `tab_info` and existing model behavior are unchanged.
- Tolerance is `1e-2 * abs(hh_ref)`, and the error message now names the first offending row.
- This also resolves the earlier `atol=1e-8` false-acceptance case (`[0, 1e-10, 1.1e-9, 2.1e-9, 3.1e-9]`), still covered by `test_non_uniform_fine_grid`.

Added `test_uniform_grid_rounded_text_precision`, which writes a real text file via `np.savetxt(..., fmt="%.6f")` and reads it back with `np.loadtxt`, so the printed-precision rounding case is now exercised through the actual text-parsing path rather than mocked arrays.

All 12 tests in `test_pairtab_preprocess.py` pass. Would appreciate a re-review when you get a chance.

deepmd/utils/pair_tab.py
```diff
@@ -57,17 +57,50 @@ def reinit(self, filename: str, rcut: float | None = None) -> None:
...
-            f"number of volumes provided in {filename} does not match guessed number of types {self.ntypes}"
+        ntypes = int(n0 + 0.1)
+        assert ntypes * (ntypes + 1) // 2 == ncol, (
+            f"number of volumes provided in {filename} does not match guessed number of types {ntypes}"
         )
 
+        self.vdata = vdata
+        self.rmin = rmin
+        self.rmax = rmax
+        self.hh = hh
```
wanghan-iapcm, line 100: This is the line that keeps the fix from closing the bug it targets, and the reason it is written this way is my comment last round, which I am withdrawing.

The check above proves every node sits within one percent of a cell of `rmin + i * hh_ref`. But what gets stored, and what every consumer then uses, is the first interval: `tab_info[1]` feeds `uu = (rr - rmin) / hh` in `pair_tab.cc` and the dpmodel/pt `_pair_tabulated_inter`, `_make_data` scales derivatives by it, `rcut_idx` and the padding `linspace` in `_check_table_upper_boundary` are computed from it. Nothing bounds how far this value may sit from `hh_ref`; the check only bounds node positions, so the first interval can be off by nearly a full percent while the table passes. A relative error in `hh` accumulates linearly in the index.

I reproduced it against this head: `linspace(0, 1, 1001)` with only row 1 moved by 9.9e-6, i.e. 0.99 of a cell, is accepted, `self.hh` comes out as 0.0010099, and at `r = 0.995` the consumer's index arithmetic lands on cell 985 while the node is 995. Ten cells off, silently, which is precisely the failure mode in the PR title. The neighbouring comment says the consumers' `rmin + i * hh` is "what must stay accurate", and it is, but it is `hh_ref` that was made accurate, not `hh`.

The fix is one line: `self.hh = hh_ref`, and drop the first-interval `hh`. I asked last round to keep them separate to avoid changing `tab_info` for existing models; having worked through it, that caution was misplaced. For a table with a round stride the two values are bit-identical, and for a printed table `hh_ref` averages out the rounding of row 1 and is the better estimate, so the change to existing models is at the printing-precision level and in the right direction. The error message already reports `hh_ref` as the step, so this also makes what the object stores agree with what it tells the user.

A test that would have caught it: a uniform grid with a single node perturbed inside the tolerance, asserting `tab.hh` against the true step with `rtol` rather than `assertAlmostEqual`.

wanghan-iapcm (changes requested):
The position-based check is the right one, and the regression test proves it: I ran the whole test file against b00aeb30 and against this head, and `test_uniform_grid_rounded_text_precision` fails on the old interval check with exactly the `%.6f` case from my earlier table, then passes here with the other eleven, including the fine-grid rejection from the bot review.

One thing still stands between the check and the PR's stated goal, and it is a consequence of advice I gave last round, so I am retracting that advice rather than just pointing at the gap. Inline below.

Two smaller things, neither blocking:

- `test_uniform_fine_grid` asserts `assertAlmostEqual(tab.hh, 1e-9)` with the default `places=7`, which passes for any `hh` below 5e-8, zero included. The rounded-text test compares to `places=6` on a value of 0.006. Neither would notice `self.hh` being wrong in the way described inline. `np.testing.assert_allclose(tab.hh, expected, rtol=1e-6)` in both would make them real, and once `self.hh` is `hh_ref` the rounded-text one can assert against `6.0 / 999` exactly.
- `doc/model/pairtab.md` still only says the file must be readable by `numpy.loadtxt`. This PR turns a silent assumption into a hard `ValueError` at model construction, so the "Table file" section should say that the first column has to be strictly increasing and uniform, and mention the one-percent tolerance, so a user who hits the error can see what is expected without reading the source.

For the record, things checked and fine: every in-tree table (`examples/water/zbl`, `examples/water/d3`, the two test fixtures) passes the new check with deviations around 1e-12 of a cell, so nothing shipped is rejected; `deserialize` bypasses `reinit`, so already-serialized models are untouched; and `n == 2` is handled.

njzjz (approved):

wanghan-iapcm (approved):
The remaining point from the last round is fixed: `reinit` now derives the step from the first and last node, validates every node against it, and stores that same value in `self.hh`, so the C++ kernel, `_make_data` and the padding all index with the quantity that was actually checked. I re-ran the test file against the previous revision (the new node-perturbation test fails there, as it should) and against this head (13 passed). The tolerance assertions now use `assert_allclose`, the documentation states the grid requirement and where the step comes from, and CI ran through on this head. Approving.

PR #5910, opened 2026-07-28
fix(pt): calibrate DPA4 bias and support PyTorch 2.13 compile

OutisLi:
## Summary

- Calibrate DPA4/SeZM output bias statistics through the complete atomic-model forward path, including model-level preprocessing and analytical contributions.
- Apply the same frame masks and intensive reductions used by the public model outputs when computing statistics, including `min_pair_dist` filtering and multitask data handling.
- Extend the PyTorch `torch.compile` compatibility range to PyTorch 2.13 and apply the Inductor symbolic-divisibility workaround to the affected releases.
- Add and update coverage for SeZM bias calibration, spin-model behavior, training statistics, and compile compatibility.

## Root cause

Bias calibration previously used an incomplete output path, so model-level contributions and output reduction semantics could be omitted from the fitted bias. Separately, PyTorch 2.13 was rejected before compilation, even though it requires the same Inductor compatibility handling as PyTorch 2.12.

## Validation

- `git diff --check` passed.
- Tests were not run locally, per request; the cherry-picked commits include the relevant test updates.

## Summary by CodeRabbit

* **New Features**
  * Added minimum pair-distance frame filtering for training and statistics, with synchronized distributed step skipping when no frames are valid.
  * Added model support for statistics-driven atomic output prediction to power “change-by-statistic” bias adjustment.
  * Introduced an atomic-output-only execution mode for SeZM models, including a safe temporary preservation of training/eval state.

* **Bug Fixes**
  * Fixed edge cases where empty samples/frames could break statistics and reduction logic, including safer intensive/masked reductions.
  * Improved SeZM spin/bridging output handling and migrated analytical bridging potential naming from InterPotential to InnerPotential.


deepmd/pt/utils/compile_compat.py
njzjz, line 1: I suggest allowing the future PyTorch version (e.g. 2.14+). Otherwise, when the new version comes out, we need to update this file again, and the user needs to update this file to use it with the new version.

OutisLi: This is not applicable since no one knows whether future version fixes this bug or not or even bring more bugs. Compile is so tricky.

deepmd/pt/model/model/sezm_model.py
```diff
@@ -3290,7 +3406,7 @@ def tf32_precision_ctx(self) -> Generator[None, None, None]:
 _A_BOHR = 0.5291772109  # Bohr radius in Å
 
 
-class InterPotential(torch.nn.Module):
+class InnerPotential(torch.nn.Module):
```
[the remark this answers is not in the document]
OutisLi: Keeping a compatibility alias is not applicable here. These DPA4/SeZM implementations are still under development and are not exported as stable APIs. We are standardizing the naming instead: the PT, dpmodel, and pt_expt implementations—including the dpmodel/pt_expt module names and the atomic-model registry type—will all use `InnerPotential`. No `InterPotential` alias will be retained.

deepmd/pt/model/atomic_model/base_atomic_model.py
```diff
@@ -596,6 +598,7 @@ def change_out_bias(
         sample_merged: Callable[[], list[dict]] | list[dict],
         stat_file_path: DPPath | None = None,
         bias_adjust_mode: str = "change-by-statistic",
+        model_forward: Callable[..., dict[str, torch.Tensor]] | None = None,
```
wanghan-iapcm: `model_forward` should not be added here, and the reason is visible in what this method does with it: nothing. It defaults it, then hands it straight to `compute_output_stats`. The atomic model never uses the value.

Meanwhile this exact signature is byte-identical in the sibling backends, which this PR does not touch:

https://github.com/deepmodeling/deepmd-kit/blob/17a57b422387f621d42b02bd692b49601c34acf2/deepmd/dpmodel/atomic_model/base_atomic_model.py#L680-L687

https://github.com/deepmodeling/deepmd-kit/blob/17a57b422387f621d42b02bd692b49601c34acf2/deepmd/pd/model/atomic_model/base_atomic_model.py#L489-L496

So pt gains a fourth parameter that dpmodel and pd do not have, on a base-class method the three have kept in lockstep. That divergence is currently latent in the worst way: dpmodel and pd simply never receive a `model_forward`, so they keep calibrating through the incomplete path this PR exists to fix. Whatever shape is chosen should land in all three, or in none.

Two things suggest the argument is avoidable. First, the model-level polymorphism is already correct - `predict_atomic_outputs_for_stat` is declared on the base model, given a generic implementation in `make_model.py`, and overridden by SeZM. That part is right. Second, the generic implementation is definitionally identical to the default this method already computes: it returns `self.atomic_model._get_forward_wrapper_func()(...)`, which is exactly what the `model_forward is None` branch falls back to. So for every model except SeZM the whole round trip - model builds a predictor, passes it down, atomic model passes it back out - reproduces the behaviour that existed before.

The codebase also already shows the idiom for injecting expensive caller-specific behaviour without widening a signature: `sample_merged` is `Callable[[], list[dict]] | list[dict]`, documented as a lazy function used because sampling is slow and memory-hungry.

The real obstacle is layering: `_get_forward_wrapper_func()` lives on the atomic model, while the term SeZM needs (the analytical pair potential) lives above it, so a plain override cannot reach it. Two ways round that keep the signature intact:

- have the model set its complete predictor on the atomic model before calling `change_out_bias`, so this method keeps using `self._get_forward_wrapper_func()` unchanged; or
- let the SeZM atomic model own the analytical contribution (or a handle to it) so its own `_get_forward_wrapper_func()` is already complete.

Either keeps `change_out_bias` at three parameters across all backends and removes the pass-through entirely.

deepmd/pt/model/model/sezm_model.py
```diff
@@ -1538,8 +1570,21 @@ def core_compute(
         with nvtx_range("SeZM/apply_out_stat"):
             fit_ret = self.atomic_model.apply_out_stat(fit_ret, atype)
 
-        # === Step 4. Apply atom mask ===
-        for key in fit_ret.keys():
+        # === Step 4. Inject analytical pair potential (edge form) ===
```
wanghan-iapcm, line 1568: Moving the analytical injection above the atom mask changes the physical energy, not just the statistics, and it leaves the exclusion half-applied.

`InnerPotential.forward` does `atom_energy.index_add_(0, dst, pair_e * 0.5)` - it deposits half of each pair energy on `dst` only, relying on both directed edges `(i,j)` and `(j,i)` being present so each atom accumulates its own half. The atom mask applied at Step 5 includes `self.atomic_model.atom_excl(atype)`.

So for a pair where `i` is atom-excluded and `j` is not: the edge depositing into `j` survives the mask, while the edge depositing into `i` is zeroed. The excluded atom ends up contributing exactly half of that pair's ZBL energy, rather than none.

To be fair, the previous ordering was also wrong in the other direction - masking first and injecting after meant an excluded atom had its fitting energy zeroed and then its full ZBL contribution added straight back. So this is not a regression from a correct state; it turns a full leak into a half leak. But the PR introduces `inter_potential_edge_mask` for exactly this class of problem and filters only `descriptor_model.exclude_types` (pair exclusion) with it, never `atom_excl` - which is where the fix belongs, since correct exclusion has to happen on edges.

This affects the output of any trained SeZM/DPA4 model combining `atom_exclude_types` with a bridging potential. `test_zbl_respects_pair_exclusion` covers `pair_exclude_types`, not `atom_exclude_types`, so nothing currently pins it.

deepmd/pt/train/training.py
```diff
@@ -1458,9 +1481,16 @@ def step(_step_id: int, task_key: str = "Default") -> None:
...
-            # DDP path in get_data() always keeps at least one frame)
-            if not input_dict:
+            # Every rank enters this collective once per optimization step,
+            # independent of the locally selected multi-task branch.
+            if (
+                self.has_min_pair_filter
+                and dist.is_available()
+                and dist.is_initialized()
+            ):
+                if not all_ranks_have_valid_frames(bool(input_dict)):
```
wanghan-iapcm: The collective itself is sound - `has_min_pair_filter` is config-derived so it is identical on every rank, `step()` runs once per iteration on all ranks, and the `MIN` all-reduce is entered unconditionally inside the branch, so no rank can diverge into a later collective. No deadlock, and the docstring rewrite in `argcheck.py` matches what the code does.

The problem is that the skip is completely silent. There is no log line, no warning, and no skipped-step counter. Because the display and logging block lives inside `step()`, a run in which every step is skipped produces no training output whatsoever and then exits normally having saved an untrained model - indistinguishable from success.

Also worth checking: `self.scheduler.step()` sits below this return, while `_step_id` keeps advancing in the outer loop. On a partially-skipped run the learning-rate schedule therefore desynchronises from the number of optimizer steps actually taken.

The behaviour this replaces guaranteed forward progress on every rank, so the new one is strictly more capable of failing quietly. A rate-limited warning plus a final "N of M steps executed" summary - loud if that ratio is zero - would make it diagnosable.

wanghan-iapcm (commented):
The direction of the bias fix is right, and I checked that rather than assuming it: routing the analytical term through the prediction removes a genuine prior double count, where ZBL was absorbed into the fitted bias and then added again at inference. Reasoning through `test_zbl_change_out_bias_is_invariant_for_self_labels`, self-generated labels include ZBL while the old stats path excluded it, so the bias would shift and the invariance assertion should fail pre-fix - that makes it a real regression test. I did not execute it against the parent, so treat that as reasoning rather than a measurement.

What I did measure: the pre-existing shared-path suites still pass against this branch - `test_atomic_model_atomic_stat.py`, `test_atomic_model_global_stat.py` and `test_polar_atomic_model_stat.py` all green, and the three failures in `test_change_bias.py` reproduce identically on the parent commit from an unrelated `libcupti` problem. So "Tests were not run locally, per request" is a process gap here, not a hidden regression. Worth running them before merge all the same, since `make_model.py`, `base_atomic_model.py` and `stat.py` are on the path of every pt model, not just SeZM.

Eight comments inline. The one on `base_atomic_model.py` is about the API shape; the others are individual defects.

One finding cannot be anchored because the file is not in this diff. `deepmd/pt/entrypoints/main.py` calls `make_stat_input(...)` with no `min_pair_dist`:

https://github.com/deepmodeling/deepmd-kit/blob/17a57b422387f621d42b02bd692b49601c34acf2/deepmd/pt/entrypoints/main.py#L620-L626

So `dp train` now calibrates on filtered data while `dp change-bias` calibrates on unfiltered data, and the same dataset plus config yields two different biases. It is structural rather than a one-line oversight: `change_bias()` only receives the frozen model's `model_params`, never the training input config, so there is no `min_pair_dist` value available to thread through. Worth deciding explicitly whether change-bias should grow the option or whether the filter belongs somewhere both entry points can see.

On `atomic_output_only`: it exists because `predict_atomic_outputs_for_stat` calls `forward_common` under `torch.no_grad()`, and `forward_common` computes force and virial through autograd, so the derivative work has to be suppressed. That is the same design choice as the `model_forward` argument, one layer down - both follow from reusing the whole model forward for a job that only needs atomic outputs. If the model instead exposed just its analytical atomic contribution, the atomic model's existing predictor plus that term would give the complete output with neither the flag nor the new argument. The flag also has a side effect worth a second look: it gates the `dens` branch as well (`get_active_mode() == "dens" and not atomic_output_only`), so a dens-mode model computes its statistics through the ener edge schema - the same "incomplete output path" divergence this PR set out to remove. If dens models are not meant to reach bias calibration, raising would be clearer than silently switching schema.

Smaller notes, not worth their own threads:

1. `SUPPORTED_COMPILE_TORCH` and `_DIVISIBILITY_REGRESSION_TORCH` are now independent tuples. Previously the `== (2, 12)` test was bound to the same literal as the gate, so they could not drift; a future `(2, 14)` added to the first alone would compile without the divisibility patch and hit `CantSplit`. Deriving the second as "everything supported except (2, 11)" would be failure-proof.

2. The `model_forward is None` fallback branch in `change_out_bias` - the one every non-SeZM model takes - is covered only by the pre-existing suites noted above, none of which are in this diff.

3. The PR bundles bias calibration with PyTorch 2.13 compile support. The commits are mostly separable (`ba3705b8a` bias, `c9cad8c99` compile), so splitting would make each independently revertable.

4. `test_change_out_bias_is_invariant_for_self_labels` and its ZBL sibling assert at `atol=1e-12`. Worth confirming that tolerance was chosen from observed residuals rather than inherited.

deepmd/pt/utils/stat.py
```diff
@@ -205,16 +292,50 @@ def model_forward_auto_batch_size(*args: Any, **kwargs: Any) -> Any:
...
+    mask: np.ndarray,
+    intensive: bool,
+) -> np.ndarray:
+    """Reduce atomic predictions with the public model reduction semantics."""
+    reduced = np.sum(prediction, axis=1)
+    if intensive:
+        atom_count = np.sum(mask, axis=1)
+        empty_frames = np.flatnonzero(atom_count == 0)
+        if empty_frames.size:
+            raise ValueError(
```
wanghan-iapcm, line 343: This turns a path that previously produced a number into a hard abort, and it is stricter than the model reduction it is meant to mirror.

Before this change the reduction was a plain `np.sum` over the atomic dimension, so a frame whose atoms are all masked - every atom of an `atom_exclude_types` species, or a fully padded frame in a mixed-`natoms` batch - still yielded a finite value. Now any such frame raises and aborts bias calibration for the entire run.

The model's own intensive reduction in `deepmd/pt/model/model/transform_output.py` divides by the mask sum unconditionally and produces `nan` in the same situation rather than raising. Since the docstring presents this path as applying "the same reductions used by the public model outputs", raising here is a stronger contract than the thing being mirrored.

Dropping such frames from the regression would preserve the old behaviour while still avoiding the `0/0`. If raising really is intended, it needs a test - no case in the diff drives `atom_count == 0`.

deepmd/pt/utils/stat.py
```diff
@@ -70,12 +141,16 @@ def make_stat_input(
...
             iterator = iter(dataloaders[i])
             numb_batches = min(nbatches, len(dataloaders[i]))
-            for _ in range(numb_batches):
-                try:
-                    stat_data = next(iterator)
-                except StopIteration:
-                    iterator = iter(dataloaders[i])
-                    stat_data = next(iterator)
+            accepted_batches = 0
+            for stat_data in iterator:
```
wanghan-iapcm: Three separate behaviour changes in this rewritten loop, none mentioned in the PR description.

The `except StopIteration: iterator = iter(...)` re-iteration fallback is gone. A dataloader that yields fewer than `numb_batches` items previously cycled to reach the requested count; it now silently produces smaller statistics instead. That is a quiet change in how many frames the bias is fitted on.

The `if accepted_batches >= numb_batches: break` test runs after `next()` has already produced `stat_data`, so once the target count is reached one extra batch is fetched and discarded per system. Harmless numerically, wasteful on a slow loader.

And the new "Skipping training system %d ... because no frame satisfies min_pair_dist=%s" message fires whenever `sys_stat` ends up empty - including when `min_pair_dist` is `0.0` and the real cause is an empty dataloader. That misattributes the failure to a filter that is switched off.

deepmd/pt/utils/compile_compat.py
```diff
@@ -51,17 +53,42 @@
 ]
 
 
+#: ``(major, minor)`` releases whose ``torch.compile`` stack is exercised by the
+#: SeZM compile-parity tests. Other releases are rejected up front rather than
+#: risking silent numerical drift or a segfault deep inside Inductor.
+SUPPORTED_COMPILE_TORCH = ((2, 11), (2, 12), (2, 13))
```
wanghan-iapcm, line 59: This comment claims coverage that does not exist. Nothing in the diff adds a CI matrix entry or workflow for 2.13, `pyproject.toml` still pins `torch==2.11.0`, and the PR description states tests were not run locally - so no 2.13 compile-parity run exists anywhere. The docstring on `patch_inductor_symbolic_divisibility` separately asserts that "2.13 retains that behaviour" about Inductor's `Mod` handling, with nothing cited behind it.

The change also inverts the safety property the comment describes. The test file previously carried an independent hard-coded `{(2,11),(2,12)}` skip set; it now derives the skip from `SUPPORTED_COMPILE_TORCH`. So adding a release to this tuple auto-enables its parity tests rather than requiring them to have passed first - which is the opposite of "rejected up front rather than risking silent numerical drift or a segfault deep inside Inductor".

CodeRabbit raised the contrary factual claim on this PR: that 2.13 already fixed the divisibility regression and the workaround is therefore unnecessary. The thread is marked resolved, but no code change was made in response - the reply addressed a separate point about forward compatibility. So the disagreement about whether 2.13 needs this patch at all is still open.

Worth noting the risk is not confined to the compile path. `apply_global_compile_patches()` runs at module scope, and `patch_inductor_symbolic_divisibility` guards only its imports - the following `SizeVarAllocator.statically_known_multiple_of` read is unguarded. That line is unchanged by this PR, so it is pre-existing, but this PR is what first exposes it to a release nobody has run.

deepmd/pt/utils/stat.py
```diff
@@ -319,10 +440,18 @@ def compute_output_stats(
     if bias_atom_e is None:
         # only get data once, sampled is a list of dict[str, torch.Tensor]
         sampled = merged() if callable(merged) else merged
+        if not sampled:
```
wanghan-iapcm, line 476: This early return is new and nothing exercises it. The tests added in `test_training.py` cover the upstream case where `make_stat_input` returns `[]` for a system, but no test calls `compute_output_stats` itself with an empty `sampled` list, so neither side of this condition is pinned.

It also deserves more than a warning. Reaching it means the entire calibration dataset was filtered away, and the function then returns `{}, {}` - so `_store_out_stat` is a no-op and training proceeds with the original bias. That is a reasonable choice, but silently keeping the pretrained bias when the user asked for recalibration is the kind of thing worth surfacing at a level someone will notice.

deepmd/pt/model/atomic_model/base_atomic_model.py
OutisLi: Addressed in fb3f3ac9d. The PT atomic change_out_bias signature is back to the shared three-argument contract. Residual calibration now lives once at the model layer and calls the polymorphic predict_atomic_outputs_for_stat method: generic models use the atomic predictor, while SeZM supplies the complete learned-plus-InnerPotential prediction. No predictor is stored on the atomic module.

deepmd/pt/model/model/sezm_model.py
OutisLi: Addressed in fb3f3ac9d. Analytical-potential validity is now built at edge level from the base edge mask, pair exclusions, and atom inclusion on both endpoints. If either endpoint is atom-excluded, both directed contributions are removed, so no half-pair ZBL energy can survive on the included atom.

deepmd/pt/train/training.py
OutisLi: Addressed in fb3f3ac9d. A configured training step now first collects a globally valid batch and only then proceeds to the optimizer. Invalid batches are retried within a rank-0-synchronized one-loader-cycle budget; exhaustion raises explicitly, and completed step IDs, scheduler, EMA, logging, and checkpoints therefore remain aligned with actual optimizer updates. Discarded batch attempts are summarized.

deepmd/pt/utils/stat.py
OutisLi: Thanks for flagging the contract difference. The fail-fast behavior is intentional, so I did not switch to dropping frames. For an intensive target, a frame with zero participating atoms has no defined mean; the public path producing NaN is not a finite reference behavior, while silently removing the frame would change the regression dataset. The helper wording now states that undefined intensive means are rejected.

deepmd/pt/utils/stat.py
OutisLi: Addressed in fb3f3ac9d. The statistics collector now checks its stopping condition before fetching, restores the StopIteration restart behavior, and bounds the valid-batch search to one declared loader cycle when filtering is active. Empty-loader and fully filtered-system messages are now distinct, so min_pair_dist=0 is no longer misreported as filtering.

deepmd/pt/utils/compile_compat.py
OutisLi: Addressed in fb3f3ac9d. The allowlist comment and compile-test comments no longer claim CI coverage, and the private SizeVarAllocator hook is read through a callable getattr guard. PyTorch 2.13 remains in the workaround set intentionally: the author has reproduced the DPA4 CantSplit path on 2.13 and verified that this workaround is required there. Future versions remain explicit opt-ins rather than inheriting the patch automatically.

deepmd/pt/utils/stat.py
OutisLi: Addressed in fb3f3ac9d. An empty calibration sample now raises ValueError instead of returning empty statistics and silently keeping the previous bias. The existing empty-statistics regression expectation was aligned with the fail-fast contract.

wanghan-iapcm (commented):
Went back through all seven threads against HEAD rather than the replies, and they hold up. The atomic `change_out_bias` signature is back to the three-argument contract that is byte-identical across the other backends, with the residual calibration moved to the model layer -- and routing it through `predict_atomic_outputs_for_stat` with `return` rather than falling through to the atomic call is right, since `_store_out_stat(..., add=True)` accumulates and running both would double-count. The `StopIteration` restart is restored with an empty-loader guard, the empty-sample case raises, and the compile allowlist no longer claims CI coverage it does not have.

On the intensive-reduction thread, I accept your reasoning and am not pressing it. A silent NaN bias is worse than an abort, and the frame it rejects -- one with no unmasked atoms -- is pathological rather than routine.

The bug-fix reflection also comes out clean. The reason existing tests missed this is specific: `pt/test_finetune.py::test_finetune_change_out_bias` does cover a ZBL model, but it calls `model.atomic_model.change_out_bias(...)`, entering at the atomic layer and bypassing the model layer where the analytical term is added -- so the incomplete path was invisible to it by construction, and every other bias test uses non-bridged models where the atomic output already is the complete output. `test_change_out_bias_is_invariant_for_self_labels` closes exactly that gap and genuinely fails pre-fix: it labels the sample with the model's own energy, so any term missing from the statistics path shows up directly as a non-zero delta bias. Pinning the side conditions too (no compile, no derivatives, training mode and caches restored, no leaked grads) is a nice touch.

Two follow-ups below, both non-blocking and both about coverage rather than correctness.

deepmd/pt/model/model/sezm_model.py
```diff
@@ -3117,6 +3204,54 @@ def reset_head_for_mode(self, mode: str) -> None:
...
+
+        atom_excl = self.atomic_model.atom_excl
+        if atom_excl is not None:
+            atom_is_present = self.atomic_model.make_atom_mask(atype)
+            safe_atype = torch.where(atom_is_present, atype, 0)
+            atom_is_included = atom_is_present & atom_excl(safe_atype).to(torch.bool)
+            atom_is_included = atom_is_included.reshape(-1)
+            keep = (
+                keep
+                & atom_is_included.index_select(0, src)
```
wanghan-iapcm, line 3249: Thanks -- this is the right shape for the fix. The mask is symmetric where it needs to be: `atom_is_included.index_select(0, src) & atom_is_included.index_select(0, dst)` tests both endpoints with the same per-node predicate, so for an excluded atom `i` both `(i,j)` and `(j,i)` drop, and the `index_add_(0, dst, pair_e * 0.5)` halving in `InnerPotential.forward` can no longer leave half a pair energy on the included atom. The pair-exclusion half is symmetric too, since `PairExcludeMask.type_mask` is explicitly symmetrized.

What is missing is a test on this branch specifically. `atom_exclude_types` appears zero times in both `source/tests/pt/model/test_sezm_model.py` and `test_sezm_spin_model.py`; the test added here, `test_zbl_respects_pair_exclusion`, sets only `pair_exclude_types` and therefore exercises the descriptor-mask line above, not these lines. The only other `atom_exclude_types` + ZBL test in the repo is in `source/tests/consistent/model/test_zbl_ener.py`, which targets `DPZBLModel` -- a separate implementation with no `InnerPotential` and no `_make_inter_potential_edge_mask` -- so it gives no coverage here either.

That leaves the exact code path that resolves the half-pair problem untested, which is the one I would most want pinned, since a one-sided regression here is silent: it produces a plausible energy that is wrong by half a pair term.

Mirroring the test you already wrote should do it: set `atom_exclude_types=[1]` instead of `pair_exclude_types`, load the ZBL model with the plain model's weights, and assert equality at 1e-12. Pre-fix that fails, because atom 0 retains half the 0-1 ZBL pair energy while atom 1 is zeroed by the atom mask.

deepmd/pt/train/training.py
```diff
@@ -763,10 +771,41 @@ def get_lr(lr_params: dict[str, Any]) -> BaseLR:
...
+                local_training_batch_attempts = max(
+                    max(1, len(self.training_dataloader[model_key]))
+                    for model_key in self.model_keys
+                )
+            else:
+                local_training_batch_attempts = max(1, len(self.training_dataloader))
+            # Rank-specific task sampling may select loaders of different
+            # lengths, but validity collectives must use one shared retry count.
+            self._training_batch_attempts = int(
+                self._broadcast_value_from_rank0(local_training_batch_attempts)
```
wanghan-iapcm, line 821: This scalar is the fix from a9b500a21, and it is a real one -- worth saying explicitly, because the divergence it repairs was introduced inside this PR and is easy to lose track of.

fb3f3ac9d stored a per-task dict of budgets and selected `max_attempts` by the rank's locally sampled `task_key`. Task sampling is rank-seeded (`deepmd/pt/entrypoints/main.py` builds `rank_seed = [rank, seed % (2**32)]`, which reaches `dp_random.seed` and then the `dp_random.choice` used for task selection), so ranks running different tasks would have taken different budgets, exited the retry loop a different number of times, and mismatched their `all_ranks_have_valid_frames` all-reduce counts -- a hang, not a wrong number. Collapsing to one broadcast max over all task loaders is the correct fix, and with it the loop-trip count is driven only by the reduced value and by a globally identical `max_attempts`, so exhaustion also raises on every rank at the same point.

Nothing locks that in. No test asserts `_training_batch_attempts` is a single broadcast scalar rather than a per-task dict, nothing covers the multitask retry path, and nothing covers the `RuntimeError` exhaustion branch. `test_invalid_batch_is_retried_without_losing_a_step` is single-rank and single-task -- it does discriminate on the scheduler behaviour (the assertion moved from `last_epoch == 1` to `== 2`), but it is blind to the shape of this value.

I appreciate that a genuine multi-rank regression test is a heavier ask than the harness in that file currently supports, so I am not asking for one. Two cheap unit tests would cover most of the regression risk without any distributed setup: one asserting `_training_batch_attempts` is an `int` equal to the max over task loaders in a multitask config, and one forcing `get_data` to always return `{}` and asserting the `RuntimeError`.

deepmd/pt/model/model/sezm_model.py
OutisLi: Addressed in f09ad2cbb. I extended the existing ZBL exclusion regression into pair- and atom-exclusion subtests. The atom case sets `atom_exclude_types=[1]`, loads the ZBL model from the plain model's weights, and asserts equal energies at `1e-12`, directly pinning that neither directed edge can leave a half-pair contribution. The targeted test passes (`1 passed, 2 subtests`).

deepmd/pt/train/training.py
OutisLi: Addressed in f09ad2cbb. I added one behavior-level multitask regression with deliberately different loader lengths, force every `get_data` call to return an empty batch, and assert that each task consumes the same global maximum number of attempts before raising `RuntimeError`. This covers both the shared scalar budget and the exhaustion path in one test, and it would fail with the previous per-task budget dict. The targeted test passes (`1 passed, 2 subtests`).

njzjz (approved):

PR #5911, opened 2026-07-28
feat(dpa1): add compressible l=2 moments across PT backends

OutisLi:
## Summary

- Add compressible `l=2` moment features across the DPA1 PyTorch backends.
- Extend the DPA1/SeAtten descriptor, Triton and CUDA graph kernels, tabulation paths, and graph energy/force handling for the new moment basis.
- Keep the regular PyTorch and `pt_expt` descriptor paths aligned with the compressed implementation.
- Correct compressed DPA1/SeAtten evaluation with `exclude_types` by honoring unsorted neighbor lists throughout tabulation.
- Update argument validation, documentation, serialization versioning, and backend-specific tests for the new functionality.

## Motivation

The DPA1 compressed path previously lacked the `l=2` moment basis and corresponding PT-backend coverage. This change supplies the missing basis and propagates it through descriptor construction, compression, graph execution, tabulation, and force/energy evaluation so compressed and regular paths can represent the same angular information.

## Validation

- All pre-commit hooks passed.
- The focused pure-Python DPA1/DPModel tests passed (`12 passed`, including 6 subtests).
- The selected compressed CPU custom-op tests passed (`29 passed`, 19 skipped).
- The extended component-basis CPU operator tests passed (`2 passed`, including 6 subtests).
- The handwritten `l=2/3/4` CUDA VJP formulas were checked numerically against PyTorch autograd with maximum absolute error below `9e-15`; CUDA compilation/runtime was not available on the macOS review host.

## Summary by CodeRabbit

* **New Features**
  * Added configurable `lmax` (1–4) for DPA1/SE-attention, expanding the moment basis size (4/9/16/25) and enabling learnable higher-order per-degree degree weighting.
  * Updated CPU/CUDA/Triton fused descriptor paths to use the expanded basis (`basis_dim`) and to thread the per-degree gain through forward/backward computations.
* **Bug Fixes**
  * Improved serialization/deserialization to preserve `lmax` and higher-order degree-gain parameters reliably across reloads.
* **Documentation**
  * Documented `lmax`/basis-dimension behavior and clarified higher-order execution constraints and experimental high-`lmax` routing.
* **Tests**
  * Added/extended unit and parity tests for higher `lmax`, serialization, and coordinate-derivative validation.


deepmd/pt/model/descriptor/se_atten.py
```diff
@@ -82,6 +85,145 @@ def tabulate_fusion_se_atten(
     torch.ops.deepmd.tabulate_fusion_se_atten = tabulate_fusion_se_atten
 
 
+_DEGREE_GAIN_INIT_STD = 0.1
+
+
+def _safe_direction(
```
wanghan-iapcm, line 91: `env_protection` does not reach this function, and for l >= 2 that reopens an r -> 0 force singularity.

Protection is applied to the radial denominator in `_compute_angular_radial` (`distance + protection`), but here `direction = diff / safe_distance` has no epsilon floor. The l=1 term is fine - it uses the classic `diff / (r + protection)^2` form and stays bounded. But the l >= 2 blocks are polynomials in the raw unit-vector components, and `d(direction)/d(diff) = (I - dir (x) dir) / safe_distance` is genuinely `O(1/r)`.

Measured on a two-neighbour system (pt, float64, `env_protection=1e-2`, gain 1), max |grad|:

```
lmax=1   1.30e6   (saturates for r <= 1e-6)
lmax=2   9.65e6   at r=1e-4
         9.68e8   at r=1e-6
         9.70e10  at r=1e-8
```

lmax=1 saturates because protection bounds it; lmax=2 keeps growing because nothing bounds the direction.

This also makes the docstring wrong. It says the function normalizes displacements "with finite first and second derivatives at zero" - true of the radial factor, not of `direction`.

`env_protection` exists for exactly this case: overlapping or near-colliding configurations during training, active learning and relaxation. Those are the runs that will hit it. Since this lives in the shared reference, every backend inherits it. Worth either flooring the direction denominator the same way, or stating plainly in the docstring that `env_protection` does not bound the angular derivative for `lmax > 1`.

deepmd/dpmodel/descriptor/dpa1.py
```diff
@@ -1023,6 +1145,11 @@ def serialize(self) -> dict:
             "trainable": self.trainable,
             "spin": None,
         }
+        if obj.lmax != 1:
+            data["lmax"] = obj.lmax
```
wanghan-iapcm, line 1149: These new keys go into `serialize()` with no version change anywhere in the PR. I grepped the whole 47-file diff: zero `@version` lines added or removed, zero `check_version_compatibility` changes, while `serialize()` gains `lmax`, `@variables/degree_gain_raw`, and - at `DescrptBlockSeAtten.serialize` - an unconditional `"trainable"`.

That is all four steps of the serialization rule missed: no bump, no widened compatibility range, no explicit deserialize fallback, in the commit that adds the fields.

Reading old archives happens to still work, but incidentally rather than by design: the new keys are only emitted when `lmax != 1`, and `lmax` defaults to 1, so `cls(**data)` falls through. Nothing records which archives may contain the new keys. The exposure is the other direction - a new `lmax=2` archive handed to an older deepmd, or to the TF loader, hits `TypeError: unexpected keyword argument 'lmax'` with no version number to explain why.

Separately, `"trainable"` is emitted unconditionally by the dpmodel block while pt's `DescrptBlockSeAtten.serialize` has no `"trainable"` key at all - so the two block formats have diverged in this PR. Whatever version scheme is chosen should land in both, along with se_atten_v2.

source/op/pt/CMakeLists.txt
```diff
@@ -12,6 +12,9 @@ option(
 # only within its own subtree, which does not cover this target.
 if(USE_CUDA_TOOLKIT AND DEEPMD_TORCH_HAS_CUDA)
   find_package(CUDAToolkit REQUIRED)
+  option(DEEPMD_ENABLE_DPA1_HIGH_LMAX
```
wanghan-iapcm, line 15: Defaulting this OFF means the new CUDA kernels are never compiled anywhere that runs. I checked `.github/workflows/` - nothing sets `DEEPMD_ENABLE_DPA1_HIGH_LMAX`, so every wheel, conda build and CI job builds without the lmax 2/3/4 specializations. `gh pr checks` confirms the shape of it: "Build C++ on CUDA" passes, while "Test C++ on CUDA" and "Test Python on CUDA" both skip.

The PR is honest about this - the high-lmax tests carry `unittest.skip("requires a build with DEEPMD_ENABLE_DPA1_HIGH_LMAX=ON")`, and the runtime paths fail loudly with a rebuild message rather than silently. That is the right failure mode.

The concern is what it means for the substantial new kernel code in `dpa1_graph_compress_kernel.cuh`, `dpa1_graph_descriptor.cu` and `dpa1_moment_basis.cuh`: it ships in the tree with no build and no execution in any pipeline. Given #5844 landed two days ago as a CPU/GPU divergence in the same tabulate breakpoint logic this PR re-templates, that is the risk worth pricing.

Worth deciding explicitly: either add one CI job that builds with the flag ON and runs the parity tests, or state in the PR that the high-lmax CUDA path is experimental and unvalidated so the next person does not read its presence as support.

wanghan-iapcm (commented):
The core of this is solid, and I verified that rather than assuming it. dpmodel, pt and pt_expt agree to 8e-16 across lmax 1-4, concat and strip, with and without `exclude_types`, and with randomized non-uniform davg/dstd; compressed matches uncompressed to 1.1e-15; rotation invariance holds to 3.3e-16; and the addition theorem `sum_m Y_lm(u) Y_lm(v) = P_l(u.v)` holds to ~2e-15 for l = 2, 3, 4. `lmax=1` really is bit-identical - the basis builder returns `rr` unchanged, degree weights are never built, the `gr[..., 1:4]` slice is a no-op at basis_dim 4, and `serialize` only emits the new keys when `lmax != 1`. The docs and the new `dpa1_moment_basis.cuh` header comment describe the maths accurately.

Eight comments inline. They fall into two groups worth separating.

Findings 1, 4, 5 and 6 touch users who never turn this feature on: an r->0 force singularity that `env_protection` no longer covers, a silent numerical change to already-frozen compressed models that use `exclude_types`, a per-call `cudaGetDeviceProperties` on the existing backward path, and a rewrite of the CPU tabulate hot loop that every SE_A/SE_ATTEN run executes. Those are the ones I would look at first.

Findings 3, 7 and 8 are one problem seen three ways: the new CUDA code is not built in CI, not compared across backends for lmax > 1, and its refactor is unverified even for the shipped lmax=1 path. That matters because #5844 (`fix(cuda): share se_a padding breakpoint`) landed two days before this PR's head and was a CPU/GPU divergence in the very `is_sorted` breakpoint logic this PR re-templates. The parity risk here is demonstrated, not hypothetical.

One finding cannot be anchored because the file is not in the diff. `source/tests/consistent/descriptor/test_dpa1.py` is untouched, so nothing in CI compares dpmodel against pt against pt_expt for lmax 2, 3 or 4:

https://github.com/deepmodeling/deepmd-kit/blob/17309f45bb1e61d2bc82fdb599ff271e3204c523/source/tests/consistent/descriptor/test_dpa1.py#L1-L20

Parity does hold today - I measured it - so this is regression protection rather than a live bug. But it is the axis this change is riskiest on, since the basis is now duplicated across three Python backends plus CUDA and Triton, and that suite exists precisely to catch such drift.

Two things I want to record as checked and clear, so they are not re-raised later. The `share_params` change in `deepmd/pt/model/descriptor/descriptor.py` is a no-op for every pre-existing descriptor block - `compress_data`/`compress_info` are `ParameterList`s and live in `_modules`, not `_parameters` - so it affects only the new `adam_degree_gain_raw`, and it closes a real gap where a trainable parameter silently failed to share across multi-task branches. And the `lmax=2` serialization round-trip test is not insensitive: corrupting or zeroing `degree_gain_raw` on the restored object shifts the output by ~1e-3 to 2e-4 relative, well above the default `rtol=1e-7`, so it does guard what it claims to.

Smaller notes, not worth their own threads:

1. `source/op/pt/CMakeLists.txt` applies `--use_fast_math` to `dpa4c_graph_compress.cu`, which does not exist anywhere in the tree. CMake silently drops the entry. Harmless today because the post-split `dpa1_graph_compress.cu` is host-only, but `--use_fast_math` changes floating-point semantics, so which translation units carry it is load-bearing for parity and this line now misleads.

2. `source/lib/src/gpu/tabulate.cu` validates `ndescrpt` with `assert()`, which is compiled out under `NDEBUG`; an unsupported value then falls into the trailing `else` (MTILE=25) and reads out of bounds. The CPU path throws via `check_se_a_basis_dimension` and the PT wrapper uses `TORCH_CHECK`. `source/lib` is reached from TF and the C++ API too, so the two entry points should behave alike.

3. `source/op/pt/dpa1_graph_common.cuh` fills `basis[4..8]` and gates its angular gradient term on `BASIS_DIM == 9`, but `dpa1_graph_descriptor.cu` carries `BASIS_DIM == 16 / == 25` sizing expressions. Safe today only because of `TORCH_CHECK(basis_dim == 4 || basis_dim == 9)`; a `static_assert(BASIS_DIM <= 9)` would make a future widening fail loudly instead of reading uninitialised registers.

4. The validation figures in `d4e0db9b0`'s commit message (106 passed, PT vs dpmodel <= 4.26e-14, compressed CUDA <= 2.25e-7) describe that commit. `5b61c1276` (the ~1370-line kernel split) and `17309f45b` landed after it and carry no validation of their own, so those numbers do not describe the head.

source/op/pt/tabulate_multi_device.cc
```diff
@@ -56,19 +63,23 @@ void TabulateFusionSeAForward(const torch::Tensor& table_tensor,
...
   const int64_t nnei = em_tensor.size(1);
+  const int64_t ndescrpt = em_tensor.size(2);
+  CheckTabulateFusionSeABasisDimension(ndescrpt);
   // compute
   if (device == "GPU") {
 #if GOOGLE_CUDA || TENSORFLOW_USE_ROCM
     deepmd::tabulate_fusion_se_a_gpu(descriptor, table, table_info, em_x, em,
-                                     two_embed, nloc, nnei, last_layer_size);
+                                     two_embed, nloc, nnei, last_layer_size,
+                                     is_sorted, ndescrpt);
```
wanghan-iapcm, line 108: This changes results for models that already exist, and it is not mentioned anywhere in the PR.

Before this change, `TabulateFusionSeAttenOp::forward` and `TabulateFusionSeAttenGradOp` called `tabulate_fusion_se_a_{cpu,gpu}` without an `is_sorted` argument, so they took the `= true` default; only grad-grad ever read `ctx->saved_data["is_sorted"]`. Now the flag is threaded into the forward and first-order-grad kernels too.

Since `se_atten.py` sets `self.is_sorted = len(self.exclude_types) == 0`, any frozen compressed DPA1 or se_atten_v2 model configured with `exclude_types` now evaluates differently from before.

I think this is a genuine fix - the sorted early-exit folds `(nnei - jj)` copies, which is wrong on an unsorted neighbour list, and it is the same family as #5438 and #5844. But it is a silent numerical change to a real user-facing configuration, it is not called out in the PR description, and nothing tests it: neither `test_compressed_descriptor_se_atten.py` nor the new `test_tabulate_fusion_se_atten.py` ever sets `exclude_types`, so `is_sorted=False` is never exercised.

Worth a line in the PR body and a compressed-plus-`exclude_types` case that would have failed before.

source/lib/src/gpu/tabulate.cu
```diff
@@ -1039,6 +1062,112 @@ __global__ void tabulate_fusion_se_r_grad_grad_fifth_order_polynomial(
...
+                                      const int nloc,
+                                      const int nnei,
+                                      const int last_layer_size,
+                                      const bool is_sorted) {
+#if GOOGLE_CUDA
+  const size_t shared_memory = sizeof(FPTYPE) * MTILE * last_layer_size;
+  int device = 0;
+  cudaDeviceProp properties;
+  DPErrcheck(cudaGetDevice(&device));
+  DPErrcheck(cudaGetDeviceProperties(&properties, device));
```
wanghan-iapcm: This query runs on every call, including the unchanged `MTILE=4` (lmax=1) path - so every backward and every force evaluation through the compressed tabulate-fusion op, which is the common production path rather than anything to do with the new feature.

`cudaGetDeviceProperties` is roughly millisecond-scale. The forward launcher has no equivalent query; only this grad path gained one, and it is new in this PR - `upstream/master`'s `tabulate.cu` has no such call.

The project already solved this once: #5758 added `deepmd::dpa1_compress_tuning::device_properties` as a cache for exactly this reason. It lives in the PT-side DPA1 compress module and is not reachable from this shared `source/lib` code, so the new code re-queries instead.

A function-local `static` cache keyed by device index would keep the behaviour and remove the per-call cost.

source/lib/src/tabulate.cc
```diff
@@ -110,8 +125,12 @@ inline void locate_xx_se_t(const FPTYPE& lower,
 }
 
 template <typename FPTYPE>
-inline FPTYPE dot(FPTYPE a[4], FPTYPE b[4]) {
-  return a[0] * b[0] + a[1] * b[1] + a[2] * b[2] + a[3] * b[3];
+inline FPTYPE dot(const FPTYPE* a, const FPTYPE* b, const int size) {
```
wanghan-iapcm: This file is a rewrite of the shared lmax=1 path, not an additive change behind a `basis_dim` switch, and that is worth being explicit about since `tabulate_fusion_se_a_cpu` and `_grad_cpu` are the innermost hot loops for the entire SE_A/SE_ATTEN family - every existing CPU training and inference run executes the new code, not just DPA1 users who opt into `lmax > 1`.

The specific changes: `dot()` went from a four-term unrolled expression to a runtime-length loop, `ll[4]` became `ll[25]`, the unrolled `out[...]` stores collapsed into `for (mm < ndescrpt)`, and the `unloop` branch folded into a `scale` factor. `ndescrpt` is a plain runtime `int` parameter defaulted to 4, not a template parameter, so no compile-time-specialised path is retained for the pre-existing width.

I could not find a numerical difference - the summation order in `dot` and the `scale` semantics look equivalent - so this is a performance concern rather than a correctness one. But losing a four-way unroll and a compile-time trip count on a per-atom-per-neighbour inner loop is the kind of regression that no test will catch and that nothing in the PR benchmarks. Templating on `ndescrpt`, or keeping a specialised `ndescrpt == 4` overload, would remove the question.

source/op/pt/dpa1_graph_compress_kernel.cuh
```diff
@@ -0,0 +1,1181 @@
...
+  constexpr int kGroups = ChannelPolicy<Width>::accumulation_groups;
+  const int lane = threadIdx.x & 31;
+  const int warp = threadIdx.x >> 5;
+  const int warps_per_block = blockDim.x / kWarpSize;
+  const long node = static_cast<long>(blockIdx.x) * warps_per_block + warp;
+  if (node >= node_count) {
+    return;
+  }
+
+  float accumulator[BasisDim][kGroups] = {};
```
wanghan-iapcm, line 420: This accumulator rewrite means the kernel split is not a pure relocation. The unrolled `accumulator0..3` became `accumulator[BasisDim][kGroups]` with templated loops - equivalent by inspection at `BasisDim == 4`, but a real code change on the path that ships today.

The problem is that nothing confirms it. `gh pr checks` shows "Build C++ on CUDA" passing while "Test C++ on CUDA" and "Test Python on CUDA" skip, so no executed test compares this against the pre-split kernel, even for lmax=1 - the configuration every current user runs.

Splitting for parallel compile is a good change and there is no historical reason to keep the single file; it was created whole only two weeks ago in #5758. The gap is verification, not design. A CPU-runnable equivalence check, or one CUDA job that actually executes the compressed parity tests at lmax=1, would close it.

source/op/pt/CMakeLists.txt
OutisLi: The high-`lmax` CUDA specializations are intentionally opt-in. I validated them locally with `DEEPMD_ENABLE_DPA1_HIGH_LMAX=ON`, but default builds keep them disabled because instantiating these templates substantially increases CUDA compilation and linking time. I also do not recommend the high-`lmax` fused CUDA path for routine production use at this point; the portable path remains the practical fallback. Enabling this experimental path in regular CI would add a large cost for limited benefit, so I am keeping both the default and CI unchanged and will keep that scope explicit in the documentation/PR description.

source/op/pt/dpa1_graph_compress_kernel.cuh
OutisLi: I checked this against the exact pre-split parent rather than only the PR base. The two kernel bodies moved by `5b61c1276` are byte-for-byte identical before and after the split: `compressed_forward_kernel` is 237 lines with SHA-256 prefix `8a241feb83c4c01b`, and `compressed_backward_kernel` is 293 lines with prefix `db20c2325bac1553`. The `accumulator[BasisDim][kGroups]` generalization happened in the preceding `d4e0db9b0` feature commit and was included in that commit's CUDA parity validation; the split itself only relocates the already-tested bodies into per-width translation units. The separate question of routinely building high-`lmax` specializations in CI is addressed in the CMake thread, but there is no post-validation accumulator rewrite to fix here.

deepmd/pt/model/descriptor/se_atten.py
OutisLi: Fixed in `152cf8f76`. The higher-order basis now uses the protected coordinates `v = diff / (r + env_protection)` consistently in PT, DPModel, compressed CUDA, and uncompressed CUDA, instead of applying protection only to the radial factor while retaining `diff / r` for the angular part. The resulting basis is expressed as norm-normalized regular solid harmonics: it is unchanged when `env_protection == 0`, while its coordinate derivatives stay bounded as `r -> 0` when protection is enabled. The CUDA VJP applies the corresponding chain rule through `v`. I also added a near-collision derivative regression and PT/DPModel basis parity coverage; an independent numerical audit of the handwritten `l=2/3/4` VJPs against PyTorch autograd had maximum absolute error below `9e-15`.

deepmd/dpmodel/descriptor/dpa1.py
OutisLi: Fixed in `152cf8f76`. High-`lmax` top-level DPA1 and `se_atten_v2` archives now emit version 4 in both DPModel and PT, while the standalone DPModel DPA1 block emits version 2. Their deserializers accept the widened version range and explicitly default a missing `lmax` to 1. Existing `lmax=1` archives retain their prior version numbers (including compressed version 3), so old archives remain readable, while older consumers reject the new high-`lmax` schema by version before trying to parse `lmax`/`degree_gain_raw`. Tests now assert the new versions and cross-backend round trip. On the `trainable` point, the PT and DPModel top-level serializers already both emit it; PT has no independent `DescrptBlockSeAtten` wire-format serializer corresponding to the standalone DPModel block, so there is no second PT block format to align.

source/op/pt/tabulate_multi_device.cc
OutisLi: Addressed in `152cf8f76`. The `is_sorted` propagation is retained because it fixes the incorrect padding fold for neighbor lists containing excluded-pair holes. I added a representative compressed DPA1 regression with `exclude_types=[(0, 1)]` that asserts `is_sorted == false` and checks compressed output against the dense path. I also updated the PR Summary to call out this user-visible numerical correction explicitly rather than presenting it as incidental plumbing.

source/lib/src/gpu/tabulate.cu
OutisLi: Fixed in `152cf8f76`. The expensive property lookup is now behind a process-local, mutex-protected cache keyed by CUDA device index, storing both the standard and opt-in shared-memory limits. Each launch still calls `cudaGetDevice` because the active device may change between calls, but `cudaGetDeviceProperties` runs only on the first access for each device instead of every backward/force evaluation, including the common `MTILE=4` path.

source/lib/src/tabulate.cc
OutisLi: Fixed in `152cf8f76`. Forward, grad, and grad-grad now dispatch once at the public entry point to implementations templated on `NDESCRPT` (`4`, `9`, `16`, or `25`). The stack arrays, dot-product loops, stores, and contractions therefore all have compile-time extents inside the hot loops, preserving a dedicated width-4 specialization for existing SE_A/SE_ATTEN workloads instead of carrying runtime-sized 25-element state through that path. The CPU custom op rebuilt successfully, and the focused component-basis and compressed parity tests passed.

njzjz (approved):

PR #5912, opened 2026-07-28
fix(pt_expt): preserve graph routing for raw checkpoints

OutisLi:
## Summary

- preserve graph-based public-forward routing when loading graph-eligible raw PyTorch checkpoints
- distinguish regular PT and PT-experimental checkpoint parameter dialects conservatively, with clear errors for ambiguous payloads
- select the NV graph builder automatically on supported CUDA hosts and otherwise use the dense graph builder; Vesin remains explicit opt-in
- reject nlist-specific controls on graph-routed models instead of silently ignoring them or changing the model's lower semantics

## Compatibility

Graph-eligible raw `.pt` checkpoints intentionally keep the source model's graph-forward semantics. Passing `neighbor_list=` or a non-default `nlist_backend` to such a checkpoint now fails fast because switching to the nlist lower can change predictions (the root cause of #5862). Use `neighbor_graph_method="ase"` for ASE-based carry-all graph construction. Explicit nlist controls remain supported for nlist-routed checkpoints and deployment artifacts.

## Checks

- targeted checkpoint-routing, dialect-detection, graph-builder selection, and graph/nlist × spin/non-spin inference tests: 14 passed, 16 subtests passed
- Ruff format and lint checks passed
- repository pre-commit hooks passed

## Summary by CodeRabbit

* **New Features**
  * Added device-aware automatic inference neighbor-graph selection via `neighbor_graph_method="auto"` (now the default) with improved neighbor-graph metadata reporting.
  * Enabled “graph-lower” execution for eligible PT checkpoints, including support for determining graph edge precision from model characteristics.
  * Improved PT checkpoint backend handling by detecting checkpoint dialects and rejecting unsupported/undecidable cases with clearer errors.
* **Bug Fixes**
  * Strengthened validation for incompatible neighbor and lower-routing configurations, with fail-fast behavior for invalid combinations.
* **Tests**
  * Expanded coverage for PT checkpoint dialect detection, neighbor-graph “auto” resolution across CPU/CUDA, and negative validation cases.

### DIFF
--- a/deepmd/backend/pt_expt.py
+++ b/deepmd/backend/pt_expt.py
@@ -1,12 +1,14 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 from collections.abc import (
     Callable,
+    Mapping,
 )
 from importlib.util import (
     find_spec,
 )
 from typing import (
     TYPE_CHECKING,
+    Any,
     ClassVar,
 )
 
@@ -25,6 +27,46 @@ if TYPE_CHECKING:
     from deepmd.utils.neighbor_stat import (
         NeighborStat,
     )
+
+
+def detect_pt_checkpoint_backend(checkpoint: Any) -> str | None:
+    """Detect the backend dialect of a raw PyTorch checkpoint.
+
+    Parameters
+    ----------
+    checkpoint : Any
+        A checkpoint payload or its unwrapped model state dictionary.
+
+    Returns
+    -------
+    str or None
+        ``"pt-expt"`` or ``"pt"`` when the parameter names identify one
+        backend unambiguously, otherwise ``None``.
+    """
+    state_dict = checkpoint
+    if isinstance(state_dict, Mapping) and "model" in state_dict:
+        state_dict = state_dict["model"]
+    if not isinstance(state_dict, Mapping):
+        return None
+
+    keys = tuple(key for key in state_dict if isinstance(key, str))
+
+    # Weight names are decisive. pt_expt DPA4 also contains ordinary
+    # torch-native ``.bias`` parameters, so bias names cannot override a
+    # clear ``.w`` versus ``.matrix`` distinction.
+    has_pt_expt_weight = any(key.endswith(".w") for key in keys)
+    has_pt_weight = any(key.endswith(".matrix") for key in keys)
+    if has_pt_expt_weight or has_pt_weight:
+        if has_pt_expt_weight == has_pt_weight:
+            return None
+        return "pt-expt" if has_pt_expt_weight else "pt"
+
+    # Bias-only state dictionaries retain the original dialect fallback.
+    has_pt_expt_bias = any(key.endswith(".b") for key in keys)
+    has_pt_bias = any(key.endswith(".bias") for key in keys)
+    if has_pt_expt_bias == has_pt_bias:
+        return None
+    return "pt-expt" if has_pt_expt_bias else "pt"
 
 
 @Backend.register("pt-expt")
@@ -51,11 +93,8 @@ class PyTorchExportableBackend(Backend):
         Returns
         -------
         - 1 for the regular `.pte` / `.pt2` suffixes (default behaviour).
-        - 2 for `.pt` files whose state-dict uses pt_expt's dpmodel
-            parameter naming (`.w`/`.b`); this outranks the legacy pt
-            backend's default suffix score (1) so pt_expt-trained `.pt`
-            checkpoints route here, while genuine pt-trained `.pt` files
-            (which use `.matrix`/`.bias`) keep going to the pt backend.
+        - 2 for `.pt` files whose state dictionary uses the pt_expt parameter
+            dialect. This outranks the pt backend's default suffix score (1).
         - 0 otherwise.
         """
         score = super().match_filename(filename)
@@ -69,21 +108,14 @@ class PyTorchExportableBackend(Backend):
 
             # weights_only=True avoids unpickling arbitrary code from an
             # untrusted .pt — sniffing only needs the dict keys.
-            sd = torch.load(filename, map_location="cpu", weights_only=True)
+            checkpoint = torch.load(filename, map_location="cpu", weights_only=True)
         except Exception:
             # Not a valid torch archive (corrupt file, wrong format, or a
             # weights_only=True restriction trip).  Surrender the claim so
             # the dispatcher falls back to the default suffix match — pt's
             # default score (1) will pick up the file under `dp --pt`.
             return 0
-        if isinstance(sd, dict) and "model" in sd:
-            sd = sd["model"]
-        keys = list(sd.keys()) if hasattr(sd, "keys") else []
-        has_pt_expt = any(k.endswith(".w") or k.endswith(".b") for k in keys)
-        has_pt = any(k.endswith(".matrix") or k.endswith(".bias") for k in keys)
-        if has_pt_expt and not has_pt:
-            return 2
-        return 0
+        return 2 if detect_pt_checkpoint_backend(checkpoint) == "pt-expt" else 0
 
     def is_available(self) -> bool:
         """Check if the backend is available.
--- a/deepmd/backend/pt_expt.py
+++ b/deepmd/backend/pt_expt.py
@@ -1,19 +1,20 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 from collections.abc import (
     Callable,
-    Mapping,
 )
 from importlib.util import (
     find_spec,
 )
 from typing import (
     TYPE_CHECKING,
-    Any,
     ClassVar,
 )
 
 from deepmd.backend.backend import (
     Backend,
+)
+from deepmd.utils.pt_checkpoint import (
+    detect_pt_checkpoint_backend,
 )
 
 if TYPE_CHECKING:
@@ -27,46 +28,6 @@ if TYPE_CHECKING:
     from deepmd.utils.neighbor_stat import (
         NeighborStat,
     )
-
-
-def detect_pt_checkpoint_backend(checkpoint: Any) -> str | None:
-    """Detect the backend dialect of a raw PyTorch checkpoint.
-
-    Parameters
-    ----------
-    checkpoint : Any
-        A checkpoint payload or its unwrapped model state dictionary.
-
-    Returns
-    -------
-    str or None
-        ``"pt-expt"`` or ``"pt"`` when the parameter names identify one
-        backend unambiguously, otherwise ``None``.
-    """
-    state_dict = checkpoint
-    if isinstance(state_dict, Mapping) and "model" in state_dict:
-        state_dict = state_dict["model"]
-    if not isinstance(state_dict, Mapping):
-        return None
-
-    keys = tuple(key for key in state_dict if isinstance(key, str))
-
-    # Weight names are decisive. pt_expt DPA4 also contains ordinary
-    # torch-native ``.bias`` parameters, so bias names cannot override a
-    # clear ``.w`` versus ``.matrix`` distinction.
-    has_pt_expt_weight = any(key.endswith(".w") for key in keys)
-    has_pt_weight = any(key.endswith(".matrix") for key in keys)
-    if has_pt_expt_weight or has_pt_weight:
-        if has_pt_expt_weight == has_pt_weight:
-            return None
-        return "pt-expt" if has_pt_expt_weight else "pt"
-
-    # Bias-only state dictionaries retain the original dialect fallback.
-    has_pt_expt_bias = any(key.endswith(".b") for key in keys)
-    has_pt_bias = any(key.endswith(".bias") for key in keys)
-    if has_pt_expt_bias == has_pt_bias:
-        return None
-    return "pt-expt" if has_pt_expt_bias else "pt"
 
 
 @Backend.register("pt-expt")
--- a/deepmd/pt_expt/infer/deep_eval.py
+++ b/deepmd/pt_expt/infer/deep_eval.py
@@ -723,9 +723,54 @@ class DeepEval(DeepEvalBackend):
         # ``communicate_extended_output``, while the graph runner emits local
         # public keys consumed directly by ``_eval_model_graph``.
         #
-        # Non-spin: (ext_coord, ext_atype, nlist, mapping, fparam, aparam)
-        # Spin:     (ext_coord, ext_atype, ext_spin, nlist, mapping, fparam, aparam)
-        if use_graph_lower:
+        # Graph:      (..., source_row_ptr, fparam, aparam, charge_spin)
+        # Graph spin: (..., source_row_ptr, spin, fparam, aparam, charge_spin)
+        # Nlist:      (ext_coord, ext_atype, nlist, mapping, fparam, aparam)
+        # Nlist spin: (ext_coord, ext_atype, ext_spin, nlist, mapping, fparam, aparam)
+        if use_graph_lower and self._is_spin:
+
+            def _eager_runner_graph_spin(
+                atype: torch.Tensor,
+                n_node: torch.Tensor,
+                n_local: torch.Tensor,
+                edge_index: torch.Tensor,
+                edge_vec: torch.Tensor,
+                edge_mask: torch.Tensor,
+                destination_order: torch.Tensor,
+                destination_row_ptr: torch.Tensor,
+                source_order: torch.Tensor,
+                source_row_ptr: torch.Tensor,
+                spin: torch.Tensor,
+                fparam: torch.Tensor | None,
+                aparam: torch.Tensor | None,
+                charge_spin: torch.Tensor | None = None,
+            ) -> dict[str, torch.Tensor]:
+                model_ret = model.forward_common_lower_graph(
+                    atype,
+                    n_node,
+                    n_local,
+                    edge_index,
+                    edge_vec,
+                    edge_mask,
+                    destination_order,
+                    destination_row_ptr,
+                    source_order,
+                    source_row_ptr,
+                    destination_sorted=True,
+                    do_atomic_virial=True,
+                    fparam=fparam,
+                    aparam=aparam,
+                    charge_spin=charge_spin,
+                    spin=spin,
+                )
+                return model._translate_eager_call(
+                    model_ret,
+                    atype,
+                    do_atomic_virial=True,
+                )
+
+            self.exported_module = _eager_runner_graph_spin
+        elif use_graph_lower:
             from deepmd.pt_expt.model.ener_model import (
                 _translate_energy_keys,
             )
--- a/deepmd/pt_expt/infer/deep_eval.py
+++ b/deepmd/pt_expt/infer/deep_eval.py
@@ -13,6 +13,9 @@ from typing import (
 import numpy as np
 import torch
 
+from deepmd.backend.pt_expt import (
+    detect_pt_checkpoint_backend,
+)
 from deepmd.dpmodel.model.transform_output import (
     communicate_extended_output,
 )
@@ -130,18 +133,6 @@ def _reshape_charge_spin(
             f"charge_spin must be reshape-compatible with ({nframes}, {dim_chg_spin}), "
             f"got shape {charge_spin_arr.shape}."
         ) from err
-
-
-def _is_pt_backend_dpa4_params(model_params: dict[str, Any]) -> bool:
-    """Return whether a training checkpoint should be loaded by the pt backend."""
-    model_type = str(model_params.get("type", "")).lower()
-    if model_type in {"sezm", "dpa4", "sezm_spin"}:
-        return True
-    descriptor = model_params.get("descriptor")
-    if isinstance(descriptor, dict):
-        descriptor_type = str(descriptor.get("type", "")).lower()
-        return descriptor_type in {"sezm", "dpa4"}
-    return False
 
 
 def _warn_legacy_edge_vec(metadata: dict) -> None:
@@ -556,9 +547,17 @@ class DeepEval(DeepEvalBackend):
 
         # Match the training resume path (training.py:712) — weights_only=True
         # avoids unpickling arbitrary code from untrusted checkpoints.
-        state_dict = torch.load(model_file, map_location=DEVICE, weights_only=True)
+        checkpoint = torch.load(model_file, map_location=DEVICE, weights_only=True)
+        checkpoint_backend = detect_pt_checkpoint_backend(checkpoint)
+        state_dict = checkpoint
         if isinstance(state_dict, dict) and "model" in state_dict:
             state_dict = state_dict["model"]
+        if checkpoint_backend == "pt":
+            raise ValueError(
+                f"Checkpoint '{model_file}' uses the regular `pt` parameter "
+                "dialect. Load it with `dp --pt`, or export it to `.pt2` / "
+                "`.pte` before loading it with `pt_expt`."
+            )
         extra = state_dict.get("_extra_state") if isinstance(state_dict, dict) else None
         if not (isinstance(extra, dict) and "model_params" in extra):
             raise ValueError(
@@ -597,13 +596,6 @@ class DeepEval(DeepEvalBackend):
                     head_state[key.replace(prefix, "model.Default.")] = value
             state_dict = head_state
             model_params = head_params
-
-        if _is_pt_backend_dpa4_params(model_params):
-            raise ValueError(
-                "DPA4/SeZM `.pt` checkpoints belong to the regular `pt` backend. "
-                "Use the `pt` backend for eager checkpoint inference, or export "
-                "the checkpoint to `.pt2` / `.pte` before loading it with `pt_expt`."
-            )
 
         model = get_model(deepcopy(model_params)).to(DEVICE)
 
--- a/deepmd/pt_expt/infer/deep_eval.py
+++ b/deepmd/pt_expt/infer/deep_eval.py
@@ -193,12 +193,14 @@ class DeepEval(DeepEvalBackend):
         nlist-form ``.pt2``, and dense-routed ``.pt`` checkpoints):
         ``"auto"`` / ``"vesin"`` / ``"native"``. Explicit non-default values
         are ignored with a warning for graph-routed artifacts.
-    neighbor_graph_method : str, default: "dense"
+    neighbor_graph_method : str, default: "auto"
         Carry-all graph builder for graph-form ``.pt2`` artifacts and
         graph-routed ``.pt`` checkpoints
-        (``metadata["lower_input_kind"] == "graph"``): ``"dense"`` / ``"ase"``
-        (backend-agnostic) or ``"vesin"`` / ``"nv"`` (on-device O(N)). A
-        non-default value on any other artifact raises at construction because
+        (``metadata["lower_input_kind"] == "graph"``): ``"auto"`` selects
+        ``"nv"`` on CUDA when nvalchemiops is available, then ``"vesin"`` when
+        available, and finally falls back to ``"dense"``. Explicit
+        ``"dense"`` / ``"ase"`` / ``"vesin"`` / ``"nv"`` choices are preserved.
+        A non-default value on any other artifact raises at construction because
         the knob would silently do nothing there; use ``nlist_backend`` for the
         nlist path instead. All builders emit the same neighbor set, so the
         choice is performance-only. Consolidating the two knobs into a single
@@ -215,14 +217,14 @@ class DeepEval(DeepEvalBackend):
         auto_batch_size: bool | int | AutoBatchSize = True,
         neighbor_list: Optional["ase.neighborlist.NewPrimitiveNeighborList"] = None,
         nlist_backend: str = "auto",
-        neighbor_graph_method: str = "dense",
+        neighbor_graph_method: str = "auto",
         **kwargs: Any,
     ) -> None:
         self.output_def = output_def
         self.model_path = model_file
         self.neighbor_list = neighbor_list
-        # Graph-lower builder selection: "dense"/"ase" (backend-agnostic) or
-        # "vesin"/"nv" (on-device O(N)).
+        # Graph-lower builder selection is resolved once after model metadata
+        # identifies the lower ABI.
         self._neighbor_graph_method = neighbor_graph_method
         self._is_pt2 = model_file.endswith(".pt2")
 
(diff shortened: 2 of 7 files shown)


deepmd/backend/pt_expt.py
```diff
@@ -27,6 +29,46 @@
...
+    keys = tuple(key for key in state_dict if isinstance(key, str))
+
+    # Weight names are decisive. pt_expt DPA4 also contains ordinary
+    # torch-native ``.bias`` parameters, so bias names cannot override a
+    # clear ``.w`` versus ``.matrix`` distinction.
+    has_pt_expt_weight = any(key.endswith(".w") for key in keys)
+    has_pt_weight = any(key.endswith(".matrix") for key in keys)
+    if has_pt_expt_weight or has_pt_weight:
+        if has_pt_expt_weight == has_pt_weight:
+            return None
```
wanghan-iapcm: Blocking.

This tie returns `None`, and the two consumers of that verdict disagree about what it means.

`match_filename` treats anything other than `"pt-expt"` as score 0, so an ambiguous `.pt` falls through to the pt backend's default suffix score of 1 and `dp test -m foo.pt` routes to pt. But `_load_pt` only rejects an explicit `"pt"` verdict - a `None` slips straight past and pt_expt proceeds to reconstruct the model. So `dp --pt-expt foo.pt`, which forces the backend and bypasses `match_filename` entirely, will happily load exactly the files the heuristic said it could not classify, and fail later inside `load_state_dict` with an opaque key-mismatch traceback.

The two paths should agree. Either `match_filename` and `_load_pt` share one authoritative verdict, or `_load_pt` should raise a clear "cannot determine the checkpoint dialect" error on `None` rather than falling through.

This is also a regression in diagnosability: the model-type-based `_is_pt_backend_dpa4_params` guard that this diff removes produced a deliberate, actionable message for precisely this situation.

`None` is reachable three ways - both `.w` and `.matrix` present, no weight keys with both bias kinds present, and any non-Mapping payload - and none of them has a test.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -268,20 +264,66 @@ def __init__(
...
             )
+        if self.metadata.get("lower_input_kind") in ("graph", "dpa1_canonical"):
+            if self.neighbor_list is not None:
+                raise ValueError(
+                    "neighbor_list only applies to nlist-routed artifacts; use "
+                    "neighbor_graph_method for this graph-routed model."
+                )
+            if nlist_backend != "auto":
+                warnings.warn(
+                    f"nlist_backend={nlist_backend!r} is ignored for graph-routed "
```
wanghan-iapcm: Blocking.

Both new guards added in this block are untested. Grepping the new test file for either message string, and for `assertWarns` or a `neighbor_list=` argument, returns nothing.

That matters more than usual here because these are the guards that define the new graph-routed contract, and this PR is what first makes them reachable for raw `.pt` checkpoints - before it, every `.pt` was `lower_input_kind="nlist"`, so neither branch could fire.

Concretely missing:

- a graph-routed checkpoint constructed with `neighbor_list=` asserting the `ValueError`;
- a graph-routed checkpoint constructed with an explicit `nlist_backend` asserting the warning;
- the negative case for each, so the guards are shown not to fire on an nlist artifact.

The first of these is the one I would most want, because it is a behaviour change rather than a new rejection - see the separate comment on the raise itself.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -268,20 +264,66 @@ def __init__(
...
         """
         if nlist_backend not in ("auto", "vesin", "native"):
             raise ValueError(
                 f"Unknown nlist_backend '{nlist_backend}'; "
                 "expected 'auto', 'vesin', or 'native'."
             )
+        if self.metadata.get("lower_input_kind") in ("graph", "dpa1_canonical"):
+            if self.neighbor_list is not None:
+                raise ValueError(
+                    "neighbor_list only applies to nlist-routed artifacts; use "
```
wanghan-iapcm: This raise turns a previously working call into a construction failure.

Before this PR `_load_pt` set `lower_input_kind = "nlist"` unconditionally, so an ASE `neighbor_list` was honoured on the dense path for every raw `.pt`. Now any graph-eligible checkpoint - dpa1, dpa2, dpa3, dpa4 - flips to `"graph"` and this raises instead.

`deepmd/calculator.py` and `deepmd/infer/deep_eval.py` both accept and forward `neighbor_list` straight into the backend constructor, so an ASE calculator user pointing at a raw pt_expt checkpoint that worked yesterday now fails at construction.

Failing fast is defensible, but it is a user-visible break that the PR body does not mention and no test covers. Two alternatives worth weighing: honour the explicit `neighbor_list` by staying on the nlist lower for that call, or keep the raise but say in the message that graph routing is newly enabled for raw checkpoints and how to opt out.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -268,20 +264,66 @@ def __init__(
         else:
             raise TypeError("auto_batch_size should be bool, int, or AutoBatchSize")
 
-    def _setup_nlist_backend(self, nlist_backend: str) -> None:
-        """Resolve the neighbor-list construction strategy from a user choice.
+    @staticmethod
+    def _resolve_neighbor_graph_method(method: str) -> str:
```
wanghan-iapcm, line 274: Two things about this resolver.

First, it changes the default for every graph-eligible artifact, not just the raw `.pt` files this PR targets. `"auto"` resolves to `nv` on CUDA with nvalchemiops, else `vesin` if importable, else `dense` - so the same model on two machines evaluates through different builders depending on what happens to be installed. The docstrings assert the builders emit the same neighbour set, but the graph lower accumulates via `segment_sum` over edges, and edge ordering after `canonicalize` is an independent implementation per builder, so ordering differences alone perturb fp64 results. Nothing in the PR asserts dense, vesin and nv agree numerically.

Second, the test meant to cover this cannot fail. `TestPtExptLoadPtGraphDPA1.setUpClass` computes `expected_graph_method` with the same `DEVICE.type == "cuda" and is_nv_available()` / `is_vesin_torch_available()` cascade that the assertion then checks, and there is no mocking anywhere in the file. It asserts the code equals a copy of itself, and on any single host it reaches only one of the three branches.

What would actually cover this: patch `is_nv_available`/`is_vesin_torch_available`/`DEVICE` to drive all three branches deterministically, and add one cross-builder equivalence test comparing dense against vesin on the same system at a tight tolerance.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -268,20 +264,66 @@ def __init__(
...
+        from deepmd.pt.utils.nv_nlist import (
+            is_nv_available,
+        )
+        from deepmd.pt_expt.utils.env import (
+            DEVICE,
+        )
+
+        if DEVICE.type == "cuda" and is_nv_available():
+            return "nv"
+        if is_vesin_torch_available():
```
wanghan-iapcm: This branch puts vesin on the default path, which contradicts the builder's own documented contract.

`deepmd/pt_expt/utils/vesin_graph_builder.py`'s module docstring says it loops over frames in Python with roughly 1 ms per frame of call overhead, that it is "intended for `nf == 1` inference and CPU use", and - explicitly - that it "is never on a default hot path: `neighbor_graph_method=None` resolves to the `dense` converter, and vesin is explicit opt-in only."

After this change, any environment with `vesin.torch` installed and without CUDA + nvalchemiops gets that per-frame Python loop by default. That includes runs under `auto_batch_size`, which deliberately batches many frames into a single `_eval_model_graph` call - exactly where the vectorised dense builder handles all frames at once and the vesin loop is worst.

That file is untouched by this PR, so its docstring is now factually wrong either way. Either prefer `dense` when `nframes > 1`, or update the contract to say vesin is now a default and re-examine whether the per-frame loop is acceptable there.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -268,20 +264,66 @@ def __init__(
...
+        if self.metadata.get("lower_input_kind") in ("graph", "dpa1_canonical"):
+            if self.neighbor_list is not None:
+                raise ValueError(
+                    "neighbor_list only applies to nlist-routed artifacts; use "
+                    "neighbor_graph_method for this graph-routed model."
+                )
+            if nlist_backend != "auto":
+                warnings.warn(
+                    f"nlist_backend={nlist_backend!r} is ignored for graph-routed "
+                    "artifacts; use neighbor_graph_method to select the graph "
```
wanghan-iapcm: This warns where the mirror knob raises, and the asymmetry undoes two earlier decisions.

#5714 added the raise for `neighbor_graph_method` on a non-graph artifact precisely because a knob that "would silently do nothing" must not be quietly accepted - that raise is still here, a few lines up. #5491 documented `nlist_backend="vesin"` as a hard request that raises when vesin is unavailable or unsupported.

After this change, a user who pins `nlist_backend="vesin"` for reproducibility on a graph-routed model silently gets `nv` or `dense` instead, with only a `UserWarning` that routine training and inference output will bury.

If the intent is that the two knobs are equally authoritative, this should raise like its sibling. If graph artifacts are meant to be permissive, the docstring for `nlist_backend` should say so explicitly, because the current pair of behaviours is not guessable.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -241,20 +237,20 @@ def __init__(
...
-        # neighbor_graph_method is consumed ONLY by graph-form .pt2 eval
-        # (_eval_model_graph); fail fast instead of silently ignoring it on
-        # nlist-form artifacts (there, the builder knob is nlist_backend).
-        if neighbor_graph_method != "dense" and getattr(self, "metadata", {}).get(
-            "lower_input_kind"
-        ) not in ("graph", "dpa1_canonical"):
+        # ``neighbor_graph_method`` is consumed only by graph-lower evaluation.
+        # Fail fast instead of silently ignoring it on nlist-form artifacts,
+        # where the corresponding builder knob is ``nlist_backend``.
+        if neighbor_graph_method not in ("auto", "dense") and getattr(
```
wanghan-iapcm: This whitelist is now inconsistent with the docstring, as a direct consequence of the default flip in this PR.

The constructor docstring for `neighbor_graph_method` still says: "A non-default value on any other artifact raises at construction because the knob would silently do nothing there."

The guard exempts both `"auto"` and `"dense"`. That was correct when `"dense"` was the default and the sole exemption. Now that the default is `"auto"`, `"dense"` is a non-default value that is accepted silently on an nlist artifact - precisely the silent no-op the docstring promises is prevented.

Either drop `"dense"` from the exemption so the guard matches the documented contract, or reword the docstring to say that both `"auto"` and `"dense"` are accepted everywhere because neither selects a non-default builder.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -589,6 +632,11 @@ def _load_pt(self, model_file: str, head: str | None = None) -> None:
         wrapper.load_state_dict(state_dict)
         model = wrapper.model["Default"].eval()
 
+        from deepmd.pt_expt.model.graph_lower import (
+            model_uses_graph_lower,
+        )
+
+        use_graph_lower = model_uses_graph_lower(model)
```
wanghan-iapcm, line 654: This gate and the export path now answer the same question differently.

Here routing is decided by `model_uses_graph_lower(model)` alone. The export path's `_resolve_lower_kind` in `deepmd/pt_expt/utils/serialization.py` requires `model_uses_graph_lower(model) and _supports_graph_export(model)`, and `_supports_graph_export` additionally rejects geo-compressed descriptors whose fused kernel fails `_fused_eligible("cuda")`.

So for a compressed DPA1 on a build without the eligible fused kernel, `dp freeze` emits an `nlist` artifact while raw `.pt` eager evaluation takes the graph lower - the same model evaluated through two different lowers depending only on file format. That is the classic pt_expt failure mode, where eager and exported paths drift apart.

The drift already bit inside this PR: the first commit picked the graph runner but the non-spin ABI, which the third commit had to repair. Routing the eager path through the same helper that computes `lower_kind` for export would make the two impossible to disagree - and would also remove the need to import the private `_graph_edge_dtype` across module boundaries.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -505,9 +547,17 @@ def _load_pt(self, model_file: str, head: str | None = None) -> None:
...
         # avoids unpickling arbitrary code from untrusted checkpoints.
-        state_dict = torch.load(model_file, map_location=DEVICE, weights_only=True)
+        checkpoint = torch.load(model_file, map_location=DEVICE, weights_only=True)
+        checkpoint_backend = detect_pt_checkpoint_backend(checkpoint)
+        state_dict = checkpoint
         if isinstance(state_dict, dict) and "model" in state_dict:
             state_dict = state_dict["model"]
+        if checkpoint_backend == "pt":
+            raise ValueError(
+                f"Checkpoint '{model_file}' uses the regular `pt` parameter "
```
wanghan-iapcm, line 564: This guard is the only thing standing between a genuine pt-dialect checkpoint and pt_expt's model reconstruction, and nothing tests that it fires.

It replaces the model-type-based `_is_pt_backend_dpa4_params` check that this diff removes. `TestBackendDispatchPt` exercises only `Backend.detect_backend_by_model` - the filename-scoring path - and never constructs a `DeepPot`/`PyTorchExptDeepEval` on a pt-dialect `.pt` to confirm this `ValueError` is raised.

A test that builds a genuine pt checkpoint and loads it through pt_expt would pin both this guard and, with an ambiguous variant, the `None` case raised in the comment on `detect_pt_checkpoint_backend`.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -268,20 +264,66 @@ def __init__(
...
+                )
+            self._neighbor_graph_method = self._resolve_neighbor_graph_method(
+                self._neighbor_graph_method
+            )
+            self._use_vesin = False
+            self._nlist_builder = None
+            return
         is_spin = bool(getattr(self, "_is_spin", False))
         # Native-spin (NeighborGraph route) graph-form artifacts never touch
         # this NLIST builder at all -- graph-form eval uses
```
wanghan-iapcm: This comment now documents an unreachable branch.

The new early-return block above handles every `lower_input_kind in ("graph", "dpa1_canonical")` case and returns before reaching here. By the time execution arrives, `lower_input_kind` can no longer be `"graph"`, so `is_native_spin_graph` a few lines below is always `False` and `unsupported = "spin models" if is_spin and not is_native_spin_graph else None` reduces to `"spin models" if is_spin else None`.

The comment is still literally true - native-spin graph artifacts really never touch this builder - but it now explains logic that cannot run, which is the kind of thing that misleads the next person editing this function into thinking the path is live. Worth deleting the dead condition, or noting that the graph-routed case returns above.

wanghan-iapcm (changes requested):
The routing fix is the right diagnosis and the weight-name dispatch is a genuine improvement. Raw `.pt` was hardcoded to `lower_input_kind="nlist"` since #5562, so a model trained on the graph lower was being evaluated through the dense path - and the old `has_pt_expt and not has_pt` rule mis-handed pt_expt DPA4 checkpoints to the pt backend, because they legitimately carry torch-native `.bias` alongside dpmodel `.w`. Making weight names decisive fixes that; I confirmed pt's SeZM/DPA4 routes through `MLPLayer` and so carries `.matrix`, and that freshly built state dicts for pt and pt_expt across dpa2 and DPA4-native-spin all classify correctly.

Two blocking comments, on the dispatch inconsistency and on the untested new guards. Eight further comments inline.

Worth stating plainly, because it frames several of the others: this PR changes the default evaluation path for every graph-eligible artifact, not only the raw `.pt` files it set out to fix. `neighbor_graph_method` flips from `"dense"` to a device-aware `"auto"`, so the builder now depends on the host and on which optional packages happen to be installed. The builders are asserted to be edge-set equivalent, but the graph lower accumulates through `segment_sum` over edges, so ordering differences alone move fp64 results - and nothing in the PR asserts that dense, vesin and nv agree. The one test that touches this is tautological (see the comment on `_resolve_neighbor_graph_method`).

A couple of things I checked and want to record as clear, so they are not re-raised:

- Virtual-atom spin models do not misroute into the native-spin graph runner. `SpinModel.__init__` calls `descriptor.disable_graph_lower()` and `_graph_lower_disabled` is a persisted buffer restored by `load_state_dict` before the check, so `model_uses_graph_lower` correctly returns False for them.
- The earlier bot concern that the native-spin fix was dead code - because `_load_pt` rejected DPA4/SeZM before that branch - was true at `7c48a6c` but is resolved at head by the final commit, which does exactly what that comment asked. The branch is reachable and has a regression test.
- CI did run on this PR: 11 passing, 2 skipping, and the single failure is a Read the Docs build, not a test. The "Checks: not run per request" note refers to local runs.

Smaller notes:

1. When no `.w` and no `.matrix` key exists, the bias-only fallback returns `"pt"` for anything carrying `.bias`, and `_load_pt` then tells the user to load it with `dp --pt`. That is safe today only because every pt_expt layer is a `NativeLayer` using `.w`/`.b`. A pt_expt model that became fully torch-native would be refused by its own backend, pointed at the wrong one. Returning `None` for "unknown" would fail soft instead.

2. `.pt` files are fully `torch.load`ed twice during dispatch - once in `match_filename` and again in `_load_pt`. Pre-existing, but this PR adds a second `detect_pt_checkpoint_backend` pass on top, and on multi-GB checkpoints backend selection alone pays a full deserialization.

3. Codecov reports 81.43% patch coverage with 13 newly-added lines uncovered, 7 of them in `deepmd/backend/pt_expt.py` - consistent with the coverage gaps in the two blocking comments.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -648,19 +696,122 @@ def _load_pt(self, model_file: str, head: str | None = None) -> None:
...
+        # graph-forward semantics; all other checkpoints use the dense lower.
+        # The dense runner emits internal keys consumed by
+        # ``communicate_extended_output``, while the graph runner emits local
+        # public keys consumed directly by ``_eval_model_graph``.
+        #
+        # Graph:      (..., source_row_ptr, fparam, aparam, charge_spin)
+        # Graph spin: (..., source_row_ptr, spin, fparam, aparam, charge_spin)
+        # Nlist:      (ext_coord, ext_atype, nlist, mapping, fparam, aparam)
+        # Nlist spin: (ext_coord, ext_atype, ext_spin, nlist, mapping, fparam, aparam)
+        if use_graph_lower and self._is_spin:
```
wanghan-iapcm: The runner dispatch flattens an orthogonal 2x2 into a linear `if`/`elif` chain, so each branch's second coordinate has to be reconstructed from what did not match earlier: `elif use_graph_lower` does not mean "graph", it means "graph and not spin", and `elif self._is_spin` means "spin and not graph".

`use_graph_lower` and `self._is_spin` are independent axes -- lower ABI (graph vs nlist) and spin-ness -- and the comment directly above already presents the space as a 2x2 table (Graph / Graph spin / Nlist / Nlist spin). The documentation is two-dimensional while the code is one-dimensional.

Nesting on the outer axis would make the code match the table:

```python
if use_graph_lower:
    if self._is_spin:
        ...  # graph spin
    else:
        ...  # graph
else:
    if self._is_spin:
        ...  # nlist spin
    else:
        ...  # nlist
```

This also places the two graph runners adjacent, where the shared `source_row_ptr` / `destination_row_ptr` signature lives, and the two nlist runners adjacent, where the shared `ext_coord, ext_atype, ..., mapping` signature lives.

Readability only -- the four branches do cover the space exhaustively and dispatch correctly. Non-blocking.

wanghan-iapcm (commented):
One readability note on the eager-runner dispatch.

source/tests/pt_expt/infer/test_deep_eval_pt_checkpoint.py
```diff
@@ -585,6 +711,80 @@ def _spin_eager_reference(model, COORD, ATYPE, SPIN, BOX):
...
+        self.assertIs(backend, Backend.get_backend("pt-expt"))
+
+        dp = DeepPot(
+            self.pt_path,
+            auto_batch_size=False,
+            neighbor_graph_method="dense",
+        )
+        self.assertTrue(dp.has_spin)
+        self.assertEqual(dp.deep_eval.metadata["lower_input_kind"], "graph")
+
```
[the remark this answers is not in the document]
OutisLi: Fixed in `5f37d97b4`. The native-spin raw-checkpoint regression now asserts `dp.deep_eval._neighbor_graph_method == "dense"`, so the explicit builder request cannot silently fall through to auto-selection.

deepmd/pt_expt/infer/deep_eval.py
```diff
@@ -13,6 +13,9 @@
 import numpy as np
 import torch
 
+from deepmd.backend.pt_expt import (
+    detect_pt_checkpoint_backend,
+)
```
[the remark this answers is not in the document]
OutisLi: Fixed in `5f37d97b4`. `detect_pt_checkpoint_backend` now lives in the neutral `deepmd.utils.pt_checkpoint` module. Both `deepmd.backend.pt_expt` and `deepmd.pt_expt.infer.deep_eval` import that utility, so inference no longer imports its backend registration module and the cycle is removed.

deepmd/backend/pt_expt.py
OutisLi: Agreed. Fixed in `5f37d97b4`: `_load_pt` now rejects both an explicit regular-PT verdict and an indeterminate verdict before `_extra_state` parsing or model reconstruction, with separate actionable messages. The detector is now a single shared utility used by dispatch and loading. I also made the bias-only fallback conservative: `.b` alone is pt_expt-specific, while `.bias` alone remains `None` because pt_expt can contain torch-native modules. A table-driven test covers mixed weights, mixed biases, non-Mapping payloads, non-string keys, and wrapped payloads; forced-loader tests cover regular PT and ambiguous checkpoints.

deepmd/pt_expt/infer/deep_eval.py
OutisLi: Agreed on the missing contract coverage. The final contract in `5f37d97b4` is stricter than the original implementation: a graph-routed checkpoint rejects both `neighbor_list=` and an explicit `nlist_backend` (the latter is now an error, not a warning), while nlist-routed checkpoints continue to accept both controls. The tests cover the two graph errors and the two nlist negative cases. I intentionally did not make `neighbor_list=` force the raw checkpoint back to nlist, because that would recreate #5862 and can produce incorrect predictions for graph-native DPA1.

deepmd/pt_expt/infer/deep_eval.py
OutisLi: Thanks for calling out the compatibility break. After rechecking #5862, I kept the fail-fast behavior rather than staying on the nlist lower: the old call was executable, but for graph-native DPA1 the padded nlist lower can produce catastrophically wrong energy/force when descriptor statistics are nonzero. The error now explains that changing lower would change inference semantics and directs ASE users to `neighbor_graph_method="ase"`; an explicitly exported nlist-form artifact remains the deliberate way to request legacy nlist semantics. I added regression coverage and documented this behavior in the PR body.

deepmd/pt_expt/infer/deep_eval.py
OutisLi: Partially agreed and addressed in `5f37d97b4`. The tautological host-dependent assertion was removed and replaced with deterministic mocked coverage of CPU, CUDA+NV, and CUDA-without-NV resolution. Auto now selects only NV on a supported CUDA host and otherwise dense; Vesin is no longer an implicit fallback. I did not add a duplicate cross-builder parity test because `source/tests/pt_expt/model/test_graph_builder_dispatch.py` already compares dense vs Vesin and dense vs NV for energy and force at tight tolerances. The raw-checkpoint regression still verifies end-to-end parity with the source public forward.

deepmd/pt_expt/infer/deep_eval.py
OutisLi: Agreed. Fixed in `5f37d97b4`: Vesin has been removed from the auto cascade and remains explicit opt-in. Auto chooses NV only on CUDA when nvalchemi-toolkit-ops is available; otherwise it uses the vectorized dense builder, with a CUDA warning that points to the NV package.

deepmd/pt_expt/infer/deep_eval.py
OutisLi: Agreed. Fixed in `5f37d97b4`: an explicit `nlist_backend` is a hard request and now raises on graph-routed models instead of being ignored with a warning. The error points to `neighbor_graph_method`; nlist-routed checkpoints retain the existing `native`/`vesin` behavior. Both sides of the contract are covered.

deepmd/pt_expt/infer/deep_eval.py
OutisLi: Agreed. With `auto` as the default, it is now the only exempt value on a nlist-routed artifact. Explicit `neighbor_graph_method="dense"` is no longer silently accepted there and is covered by the nlist checkpoint contract test.

deepmd/pt_expt/infer/deep_eval.py
OutisLi: Thanks; I addressed the shared private-helper concern but intentionally did not reuse `_supports_graph_export` for raw eager routing. These gates answer different questions: `_supports_graph_export` says whether a graph lower can be packaged through `torch.export` (notably for geo-compressed DPA1 without an eligible fused CUDA operator), whereas a raw `.pt` runs eagerly and must reproduce the source model's public `forward`. Applying the exportability gate here would force that raw checkpoint back to nlist and reintroduce the correctness failure in #5862. In `5f37d97b4`, `graph_edge_dtype` was moved to the public shared `model.graph_lower` helper used by both serialization and DeepEval, while eager lower selection deliberately remains `model_uses_graph_lower(model)`.

deepmd/pt_expt/infer/deep_eval.py
OutisLi: Agreed. `5f37d97b4` adds forced-pt_expt loader coverage for both a regular-PT `.matrix` checkpoint and an ambiguous mixed `.w`/`.matrix` checkpoint; both fail before model reconstruction with the intended message. The guard depends only on state-dict parameter names, so the tests use minimal representative payloads rather than constructing an unrelated full PT model. The detector's table test separately covers every `None` path.

deepmd/pt_expt/infer/deep_eval.py
OutisLi: Agreed. Removed the unreachable `is_native_spin_graph` condition in `5f37d97b4`. The graph branch returns first; the remaining spin branch now directly documents and enforces the virtual-atom nlist case.

deepmd/pt_expt/infer/deep_eval.py
OutisLi: Agreed. The runner selection in `5f37d97b4` is now nested first by lower ABI (graph vs nlist), then by spin, matching the documented 2×2. Targeted end-to-end tests execute graph/non-spin, graph/native-spin, nlist/non-spin, and nlist/spin runners.

OutisLi:
Follow-up to the review summary, implemented in `5f37d97b4`:

- The bias-only fallback is now conservative: `.b` alone identifies pt_expt, while `.bias` alone returns `None` because torch-native modules can occur in pt_expt checkpoints. Weight names remain decisive, so pt_expt DPA4 (`.w` plus ordinary `.bias`) still routes correctly.
- The double `torch.load` during automatic backend dispatch remains unchanged. It predates this fix, and eliminating it cleanly requires the dispatcher to hand the loaded payload to the selected backend (or another explicit ownership/lifetime design); a process-level checkpoint cache would add memory and invalidation risks and is outside this routing-correctness PR.
- The previously uncovered verdict and guard branches now have focused coverage. Together with the four graph/nlist × spin/non-spin runner checks, the targeted run completed with 14 passed tests and 16 passed subtests.

The PR description now also records the graph/nlist compatibility contract and directs ASE graph users to `neighbor_graph_method="ase"`.

wanghan-iapcm (approved):
Verified all eleven at `5f37d97b4`. The nine you agreed with are genuinely fixed, and I accept both of the ones you pushed back on.

The routing foundation is now right. Moving the detector into [`deepmd/utils/pt_checkpoint.py`](https://github.com/deepmodeling/deepmd-kit/blob/5f37d97b495cc8516e4ea8a411cf9419c37167d6/deepmd/utils/pt_checkpoint.py) so dispatch and loading share one verdict was the important part, and `_load_pt` rejects the explicit `pt` verdict and the indeterminate one separately, before `_extra_state` parsing, so a forced `dp --pt-expt foo.pt` no longer reaches `load_state_dict` with an opaque failure. The bias asymmetry is better than what I asked for: treating a lone `.b` as pt_expt-specific while leaving a lone `.bias` unclassified is conservative in the safe direction, since pt_expt models legitimately contain torch-native modules.

The rest check out as described: Vesin is out of the auto cascade and `auto` now resolves to NV only on a CUDA host with the package present; an explicit `nlist_backend` on a graph-routed model raises rather than being ignored; `dense` is no longer silently accepted on an nlist artifact now that `auto` is the default; the unreachable `is_native_spin_graph` condition is gone; and the runner dispatch nests on lower ABI then spin, with all four quadrants covered by test classes. The mocked CPU / CUDA+NV / CUDA-without-NV resolution test replaces the tautological host-dependent assertion, which was the actual problem there.

On the two you did not take:

Keeping `neighbor_list=` as a hard failure is the right call. My concern was that a working call becomes a construction error, but the old call could produce badly wrong energies for graph-native DPA1 with nonzero descriptor statistics, and failing loudly beats returning plausible garbage. The error now names `neighbor_graph_method='ase'` and the exported nlist artifact as the two real alternatives, and the break is written down in the PR body, which is what I was actually asking for.

I also accept not reusing `_supports_graph_export` for eager routing. Exportability and reproducing the source model's public `forward` are different questions, and applying the export gate here would push raw checkpoints back onto the nlist lower and reintroduce #5862. What remains is that a compressed DPA1 can be frozen to an nlist artifact while eager evaluation of the same model takes the graph lower; I read that as a limitation of the export path rather than something this PR should carry.

On coverage: `TestPtExptLoadPtGraphDPA1::test_eval_matches_public_forward` is the test that pins the original defect, since it compares a raw `.pt` evaluation against the source model's public forward for a graph-native descriptor -- the comparison nothing made before, which is why the unconditional `lower_input_kind = "nlist"` went unnoticed.

PR #5913, opened 2026-07-28
perf(pt_expt): use scalable graph builders during training

OutisLi:
## Summary

- resolve one concrete scalable graph backend for each PT-expt training run
- reuse the selected backend across eager, compiled, and validation paths
- support HybridMuon together with the shared compiler compatibility safeguards and Inductor options
- validate graph-builder dispatch and training integration

## Checks

Not run per request; the changes were tested separately on another machine.

## Summary by CodeRabbit

* **New Features**
  * Added configurable `training.neighbor_graph_method` (`auto`, `dense`, `nv`) with consistent runtime selection for graph-eligible energy models.
  * Expanded training optimizer configuration support (including `Adam`, `AdamW`, and `HybridMuon`).
* **Bug Fixes**
  * Improved NV neighbor-list capacity estimation and neighbor-graph dispatch/selection for eager, compiled, and validation paths.
  * Updated HybridMuon bias routing for more bias-like parameter names.
* **Documentation**
  * Documented `training.neighbor_graph_method` behavior and resolution timing.
* **Tests**
  * Added method-resolution, CPU-CUDA validation, and NV-vs-dense equivalence coverage (plus compile-support gating for compile tests).


deepmd/pt_expt/train/training.py
```diff
@@ -1576,22 +1585,48 @@ def _make_sample(
...
         # Optimiser -----------------------------------------------------------
-        opt_type = training_params.get("opt_type", "Adam")
+        opt_type = optimizer_params.get("type", "Adam")
         # LambdaLR multiplies each param group's initial learning rate by the
         # lambda value.  Warmup schedules legitimately return zero at step 0,
         # so use the nonzero schedule base as the denominator and let the
         # lambda initialize the optimizer to the requested warmup value.
         initial_lr = float(self.lr_schedule.start_lr)
+        adam_betas = (
+            float(optimizer_params["adam_beta1"]),
```
wanghan-iapcm, line 1638: `optimizer_params["adam_beta1"]` / `["adam_beta2"]` / `["weight_decay"]` are indexed unconditionally here, before the `opt_type` dispatch below. The `else: raise ValueError(f"Unsupported optimizer type: {opt_type}")` at the end of that chain is therefore unreachable for any type that does not carry the Adam keys.

Concretely, `optimizer_lkf()` in `deepmd/utils/argcheck.py` contributes only the `kf_*` keys, and `convert_optimizer_v31_to_v32` backfills `adam_beta1/2` and `weight_decay` only when `optimizer_cfg["type"] in ("Adam", "AdamW")`. So a config with `optimizer.type: LKF` on pt_expt dies with an opaque `KeyError: 'adam_beta1'` instead of the intended message naming the unsupported optimizer. A second path to the same `KeyError`: any caller that runs `normalize()` without `update_deepmd_input()` -- the `optimizer` section is created by `convert_optimizer_v31_to_v32`, not by `normalize()` -- where the old `training_params.get("opt_type", "Adam")` code silently worked.

The PT twin handles this by special-casing LKF before touching the Adam keys (`deepmd/pt/train/training.py`, `if self.opt_type == "LKF": ... else: adam_betas = ...`). Moving the dispatch first, or reading via `.get()` inside each branch, restores the clear error.

Worth noting the diff also fixes a real latent bug in passing: after `convert_optimizer_v31_to_v32` moves `opt_type` out of `training`, the old `training_params.get("opt_type", "Adam")` was always `"Adam"`, so pt_expt silently ignored the user's chosen optimizer.

deepmd/pt_expt/train/training.py
```diff
@@ -1576,22 +1585,48 @@ def _make_sample(
...
+            )
         elif opt_type == "AdamW":
-            weight_decay = training_params.get("weight_decay", 0.001)
             self.optimizer = torch.optim.AdamW(
                 self.wrapper.parameters(),
                 lr=initial_lr,
+                betas=adam_betas,
                 weight_decay=weight_decay,
             )
+        elif opt_type == "HybridMuon":
```
wanghan-iapcm: This new branch constructs `HybridMuonOptimizer` with nine config-driven kwargs (`momentum`, `lr_adjust`, `lr_adjust_coeff`, `muon_mode`, `enable_gram`, `flash_muon`, `magma_muon`, plus `adam_betas`/`weight_decay`), and the `AdamW` branch above changes behaviour too (`weight_decay` moves from `training_params` to `optimizer_params`, `betas` added). Neither has any pt_expt test: `HybridMuon`, `AdamW` and `opt_type` return no hits anywhere under `source/tests/pt_expt/`, and the tests this PR adds to `source/tests/pt_expt/test_training.py` cover only `neighbor_graph_method`. `source/tests/pt/test_hybrid_muon.py` tests the optimizer's internal routing, not that the pt_expt trainer wires config into it correctly.

CLAUDE.md: "UTs should cover all code paths, including both branches of boolean conditions ... When adding a new feature or API, provide tests that exercise every reachable code path."

A wrong kwarg name, or a key absent from a hand-built config, surfaces only at a live HybridMuon run. The gap is cheap to close -- a construct-the-trainer-and-assert-isinstance test in the shape of the `test_trainer_installs_resolved_graph_method` you just added would do it for both branches.

deepmd/pt_expt/utils/graph_builder.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+    if requested == "dense":
+        return "dense"
+
+    from deepmd.pt.utils.nv_nlist import (
+        is_nv_available,
+    )
+
+    if requested == "auto":
+        return "nv" if device.type == "cuda" and is_nv_available() else "dense"
```
wanghan-iapcm: Making `auto` the default means the neighbor graph used for **training** is chosen by what happens to be installed on the machine: a CUDA box with nvalchemi-toolkit-ops trains through the NV builder, anything else through dense. Edge ordering after canonicalization is an independent implementation per builder and the graph lower accumulates via `segment_sum`, so ordering differences alone perturb fp64 results -- and here that feeds gradient accumulation across the whole run, not a single eval call.

What makes this hard to accept as a default is the coverage situation. `source/tests/pt_expt/utils/test_nv_graph_builder.py` carries a module-level `pytest.mark.skipif(not (torch.cuda.is_available() and is_nv_available()))`, so CPU CI has zero coverage of the NV path, and "Test Python on CUDA" is label-gated and was skipped on this PR. The default therefore selects, on exactly the machines people train on, the one builder no routinely-running job exercises.

This PR also deletes the two notes that documented that gap -- the `Notes` block in `build_neighbor_graph_nv` and the trailing NOTE at the end of `source/tests/common/dpmodel/test_neighbor_graph_builder.py` ("no local oracle set-equality test for pair_excl ... validation is deferred to GPU box tests"). The replacement `test_nv_pair_exclusion_matches_dense` is real coverage, but it is CUDA-gated, so the deleted notes' claim is still true; removing them reads as closing the gap when it has only moved.

Suggest defaulting to `dense` and keeping `nv` explicit opt-in until a GPU job actually runs these tests. Same concern I raised on the inference-side knob in #5912; training is the higher-stakes instance of it.

wanghan-iapcm (commented):
Reviewed the graph-builder dispatch, the optimizer-config move and the compile-contract alignment. Three findings below; the dispatch plumbing itself checks out -- the attribute is installed after every model-replacing step (finetune/init_model/change-bias) and before `_compile_model` and `FullValidator`, and `getattr(..., "dense")` preserves prior behaviour on CPU.

deepmd/pt_expt/train/training.py
```diff
@@ -1793,9 +1828,14 @@ def _make_sample(
...
                 )
 
+        self._configure_neighbor_graph_method(
+            training_params.get("neighbor_graph_method", "auto")
+        )
+
         # torch.compile -------------------------------------------------------
         self.enable_compile = training_params.get("enable_compile", False)
         if self.enable_compile:
+            check_compile_torch_version()
```
wanghan-iapcm, line 1888: `check_compile_torch_version()` raises `RuntimeError` unless torch is 2.11.x or 2.12.x, and here it fires whenever a user sets the ordinary `training.enable_compile: true`. But `pyproject.toml` has `pin_pytorch_gpu = ["torch==2.10.0"]` (the CPU pin is 2.11.0), so on the repo's own pinned GPU torch this turns pt_expt compiled training from working into a hard failure.

The comparison with the existing callers is what makes me want to ask rather than just note it. On master `check_compile_torch_version` is called only from `deepmd/pt/model/model/sezm_model.py`, behind a default-off opt-in; this call site is reached by a plain training config. `source/tests/pt_expt/test_training.py` references `enable_compile` about 29 times with no version skip, and "Test Python on CUDA" is label-gated and was skipped on this PR, so nothing here would have caught it -- CPU CI passes on the 2.11.0 pin.

So: is 2.11/2.12 genuinely required for the pt_expt compiled path, or is the gate being inherited wholesale from the PT SeZM contract? If it is genuinely required, `pin_pytorch_gpu` should move with it, otherwise GPU users are pinned to a torch the feature rejects. If it is not, the gate probably wants to be narrower than every `enable_compile` run.

Non-blocking, but I would rather have the answer before this lands than discover it on a GPU box.

wanghan-iapcm (commented):
One follow-up question on the compile version gate.

deepmd/pt_expt/utils/graph_builder.py
wanghan-iapcm: To be clear about the weight of this one: it is a preference about the default, not a defect. The builders are asserted set-equivalent, so any divergence is summation-order noise rather than a correctness bug, and auto-detecting an accelerated backend when the hardware supports it is a normal pattern. Treat it as a discussion point -- the CI-coverage observation is the part I would still like addressed either way.

OutisLi:
Regarding the CodeRabbit capacity-comment nit: this is not applicable. Upstream `estimate_max_neighbors` declares `atomic_density: float = 0.2`, while this call explicitly passes `0.25`, which is exactly a 25% margin over that baseline. The existing comment already states this accurately, so no code change is needed.

Upstream source: https://github.com/NVIDIA/nvalchemi-toolkit-ops/blob/main/nvalchemiops/neighbors/neighbor_utils.py

deepmd/pt_expt/train/training.py
OutisLi: Good catch. Fixed in 534a52e75: `optimizer_params["type"]` is now validated against the three optimizers supported by pt_expt before any Adam-family field is read, so `LKF`/`AdaMuon` fail with `ValueError("Unsupported optimizer type: ...")` rather than an opaque `KeyError`. `TestTraining.test_unsupported_optimizer_has_clear_error` covers the LKF regression.

The supported training-config path is `update_deepmd_input()` followed by `normalize()`, which materializes the selected optimizer variant and its defaults. I intentionally did not duplicate those defaults inside `Trainer`; if `normalize()` alone is expected to synthesize a completely absent `optimizer` block, that belongs at the shared argcheck/compatibility boundary rather than in backend-local `.get()` fallbacks. The migrated `optimizer.type` is now the sole runtime selector, so the old silent fallback to Adam is gone as well.

deepmd/pt_expt/train/training.py
OutisLi: Agreed. Added `TestTraining.test_supported_optimizers_construct` in 534a52e75. It runs the normal compatibility-and-normalization path for both `AdamW` and `HybridMuon`, constructs a pt_expt trainer, and asserts the concrete optimizer class. For HybridMuon this exercises all normalized config-driven constructor arguments, so a missing key or incorrect keyword fails during construction. I kept this as the focused construction test you suggested rather than adding a redundant training loop; it passes locally.

deepmd/pt_expt/utils/graph_builder.py
OutisLi: Thanks for separating the default-policy preference from the correctness concern. I am keeping `auto` → NV on CUDA intentionally, but have addressed the actionable CI-coverage point.

The `Test CUDA` label is active on this PR, and the PR-specific run https://github.com/deepmodeling/deepmd-kit/actions/runs/30471465186 is currently executing on HEAD `534a52e75`; both the Python and C++ CUDA jobs are in progress. The Python job runs the full `source/tests` suite with the NV package installed. The dedicated NV suite checks dense-set equivalence for PBC and non-PBC inputs, multi-frame batching, differentiable edge vectors, virtual-atom exclusion, pair exclusion, and model energy/force equivalence. CPU CI still covers only policy resolution and fallback; it is not being presented as an oracle for the NV kernels. The old “validation deferred” notes were removed because this concrete CUDA oracle suite now exists and is being run here.

I also added an explicit warning for `auto` on CUDA when the NV package is unavailable, including the installation command, so the fallback is no longer silent.

deepmd/pt_expt/train/training.py
OutisLi: The shared gate is intentional. pt_expt compiled training uses the same `make_fx`/Dynamo/Inductor contract and private workarounds as the PT compile path, so running it on an unvalidated Torch release is not a supported fallback. `training.enable_compile` defaults to `false`: ordinary eager CUDA training on the pinned 2.10 remains supported, while explicitly enabling compile on 2.10 now fails fast by design.

I did not move the repository-wide GPU pin merely to exercise this opt-in feature. Instead, 534a52e75 decorates every pt_expt compile-only test with the same production `check_compile_torch_version()` gate. Therefore the 2.10 CUDA job still runs the eager/NV coverage and skips only compile-specific tests with the production reason.

PyTorch 2.13 is explicitly supported by #5910, which adds 2.13 to the shared allowlist and applies the required symbolic-divisibility workaround on 2.13. This PR deliberately does not duplicate that independent change; once #5910 lands, these tests use the same shared gate and run on 2.13 automatically.

deepmd/pt_expt/model/make_model.py
```diff
@@ -719,7 +672,7 @@ def _resolve_graph_method(
             if "energy" not in self.atomic_output_def().keys():
                 return None
             if self.mixed_types() and self.atomic_model.uses_graph_lower():
-                return "dense"
+                return getattr(self, "neighbor_graph_method", "dense")
```
wanghan-iapcm, line 675: This line is the regression. On master `_resolve_graph_method` returned `"dense"` unconditionally for graph-eligible models, so training never reached the NV builder; with `neighbor_graph_method` defaulting to `auto`, a CUDA box with `nvalchemi-toolkit-ops` installed now routes every training step through `build_neighbor_graph_nv`. That path is broken for multi-frame batches:

```
FAILED source/tests/pt_expt/test_finetune.py::TestEnergyModelDPA1::test_finetune_change_type
  RuntimeError: shape '[1, 3, 3]' is invalid for input of size 18
```

from `Test Python on CUDA` on this commit. The traceback runs `train_step` -> `_call_common_graph` -> `build_neighbor_graph_for_method` -> `build_neighbor_graph_nv` -> `nv_search_matrix`, and dies on `box.reshape(nf, 3, 3)` with `nf=1` against a box carrying 2 frames (18 = 2x9). The system is `natoms=192, bch_sz=2` -- an ordinary batch, not an edge case.

`source/tests/pt_expt/test_finetune.py` is untouched by this PR and the last master CUDA runs are green, so this is the routing change, not a pre-existing failure.

Since graph-eligible models (dpa4, dpa1 with `attn_layer=0`) with `batch_size > 1` are the normal case, this would break most real training runs on a CUDA machine that happens to have the NV package installed -- which is exactly the "chosen by what is installed on the machine" property that made me uneasy about the default. I do not think the default is the core problem though; the builder should be fixed either way.

wanghan-iapcm (commented):
Thanks -- the four points from the previous round check out against HEAD `534a52e75`, with one exception that the CI on this very commit now settles.

**Addressed.** The optimizer dispatch guard is in the right place: `opt_type` is validated against the three supported optimizers at [training.py#L1589-L1592](https://github.com/deepmodeling/deepmd-kit/blob/534a52e758ba022e051320d78a641efae4ed72b3/deepmd/pt_expt/train/training.py#L1589-L1592), before `adam_beta1` is read at L1600, so `LKF` raises `ValueError` rather than `KeyError`. `test_unsupported_optimizer_has_clear_error` genuinely fails pre-fix (the old code raised `KeyError`, not `ValueError`). `test_supported_optimizers_construct` covers both new branches through the real `update_deepmd_input` + `normalize` path, and `optimizer_hybrid_muon()` supplies every one of the nine keys the HybridMuon branch reads, so a renamed kwarg or missing key fails construction. Both tests ran and passed in the CPU Python jobs on this commit. I accept your reasoning on not adding `.get()` fallbacks for a config that skipped `update_deepmd_input()`.

**The compile gate.** Your answer resolves my question -- the gate is deliberate and inherited on purpose. Noting the consequence rather than contesting it, in the inline comment below.

**The `auto` default -- this is the one I would like resolved before merge.** I flagged the CI-coverage gap as the actionable half of that thread, and the run you pointed to has since finished: `Test Python on CUDA` [failed](https://github.com/deepmodeling/deepmd-kit/actions/runs/30471465186/job/90642482403) on `534a52e75`, in the NV path this PR newly makes the default. Details inline. (The three `Test C++` failures are the unrelated paddle-inference tarball extraction problem, not this PR.)

deepmd/pt_expt/utils/nv_graph_builder.py
```diff
@@ -301,9 +295,11 @@ def build_neighbor_graph_nv(
...
     # in-cell positions, so BOTH the search and the edge_vec recomputation use
     # the normalized coords; S then matches the coords the search actually saw.
+    # The 0.25 density preserves a 25% margin over the estimator's 0.2
+    # baseline without using the safety_factor argument deprecated in Ops 0.4.
     initial_capacity = max(
         64,
-        estimate_max_neighbors(float(rcut), safety_factor=1.25),
+        estimate_max_neighbors(float(rcut), atomic_density=0.25),
     )
     coord, cell, neighbor_matrix, num_neighbors, shifts = nv_search_matrix(
```
wanghan-iapcm, line 303: Root cause of the CUDA training failure (flagging here since this is the nearest changed line; the defect is at [L272-L273](https://github.com/deepmodeling/deepmd-kit/blob/534a52e758ba022e051320d78a641efae4ed72b3/deepmd/pt_expt/utils/nv_graph_builder.py#L271-L274)):

```python
nf = coord.shape[0] if coord.ndim == 3 else 1
coord = coord.reshape(nf, -1, 3)
```

`nf` is inferred from `coord` alone and then imposed on `box` via `box.reshape(nf, 3, 3)` in `nv_search_matrix`. The docstring above promises `coord` may be `(nf, nloc*3)`, and `_call_common_graph` passes `cc` straight through from `_input_type_cast`. For the 2-D form `coord.ndim == 2`, so `nf` collapses to 1, every frame is folded into one, and the reshape of a multi-frame `box` raises. For a single frame the wrong inference is invisible, which is why this stayed dormant.

Q1, why no test caught it: every case in `source/tests/pt_expt/utils/test_nv_graph_builder.py` passes a 3-D `coord`, including the multi-frame one -- [`test_nv_batches_frames_without_python_loop`](https://github.com/deepmodeling/deepmd-kit/blob/534a52e758ba022e051320d78a641efae4ed72b3/source/tests/pt_expt/utils/test_nv_graph_builder.py#L62-L79) uses `rng.random((3, 5, 3))`. So `coord.ndim == 3` there and `nf` is always correct. The untested cell is **NV builder x 2-D `(nf, nloc*3)` coord x nframes > 1** -- the exact shape the training caller uses. The suite covers multi-frame and it covers the builder, but never their intersection with the flat coord form.

Deriving `nf` from `box` when it is present, or asserting the two agree, would both fix it and turn a silent frame-collapse into a loud error. Whichever way it goes, a regression test crossing that cell (2-D coord, nframes > 1) is what proves it -- it fails today and passes after.

source/tests/pt_expt/compile_utils.py
```diff
@@ -0,0 +1,20 @@
...
+)
+
+try:
+    check_compile_torch_version()
+except RuntimeError as error:
+    _COMPILE_SUPPORT_ERROR = str(error)
+else:
+    _COMPILE_SUPPORT_ERROR = ""
+
+REQUIRES_SUPPORTED_COMPILE = unittest.skipIf(
```
wanghan-iapcm, line 17: Consequential, not blocking, and a direct result of the answer you gave on the version gate -- worth stating so it is a recorded tradeoff rather than a surprise later.

`pin_pytorch_gpu = ["torch==2.10.0"]` and `check_compile_torch_version()` admits only 2.11/2.12, so this decorator now skips **every** pt_expt compile test on the GPU job. The CPU pin is 2.11.0, so compile itself stays covered -- but only on CPU.

The part that gives me pause is that some of the compile machinery is CUDA-specific by construction. [`patch_inductor_force_int64_indexing`](https://github.com/deepmodeling/deepmd-kit/blob/534a52e758ba022e051320d78a641efae4ed72b3/deepmd/pt/utils/compile_compat.py#L99-L112) exists precisely because a wrapped int32 index "surfaces asynchronously as a CUDA illegal memory access"; it is applied on every supported version and can only be exercised on a GPU. After this change nothing in CI exercises it. Same for `triton.max_tiles=1`.

So the net position is: compiled pt_expt training is unsupported on the pinned GPU torch, and untested on any GPU torch. That seems acceptable only as long as #5910 lands and moves the pin -- if that slips, this is a coverage hole nobody is watching. Would be worth a line in the PR description or an issue so it is tracked.

OutisLi: You are right that there is currently no GPU CI coverage of pt_expt compilation on a supported PyTorch release. To clarify my previous reply: this gap was not introduced by this PR. The repository-wide GPU job was already pinned to PyTorch 2.10 while the compile contract supported 2.11/2.12 (with 2.13 support added in #5910); this PR makes that pre-existing mismatch explicit by skipping compile-only tests under the unsupported 2.10 configuration.

We cannot raise the repository-wide GPU pin to 2.11 or newer at present because of unrelated CI compatibility problems, and changing that pin is outside the scope of this PR. The compile path itself has been tested locally on CUDA with PyTorch 2.11, 2.12, and 2.13, and those tests pass.

I have opened #5935 to track adding focused GPU CI coverage on a supported PyTorch release, including the CUDA-specific int64-indexing and Triton-option paths: https://github.com/deepmodeling/deepmd-kit/issues/5935

deepmd/pt_expt/model/make_model.py
OutisLi: Agreed. Fixed in a48021e0b without changing the `auto` policy. The NV builder now derives `(nf, nloc)` from `atype`, which is the authoritative frame/atom contract, and reshapes coordinates to exactly `(nf, nloc, 3)`. A flattened `(nf, nloc * 3)` training batch therefore keeps all frames instead of collapsing to `nf=1`.

I also changed the existing multi-frame NV test to use the flattened coordinate layout produced by training, so it covers this exact regression without adding another CUDA test case. The new push will re-exercise the unchanged finetune test through the default NV route on CUDA.

deepmd/pt_expt/utils/nv_graph_builder.py
OutisLi: Agreed on the root cause. Fixed in a48021e0b by deriving both `nf` and `nloc` from `atype.shape[:2]` and then reshaping `coord` to `(nf, nloc, 3)`. I used `atype` rather than `box` because it remains authoritative for both periodic and non-periodic inputs (`box` may be `None`), and it matches the existing dense builder's shape normalization.

`test_nv_batches_frames_without_python_loop` now feeds a 2-D `(nf, nloc * 3)` coordinate tensor with multiple frames while retaining the dense builder as the edge-set oracle. That crosses the previously missing test cell directly.

deepmd/pt_expt/utils/graph_builder.py
OutisLi: Follow-up after the CUDA run: it correctly exposed the missing flattened multi-frame case in the NV builder. Commit a48021e0b fixes the shape contract by deriving `(nf, nloc)` from `atype`, and the existing multi-frame NV test now uses the flattened training layout. This addresses the concrete correctness issue while preserving the intentional `auto` → NV-on-CUDA default.

wanghan-iapcm (approved):
The NV multi-frame regression is fixed, and the fix is better than what I suggested.

[`nv_graph_builder.py`](https://github.com/deepmodeling/deepmd-kit/blob/dc3abd5643e665a05bf81baa43de0cb17a091dcc/deepmd/pt_expt/utils/nv_graph_builder.py#L270-L274) now derives the shape from `atype`:

```python
nf, nloc = atype.shape[:2]
coord = coord.reshape(nf, nloc, 3)
```

`atype` is `(nf, nloc)` in both layouts, so it is unambiguous where `coord.ndim` was not. I had proposed deriving `nf` from `box` or asserting the two agree; keying off `atype` is cleaner and also fixes `nloc`, which the old code inferred from a coordinate array that had already collapsed. `nv_search_matrix`'s own `nf = coord.shape[0]` is fine now, since the caller guarantees the reshape its docstring assumes.

Reusing the existing multi-frame test rather than adding a CUDA case was the right economy, and the new layout crosses the cell that was missing: `rng.random((3, 5, 3)).reshape(3, -1)` is the flattened training form with three frames, which is what `_call_common_graph` passes. On the old code that collapses to `nf=1` and the three-frame box fails to reshape, so it fails before the fix and passes after.

Approving on the code. Worth noting for whoever merges: the `Test CUDA` label has come off, so `Test Python on CUDA` is currently skipping and neither this fix nor its test has actually executed -- `Pass testing on CUDA` reports green because it aggregates over skipped jobs. The merge queue does run `test_cuda.yml`, so the finetune case that originally surfaced this will be exercised there before merge; I mention it only because the CUDA run is the only thing that can see this code path, and it would be easy to read the current green as coverage.

Issue #5636, opened 2026-06-30
[Code scan] Preserve per-frame find flags when collating LMDB batches


njzjz:
Found during a Codex global scan of `deepmodeling/deepmd-kit` at commit `73de44b1f94471b2e3bdb6b11f57b34d7bc791bb`.

## Problem

`collate_lmdb_frames()` takes every `find_*` flag from the first frame and applies it to the whole batch.

Evidence:

- `LmdbDataReader.__getitem__()` sets `find_<key>` per frame when registered data requirements are present or missing: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/utils/lmdb_data.py#L648-L689
- It also sets per-frame `find_fparam`, `find_aparam`, `find_spin`, and `find_charge_spin`: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/utils/lmdb_data.py#L696-L701
- The collator documents that `find_*` flags are taken from the first frame: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/utils/lmdb_data.py#L811-L821
- The implementation copies `frames[0][key]` for all `find_*` keys: https://github.com/deepmodeling/deepmd-kit/blob/73de44b1f94471b2e3bdb6b11f57b34d7bc791bb/deepmd/dpmodel/utils/lmdb_data.py#L831-L841

## Impact

If optional labels vary within a batch, valid labels in later frames can be ignored, or default-filled labels in later frames can be treated as real data. This can bias losses and validation metrics for partially labeled LMDB datasets.

## Suggested Fix

Stack `find_*` flags per frame, or batch/group frames by label availability before collation. Add a two-frame LMDB batch test where only one frame has a label such as `energy` or `force`, and assert the loss mask uses exactly that frame.

PR #5914, opened 2026-07-28
perf(lmdb): parallelize batch decoding and support epoch-based schedules

OutisLi:
## Summary

- parallelize LMDB batch decoding across PT and PT-expt backends with bounded worker resources
- add LMDB worker configuration and strengthen dataset/resource lifecycle handling
- resolve backend-independent epoch-to-step schedules, including multi-task probabilities and distributed rank synchronization
- cover LMDB decoding, scheduling, multitask, and distributed-training behavior with regression tests

## Checks

No test suite was run per request; the changes were tested separately on another machine. Repository pre-commit style hooks passed while creating the cherry-pick commit.

## Summary by CodeRabbit

- **New Features**
  - Unified step scheduling for single- and multi-task training, deriving total steps and task sampling probabilities from config.
  - Faster LMDB training input pipeline with deterministic parallel batch iteration and config-driven decoding, plus a new LMDB batch loader for training.
- **Bug Fixes**
  - Pinned epoch-to-step/run-length decisions for consistent distributed behavior across ranks.
  - Improved LMDB safety with reliable decoder/loader cleanup on failure and safer dataset merging/iterator lifecycle.
- **Documentation**
  - Documented `DP_LMDB_NUM_WORKERS` override and auto-selection behavior.
- **Tests**
  - Expanded coverage for step scheduling, LMDB decoding/batching correctness, distributed epoch consistency, and shutdown robustness.


wanghan-iapcm (commented):
Reviewed the parallel-decode path, the new shared step schedule, and the worker sizing.

One thing worth recording up front, since it was the property I most expected to break: batch contents and ordering are independent of worker count. `_submit` slices into contiguous chunks in worker order, `[f.result() for f in ...]` preserves submission order, and `_merge_lmdb_chunks` concatenates in list order -- so `DP_LMDB_NUM_WORKERS` is a pure performance knob and does not perturb training numerics. Three separate passes over the code agreed on that. The `merge_lmdb` refcount fix (`try/finally: _close_lmdb(src_path)` replacing a direct `src_env.close()` on a refcounted handle) and the `pin_memory=DEVICE.type != "cpu"` correction are both genuine fixes in passing.

Four findings below plus two coverage asks. The two correctness ones (the epoch counter and the pt_expt epoch length) are both confined to DDP, which is also why no existing test catches them -- the added distributed tests use mocked `epoch_length` callables and never cross an epoch boundary at `world_size > 1`.

None of this is blocking from my side; the rebase question is the one I would settle first.

deepmd/dpmodel/utils/lmdb_data.py
```diff
@@ -161,6 +176,603 @@ def _remap_atom_types(atype: np.ndarray, type_remap: np.ndarray) -> np.ndarray:
...
+                    coord,
+                    box,
+                    atype,
+                    stop_below=float(default),
+                )
+            ],
+            dtype=_resolve_frame_dtype(config, "min_pair_dist"),
+        )
+
+    structural_keys = frozenset(
```
wanghan-iapcm, line 402: This PR's merge base is `2f419f4e3`, which predates `e37cc0095` -- "fix(data): preserve LMDB label availability across batches (#5839)", Fixes #5636, merged into master on 2026-07-28. That fix changed this same file by +408 lines, and six files overlap between the two changes. GitHub currently reports the PR as CONFLICTING/DIRTY, so the collision is visible rather than silent -- but the two changes are effectively disjoint rewrites of the same file, which makes the conflict resolution the risky part rather than the conflict itself.

Concretely, #5839 hoisted the structural-key set to a module constant precisely so the decoder and the availability-signature builder would share one definition; this line re-inlines it as a local frozenset. #5839 also added `get_find_signature()` / `group_indices_by_find_signature()` and partitioned batches by `(nloc, find_* signature)` rather than by `nloc` alone, plus a `_find_signature_cache` invalidated inside `add_data_requirement`. None of that vocabulary exists in this PR's version of the file, and the rewritten `add_data_requirement` here has no cache invalidation.

That matters because this PR's new `decode_lmdb_batch` raises on a batch whose frames disagree on `find_*`. If the resolution keeps this PR's nloc-only grouping, the exact #5636 dataset (only some frames carrying `virial`) now produces a batch that mixes `find_virial=1` and `find_virial=0` frames and training aborts with `ValueError` on data that works on today's master. If instead #5839's sampler survives but this `add_data_requirement` wins, the signature cache is never invalidated after new requirements register, and grouping is computed from a stale signature.

Could you rebase onto current master and re-verify rather than resolving this file textually? The two designs need reconciling by hand.

deepmd/dpmodel/utils/lmdb_data.py
```diff
@@ -161,6 +176,603 @@ def _remap_atom_types(atype: np.ndarray, type_remap: np.ndarray) -> np.ndarray:
...
+                _decode_lmdb_worker_chunk,
+                self._reader.lmdb_path,
+                self._reader.frame_format,
+                decode_config,
+                chunk,
+            )
+            for chunk in chunks
+        ]
+
+    def _parallel_eligible(self, indices: list[int]) -> bool:
```
wanghan-iapcm: `_parallel_eligible` requires `len(indices) >= self._num_workers`. Combined with the defaults, that condition appears to be false for essentially every batch, so the parallel decode path this PR adds never runs:

- `get_lmdb_num_workers()` auto-resolves to `min(32, cpus // local_world_size, 64 // local_world_size)` -- 8 to 32 on any real machine
- `training_data.batch_size` defaults to `"auto"`, whose rule is a 32-atom budget, and `get_batch_size_for_nloc` returns `ceil(32 / nloc)` -- i.e. **1 frame** for any system with `nloc >= 32`

So `len(indices)` is 1 while `num_workers` is 8-32. Every batch takes the `_deferred_indices` branch in `_schedule`, and that branch stores indices rather than a future, so `__next__` decodes it synchronously -- meaning there is no prefetch overlap either, not just no parallelism. `doc/env.md` documents the small-batch synchronous fallback, so that rule is clearly intentional; the issue is that with the auto worker count it swallows the default configuration whole, and the same paragraph's claim that "the next batch is prefetched while the current batch is consumed" does not hold on that path.

`_submit` already clamps `workers = min(self._num_workers, len(original_keys))`, so the strict `>=` gate is not needed for safety -- a threshold like `len(indices) >= 2`, or one based on chunk size rather than worker count, would let the pool actually engage.

Worth confirming against the configuration you benchmarked on: if that run set an explicit large `batch_size`, the speedup would be real there and absent for default users.

deepmd/dpmodel/utils/lmdb_data.py
```diff
@@ -161,6 +176,603 @@ def _remap_atom_types(atype: np.ndarray, type_remap: np.ndarray) -> np.ndarray:
...
+            if self._parallel_eligible(indices):
+                chunks = [future.result() for future in self._submit(indices)]
+                batch = _merge_lmdb_chunks(chunks)
+            else:
+                batch = self._reader.decode_batch(indices)
+
+        self._schedule(self._next_indices())
+        return batch
+
+    def _next_indices(self) -> list[int]:
```
wanghan-iapcm, line 881: `_next_indices` advances `self._epoch` on **this rank's** `StopIteration` and `_iter_epoch` then calls `sampler.set_epoch(self._epoch)`. That is safe for a single process, but not for `DistributedSameNlocBatchSampler`, whose contract is that all ranks build the same deterministic global list from `seed + epoch` and each take a disjoint stride of it.

`DistributedSameNlocBatchSampler.__len__` is `max(0, (total + world_size - 1 - rank) // world_size)`, which differs by one across ranks whenever `total % world_size != 0`. With 10 global batches on 3 ranks the lengths are 4/3/3: ranks 1 and 2 exhaust first, bump `_epoch` and reseed, while rank 0 is still finishing epoch 0. From then on the ranks are permanently phase-shifted -- at the same optimizer step they draw from lists built with different epochs, and the stride is only disjoint *within one list*, so ranks process overlapping frames while others are skipped, with the drift accumulating each epoch.

This is newly reachable: before this PR `set_epoch` had no production caller (only tests), so every rank sat at epoch 0 and the shared-epoch invariant held trivially. The prefetch also fires the increment one step before consumption, so it isn't reproducible from the training-step index alone.

Deriving the epoch from a rank-synchronized quantity, or broadcasting it the way this PR already broadcasts the epoch lengths, would close it. None of the added DDP tests cross an epoch boundary at `world_size > 1`.

deepmd/pt_expt/train/training.py
```diff
@@ -2080,6 +2082,33 @@ def get_data(
...
+        and ``sys_probs[i]``, the probability of drawing from that system, from
+        which ``compute_total_numb_batch`` derives the dataset-wide epoch
+        length ``ceil(max_i(nbatches[i] / sys_probs[i]))``.  Unlike the pt
+        backend, whose loader shards the systems, every rank here holds the
+        complete data system and samples it independently, so the ranks jointly
+        cover one epoch after ``total / world_size`` steps each.
+        """
+        data = self.training_data_by_task[model_key]
+        total = compute_total_numb_batch(data.nbatches, data.sys_probs)
+        return int(np.ceil(total / self.world_size))
```
wanghan-iapcm, line 2155: This divides unconditionally by `world_size`, and the docstring justifies it with "every rank here holds the complete data system and samples it independently, so the ranks jointly cover one epoch after `total / world_size` steps each."

That premise holds for `DeepmdDataSystem`, but not for `LmdbDataSystem`. `pt_expt/utils/lmdb_dataset.py` constructs a plain `SameNlocBatchSampler(...)` with no rank/world_size argument -- unlike the pt backend, which swaps in `DistributedSameNlocBatchSampler` when `world_size > 1` -- and `pt_expt/entrypoints/main.py` passes the same `training.seed` to every rank. `SameNlocBatchSampler.__iter__` derives its RNG purely from `seed + epoch`, so under DDP every pt_expt rank walks the identical deterministic batch list.

So the ranks do not jointly cover an epoch; they each repeat the same one. With `numb_epoch: 1` on 4 GPUs and 4000 batches, `num_steps` resolves to 1000, all four ranks consume the same first 1000 batches, three quarters of the dataset is never seen, and the run reports one epoch. The requested run length is silently divided by `world_size`.

Either shard `LmdbDataSystem` per rank as the pt backend does, or make the division conditional on the data system actually being replicated-and-independently-sampled.

deepmd/dpmodel/utils/lmdb_data.py
```diff
@@ -161,6 +176,603 @@ def _remap_atom_types(atype: np.ndarray, type_remap: np.ndarray) -> np.ndarray:
...
+            if field == "type" or value is None:
+                continue
+            if field == "fid":
+                batch[field][row] = value
+            else:
+                destination = batch[field]
+                array = np.asarray(value)
+                if destination.shape[1:] != array.shape:
+                    raise ValueError(
+                        f"LMDB field {field!r} changes shape within one batch: "
```
wanghan-iapcm, line 536: This is one of four consistency guards the PR adds that no test exercises -- the other three being "LMDB frames in one same-nloc batch expose inconsistent fields" just above, and "LMDB worker chunks expose inconsistent fields" / "LMDB field availability changes across worker chunks" in `_merge_lmdb_chunks`. Only the sibling `find_*` mismatch is covered, by `test_batch_rejects_mixed_label_availability`.

These four are the merge step's only protection: without them a heterogeneous chunk becomes an opaque `np.concatenate` shape error or, worse, wrong data merged into the batch. Since chunked parallel decode is the new and most error-prone plumbing in this PR, its guards are where a regression would land.

CLAUDE.md asks for tests that "exercise every reachable code path", and these are cheap to reach -- hand `_merge_lmdb_chunks` two chunks that disagree on a field, and `decode_lmdb_batch` two frames that disagree on a shape.

deepmd/env.py
```diff
@@ -143,6 +144,58 @@ def get_default_nthreads() -> tuple[int, int]:
     )
 
 
+def get_lmdb_num_workers() -> int:
```
wanghan-iapcm, line 147: `get_lmdb_num_workers()` is new public API in this PR, and `source/tests/pt_expt/utils/test_env.py` covers three cases: a valid override, a negative override, and one CPU-affinity path with `LOCAL_WORLD_SIZE=4`. Five reachable branches have no test:

- a non-numeric `DP_LMDB_NUM_WORKERS` (a distinct `except` clause from the negative-value check)
- the `os.sched_getaffinity` `AttributeError` fallback to `os.cpu_count()`
- `LOCAL_WORLD_SIZE` unset, i.e. the `"1"` default -- the single-process case, which is the common one
- `LOCAL_WORLD_SIZE` non-integer
- `LOCAL_WORLD_SIZE` zero or negative

All are reachable with `monkeypatch.setenv` alone, no special hardware. CLAUDE.md: "UTs should cover all code paths, including both branches of boolean conditions ... When adding a new feature or API, provide tests that exercise every reachable code path."

deepmd/dpmodel/utils/lmdb_data.py
OutisLi: Addressed. I merged current upstream/master in da59667ef and re-verified the #5839 behavior. The resulting code retains the shared structural-key definition, groups batches by atom count plus label-availability signature, and invalidates the signature cache when requirements change. e611ca5c0 also makes grouping and collation use the same availability rule. Thanks for flagging the overlap.

deepmd/dpmodel/utils/lmdb_data.py
OutisLi: Confirmed. The benchmarked workloads use batches large enough to enter parallel decode; smaller auto batches intentionally remain synchronous because process scheduling and IPC are not amortized there. I am keeping this threshold unchanged.

deepmd/dpmodel/utils/lmdb_data.py
OutisLi: Fixed in 0ca3b5f95. The deterministic global batch list is now padded to a multiple of world_size before strided partitioning, so every rank yields ceil(total_batches / world_size) batches and crosses the epoch boundary together. Padding duplicates at most world_size minus one batches, and the non-divisible case now has a regression test.

deepmd/pt_expt/train/training.py
OutisLi: Fixed in 0ca3b5f95. get_trainer now passes the initialized rank and world_size into the LMDB training data system, which selects DistributedSameNlocBatchSampler for multi-rank runs. nbatches continues to report the global count, so the existing ceil(total / world_size) epoch length matches the local sampler length. Validation data remains unsharded.

deepmd/dpmodel/utils/lmdb_data.py
OutisLi: Covered in 0ca3b5f95 with one compact subtest. It exercises inconsistent frame fields, per-field shape changes, inconsistent worker-chunk fields, and availability changes across chunks.

deepmd/env.py
OutisLi: Covered in 0ca3b5f95. The tests now include non-numeric and negative worker overrides, the affinity-to-cpu_count fallback with LOCAL_WORLD_SIZE unset, non-integer LOCAL_WORLD_SIZE, and non-positive LOCAL_WORLD_SIZE, while retaining the explicit affinity-budget case.

wanghan-iapcm (approved):
Verified all six threads against the current head rather than the replies.

The rebase reconciled cleanly: #5839 is now in the history, its structural-key hoisting and availability-signature grouping survive, and `add_data_requirement` still invalidates the signature cache. Handling `refresh_batch_count()` on the distributed sampler was a good catch that I had not asked for -- the availability partitioning changes the batch count, so the sharding fix and the rebase genuinely interact there.

The two distributed fixes are the substantive ones and both are right. `DistributedSameNlocBatchSampler.__len__` is now rank-independent with deterministic padding in `_partition_batches`, so ranks cross the epoch boundary together instead of drifting apart when the batch count is not divisible by the world size. And plumbing rank/world_size through to `LmdbDataSystem` fixes the pt_expt side properly -- keeping `nbatches` as the global count so `ceil(total / world_size)` still matches the local sampler length is exactly the invariant that keeps the epoch arithmetic honest.

`test_distributed_batches_are_sharded_with_equal_epoch_lengths` is a real regression test, not a placeholder: three global batches on two ranks is the non-divisible case, and asserting `len(sampler) == [2, 2]` fails on the old rank-dependent length, while the disjointness assertion fails on the old shared-seed sampler where both ranks drew identical batches. One test that fails for both reasons before the fix.

The four merge guards and the five worker-count branches are now exercised.

On the parallel-decode threshold: your reasoning is fine and I am not pressing it. One leftover, non-blocking -- `doc/env.md` still says "The next batch is prefetched while the current batch is consumed" one sentence before explaining that small batches decode synchronously, and the deferred path stores indices rather than a future, so there is no prefetch there either. Now that the synchronous fallback is a deliberate choice rather than an oversight, that sentence would be worth qualifying whenever you next touch the file.

deepmd/pt_expt/utils/lmdb_dataset.py
```diff
@@ -132,7 +164,7 @@ def get_stat_batch(self, sys_idx: int) -> dict[str, Any]:
             start = 0
         stop = min(start + batch_size, len(group_indices))
         self._stat_offsets[sys_idx] = stop
-        return self._collate_indices(group_indices[start:stop])
+        return self._reader.decode_batch(group_indices[start:stop])
```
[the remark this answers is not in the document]
OutisLi: Fixed in 0d188183f. Statistics now reuse `collect_lmdb_sampling_groups()`, the same stable `(nloc, find-signature)` partition used by training iteration and batch-count logic. `LmdbDataSystem` rebuilds its statistical systems and offsets after data requirements are registered, so `make_stat_input()` cannot combine incompatible availability flags. I also extended the partial-label regression to cover complementary energy/force labels through the statistics path.

deepmd/dpmodel/utils/lmdb_data.py
```diff
@@ -195,6 +215,738 @@ def _remap_atom_types(atype: np.ndarray, type_remap: np.ndarray) -> np.ndarray:
...
+        return self._closed
+
+    def close(self) -> None:
+        """Cancel prefetched work and release the shared decoder pool."""
+        if self._closed:
+            return
+        self._closed = True
+        if self._pending is not None:
+            for future in self._pending.futures:
+                future.cancel()
```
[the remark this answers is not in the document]
OutisLi: Fixed in 0d188183f. `_collect()` and `close()` now share `_wait_for_decoder()`. Closing cancels queued futures, monitors any already-running prefetch, and calls `_lose_the_pool()` when a decoder exits or the futures report `BrokenProcessPool`; the final release then dismantles the unhealthy executor instead of waiting on a partial result. Added deterministic stalled-prefetch coverage and verified the real lost-pool shutdown and shared-pool paths.

njzjz (approved):

PR #5918, opened 2026-07-28
fix(cmake): verify paddle inference download integrity

njzjz:
## Summary

The `paddle_inference.tgz` download in `source/CMakeLists.txt` used `file(DOWNLOAD)` without any integrity check. A truncated or corrupted archive (common with the ~200 MB – 1.6 GB files served from `paddle-qa.bj.bcebos.com`) would silently proceed to extraction, which itself ignored its exit code, leaving an empty/partial `paddle_inference_install_dir`. This produced confusing downstream "missing header/library" errors unrelated to the real cause, and `file(REMOVE ${TGZ_FILE})` even deleted the evidence.

## Changes

Add `EXPECTED_HASH SHA256=...` to the `file(DOWNLOAD)` call for all three variants so CMake fails fast with a clear message on checksum mismatch:

| Variant   | SHA256 |
|-----------|--------|
| CUDA 11.8 | `89963f601a3469270080cca076327da1e6c5e8f0c0ee3aa5e0e0371b2b8c2cb7` |
| CUDA 12.6 | `5318eb879f4930d4b08e52c5e040c0b4b96d4afae62bf57bcae031fff0396551` |
| CPU       | `6adeccb3920e92ddea3e7f36eee770bc10f7b51bab63ed96b9bbade47951de30` |

This is a minimal change; retry logic and timeout handling are intentionally left for a follow-up.

---

Coding agent: opencode
opencode version: 1.18.8
Model: ustc/glm-5.2
Reasoning effort: max

## Summary by CodeRabbit

* **Bug Fixes**
  * Added stronger integrity checks for Paddle inference library downloads using SHA-256 verification.
  * Improved handling of download and extraction failures by surfacing detailed status and failing configuration when they don’t succeed.


source/CMakeLists.txt
```diff
@@ -53,6 +53,8 @@ if(ENABLE_PADDLE)
         set(DOWNLOAD_URL
             "https://paddle-qa.bj.bcebos.com/paddle-pipeline/GITHUB_Docker_Compile_Test_Cuda118_cudnn897_Trt8616_D1/6ed5dd3833c32c3b21e14b1fb1a71f5a535a0fcc/paddle_inference.tgz"
         )
+        set(DOWNLOAD_SHA256
+            "89963f601a3469270080cca076327da1e6c5e8f0c0ee3aa5e0e0371b2b8c2cb7")
```
wanghan-iapcm, line 57: How were these two CUDA hashes obtained?

The CPU hash is already proven correct by this PR's own CI, and I checked that this is genuine coverage rather than a formality: `.github/workflows/test_cc.yml` runs two matrix entries with `enable_paddle: true`, and `source/install/test_cc_local.sh` sets `PADDLE_INFERENCE_DIR` only as a shell variable -- it never passes `-DPADDLE_INFERENCE_DIR` to `cmake`. So `if(NOT DEFINED PADDLE_INFERENCE_DIR)` is true on the runner, the download and the new `EXPECTED_HASH` really execute, and a later step copies `.so` files out of `paddle_inference_install_dir`, which would fail outright on an empty extraction. Both jobs are green here, so `6adeccb3...` is confirmed.

Neither CUDA hash has that backing. `test_cuda.yml` is gated on the `Test CUDA` label / workflow_dispatch / merge_group and was skipped for this PR, so `5318eb87...` (CUDA 12.6) is unverified on this run. And nothing in CI ever configures a CUDA 11 toolkit -- `build_cc.yml` only has a cuda120 entry and does not enable Paddle at all -- so `89963f60...` is exercised by no job, on this PR or any future one.

That asymmetry is what concerns me. A wrong constant here is not a subtle bug; it is a hard configure failure for every affected source build, and for CUDA 11.8 there is no point at which we would find out before a user does. I tried to recompute all three locally, but the bucket throttles to a few tens of kB/s from here and 1.0 GB / 1.6 GB were not retrievable in reasonable time.

Applying the `Test CUDA` label would cover 12.6. For 11.8, could you say how the value was derived -- computed from an actual download, or produced some other way? If it came out of the model rather than off the wire, it would be worth someone recomputing it out of band before this merges.

source/CMakeLists.txt
```diff
@@ -75,10 +79,13 @@ if(ENABLE_PADDLE)
...
           "https://paddle-qa.bj.bcebos.com/paddle-pipeline/GITHUB_Docker_Compile_Test_Cpu_Mkl_Avx_D1/6ed5dd3833c32c3b21e14b1fb1a71f5a535a0fcc/paddle_inference.tgz"
       )
+      set(DOWNLOAD_SHA256
+          "6adeccb3920e92ddea3e7f36eee770bc10f7b51bab63ed96b9bbade47951de30")
     endif()
     set(TGZ_FILE "${CMAKE_BINARY_DIR}/paddle_inference.tgz")
     set(EXTRACTED_DIR "${CMAKE_BINARY_DIR}/paddle_inference_install_dir")
-    file(DOWNLOAD ${DOWNLOAD_URL} ${TGZ_FILE})
+    file(DOWNLOAD ${DOWNLOAD_URL} ${TGZ_FILE}
+         EXPECTED_HASH SHA256=${DOWNLOAD_SHA256})
```
wanghan-iapcm: A hash mismatch does not stop the configure where it happens, which undercuts the "fails fast" goal.

`file(DOWNLOAD ... EXPECTED_HASH ...)` reports a mismatch at `SEND_ERROR` severity, not `FATAL_ERROR`. CMake prints `file DOWNLOAD HASH mismatch` and then keeps executing this file, failing only at the very end with `Configuring incomplete, errors occurred!`. In this block that means everything after the download still runs: the `execute_process(... tar -xzvf ...)` on the corrupt tarball (exit code unchecked), the `file(REMOVE ${TGZ_FILE})` that destroys the evidence, and the `set(PADDLE_INFERENCE_DIR ${EXTRACTED_DIR} CACHE PATH ...)`.

The cache write is the part that bites. The natural reaction to a checksum error is to re-run `cmake` in the same build directory -- and on that second run `if(NOT DEFINED PADDLE_INFERENCE_DIR)` is now false, so the entire download block is skipped, configure reports success, and the build proceeds against a directory that does not exist. I reproduced this in a minimal project mirroring the same statement ordering: run 1 exits 1 with the mismatch but still runs the extraction and writes the cache entry; run 2 prints `Configuring done` / `Generating done` and exits 0 with the target directory absent.

This is pre-existing structure, but before this PR the failure was essentially unreachable by hash, and adding the check is exactly what makes it fire. Capturing `STATUS` and raising `message(FATAL_ERROR ...)` before the extraction, remove and cache-set would close it, and would also cover the plain network-failure case that currently produces a 0-byte file in complete silence. That subsumes the extraction exit-code point Copilot already raised, so it may be simplest to handle both together.

Worth noting for whoever picks this up: the PR body scopes retry and timeout handling to a follow-up, and I have no objection to that -- but the abort-in-place behaviour is arguably part of this change rather than the follow-up, since without it the checksum still does not reliably stop a bad build.

wanghan-iapcm (commented):
The premise checks out. I confirmed with cmake 3.31.6 that `file(DOWNLOAD)` without `STATUS` or `EXPECTED_HASH` does not fail on a bad URL -- it writes a 0-byte file, exits 0, and configure continues -- so the pre-existing behaviour really was a silent failure that surfaced later as unrelated missing-header errors, with `file(REMOVE)` having already deleted the evidence. `DOWNLOAD_SHA256` is set on every branch that can reach the download; the remaining `else()` is a `FATAL_ERROR`, so there is no path that reaches `file(DOWNLOAD)` with the variable unset. The direction is right and the change is minimal.

Two things below. Neither is a blocker on its own, but the first is the one I would want settled before merge.

On testing: there is no harness in this repo for asserting cmake configure behaviour, so the only coverage this code has ever had is the happy path -- CPU variant, network fine, bytes intact -- where a missing integrity check is invisible by construction. That is why the gap survived since the block was introduced. It is also why the incomplete abort in the second comment went unnoticed here: green CI shows the good path still works, not that a corrupt download now stops the build.

One piece of context that argues in favour of this change, from checking the objects directly: the CUDA 12.6 tarball has `Last-Modified: 2025-09-13`, roughly five months after the URL was pinned to that Paddle commit id, with no repo commit touching the URL in between. The path component is a Paddle source commit, not a digest of the artifact, and `paddle-qa.bj.bcebos.com/paddle-pipeline/` is a QA bucket with no immutability guarantee -- so the same key demonstrably gets overwritten in place. Detecting that silently-swapped content is precisely what a pinned hash buys. The flip side is that a future re-upload becomes a hard configure failure with no way to override the expected hash short of `-DPADDLE_INFERENCE_DIR`; exposing the hash as a cache variable would give an escape hatch. Fine either way, but worth being a deliberate choice.

source/CMakeLists.txt
njzjz: Addressed in 5d89c0415. The download now captures `STATUS` and raises `FATAL_ERROR` before extraction, removal, or cache mutation. Extraction likewise checks `RESULT_VARIABLE`, reports stderr, and fails before cleanup/cache mutation. I also set its working directory explicitly to `${CMAKE_BINARY_DIR}`, matching the cached extraction path.

Validation covered the failure and retry behavior directly:

- a wrong hash made two consecutive configures in the same build directory fail;
- neither attempt cached `PADDLE_INFERENCE_DIR`, and the failed artifact remained available;
- a correctly hashed non-archive failed immediately during extraction and preserved the artifact;
- the real CPU artifact completed a fresh configure successfully and extracted into the build directory.

This also addresses the extraction-status concern from the Copilot review.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

source/CMakeLists.txt
njzjz: Verified from full downloads of the published artifacts on 2026-07-30; these values were not model-derived. I streamed each URL directly through `sha256sum`:

- CUDA 11.8: 1,076,737,733 bytes, `89963f601a3469270080cca076327da1e6c5e8f0c0ee3aa5e0e0371b2b8c2cb7`
- CUDA 12.6: 1,621,970,010 bytes, `5318eb879f4930d4b08e52c5e040c0b4b96d4afae62bf57bcae031fff0396551`

Both recomputed hashes match the constants in this PR. The CPU artifact also completed a fresh local Paddle CMake configure after the follow-up fix.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

source/CMakeLists.txt
```diff
@@ -75,13 +79,38 @@ if(ENABLE_PADDLE)
...
+      EXPECTED_HASH SHA256=${DOWNLOAD_SHA256}
+      STATUS PADDLE_DOWNLOAD_STATUS)
+    list(GET PADDLE_DOWNLOAD_STATUS 0 PADDLE_DOWNLOAD_STATUS_CODE)
+    list(GET PADDLE_DOWNLOAD_STATUS 1 PADDLE_DOWNLOAD_STATUS_MESSAGE)
+    if(NOT PADDLE_DOWNLOAD_STATUS_CODE EQUAL 0)
+      message(
+        FATAL_ERROR
+          "Failed to download Paddle inference library from ${DOWNLOAD_URL} (status ${PADDLE_DOWNLOAD_STATUS_CODE}): ${PADDLE_DOWNLOAD_STATUS_MESSAGE}"
+      )
+    endif()
```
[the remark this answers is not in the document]
njzjz: Good catch on the cross-version behavior. Addressed in 6eabd5466 by recomputing the archive with `file(SHA256 ...)` and comparing it with `DOWNLOAD_SHA256` before extraction, cleanup, or cache mutation.

I validated the specific missed-status scenario by making `file(DOWNLOAD)` pass its transport and `EXPECTED_HASH` checks while deliberately setting `DOWNLOAD_SHA256` to a different value. Two consecutive configures in the same build directory both stopped at the new explicit comparison, preserved the downloaded artifact, and left `PADDLE_INFERENCE_DIR` uncached. A fresh configure using the real CPU artifact also completed successfully.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

wanghan-iapcm (approved):
Both points are resolved.

The download hardening does the thing that was easy to miss. Capturing `STATUS` alone would not have caught a checksum failure, because `STATUS` reflects the transfer rather than the `EXPECTED_HASH` comparison -- so adding the explicit [`file(SHA256)` check](https://github.com/deepmodeling/deepmd-kit/blob/6eabd546657b5e5fbe2b390aa23be599f53c5826/source/CMakeLists.txt#L99-L107) is what actually closes it, and the comment records why both are there. The ordering is right too: download, status check, hash check, extraction with `RESULT_VARIABLE`, and only then `file(REMOVE)` and the cache write. That breaks every link in the chain I described, including the one that mattered most, where a poisoned `PADDLE_INFERENCE_DIR` in the cache made the second `cmake` run skip the block entirely and configure against a corrupt tree.

The `STATUS` check also covers the separate case where a failed download leaves a zero-byte file and exits successfully.

I also liked that the validation exercised the failure modes rather than the happy path -- two consecutive failing configures in the same build directory with no cache write and the artifact preserved, plus a correctly-hashed non-archive failing at extraction.

On the hashes: thanks for the provenance. Streaming both URLs through `sha256sum` with the byte counts is exactly what I was asking for. For the record, I have not re-derived those two digests independently -- I tried fetching those artifacts earlier and the transfer throttled badly -- so the CUDA constants rest on your reported downloads, while the CPU one is the one with CI behind it. That seems a reasonable place to land given the alternative is gating this on a `Test CUDA` run that would not exercise the download anyway.

Unrelated to this change: the failing `docs/readthedocs.org` check is infrastructure, since this PR touches only `source/CMakeLists.txt`.

PR #5929, opened 2026-07-29
docs(backend): document backend support and workflows

### LINKED ISSUE #5919: docs(backend): document TF2 eager and PyTorch-Exportable backends
Parent issue: #5917 (checklist items 1–2)

## Scope

Two backends shipped since v3.1.3 have no user-facing documentation:

1. **TF2 (TensorFlow 2 eager) backend** — registered as `tf2`/`tensorflow2` in `deepmd/backend/tf2.py:32` (`name = "TensorFlow2"`, model suffix `.savedmodeltf`). Implements train (incl. multi-task, finetune, TensorBoard), freeze, compress, test via `deepmd/tf2/entrypoints/` (`train.py`, `freeze.py`, `compress.py`, `main.py`). Related commits: `73de44b1f`, `26ad0fb83`.
2. **`pt-expt` (PyTorch-Exportable) backend** — registered as `pt-expt`/`pytorch-exportable` in `deepmd/backend/pt_expt.py:35` (`name = "PyTorch-Exportable"`, suffixes `.pte`, `.pt2`, checkpoints `.pt` with dpmodel parameter naming). Implements train/multi-task/finetune/change-bias/compress via `deepmd/pt_expt/entrypoints/`. Distinct capabilities: torch.export/AOTInductor `.pt2` export (`dp freeze` with `--lower-kind nlist|graph`), LMDB datasets, pluggable neighbor lists (vesin). Related commits: `9bd1f16b1`, `2cbef365d`, `3f91293b8`, `23622361d`.

Neither backend appears in `doc/backend.md`; there are no tabs for them in `doc/train/training.md`, `doc/freeze/freeze.md`, or `doc/freeze/compress.md`. Note the CLI flag is `--pt-expt` (NOT `--pt_expt`); the Python package dir is `deepmd/pt_expt/`.

### 1. `doc/backend.md`

- Update the intro (lines 5–8): DeePMD-kit now supports seven backends — TensorFlow (v1 graph), TensorFlow 2 (eager), PyTorch, PyTorch-Exportable, JAX, Paddle, and DP.
- Add `### TensorFlow 2 {{ tensorflow_icon }}` after the existing TensorFlow section:
  - Model filename extension `.savedmodeltf` (a TF SavedModel directory); checkpoint stored as a `.tf2` checkpoint directory (verify exact spelling against `deepmd/tf2/entrypoints/freeze.py`).
  - Eager-mode backend using the TF 2 API (unlike the default TF backend, which is v1 graph mode); optional `tf.function` JIT compilation controlled by the `DP_JIT` environment variable (cross-ref `doc/env.md`).
  - Supports training (incl. multi-task and finetune), freeze, compress, and test via `dp --tf2 <command>`.
- Add `### PyTorch-Exportable {{ pytorch_icon }}` after the PyTorch section:
  - Model filename extensions `.pt2` (torch.export / AOTInductor artifact) and `.pte`; checkpoint extension `.pt` using dpmodel parameter naming (`.w`/`.b`), which is auto-routed to this backend (see `deepmd/backend/pt_expt.py` `match_filename`).
  - Select with `dp --pt-expt` (alias `--pytorch-exportable`).
  - Supports LMDB training datasets, vesin neighbor lists for Python/ASE inference (cross-ref to the inference docs), and graph-native export via `--lower-kind graph` (cross-ref `doc/model/dpa2.md` / `doc/model/dpa4.md`).
- Update **Switch the backend → Training** (line 67): mention `dp --tf2`, `dp --pt-expt`, and `dp --jax` in addition to `--tf`/`--pt`/`--pd`.
- Update **Inference** (line 71): backend detection also covers `.savedmodeltf`, `.pt2`, `.pte`.

### 2. `doc/train/training.md`

Add two tabs to the tab-set (after PyTorch, keep TF/PyTorch order stable):

- `TensorFlow 2 {{ tensorflow_icon }}`: `dp --tf2 train input.json`
- `PyTorch-Exportable {{ pytorch_icon }}`: `dp --pt-expt train input.json`

### 3. `doc/freeze/freeze.md`

Add two tabs:

- `TensorFlow 2 {{ tensorflow_icon }}`: verify the exact command from `deepmd/tf2/entrypoints/freeze.py` (expected `dp --tf2 freeze -o model.savedmodeltf` run in the training folder).
- `PyTorch-Exportable {{ pytorch_icon }}`: `dp --pt-expt freeze -o model.pt2`; mention `--lower-kind nlist|graph` (default `nlist`) and `--head` for multi-task models if supported (verify in code).

### 4. `doc/freeze/compress.md`

- Update the supported-backends note at line 1–4 to add TensorFlow 2 and PyTorch-Exportable icons.
- Add two tabs: `dp --tf2 compress` and `dp --pt-expt compress` — verify exact CLI semantics (input/output model kinds) from `deepmd/tf2/entrypoints/compress.py` and `deepmd/pt_expt/entrypoints/compress.py` before writing; TF2 accepts a `.tf2` checkpoint directory or checkpoint prefix as input (see the `-i` help text in `deepmd/main.py`).

(dpmodel/JAX compression tabs are handled by the sibling sub-issue to avoid conflicts.)

## Constraints

- All commands shown MUST be verified against the implementation (entrypoint modules listed above); do not guess.
- Follow the existing MyST tab-set syntax (`::::{tab-set}` / `:::{tab-item}`) and backend icon roles exactly as used in neighboring tabs.
- Keep the TF v1 backend description unchanged.

## Verification

- `grep -n "tf2" doc/backend.md doc/train/training.md doc/freeze/freeze.md doc/freeze/compress.md` shows the new content.
- `grep -n "pt-expt" doc/backend.md` shows the new section and consistent flag spelling.

---

Parent issue: #5917

Coding agent: opencode
opencode version: 1.18.9
Model: ustc/k3
Reasoning effort: max

### LINKED ISSUE #5920: docs(jax): document JAX training and freeze --hessian; fix stale Hessian freeze claims
Parent issue: #5917 (checklist items 3, 6, 18)

## Scope

1. **JAX training (`dp --jax train`)** was added in `20874166a` (`feat(jax): add training`, entrypoint `deepmd/jax/entrypoints/train.py`), but `doc/train/training.md` has no JAX tab and `doc/backend.md` makes no positive statement about JAX training.
2. **`dp freeze --hessian` (JAX)** was added in `44131151f` (`deepmd/main.py:354`: `(Supported backend: JAX) Add the Hessian to the frozen model output.`), but `doc/freeze/freeze.md` does not mention `--hessian` and has no JAX tab.
3. **Stale Hessian claims**: `doc/model/train-energy-hessian.md:93-94` warns "A model trained with Hessian cannot be frozen", and `doc/model/overall.md:60-61` states frozen models cannot output Hessians due to torch jit — both are now incorrect for the JAX backend.

### 1. `doc/train/training.md`

Add a `JAX {{ jax_icon }}` tab to the training tab-set:

```bash
$ dp --jax train input.json
```

Add one sentence noting the checkpoints use the `.jax` extension (per `doc/backend.md`) and that JAX training was introduced after v3.1.3. Verify the supported workflow against `deepmd/jax/entrypoints/train.py` (e.g. any limitations vs. other backends — read the entrypoint before writing).

### 2. `doc/freeze/freeze.md`

Add a `JAX {{ jax_icon }}` tab:

- Basic: `dp --jax freeze -o model.savedmodel` (confirm supported output formats — `.savedmodel` for C++ inference, `.jax`, `.xlo` — from `deepmd/jax/entrypoints/freeze.py`).
- Document `--hessian`: `dp --jax freeze -o model.savedmodel --hessian` keeps Hessian output in the frozen model (JAX backend only).

### 3. `doc/model/train-energy-hessian.md` (lines 91–110)

Rewrite the warning so it is backend-scoped, e.g.:

> [!WARNING]
> With the PyTorch backend, a model trained with Hessian cannot be frozen with Hessian output (torch jit restriction): freezing treats it as a standard energy model. With the JAX backend, Hessian models can be frozen using `dp --jax freeze --hessian`, and the frozen model retains Hessian output.

### 4. `doc/model/overall.md` (lines 60–61)

Similarly scope the torch-jit restriction to PyTorch and add that the JAX backend supports freezing models with Hessian output via `dp freeze --hessian`.

## Constraints

- Verify every command/flag against `deepmd/jax/entrypoints/freeze.py` and `deepmd/main.py` before documenting.
- Keep existing tab order; JAX tab goes after Paddle.

## Verification

- `grep -n "jax" doc/train/training.md doc/freeze/freeze.md` shows new tabs.
- `grep -n -i "hessian" doc/freeze/freeze.md` shows `--hessian` documented.
- The two stale warnings are updated and no longer blanket-claim Hessian models cannot be frozen.

---

Parent issue: #5917

Coding agent: opencode
opencode version: 1.18.9
Model: ustc/k3
Reasoning effort: max

### LINKED ISSUE #5923: docs(compress): document dpmodel and JAX model compression routes
Parent issue: #5917 (checklist item 8)

## Scope

`f143171cf` (`feat(dpmodel): add descriptor compression`) added descriptor compression for the DP (dpmodel) and JAX backends:

- `deepmd/dpmodel/entrypoints/compress.py` and `compress_common.py`;
- CLI: `dp --dp compress -i model.dp -o model-compress.dp` and `dp --jax compress -i frozen_model.hlo -o compressed_model.hlo` (see example in `deepmd/main.py:608`; the `-i` help text at `deepmd/main.py:617` lists accepted input kinds per backend: TF `.pb`, TF2 `.tf2` checkpoint dir/prefix, PyTorch `.pth`, DPModel `.dp`, JAX `.hlo`/`.jax`).

`doc/freeze/compress.md:4` still says "Supported backends: TensorFlow, PyTorch" and only shows `dp compress` / `dp --pt compress`.

### `doc/freeze/compress.md`

Read `deepmd/dpmodel/entrypoints/compress.py`, `compress_common.py`, and the JAX entrypoint (`deepmd/jax/entrypoints/compress.py`) before writing.

- Update the title line and supported-backends note (lines 1–4) to add DP {{ dpmodel_icon }} and JAX {{ jax_icon }} (icons as used elsewhere, e.g. `doc/backend.md`).
- Add two tabs to the Instructions tab-set:
  - `DP {{ dpmodel_icon }}`: `dp --dp compress -i model.dp -o model-compress.dp` (verify exact input/output kinds from the entrypoint).
  - `JAX {{ jax_icon }}`: `dp --jax compress -i frozen_model.hlo -o compressed_model.hlo` (verify `.hlo` vs `.jax` handling).
- Add a short paragraph noting the dpmodel implementation is the reference path also used by the TF2 and pt-expt compress routes (verify the import structure, e.g. `deepmd/tf2/entrypoints/compress.py`, to state this accurately).
- Note any limitations specific to dpmodel/JAX compression (e.g. supported descriptors, extrapolation/check-frequency semantics) only as implemented — do not copy TF-specific claims blindly.

## Constraints

- TF2 and pt-expt compress tabs are handled by the sibling sub-issue "document TF2 eager and pt-expt backends"; this issue only adds the DP and JAX tabs plus the backends note. If that sub-issue is not merged yet when this is implemented, still update the note line comprehensively (TF, TF2, PT, pt-expt, DP, JAX) to avoid a second touch.
- Verify every command against the entrypoints; do not guess.

## Verification

- `grep -n "dp --dp compress\|dp --jax compress" doc/freeze/compress.md` shows the new tabs.
- Supported-backends note includes DP and JAX.

---

Parent issue: #5917

Coding agent: opencode
opencode version: 1.18.9
Model: ustc/k3
Reasoning effort: max

njzjz:
## Summary

- document TensorFlow 2 and PyTorch-Exportable backend selection, training, freezing, compression, formats, and inference routing;
- fold the JAX training and Hessian-freeze documentation from #5930 into this PR;
- update model and descriptor backend icons after checking the current TF2, PyTorch-Exportable, and JAX registrations;
- distinguish JAX non-energy fitting/model availability from its current energy-only training loss support;
- document the DP and JAX compression routes requested by #5923.

Closes #5919.
Closes #5920.
Closes #5923.
Supersedes #5930.

## Validation

- `ruff format .`
- `ruff check .`
- `uvx pre-commit run mdformat --files ...` for all changed Markdown files
- `dp --dp compress --help`
- `dp --jax compress --help`
- `git diff --check`

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Documentation**
  * Expanded backend guidance for training, freezing, compression, conversion, and inference, including TensorFlow 2, PyTorch-Exportable, JAX, Paddle, and DP.
  * Added backend selection commands, workflow examples, multi-task checkpoint selection, and Hessian-enabled JAX freezing.
  * Documented backend-specific file formats, export artifacts, ABI selection, optional TensorFlow 2 JIT compilation, and feature limitations.
  * Updated model pages with corrected backend support, spin and graph-native inference guidance, and standardized command syntax.


doc/model/pairtab.md
```diff
@@ -1,7 +1,12 @@
-# Interpolation or combination with a pairwise potential {{ tensorflow_icon }} {{ pytorch_icon }} {{ dpmodel_icon }}
+# Interpolation or combination with a pairwise potential {{ tensorflow_icon }} {{ pytorch_icon }} {{ jax_icon }} {{ paddle_icon }} {{ dpmodel_icon }}
 
 > [!NOTE]
-> **Supported backends**: TensorFlow {{ tensorflow_icon }}, PyTorch {{ pytorch_icon }}, DP {{ dpmodel_icon }}
+> **Short-range interpolation backends**: TensorFlow and TensorFlow 2
+> {{ tensorflow_icon }}, PyTorch and PyTorch-Exportable {{ pytorch_icon }}, JAX
+> {{ jax_icon }}, Paddle {{ paddle_icon }}, DP {{ dpmodel_icon }}
```
wanghan-iapcm: Paddle does not belong on this line, or in the page title.

There is no pairtab/ZBL implementation anywhere under `deepmd/pd/` at this head. The whole Paddle atomic-model tree is `base_atomic_model.py`, `dp_atomic_model.py`, `energy_atomic_model.py` -- no `pairtab_atomic_model.py`, no `dp_zbl_model.py`, and a recursive tree search for `zbl` or `pairtab` under `deepmd/pd/` returns nothing. Every other backend you list here does have both files. `get_model` in [`deepmd/pd/model/model/__init__.py`](https://github.com/deepmodeling/deepmd-kit/blob/bd78747e5fb790fdb9c06e96d24f0d4be8da3ff7/deepmd/pd/model/model/__init__.py) special-cases only `"standard"` and otherwise dispatches to `BaseModel.get_class_by_type(model_type)`, so `"type": "zbl"` fails to resolve on Paddle.

The history is consistent with that: the narrow list this PR replaces was set deliberately in #3435 (PyTorch + DP only), and `source/tests/pd/model/test_permutation.py` still carries `@unittest.skip("Skip for not implemented yet")` on `TestEnergyModelZBL` from the original Paddle port. So a user following the Paddle icon here hits a hard failure.

The JAX and TF2 half of the broadening is fine -- `deepmd/jax/model/dp_zbl_model.py` and `deepmd/tf2/model/dp_zbl_model.py` both exist.

Separately, on the `Linear-combination backends` line just below: splitting the old single note dropped DP, but [`deepmd/dpmodel/model/dp_linear_model.py#L28-L32`](https://github.com/deepmodeling/deepmd-kit/blob/bd78747e5fb790fdb9c06e96d24f0d4be8da3ff7/deepmd/dpmodel/model/dp_linear_model.py#L28-L32) carries `@BaseModel.register("linear_ener")` and `@BaseModel.register("linear")` on a full `LinearEnergyModel`, and the pre-PR note did list DP. Same omission in [linear.md](https://github.com/deepmodeling/deepmd-kit/blob/bd78747e5fb790fdb9c06e96d24f0d4be8da3ff7/doc/model/linear.md). I scored this one lower than the Paddle claim since it understates rather than overstates support, but it looks like collateral from the split rather than a decision.

doc/freeze/compress.md
```diff
@@ -70,19 +72,66 @@ dp compress -i graph.pb -o graph-compress.pb
...
+dp --pt-expt compress -i model.pte -o model-compress.pte
+```
+
+Use matching `.pte` or `.pt2` suffixes to preserve the exported model form.
+:::
+
+:::{tab-item} JAX {{ jax_icon }}
+
+```bash
+dp --jax compress -i frozen_model.hlo -o compressed_model.hlo
```
wanghan-iapcm: This example is the case that fails.

`deserialize_to_file` calls `_check_compressed_hlo_exportable(data)` on the `.hlo` write path at [`deepmd/jax/utils/serialization.py#L254-L256`](https://github.com/deepmodeling/deepmd-kit/blob/bd78747e5fb790fdb9c06e96d24f0d4be8da3ff7/deepmd/jax/utils/serialization.py#L254-L256), and that helper hard-raises for a compressed `se_e2_a`/`se_a`/`dpa1`/`se_atten` descriptor with `type_one_side=False`, telling the user to "write a `.jax` checkpoint instead" ([L175-L185](https://github.com/deepmodeling/deepmd-kit/blob/bd78747e5fb790fdb9c06e96d24f0d4be8da3ff7/deepmd/jax/utils/serialization.py#L175-L185)). `type_one_side` defaults to `False` for `se_e2_a`, so copy-pasting this line for a default-configuration model -- the most common case, and the descriptor used in the examples elsewhere in this file -- deterministically raises `ValueError`.

The prose above is fine as a statement about accepted suffixes; it is the headline example that misleads, and it does so in the direction of "use the same suffix for input and output", which is exactly the advice that triggers the error.

Either make the example `-i frozen_model.jax -o compressed_model.jax`, or keep the `.hlo` example and add the `type_one_side=True` precondition next to it.

wanghan-iapcm (commented):
This is a large and mostly accurate sweep -- I checked the great majority of the new support claims against backend registration at head and they hold up, including the ones that are easy to get wrong. Specifically confirmed correct: every `dp --tf2` / `dp --pt-expt` / `dp --jax` flag against the aliases in `deepmd/backend/*.py`; the `dp --pt_expt` to `dp --pt-expt` and `.xlo` to `.hlo` corrections (both were real doc bugs); all the frozen-model suffixes against each backend's `suffixes`; `dp --jax freeze --hessian`; the JAX energy-loss-only and no-`--init-frz-model` caveats; the TF2 and Paddle additions on dpa2/dpa3/se_e3_tebd; and the DPA4 note, where JAX really does have `deepmd/jax/descriptor/dpa4.py` but no `jax/model/dpa4_model.py`, so "descriptor-only" is exactly right.

Two claims do not hold. Both are inline.

One smaller thing I did not raise inline. The Hessian rewording in [overall.md](https://github.com/deepmodeling/deepmd-kit/blob/bd78747e5fb790fdb9c06e96d24f0d4be8da3ff7/doc/model/overall.md) and [train-energy-hessian.md](https://github.com/deepmodeling/deepmd-kit/blob/bd78747e5fb790fdb9c06e96d24f0d4be8da3ff7/doc/model/train-energy-hessian.md) is not wrong -- it names `dp --pt freeze` and the TorchScript route specifically -- but `hessian_mode` is honored only at [`deepmd/pt/model/model/__init__.py#L371-L374`](https://github.com/deepmodeling/deepmd-kit/blob/bd78747e5fb790fdb9c06e96d24f0d4be8da3ff7/deepmd/pt/model/model/__init__.py#L371-L374), and neither `pt_expt/model/get_model.py` nor `dpmodel/model/model.py` mention it, so PyTorch-Exportable cannot construct a Hessian model at all. Since this PR establishes that `{{ pytorch_icon }}` covers both, the page is now silent on pt_expt where it used to be blanket. Worth a clause if you touch these lines again; not worth a round trip on its own.

OutisLi (changes requested):
The current DPA4 native-spin multi-rank guidance contradicts both the implementation and itself. The graph-route/native-spin section of doc/model/dpa4.md says that native-spin .pt2 archives carry no with-comm artifact, have has_comm_artifact=false, and are single-rank only. At this head, _needs_with_comm_artifact explicitly admits a graph-lowered NativeSpinModelKind whose descriptor communicates across ranks; test_native_spin_needs_with_comm_on_the_graph_lower_only asserts that behavior; and DeepSpinPTExpt loads and dispatches forward_lower_with_comm.pt2 for multi-rank graph native spin. The same page also correctly states near the earlier multi-rank overview and the Limitations section that pt_expt native spin supports multi-rank. Please remove or update the obsolete single-rank paragraphs so this page documents one contract that matches the current export and C++ runtime.

OutisLi (changes requested):
The documented PyTorch-Exportable compression workflow does not actually put a compressed model into the executable artifact for the ordinary .pte example. Except for the special fused DPA1 graph .pt2 branch, enable_compression passes the original uncompressed model dictionary to deserialize_to_file and stores the compressed dictionary only as model_json_override. DeepEval reconstructs that model.json object for metadata and introspection, but energy inference runs exported_module or _pt2_runner, so the tabulated model is not executed. I reproduced the documented se_e2_a-style .pte workflow locally: the input and output exported graphs were identical (711 nodes each), the output executable contained no tabulate_fusion operator, while only model.json contained the compression state. Therefore model-compress.pte provides no compression speed or memory benefit even though this page and backend.md present pt_expt compression as generally supported. Please either export the genuinely compressed executable or document the narrow supported case instead of advertising the ordinary .pte workflow as compressed.

doc/backend.md
```diff
@@ -23,17 +42,50 @@ DeePMD-kit does not use the TensorFlow v2 API but uses the TensorFlow v1 API (`t
...
+(`lower_input_kind: graph`). All variants are loaded for inference by the
+PyTorch-Exportable runtime, which reads this metadata to select the correct
+input path. The `--lower-kind` option controls only the PyTorch-Exportable
+freeze route; see the [DPA4 export documentation](model/dpa4.md#freeze-to-pt2)
+for the separate PyTorch route.
+
 ### JAX {{ jax_icon }}
 
-- Model filename extension: `.xlo`, `.savedmodel`
+- Model filename extensions: `.hlo`, `.jax`, `.savedmodel`
```
OutisLi: `.jax` is accepted by serialization/freeze, but it is not a usable model file for the normal inference/test route described on this page. `JAXBackend` detects the suffix, while `deepmd/jax/infer/deep_eval.py::DeepEval.__init__` only handles `.hlo` and `.savedmodel`; `.jax` reaches `ValueError("Unsupported file extension")`. I also reproduced this with a minimal real model: `serialize_from_file()` round-trips successfully, but `DeepPot(path)` fails at construction. Thus a user following `dp --jax freeze ... -o model.jax` gets an artifact that cannot be consumed by `dp test`/`DeepPot`. Please distinguish `.jax` as a lossless checkpoint/serialization (and JAX-MD) format from DeepEval model formats, or add `.jax` inference support before advertising it as a model format.

OutisLi (changes requested):
The newly documented JAX `.jax` frozen-model route is not usable by the normal DeePMD inference/test interface. Please correct the format contract or implement the missing loader support.

doc/freeze/compress.md
```diff
@@ -70,19 +72,66 @@ dp compress -i graph.pb -o graph-compress.pb
 ```
 :::
 
+:::{tab-item} TensorFlow 2 {{ tensorflow_icon }}
+
+```bash
+dp --tf2 compress -i model.ckpt.tf2 -o model-compress.savedmodeltf
```
OutisLi, line 78: This advertised command fails for the normal default `se_e2_a` configuration. After input argcheck normalization, `type_one_side` defaults to `False`. The two-sided compressed branch in `deepmd/dpmodel/descriptor/se_e2_a.py::_call_compressed` derives per-type slice boundaries with `int(xp.sum(...))`; during TF2 SavedModel export this runs under `tf.function` with symbolic atom types and raises `TypeError: cannot convert a TensorFlow graph tensor to a Python scalar`. I reproduced this on the current head with a controlled pair: `type_one_side=True` exports successfully, while `type_one_side=False` fails at that exact conversion. Thus this is not an unusual model corner case: the normalized default model cannot follow the documented TF2 compression workflow. Please make the two-sided compressed path graph-safe, or accurately restrict the documented support until it is implemented.

OutisLi (changes requested):
The newly documented TensorFlow 2 compression workflow fails for the normalized default `se_e2_a` configuration. The implementation or the documented support contract needs to be corrected.

doc/freeze/freeze.md
```diff
@@ -8,17 +8,30 @@ To freeze a model, typically one does
...
 ```
 
 in the folder where the model is trained. The output model is called `model.pb`.
 The idea and part of our code are from [Morgan](https://blog.metaflow.fr/tensorflow-how-to-freeze-a-model-and-serve-it-with-a-python-api-d4f3596b3adc).
 :::
 
+:::{tab-item} TensorFlow 2 {{ tensorflow_icon }}
+
+```bash
+dp --tf2 freeze -c model.ckpt -o model.savedmodeltf
```
OutisLi, line 21: This ordinary freeze workflow fails even before compression for the normalized default `se_e2_a` model. Input argcheck sets `type_one_side=False`; the uncompressed two-sided branch in `deepmd/dpmodel/descriptor/se_e2_a.py` computes per-center-type slice boundaries using `int(xp.sum(...))`. SavedModel export traces that branch under `tf.function`, where the atom types are symbolic, and raises `TypeError: cannot convert a TensorFlow graph tensor to a Python scalar`. I verified a controlled pair on the current head: an uncompressed `type_one_side=True` model exports, while the otherwise identical `type_one_side=False` model fails. Since the freeze entrypoint restores the checkpoint and then calls this same `deserialize_to_file()` export path, a normally normalized training checkpoint cannot follow the command documented here. Please make the two-sided path graph-safe, or accurately qualify TF2 freeze support until it is usable.

OutisLi (changes requested):
The newly documented TensorFlow 2 freeze workflow also fails for an uncompressed, normalized-default `se_e2_a` model. This is broader than the compression failure reported separately.

doc/model/train-fitting-dos.md
```diff
@@ -1,7 +1,14 @@
 # Fit electronic density of states (DOS) {{ tensorflow_icon }} {{ pytorch_icon }} {{ jax_icon }} {{ dpmodel_icon }}
 
 > [!NOTE]
-> **Supported backends**: TensorFlow {{ tensorflow_icon }}, PyTorch {{ pytorch_icon }}, JAX {{ jax_icon }}, DP {{ dpmodel_icon }}
+> **Supported backends**: TensorFlow and TensorFlow 2
```
OutisLi, line 4: TensorFlow 2 DOS is not usable through the frozen-model testing/inference workflow advertised by this PR. I verified this independently of the previously reported `se_e2_a` issue by using `type_one_side=True`: a TF2 DOS model with `numb_dos=2` exports successfully to `.savedmodeltf`, but `DeepDOS(...).eval()` fails with `ValueError: cannot reshape array of size 2 into shape (1,0)`. The TF2 SavedModel exporter does not persist a `get_numb_dos` method, and `deepmd/tf2/infer/deep_eval.py::get_numb_dos()` is hard-coded to return `0`, which `DeepDOS` then uses for its reshape. Please either preserve this metadata and make frozen TF2 DOS models work with `dp test`/`DeepDOS`, or qualify the TF2 support statement so it does not imply a working end-to-end workflow.

OutisLi (changes requested):
The newly documented TensorFlow 2 DOS workflow cannot currently be completed after freezing: the SavedModel exports, but `dp test`/`DeepDOS` cannot recover its DOS width.

doc/freeze/compress.md
njzjz: Addressed in 21363e267: changed the primary JAX compression example to a lossless .jax input/output pair and documented the type_one_side requirement for compressed .hlo export.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/freeze/compress.md
njzjz: Addressed in 21363e267: the TensorFlow 2 compression workflow now states that se_e2_a/se_a SavedModel export requires type_one_side: true and that the normalized false default is not graph-traceable.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/model/pairtab.md
njzjz: Addressed in 21363e267: removed Paddle from the pairtab/ZBL support claims and restored DP on both the linear-combination note and linear-model page.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/model/train-fitting-dos.md
njzjz: Addressed in 21363e267: TensorFlow 2 DOS is now described as trainable, but frozen .savedmodeltf inference through dp test or DeepDOS is explicitly unsupported until numb_dos is preserved.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/freeze/freeze.md
njzjz: Addressed in 21363e267: the TensorFlow 2 freeze instructions now carry the same explicit type_one_side: true precondition for se_e2_a/se_a SavedModel export.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/backend.md
njzjz: Addressed in 21363e267: .hlo and .savedmodel are now identified as DeepEval model formats, while .jax is documented as lossless checkpoint/serialization for round-tripping and JAX-MD, not dp test or DeepPot.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

njzjz:
Addressed the two changes-requested reviews and the Hessian non-blocking note in 21363e267:

- DPA4 native-spin documentation now matches the implemented graph multi-rank contract: with-comm artifact selection, per-block border_op feature exchange, and LAMMPS sp ghost-spin forwarding. It also distinguishes the single-rank ZBL bridging limitation.
- PyTorch-Exportable compression is no longer advertised generally. The docs now restrict executable compression to fused-CUDA-eligible graph DPA1/se_atten strip .pt2 exports and state that other outputs update model.json without replacing the executable inference graph.
- The Hessian pages now state that PyTorch-Exportable cannot construct a Hessian model for freezing or inference.
- The six line-specific backend workflow findings were also updated, replied to inline, and resolved.

Validation:
- ruff format .: 1832 files unchanged
- ruff check .: passed
- git diff --check: passed
- pytest source/tests/common/test_doc_train_input.py -q: 6 passed
- Full Sphinx parsing was unavailable in this environment because the configured exhale extension is not installed.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/freeze/freeze.md
```diff
@@ -15,6 +15,19 @@ in the folder where the model is trained. The output model is called `model.pb`.
...
+
+```bash
+$ dp --tf2 freeze -o model.savedmodeltf
+```
+
+Run the command in the training folder. By default, it reads the
+`model.ckpt.tf2` checkpoint directory and writes the TensorFlow SavedModel to
+`model.savedmodeltf`. Use `-c` to select another checkpoint directory or
+checkpoint prefix. For a multi-task checkpoint, select a branch with
+`--head CHOSEN_BRANCH`.
```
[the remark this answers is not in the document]
njzjz: Addressed on the current branch: the TensorFlow 2 freeze section now says that -c names the prefix whose .tf2 path is checked, while omitting -c defaults to the current directory.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/freeze/compress.md
```diff
@@ -1,7 +1,8 @@
 # Compress a model {{ tensorflow_icon }} {{ pytorch_icon }}
 
 > [!NOTE]
-> **Supported backends**: TensorFlow {{ tensorflow_icon }}, PyTorch {{ pytorch_icon }}
+> **Supported backends**: TensorFlow and TensorFlow 2 {{ tensorflow_icon }},
+> PyTorch and PyTorch-Exportable {{ pytorch_icon }}
```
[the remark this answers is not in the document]
njzjz: Addressed on the current branch: the page title and backend note include JAX and DP, and the note is phrased as Backends covered below.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/freeze/freeze.md
```diff
@@ -33,6 +46,19 @@ $ dp --pt freeze -o model_branch1.pth --head CHOSEN_BRANCH
...
+
+```bash
+$ dp --pt-expt freeze -c model.ckpt.pt -o model
+```
+
+The backend writes `.pte` for the dense neighbor-list lower form and `.pt2` for
+the graph lower form. A suffixless output lets DeePMD-kit select the matching
+extension. Use `--lower-kind nlist` or `--lower-kind graph` to request a form;
+graph-capable DPA models may select the graph form automatically. In multi-task
+mode, select a model branch with `--head CHOSEN_BRANCH`.
```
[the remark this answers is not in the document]
njzjz: Addressed on the current branch: the PyTorch-Exportable freeze section now states that graph requires an eligible model and that a graph-capable DPA model may override a requested nlist lower with graph while warning.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/model/train-fitting-dos.md
```diff
@@ -90,13 +99,27 @@ dp --tf train input.json
...
 
 ```bash
 dp --pt train input.json
 ```
 :::
 
+:::{tab-item} PyTorch-Exportable {{ pytorch_icon }}
+
+```bash
+dp --pt-expt train input.json
```
OutisLi: PyTorch-Exportable DOS is not usable through the exported-model testing/inference workflow. I verified this independently of the `se_e2_a` default issue by using `type_one_side=True`: a pt_expt DOS model configured with `numb_dos=2` reports `get_numb_dos()==0`, exports successfully to `.pte`, and then `DeepDOS(...).eval()` fails with `ValueError: cannot reshape array of size 2 into shape (1,0)`. `deepmd/pt_expt/model/dos_model.py::DOSModel` does not override `get_numb_dos()` as the dpmodel DOS class does, and `deepmd/pt_expt/infer/deep_eval.py::get_numb_dos()` is also hard-coded to return `0`. Please make the exported pt_expt DOS model preserve/report its output width so the newly documented workflow works, or qualify the support statement.

OutisLi (changes requested):
The newly documented PyTorch-Exportable DOS workflow cannot be completed after export: the model/evaluator report a DOS width of zero, so `dp test`/`DeepDOS` fails.

doc/freeze/compress.md
```diff
@@ -70,19 +72,78 @@ dp compress -i graph.pb -o graph-compress.pb
...
+:::{tab-item} TensorFlow 2 {{ tensorflow_icon }}
+
+```bash
+dp --tf2 compress -i model.ckpt.tf2 -o model-compress.savedmodeltf
+```
+
+TensorFlow 2 compression reads a `.tf2` training checkpoint directory or a
+checkpoint prefix and writes a compressed `.savedmodeltf` model. For
+`se_e2_a`/`se_a`, SavedModel export currently requires
+`type_one_side: true`; the normalized default, `false`, is not graph-traceable.
```
njzjz: It's better to write it on the descriptor page. Btw, JAX and other backends may have the same problem.

doc/freeze/compress.md
```diff
@@ -70,19 +72,78 @@ dp compress -i graph.pb -o graph-compress.pb
...
+
+```bash
+dp --pt-expt compress -i dpa1-graph.pt2 -o dpa1-graph-compress.pt2
+```
+
+This executable compression path is currently limited to graph-lowered
+DPA1/`se_atten` strip models that are eligible for the fused CUDA table
+operator: `attn_layer: 0`, no excluded type pairs, float32 descriptor
+statistics and tables, `neuron[-1] <= 256`, and
+`axis_neuron <= min(16, neuron[-1])`. Their `.pt2` export embeds the tabulated
```
njzjz: It's better to document it on the specific descriptor page.

doc/freeze/compress.md
```diff
@@ -70,19 +72,78 @@ dp compress -i graph.pb -o graph-compress.pb
...
+
+```bash
+dp --jax compress -i frozen_model.jax -o compressed_model.jax
+```
+
+JAX compression accepts `.jax` and `.hlo` inputs. Use `.jax` for the general,
+lossless compressed serialization path. A compressed `.hlo` can be written
+only when the descriptor is StableHLO-exportable; compressed `se_e2_a`, `se_a`,
+`dpa1`, and `se_atten` descriptors with `type_one_side: false` must remain
+`.jax` or be rebuilt with `type_one_side: true`.
```
njzjz: It's better to document it on the specific descriptor page.

doc/freeze/compress.md
```diff
@@ -70,19 +72,78 @@ dp compress -i graph.pb -o graph-compress.pb
...
+DP compression accepts native `.dp` and `.yaml` models.
+:::
+
 ::::
 
-where `-i` gives the original frozen model, `-o` gives the compressed model. Several other command line options can be passed to `dp compress`, which can be checked with
+where `-i` gives the original frozen model, `-o` gives the compressed model.
+The DP, JAX, and TensorFlow 2 entrypoints share the native-model compression
+helpers for resolving the minimum neighbor distance and tabulating the
+descriptor's embedding networks. The limited PyTorch-Exportable graph-DPA1
```
njzjz: It's better to document it on the specific descriptor page.

doc/freeze/freeze.md
```diff
@@ -8,17 +8,32 @@ To freeze a model, typically one does
...
+```bash
+dp --tf2 freeze -c model.ckpt -o model.savedmodeltf
+```
+
+When `-c` names a checkpoint prefix, the backend also checks the corresponding
+path with `.tf2` appended, so the example reads `model.ckpt.tf2` and writes the
+TensorFlow SavedModel to `model.savedmodeltf`. If `-c` is omitted, it defaults
+to the current directory. For a multi-task checkpoint, select a branch with
+`--head CHOSEN_BRANCH`. SavedModel export currently requires graph-traceable
+model code; an `se_e2_a`/`se_a` descriptor must set `type_one_side: true`
```
njzjz: It's better to document it on the specific descriptor page.

doc/model/train-fitting-dos.md
njzjz: It seems to be better to remove it at this time, and add an issue to track it if there is not.

doc/train/training.md
```diff
@@ -13,30 +13,55 @@ After switching to that directory, the training can be invoked by
...
+```
+:::
+
+:::{tab-item} JAX {{ jax_icon }}
+
+```bash
+dp --jax train input.json
 ```
+
+JAX training, introduced after DeePMD-kit v3.1.3, writes checkpoints with the
```
njzjz:
```suggestion
JAX training, introduced in v3.2.0, writes checkpoints with the
```

doc/backend.md
```diff
@@ -15,6 +20,24 @@ In the documentation, TensorFlow {{ tensorflow_icon }}, PyTorch {{ pytorch_icon
...
+### TensorFlow 2 {{ tensorflow_icon }}
+
+- Model filename extension: `.savedmodeltf`
+- Checkpoint directory extension: `.tf2`
+
+The TensorFlow 2 backend uses the TensorFlow v2 eager API. Select it with
+`dp --tf2` (alias `dp --tensorflow2`). It supports training, including
+multi-task training and fine-tuning. Freezing, compression, and testing use a
+`.savedmodeltf` export and therefore require a graph-traceable model. In
+particular, `se_e2_a`/`se_a` descriptors currently require
```
njzjz: It's better to document it on the specific descriptor page.

doc/backend.md
```diff
@@ -15,6 +20,24 @@ In the documentation, TensorFlow {{ tensorflow_icon }}, PyTorch {{ pytorch_icon
...
+`type_one_side: true`; the normalized default, `false`, cannot be exported.
+Frozen TensorFlow 2 DOS models also cannot yet be tested because the SavedModel
+does not preserve `numb_dos`. Training stores checkpoints in a directory named
```
njzjz: It's better to document it on the specific descriptor page.

doc/backend.md
```diff
@@ -15,6 +20,24 @@ In the documentation, TensorFlow {{ tensorflow_icon }}, PyTorch {{ pytorch_icon
...
+after the `save_ckpt` prefix with `.tf2` appended, such as `model.ckpt.tf2`.
+
+Setting [`DP_JIT`](env.md#envvar-DP_JIT) enables optional `tf.function` JIT
```
njzjz: There is an option in the training parameter that has the same effect.

doc/backend.md
```diff
@@ -23,17 +46,56 @@ DeePMD-kit does not use the TensorFlow v2 API but uses the TensorFlow v1 API (`t
...
+### PyTorch-Exportable {{ pytorch_icon }}
+
+- Model filename extensions: `.pte`, `.pt2`
+- Checkpoint filename extension: `.pt`
+
+Select this backend with `dp --pt-expt` (alias
+`dp --pytorch-exportable`). It uses PyTorch with the backend-independent model
+implementation and supports training, including multi-task training and
+fine-tuning, freezing, change-bias, and testing. Executable compression is
+currently limited to fused-CUDA-eligible, graph-lowered DPA1/`se_atten` strip
```
njzjz: It's better to document it on the specific descriptor page.

doc/backend.md
```diff
@@ -15,6 +20,24 @@ In the documentation, TensorFlow {{ tensorflow_icon }}, PyTorch {{ pytorch_icon
...
+particular, `se_e2_a`/`se_a` descriptors currently require
+`type_one_side: true`; the normalized default, `false`, cannot be exported.
+Frozen TensorFlow 2 DOS models also cannot yet be tested because the SavedModel
+does not preserve `numb_dos`. Training stores checkpoints in a directory named
+after the `save_ckpt` prefix with `.tf2` appended, such as `model.ckpt.tf2`.
+
+Setting [`DP_JIT`](env.md#envvar-DP_JIT) enables optional `tf.function` JIT
+compilation; depending on the workload, this may improve or reduce performance.
+
 ### PyTorch {{ pytorch_icon }}
```
njzjz: Please document that this backend will be deprecated and replaced by PyTorch-Exportable, due to the deprecation of TorchScript. I suggest renaming the backend from `PyTorch` to `PyTorch-TorchScript`. (also change other documentation)

doc/backend.md
```diff
@@ -23,17 +46,56 @@ DeePMD-kit does not use the TensorFlow v2 API but uses the TensorFlow v1 API (`t
...
+the exported inference graph with the tabulated model. Training can read LMDB
+datasets, and Python inference can use the optional vesin neighbor-list
+implementation.
+
+Freezing exports a `torch.export` model. The dense neighbor-list lower form
+normally uses `.pte`, while the graph lower form uses an AOTInductor `.pt2`
+package. Use `--lower-kind graph` to request graph-native export for an eligible
+model; graph-capable DPA models may select that form automatically. The `.pt`
+checkpoint format uses DP-model parameter names ending in `.w` and `.b`, which
+allows DeePMD-kit to distinguish it from a regular PyTorch checkpoint.
```
njzjz: What is "regular"?

doc/model/train-fitting-dos.md
njzjz: Addressed in f9c46aa23: removed PyTorch-Exportable from the DOS supported-backend list and training tabs, documented the missing output-width limitation, and opened #5949 to track the implementation fix.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/freeze/compress.md
njzjz: Addressed in f9c46aa23: moved the TensorFlow 2 `se_e2_a`/`se_a` SavedModel export restriction to the descriptor page. That page also records the corresponding JAX `type_one_side` restriction, while the compression overview now stays descriptor-agnostic.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/freeze/compress.md
njzjz: Addressed in f9c46aa23: moved the PyTorch-Exportable fused-CUDA eligibility rules and executable-artifact behavior to the DPA-1 descriptor compression section. The workflow page now links to that canonical section.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/freeze/compress.md
njzjz: Addressed in f9c46aa23: moved the JAX compressed `.jax`/StableHLO `.hlo` restrictions to the `se_e2_a` and DPA-1 descriptor pages. The generic compression page now only describes the available formats.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/freeze/compress.md
njzjz: Addressed in f9c46aa23: moved the PyTorch-Exportable DPA-1 table-stride and minimum-neighbor-distance behavior to the DPA-1 compression section. The workflow page retains only the backend-independent training-script fallback.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/freeze/freeze.md
njzjz: Addressed in f9c46aa23: moved the TensorFlow 2 `type_one_side` export restriction to the `se_e2_a` descriptor page. The freeze overview now states only the general graph-traceability requirement.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/train/training.md
njzjz: Applied in f9c46aa23: the JAX training text now says it was introduced in DeePMD-kit v3.2.0.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/backend.md
njzjz: Addressed in f9c46aa23: moved the TensorFlow 2 `se_e2_a`/`se_a` export limitation from the backend overview to the descriptor page, alongside the JAX backend-specific restriction.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/backend.md
njzjz: Addressed in f9c46aa23: removed the DOS limitation from the backend overview and kept the TensorFlow 2 and PyTorch-Exportable output-width limitations on the DOS model page, with #5949 tracking the latter.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/backend.md
njzjz: Addressed in f9c46aa23: documented `training.enable_compile` as the training configuration option and clarified that `DP_JIT` supplies the same model-level default and also applies to SavedModel export.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/backend.md
njzjz: Addressed in f9c46aa23: moved the PyTorch-Exportable executable-compression limitations to the DPA-1 descriptor page. The backend overview now directs readers to descriptor-specific compression requirements.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/backend.md
njzjz: Addressed in f9c46aa23: renamed the legacy backend to PyTorch-TorchScript across the backend reference and backend-specific documentation, and documented that it will be deprecated and replaced by PyTorch-Exportable because TorchScript is deprecated.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

doc/backend.md
njzjz: Addressed in f9c46aa23: replaced “regular PyTorch checkpoint” with the precise PyTorch-TorchScript checkpoint contract and documented its `.matrix`/`.bias` parameter suffixes versus PyTorch-Exportable `.w`/`.b` suffixes.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

wanghan-iapcm (dismissed):
Both points are addressed, and I checked the files at `3bc4ff06a` rather than taking the replies at their word.

On the pairtab page: Paddle is gone from the title and from both notes, and DP is back on the linear-combination note and on `linear.md`. Since the fix introduced a claim I had not checked in the first round, I verified it too — the new "PyTorch-Exportable" on both lines holds, because `deepmd/pt_expt/model/` carries both `dp_zbl_model.py` and `dp_linear_model.py`. The asymmetry between the two notes is also right rather than an oversight: JAX and TF2 appear on the short-range interpolation line, where each has `pairtab_atomic_model.py` and `dp_zbl_model.py`, but not on the linear-combination line, where neither has a `dp_linear_model.py` — only `linear_atomic_model.py`. And Paddle genuinely has none of them.

On the compression page: switching the headline example to `-i frozen_model.jax -o compressed_model.jax` removes the case that would have raised. The prose now points at `.jax` as the general lossless path and defers the StableHLO question to the descriptor pages; I confirmed that deferral is not a dead pointer, since `train-se-e2-a.md` and `train-se-atten.md` both state the `type_one_side` requirement for `.hlo` export. One small thing for the record only: your reply said the `type_one_side` requirement was documented for compressed `.hlo` export, and in `compress.md` itself it is referred out rather than stated. That is the better placement given the requirement is descriptor-specific, so I am not asking for a change — just noting the wording difference in case you expected it inline.

The Hessian point from my earlier review body is resolved as well. `train-energy-hessian.md` now says "Supported backends: PyTorch-TorchScript, JAX" and labels the tabs the same way, instead of the blanket PyTorch icon. That matches where `hessian_mode` is actually honored — `deepmd/pt/`, `deepmd/jax/`, and `deepmd/dpmodel/model/make_hessian_model.py`, with nothing on the pt_expt side — so the page no longer implies coverage that does not exist.

wanghan-iapcm (dismissed):
Both points are addressed, and I checked the files at `3bc4ff06a` rather than taking the replies at their word.

On the pairtab page: Paddle is gone from the title and from both notes, and DP is back on the linear-combination note and on `linear.md`. Since the fix introduced a claim I had not checked in the first round, I verified it too — the new "PyTorch-Exportable" on both lines holds, because `deepmd/pt_expt/model/` carries both `dp_zbl_model.py` and `dp_linear_model.py`. The asymmetry between the two notes is also right rather than an oversight: JAX and TF2 appear on the short-range interpolation line, where each has `pairtab_atomic_model.py` and `dp_zbl_model.py`, but not on the linear-combination line, where neither has a `dp_linear_model.py` — only `linear_atomic_model.py`. And Paddle genuinely has none of them.

On the compression page: switching the headline example to `-i frozen_model.jax -o compressed_model.jax` removes the case that would have raised. The prose now points at `.jax` as the general lossless path and defers the StableHLO question to the descriptor pages; I confirmed that deferral is not a dead pointer, since `train-se-e2-a.md` and `train-se-atten.md` both state the `type_one_side` requirement for `.hlo` export. One small thing for the record only: your reply said the `type_one_side` requirement was documented for compressed `.hlo` export, and in `compress.md` itself it is referred out rather than stated. That is the better placement given the requirement is descriptor-specific, so I am not asking for a change — just noting the wording difference in case you expected it inline.

The Hessian point from my earlier review body is resolved as well. `train-energy-hessian.md` now says "Supported backends: PyTorch-TorchScript, JAX" and labels the tabs the same way, instead of the blanket PyTorch icon. That matches where `hessian_mode` is actually honored — `deepmd/pt/`, `deepmd/jax/`, and `deepmd/dpmodel/model/make_hessian_model.py`, with nothing on the pt_expt side — so the page no longer implies coverage that does not exist.

wanghan-iapcm (approved):
Both points are addressed, and I checked the files at `3bc4ff06a` rather than taking the replies at their word.

On the pairtab page: Paddle is gone from the title and from both notes, and DP is back on the linear-combination note and on `linear.md`. Since the fix introduced a claim I had not checked in the first round, I verified it too — the new "PyTorch-Exportable" on both lines holds, because `deepmd/pt_expt/model/` carries both `dp_zbl_model.py` and `dp_linear_model.py`. The asymmetry between the two notes is also right rather than an oversight: JAX and TF2 appear on the short-range interpolation line, where each has `pairtab_atomic_model.py` and `dp_zbl_model.py`, but not on the linear-combination line, where neither has a `dp_linear_model.py` — only `linear_atomic_model.py`. And Paddle genuinely has none of them.

On the compression page: switching the headline example to `-i frozen_model.jax -o compressed_model.jax` removes the case that would have raised. The prose now points at `.jax` as the general lossless path and defers the StableHLO question to the descriptor pages; I confirmed that deferral is not a dead pointer, since `train-se-e2-a.md` and `train-se-atten.md` both state the `type_one_side` requirement for `.hlo` export. One small thing for the record only: your reply said the `type_one_side` requirement was documented for compressed `.hlo` export, and in `compress.md` itself it is referred out rather than stated. That is the better placement given the requirement is descriptor-specific, so I am not asking for a change — just noting the wording difference in case you expected it inline.

The Hessian point from my earlier review body is resolved as well. `train-energy-hessian.md` now says "Supported backends: PyTorch-TorchScript, JAX" and labels the tabs the same way, instead of the blanket PyTorch icon. That matches where `hessian_mode` is actually honored — `deepmd/pt/`, `deepmd/jax/`, and `deepmd/dpmodel/model/make_hessian_model.py`, with nothing on the pt_expt side — so the page no longer implies coverage that does not exist.

doc/model/dpa4.md
```diff
@@ -17,7 +22,8 @@ Reference: [DPA4 paper](https://arxiv.org/abs/2606.02419).
 
 ## Quick start
 
-DPA4 is a PyTorch-only model. Train it with the standard `dp --pt` workflow:
+DPA4 is available only through the PyTorch framework backends. Train a
+PyTorch-TorchScript model with the standard `dp --pt` workflow:
```
OutisLi: DPA4 is not a pytorch-torchscript model

doc/model/dpa4.md
```diff
@@ -408,14 +414,14 @@ once the target system and batch size fit.
 
 ### Freeze to `.pt2`
 
-DPA4/SeZM checkpoints use the PyTorch `.pt2` (AOTInductor) export path; the
+DPA4/SeZM checkpoints use the PyTorch-TorchScript `.pt2` (AOTInductor) export path; the
```
OutisLi: Same here. SeZM/DPA4 is realized in pt backend, but the deployment path is built on the pt export backend

iProzd (approved):
Re-reviewed the current head after the DPA4/SeZM follow-up. The documentation now correctly separates the `dp --pt` training route from the AOTInductor `.pt2` export/runtime path and explicitly states that DPA4/SeZM does not use ordinary TorchScript. The prior backend-contract concerns are resolved, and I found no new blocking issue.

doc/backend.md
```diff
@@ -15,25 +21,88 @@ In the documentation, TensorFlow {{ tensorflow_icon }}, PyTorch {{ pytorch_icon
...
+The `.pt2` suffix identifies an AOTInductor package, but not its lower-input
+ABI. A DPA4/SeZM model frozen with `dp --pt freeze` normally uses the legacy
+`edge_vec` ABI (`lower_input_kind: edge_vec`); its deepspin virtual-atom variant
+uses the dense `nlist` ABI instead. A graph model frozen with
+`dp --pt-expt freeze --lower-kind graph` uses the NeighborGraph ABI
+(`lower_input_kind: graph`). All variants are loaded for inference by the
+PyTorch-Exportable runtime, which reads this metadata to select the correct
+input path. The `--lower-kind` option controls only the PyTorch-Exportable
+freeze route; see the [DPA4 export documentation](model/dpa4.md#freeze-to-pt2)
+for the separate PyTorch-TorchScript route.
```
OutisLi: [P2] Do not call the DPA4 export route TorchScript

The latest commit correctly states in `dpa4.md` that DPA4/SeZM bypasses TorchScript and exports through AOTInductor, but this overview still labels the same `dp --pt freeze` path as the “PyTorch-TorchScript route.” That leaves the documentation internally contradictory and misclassifies the artifact. Please describe it as the separate PyTorch AOTInductor export route (or equivalent).

OutisLi (changes requested):
One remaining backend-classification error is noted inline.

OutisLi (commented):
One additional backend-workflow omission is noted inline.

doc/model/change-bias.md
```diff
@@ -1,7 +1,7 @@
 # Change the model output bias for trained model {{ tensorflow_icon }} {{ pytorch_icon }}
 
 > [!NOTE]
-> **Supported backends**: TensorFlow {{ tensorflow_icon }}, PyTorch {{ pytorch_icon }}
+> **Supported backends**: TensorFlow {{ tensorflow_icon }}, PyTorch-TorchScript {{ pytorch_icon }}
```
OutisLi: [P2] Document the supported PyTorch-Exportable change-bias route

This support list and the examples below omit PyTorch-Exportable, even though `doc/backend.md` explicitly advertises change-bias for that backend and `deepmd/pt_expt/entrypoints/main.py` implements `dp --pt-expt change-bias` for `.pt`, `.pte`, and `.pt2` inputs, with dedicated tests. Please document that route here so the backend overview and workflow page agree, and state its current limitation that multi-task checkpoints are not supported.

doc/train/parallel-training.md
```diff
@@ -92,14 +92,14 @@ optional arguments:
                         master)
 ```
 
-## PyTorch Implementation {{ pytorch_icon }}
+## PyTorch-TorchScript Implementation {{ pytorch_icon }}
```
OutisLi: [P2] Document the existing PyTorch-Exportable DDP route separately

This rename leaves the parallel-training guide with only the PyTorch-TorchScript route, although `deepmd/pt_expt/entrypoints/main.py::setup_run()` initializes `torch.distributed` under `torchrun`/`srun` and the pt_expt trainer wraps both single-task and multi-task models in `DistributedDataParallel`. Please add a separate PyTorch-Exportable DDP subsection and a `torchrun ... dp --pt-expt train input.json` example. It should remain separate from the ZeRO/FSDP2 subsection below, because those options are specific to the PyTorch-TorchScript trainer rather than pt_expt.

OutisLi (commented):
One additional PyTorch-Exportable training-workflow omission is noted inline.

doc/model/dpa2.md
```diff
@@ -96,13 +98,13 @@ The performance improvement will be limited if other parts are more expensive.
...
+> **Default route change in pt_expt (eager & training).** For a graph-eligible DPA-2 descriptor, the pt_expt backend now defaults to the carry-all graph route not only for `--lower-kind graph` freezing but also in **eager inference/evaluation and in (compiled) training** (`neighbor_graph_method=None` resolves to the graph). This changes the numerical behavior of existing pt_expt configurations relative to the dense neighbor-list route (by the amounts described above — negligible at non-binding `sel`). The other backends (dpmodel/PyTorch-TorchScript/Paddle/TensorFlow/JAX) are unaffected: they keep the dense route as their only path.
```
OutisLi, line 107: [P1] Remove the unavailable dense fallback later in this section

This new default-route description conflicts with the final paragraph, which tells users to avoid the empty-rank restriction by using “the dense (`nlist`, the default) artifact.” For the graph-eligible DPA2 discussed here, `deepmd/pt_expt/entrypoints/main.py::freeze()` overrides every non-graph lower request—including an explicit `--lower-kind nlist`—to `graph` whenever `model_uses_graph_lower(m)` is true. Thus `nlist` is neither the default nor an alternative artifact selectable for the same model; a dense artifact requires changing the model into a non-graph-eligible configuration. Please remove that fallback advice and describe the forced graph behavior consistently.

OutisLi (commented):
One additional graph-freeze behavior mismatch is noted inline.

doc/backend.md
```diff
@@ -15,25 +21,88 @@ In the documentation, TensorFlow {{ tensorflow_icon }}, PyTorch {{ pytorch_icon
...
+
+### PyTorch-Exportable {{ pytorch_icon }}
+
+- Model filename extensions: `.pte`, `.pt2`
+- Checkpoint filename extension: `.pt`
+
+Select this backend with `dp --pt-expt` (alias
+`dp --pytorch-exportable`). It uses PyTorch with the backend-independent model
+implementation and supports training, including multi-task training and
+fine-tuning, freezing, change-bias, and testing. Compression support and export
```
OutisLi, line 65: [P2] Add the advertised TF2 and pt_expt fine-tuning workflows

This overview now advertises fine-tuning for TensorFlow 2 and PyTorch-Exportable, but the dedicated `doc/train/finetuning.md` page still contains only the legacy TensorFlow, `dp --pt`, and Paddle workflows. Both new backends have implemented fine-tuning entrypoints, and pt_expt has coverage for single-task, multi-task, DDP, and `.pt`/`.pte`/`.pt2` sources. Please update the dedicated guide with executable `dp --tf2 train --finetune ...` and `dp --pt-expt train --finetune ...` routes and their backend-specific accepted formats/limitations, so this capability claim leads to a complete public contract.

OutisLi (commented):
One additional backend-workflow documentation gap is noted inline.

doc/model/train-energy-spin.md
```diff
@@ -11,7 +12,7 @@ keeping other sections the same as the normal energy model's input script.
 >
 > - In the TensorFlow backend, the `se_e2_a` descriptor will treat those atom types with spin as new (virtual) types,
 >   and duplicate their corresponding selected numbers of neighbors ({ref}`sel <model[standard]/descriptor[se_e2_a]/sel>`) from their real atom types.
-> - In the PyTorch backend, if spin settings are added, all the types (with or without spin) will have their virtual types.
+> - In the PyTorch-TorchScript and PyTorch-Exportable backends, if spin settings are added, all the types (with or without spin) will have their virtual types.
```
OutisLi: [P1] Limit the virtual-type statement to the deepspin scheme

This is true for the default `deepspin` virtual-atom scheme, but not for DPA4/SeZM with `model.spin.scheme: native`. The native path passes the per-atom spin vector directly into the descriptor and introduces neither virtual atoms nor a doubled type map. Moreover, native DPA4 is graph-only, and graph neighbor construction is completely `sel`-free: every edge within `rcut` is retained, so `sel` neither truncates nor sizes its neighbor set. The advice to extend `sel` therefore applies only to dense virtual-atom routes, not to native DPA4. The DPA4 page documents this distinction and links back to this page as the general spin-training workflow. Please qualify this paragraph as `deepspin`-specific, mention the DPA4 native exception, and include DPA4 in the scheme-aware supported-descriptor discussion below so the two pages expose one consistent spin contract.

OutisLi (commented):
One additional spin-scheme semantic mismatch is noted inline.

OutisLi (changes requested):
The native-spin section still documents a lower-kind choice that the current pt_expt freeze interface does not provide. It says that a plain-energy DPA4/SeZM descriptor can freeze to either the dense or graph lower and that `--lower-kind auto` is the default. However, the CLI accepts only `nlist` and `graph`, defaults to `nlist`, and `freeze()` overrides that default/request to `graph` whenever `model_uses_graph_lower(m)` is true. Since an ordinary DPA4 descriptor reports `uses_graph_lower() == True`, it is graph-frozen as well; `auto` is not a valid value. Please make this paragraph match the current forced-graph policy and remove the nonexistent `auto` option.

OutisLi (changes requested):
The earlier correction that distinguishes JAX .jax serialization from DeepEval-compatible frozen formats is still incomplete in the Hessian documentation. train-energy-hessian.md says backend checkpoints can be tested directly and gives dp --jax test -m model.ckpt.jax, but test is a common entrypoint and the JAX DeepEval constructor accepts only .hlo and .savedmodel; a .jax path raises ValueError(Unsupported file extension). Please remove the direct .jax test workflow and use the already documented freeze --hessian to .hlo/.savedmodel route instead, unless .jax DeepEval loading is implemented.

OutisLi (changes requested):
The .pt2 lower-input ABI overview is incomplete: it enumerates edge_vec, nlist, and graph and then describes these as all variants, but the documented pt_expt DPA1 compression workflow can produce a fourth production ABI, lower_input_kind=dpa1_canonical, whenever canonical_model_eligible(model) is true. DeepPotPTExpt also has a distinct lower_input_is_canonical_ dispatch branch for it. Please document dpa1_canonical as the compact compressed-DPA1 ABI in both the ABI overview and the later extension-based inference-routing summary instead of subsuming it under NeighborGraph.

OutisLi (approved):

PR #5939, opened 2026-07-30
feat: SFPG cross-rank completion — capabilities, ZBL bridging and spin+ZBL multi-rank

### LINKED ISSUE #5906: DPA4/SeZM zone-bridging (SFPG) gate is incomplete under domain decomposition; blocks ZBL and spin+ZBL multi-rank
## Summary

The DPA4/SeZM zone-bridging gate (Source Freeze Propagation Gate, SFPG) is computed from a per-node reduction over a node's **outgoing** edge set. Under MPI domain decomposition no single rank observes that full set, so the gate is incomplete on every rank. This blocks multi-rank inference for ZBL-bridged DPA4 models, and — separately — for the combination of native spin + ZBL.

Native-spin multi-rank on the NeighborGraph route is already working (PR #5884); bridging is the one remaining graph-route feature that cannot go multi-rank without a new communication primitive.

## Why the gate is rank-incomplete

`compute_edge_src_gate` computes, per node *j*:

```
w_e   = bridging_switch(r_e)              in [0, 1]
eta_j = prod over { e : src_e == j } w_e  in [0, 1]
gate_e = eta_{src_e}
```

- dpmodel/pt_expt: `deepmd/dpmodel/descriptor/dpa4_nn/edge_cache.py:127` (reduction at lines 209 and 218, keyed by `src`)
- pt: `deepmd/pt/model/descriptor/sezm_nn/edge_cache.py:113` (same structure)

The reduction key is `src` — the **neighbor** side of the edge. But a rank's edge list only contains edges whose `dst` (center) is an owned atom, which is the LAMMPS neighbor-list contract. An owned boundary atom *j* is a neighbor of centers owned by other ranks, and those edges exist only on those ranks. So `eta_j` is a partial product on every rank, for owned atoms as well as ghosts.

This is structurally different from every exchange the descriptor already does. `deepmd_export::border_op` broadcasts an owner's already-complete value onto its ghost copies; here **no rank ever holds a complete value to broadcast**. The partial products must first be combined across ranks.

## When it produces wrong numbers

Edges at `r >= r_outer` contribute `w = 1` (`log w = 0`), so omitting them is harmless. The error appears only when a **close contact inside the bridging window straddles a rank boundary**: atom *j* within `r_outer` of a center *i* owned by another rank. Both ranks see that pair, but with opposite orientation — *j*'s rank holds `(src=i, dst=j)`, while `eta_j` needs `(src=j, dst=i)`, which lives only on *i*'s rank.

The result is that `eta_j` fails to freeze (or freezes too weakly) for exactly the cross-boundary close pairs that ZBL bridging exists to handle. Any test that does not place a sub-`r_outer` pair across a subdomain boundary passes vacuously.

## Current state per backend

**pt_expt — safe, fails fast.** `DescrptDPA4.has_message_passing_across_ranks()` returns `self.bridging_switch is None` (`deepmd/dpmodel/descriptor/dpa4.py:2280-2294`), so a bridged model freezes with `has_comm_artifact=false` (asserted in `source/tests/infer/gen_dpa4_zbl.py:155`) and the C++ inference path rejects multi-rank. Wrong answers are not reachable; multi-rank is simply unsupported.

**pt — no guard.** `DescrptSeZM.forward` accepts `comm_dict` (`deepmd/pt/model/descriptor/sezm.py:1472`) and then builds the edge cache with `bridging_switch=self.bridging_switch` unconditionally (`sezm.py:1526`). There is no bridging-vs-parallel rejection on that path, and `comm_dict` reaches the descriptor through the ordinary model plumbing (`deepmd/pt/model/model/make_model.py:402`). A multi-rank run of a bridged `.pth` would therefore compute a wrong gate at subdomain boundaries with no diagnostic.

Caveat on the pt claim: this is established by reading the code, not by a failing test. It is reachable by construction, but has not been demonstrated end-to-end.

## Proposed fix: reverse-accumulate, then forward-broadcast

Three phases per forward pass:

1. **Local partials.** Scatter-sum `log w` and the zero-count into a per-node buffer spanning owned *and* ghost rows, so a ghost row accumulates the contributions of edges where that ghost is the `src`.
2. **Reverse-accumulate.** Send ghost rows to their owning rank and `+=` them into the owner's row — the transpose of `border_op`'s forward, walking the same LAMMPS swap schedule backwards. Log-products are additive and each edge lives on exactly one rank, so after this every owner row holds the complete global value with nothing double-counted.
3. **Forward-broadcast.** A plain `border_op` copies the completed owner values back onto ghost copies.

Then `eta = exp(log_eta)` (with the zero-count hard-freeze rule) is correct everywhere and `gate_e = take(eta, src)` is unchanged.

The two per-node scalars (`log_eta`, `zero_count`) pack into one `(N, 2)` tensor, so the cost is one reverse plus one forward exchange of a small tensor per step — negligible next to the per-block feature exchanges.

**Much of the primitive already exists.** `deepmd_export::border_op_backward` is registered as a standalone op with CPU and CUDA implementations (`source/op/pt/comm.cc:650, 659-668`), and its kernel is exactly the ghost-to-owner `index_add_` reverse walk (`comm.cc:322-386`), including the ghost-row zeroing (`comm.cc:400-402`) that phase 3 then refills. The remaining work is to expose it as a first-class **forward** op that `make_fx` can trace into the `.pt2`, pair it with the transposed autograd (its backward is `border_op`'s forward, so gradients from the gate cross ranks correctly), add the fake/meta registration and the Python fallback, and thread `comm_dict` into the edge-cache build — today `comm_dict` only reaches the interaction blocks, while the SFPG runs before block 0.

Rejected alternative: widening the halo so each rank sees every relevant node's outgoing edges. That needs a 2×rcut communication skin plus ghost-ghost edges, which LAMMPS neighbor lists do not give pair styles.

## Scope

**Task 2 — ZBL bridging multi-rank.** The primitive above, wired at the edge-cache build in both dpmodel/pt_expt and pt (one owner per backend, same seam), lifting the `has_message_passing_across_ranks` exclusion, plus C++/LAMMPS enablement.

**Task 3 — native spin + ZBL multi-rank.** Expected to follow from Task 2 with wiring and verification only: native-spin multi-rank on the graph route is already verified, so the bridging gate is the only missing piece of the combination.

**Task 4 — export-time questions must be atomic-model capabilities, not reach-through.** *(prerequisite of Task 2 — same seam, not a separate concern)*

The general statement: the export/serialization layer asks questions the atomic-model interface does not expose, so it substitutes structural knowledge of a concrete model — either the wrapper's TYPE (`isinstance`) or its INTERNALS (`.descriptor`). Both assume "one atomic model == one descriptor, of a class I recognise", which every composition falsifies, and both fail SILENTLY: `isinstance` returning False and `getattr(..., None)` returning `None` each land on a plausible default instead of an error.

Model-time questions already do this correctly — `uses_graph_lower()`, `supports_native_spin()`, `has_chg_spin_ebd()`, `get_dim_chg_spin()` are capabilities on `BaseAtomicModel` that `LinearEnergyAtomicModel` aggregates (`all` / `any` / `max` as the semantics require). Export-time questions were never migrated.

Four capabilities need promoting to the atomic model with composition aggregation, after which every `isinstance` and every `.descriptor` lookup below becomes unreachable rather than merely fixed:

| # | Question | Asked today at | How |
|---|---|---|---|
| 1 | needs cross-rank message passing | `serialization.py:193` (`isinstance(LinearEnergyAtomicModel)`) **and** `:200` (`desc.has_message_passing_across_ranks()`) | already a capability, but on the DESCRIPTOR; aggregate: ANY child needs it, EVERY child supports edge-parallel |
| 2 | uses compact edge pairs (torch>=2.6 unbacked-SymInt guard) | `serialization.py:248` (`desc.uses_compact_edge_pairs()`) | a real capability asked of the wrong object |
| 3 | graph edge dtype | `serialization.py:1027` — reads `desc.se_atten.mean.dtype`, THREE levels deep into a DPA1-specific internal | no capability exists; add one |
| 4 | supports graph export | `serialization.py:1047` — reads `desc.geo_compress` and the PRIVATE `desc._fused_eligible` | no public capability exists at all |

Item 1 is the one that blocks Task 2: `_needs_with_comm_artifact` denies a composition through BOTH the `isinstance` check and the `desc is None -> return False` fallthrough immediately after it, so deleting only the former changes nothing, and fixing the SFPG gate alone would stay invisible at export. Reported and reproduced by @OutisLi in the #5884 review (at `2f4e63dd8`) with a two-DPA2 `linear_ener` model that is graph-eligible, whose children both report `has_message_passing_across_ranks() == True`, and which still gets no with-comm artifact.

Items 2-4 are latent rather than live: for DPA4 every default they fall back on happens to be correct (DPA4 has no `geo_compress`, and `uses_compact_edge_pairs()` is False), so bridged DPA4 works today. A bridged `geo_compress` DPA1/DPA2 would silently get fp64 edges and no torch-version guard — a broken artifact, not an error.

Fixing these one function at a time invites the next helper to reach through `.descriptor` again, which is how this accumulated; the interface completion is the actual fix.

Regressions: a two-DPA2 `linear_ener` asserting the with-comm artifact **is** requested (unblocked by item 1 alone), and a bridged composition asserting it is **not**, until Task 2 lands.

Also stale and worth correcting in the same pass: the comment at `serialization.py:248` says models without a single descriptor "take the dense route anyway". Since deepmodeling/deepmd-kit@881e2087a, graph-capable models — bridged ones included — are auto-resolved ONTO the graph route, so the comment now documents the opposite of the behaviour.

Deferred out of #5884 rather than fixed there.

**Interim hardening (independent of the above).** pt's parallel path should get the same rejection pt_expt already has — a `comm_dict is not None and self.bridging_switch is not None` guard in `DescrptSeZM.forward` — so the silent-wrong-answer window closes now rather than at fix time.

## Verification

The regression test both backends need is a **cross-boundary close pair**: a fixture with a sub-`r_outer` contact positioned so that a `processors 2 1 1` decomposition splits it, compared against the 1-rank result. Without that geometry the test is vacuous.

The bisection ladder that resolved the native-spin multi-rank bug in #5884 applies here too, each level exonerated before descending: eager self-comm parity (single-rank `comm_dict` self-send, no MPI runtime) → `make_fx` trace → locally compiled AOTI artifact → C++ single-rank forced through the with-comm branch → per-rank tensor dumps in a real 2-rank run.

## Out of scope for #5884

PR #5884 delivers the graph-native DPA4 port, multi-rank for non-bridged models, native spin (single and multi-rank), and charge-spin/bridging on the graph route for single-rank. The bridged multi-rank cases above are tracked here and deliberately left to a follow-up.

wanghan-iapcm:
Closes #5906.

The DPA4/SeZM Source Freeze Propagation Gate computes each node's `eta_j = prod over outgoing edges of w(r_e)`; under MPI domain decomposition a rank only holds edges with owned destinations, so the src-keyed per-node partials are rank-incomplete and bridged models were single-rank only. This PR completes the gate across ranks and, as a prerequisite, promotes the export-time questions to atomic-model capabilities so compositions answer by aggregation.

## Phase 1 — capability aggregation (issue Task 4)

- Split the conflated descriptor capability: `has_message_passing_across_ranks` (needs the per-block exchange; unconditionally true for SeZM) vs the new `supports_edge_parallel` (can run under domain decomposition).
- Six capabilities on `BaseAtomicModel` with concrete defaults, descriptor delegation on `DPAtomicModel`, and any/all aggregation on `LinearEnergyAtomicModel`: `has_message_passing_across_ranks` (any), `supports_edge_parallel` (all), `dense_lower_supports_comm` (all), `uses_compact_edge_pairs` (any), `graph_edge_dtype` (float32 iff all children), `supports_graph_export` (all).
- `forward_lower_graph_exportable_with_comm` hoisted from `EnergyModel` into `make_model` (one owner, next to the non-comm twin) so `LinearEnergyModel` compositions can export it.
- The four `serialization.py` helpers now consult the atomic model — no `isinstance`-on-concrete-model checks, no `.descriptor` walks. Regression fixed: a linear composition of two DPA2 children now gets its with-comm artifact (previously denied by wrapper type).
- Composition-safe reach-ins outside serialization: `.pt`-checkpoint eval no longer crashes on compositions (`ntypes` via the model API), `enable_compile` degrades gracefully, and pt_expt `get_standard_model` honors `bridging_method` like its dpmodel twin (`_compose_bridging` is the single composition owner).

## Phase 2 — SFPG cross-rank completion (issue Tasks 2 and 3)

No new communication machinery: the fix is one extra invocation of the existing `deepmd_export::border_op_backward` + `border_op` pair (they are exact transposes, `R = B^T`) on an `(N, 2)` `[log_eta, zero_count]` tensor before the gate is applied — reverse-accumulate ghost partials into owners, then broadcast the completed values back. Zero C++ changes.

- `border_op_backward` gains autograd (its gradient is `border_op`'s forward), so gate gradients cross ranks.
- dpmodel: `compute_edge_src_gate` packs the partials through an optional `node_partial_exchange` hook; the dpmodel `_gate_partial_exchange` raises (single-process reference), the pt_expt subclass implements it on the border-op pair.
- pt backend wired the same way. The red run of the new pt parity test demonstrated the issue's claim and more: pt's bridged parallel path did not just compute a silently wrong gate — it crashed outright (the ZBL injection indexed per-local types with extended ghost `src` indices); fixed by reading extended types.
- Gates flipped: `supports_edge_parallel` is now `True` for bridged SeZM in both backends; bridged (and spin+ZBL) graph freezes embed the nested `forward_lower_with_comm.pt2`.

## Verification

Anti-vacuous discipline throughout: every parity test places a sub-`r_outer` pair ACROSS the periodic/rank boundary (without it every cross-rank gate contribution is `log w = 0`), covers both bridging channels (hard-freeze `zero_count` at 0.4 Å, transition-zone `log_eta` at ~1 Å), and carries an identity-exchange ablation that must diverge.

- Eager self-comm parity vs the folded reference at rtol/atol 1e-12 (energy, force, and force_mag for the spin variant), pt and pt_expt.
- make_fx traces both border ops symbolically (21-input with-comm ABI unchanged); freeze embeds the nested artifact for ZBL and spin+ZBL compositions.
- LAMMPS end-to-end on a Tesla T4: 2-rank vs 1-rank close-pair parity for ZBL (`pair_style deepmd`) and spin+ZBL (`pair_style deepspin`, incl. magnetic forces) — the spin+ZBL variant gets its first LAMMPS file. All 24 `*Dpa4Zbl*` C++ gtests pass (CPU + T4).
- Variant-alignment coverage: ZBL empty-rank fail-fast twin, the first test of the DeepSpin owned-empty phantom path, charge-spin through `pair_style deepspin`, and default-CLI `dp freeze` resolution (nlist→graph auto-override + with-comm artifact) for both compositions.

## Known limitations

1. pt eager multi-rank bridging has no true-MPI pt test (no pt `.pth` LAMMPS ZBL fixtures exist); its parity rung is self-comm.
2. `graph_edge_dtype` composition rule (float32 iff ALL children) is conservative; fp64 is the universal ABI.
3. `supports_graph_export` keeps the hardcoded `"cuda"` probe inside pt_expt DPA1 (capability promoted; probe internals unchanged).
4. The `NativeSpinModelKind` marker-base check in `_needs_with_comm_artifact` remains (a cross-backend family test, not a concrete-type reach-through).
5. Model-deviation coverage stays absent for all dpa4 variants (pre-existing; Python `model_devi` has no spin support at all).
6. DeepPot vs DeepSpin empty-rank designs deliberately differ (fail-fast vs phantom-pad, PR #5485); both are now pinned per variant, not unified.
7. Found while testing, left for a separate fix: `source/api_c/include/deepmd.hpp` uses `&vec[0]` on possibly-empty vectors (~33 sites) — undefined behavior that SIGABRTs under `_GLIBCXX_ASSERTIONS` before the empty-rank guard's message can fire (benign on non-hardened builds).

## Summary by CodeRabbit

* **New Features**
  * Added multi-rank inference for bridged DPA4/SeZM models, including native-spin and ZBL configurations.
  * Improved graph export detection, metadata, edge precision, and communication-aware export.
  * Added atomic-output-only inference for statistics workflows.
  * Standard model loading now preserves bridging configurations.

* **Bug Fixes**
  * Improved handling of atom types, ghost atoms, empty MPI ranks, charge-spin inputs, and cross-rank calculations.

* **Documentation**
  * Updated DPA4 and native-spin documentation for expanded multi-rank and graph export support.

* **Tests**
  * Added regression coverage for MPI parity, graph exports, bridging, charge-spin behavior, and capability reporting.


doc/model/dpa4.md
```diff
@@ -595,18 +592,9 @@ inference](#multi-gpu-mpi-inference).
...
-- **Charge-spin FiLM conditioning.** Combining `add_chg_spin_ebd` with
-  `spin.scheme: native` is rejected at model-construction time; use one or
-  the other.
-- **ZBL zone bridging.** Combining `bridging_method: ZBL` with
-  `spin.scheme: native` is not supported on the pt_expt backend (`bridging_method`
-  is rejected there independently of spin -- see [Zone bridging
-  (ZBL)](#zone-bridging-zbl)).
+graph route: none of the earlier restrictions remain. Multi-rank inference,
+charge-spin FiLM conditioning (`add_chg_spin_ebd`), and ZBL zone bridging
+(`bridging_method: ZBL`) all combine freely with `spin.scheme: native`.
```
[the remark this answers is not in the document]
wanghan-iapcm: Fixed in [d855cf5](https://github.com/wanghan-iapcm/deepmd-kit/commit/d855cf5715993a243919b26ed950d7ea6df65cdf)

source/tests/infer/gen_dpa4_spin_zbl.py
```diff
@@ -354,24 +354,20 @@ def _check_metadata(pt2_path: str) -> None:
...
+        f"{md.get('has_comm_artifact')!r}, expected True; the SFPG cross-rank "
+        f"completion (issue #5906) makes bridged native-spin models "
+        f"multi-rank capable."
     )
-    assert "model/extra/forward_lower_with_comm.pt2" not in names, (
-        f"{pt2_path}: a nested forward_lower_with_comm.pt2 was exported for a "
-        f"bridged model; see above -- the archive must not carry one."
+    assert "model/extra/forward_lower_with_comm.pt2" in names, (
+        f"{pt2_path}: the nested forward_lower_with_comm.pt2 is missing from "
+        f"a bridged native-spin archive; multi-rank dispatch would fail."
```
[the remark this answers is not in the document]
wanghan-iapcm: Fixed in [d855cf5](https://github.com/wanghan-iapcm/deepmd-kit/commit/d855cf5715993a243919b26ed950d7ea6df65cdf)

source/tests/pt/model/test_sezm_parallel_bridging_parity.py
```diff
@@ -0,0 +1,207 @@
...
+            sysm["coord"],
+            sysm["atype"],
+            sysm["edge_scatter_index"],
+            sysm["edge_vec"],
+            sysm["edge_scatter_index"],
+            sysm["edge_mask"],
+            do_atomic_virial=True,
+            comm_dict=comm,
+            extended_atype=sysm["extended_atype"],
+        )
```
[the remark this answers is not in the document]
wanghan-iapcm: This is intentional and mirrors the established harness in `test_sezm_parallel.py::TestSeZMSelfCommParity._assert_parity`: the parallel (comm_dict) path indexes the EXTENDED node set directly, so the extended `edge_scatter_index` deliberately doubles as the message-passing `edge_index` (its ghost src rows are the whole point — they carry the rank-incomplete SFPG partials that the exchange completes). Passing the folded `edge_index` would collapse ghost srcs onto owners and make the test vacuous for the cross-rank gate. Connectivity is not discarded: the parity holds at rtol 1e-8 against the folded reference with the real exchange, and the identity-exchange ablation in this file diverges by >1e-6 — both would be impossible if the graph inputs were malformed.

source/tests/pt_expt/model/test_dpa4_zbl_parallel.py
```diff
@@ -0,0 +1,427 @@
...
+            1.0,  # log_eta channel across the boundary
+        ],
+    )
+    def test_self_comm_matches_folded_reference(self, gap: float) -> None:
+        coord = _close_pair_coords(gap)
+        e_ref, f_ref, fm_ref = self._run_folded(coord)
+        e_par, f_par, fm_par = self._run_self_comm(coord)
+        np.testing.assert_allclose(e_par, e_ref, rtol=1e-12, atol=1e-12)
+        np.testing.assert_allclose(f_par, f_ref, rtol=1e-12, atol=1e-12)
+        np.testing.assert_allclose(fm_par, fm_ref, rtol=1e-12, atol=1e-12)
```
OutisLi, line 409: [P1] Fix native-spin `force_mag` parity before enabling multi-rank

This newly added regression fails deterministically on the current HEAD for `gap=1.0`: the energy and coordinate-force assertions immediately above pass, but `force_mag` mismatches 6/12 elements (`max abs = 3.50371e-3`, `max rel = 7.4047e-2`). The `gap=0.4` hard-freeze case passes. I reproduced this against the PR worktree source both normally and with single-thread settings, so this is not reduction-order noise. Since `force_mag = -dE/dspin` and `_run_self_comm` already folds owner and ghost rows back through `mapping`, the advertised native-spin + ZBL multi-rank magnetic-force contract is not established. Please fix the ghost-spin backward/folding discrepancy, or correct this parity harness with evidence that the production multi-rank result is valid, before marking this variant multi-rank capable.

OutisLi (dismissed):
The native-spin + ZBL multi-rank magnetic-force contract is not yet validated: the PR's own transition-zone self-comm parity case fails reproducibly on the current HEAD.

OutisLi (commented):
The existing graph-freeze regressions still pin the obsolete single-rank composition contract and need to be updated together with this gate change.

source/tests/pt_expt/model/test_zbl_bridging.py
```diff
@@ -220,15 +220,15 @@ def test_serialize_roundtrip(self) -> None:
...
+        (issue #5906 Task 2): the graph with-comm artifact is compiled.
+        """
         from deepmd.pt_expt.utils.serialization import (
             _needs_with_comm_artifact,
         )
 
-        assert (
-            _needs_with_comm_artifact(self.pt_expt_model, lower_kind="graph") is False
-        )
+        assert _needs_with_comm_artifact(self.pt_expt_model, lower_kind="graph") is True
```
OutisLi, line 231: [P1] Update the existing freeze tests to the new with-comm contract

This new `True` contract conflicts with two executable regressions later in this same file: `test_graph_freeze_and_deep_eval_parity` still asserts `md["has_comm_artifact"] is False` at line 418, and `test_native_spin_with_bridging_graph_freeze_and_deep_eval` does the same at line 653. Their adjacent comments also still say compositions forbid a with-comm artifact. I verified `_collect_metadata(..., lower_kind="graph")` now returns `True` for both ZBL and native-spin + ZBL. These AOT tests are skipped only when `CI=true`, so a normal local full-suite run will compile the archive and then deterministically fail on the obsolete assertions. Please update both tests and their comments together with this capability flip.

deepmd/dpmodel/descriptor/make_base_descriptor.py
```diff
@@ -209,6 +220,32 @@ def uses_compact_edge_pairs(self) -> bool:
...
 
+        def dense_lower_supports_comm(self) -> bool:
+            """Whether the DENSE (nlist) lower implements comm_dict exchange.
+
+            Default ``True`` — dense comm is the production multi-rank path
+            for dpa2/dpa3 (previously this was probed by method absence in
+            the freeze machinery); a descriptor whose dense adapter raises
+            on ``comm_dict`` (DPA4) overrides to ``False``.
+            """
+            return True
```
OutisLi, line 231: [P1] Aggregate the dense-comm capability through hybrid descriptors

This default is incorrect for a `DescrptHybrid` containing DPA4. `DescrptHybrid` already aggregates `has_message_passing_across_ranks()` over its children, but it inherits `dense_lower_supports_comm() == True` here instead of requiring every child to support it. I reproduced this with a valid `hybrid([dpa4, se_e2_a])` model: it reports `uses_graph_lower=False`, `has_message_passing_across_ranks=True`, and `dense_lower_supports_comm=True`, so `_needs_with_comm_artifact(model, "nlist")` returns `True`. The resulting dense with-comm trace then deterministically raises DPA4's `NotImplementedError` as soon as `comm_dict` is passed. Consequently an auto/nlist freeze tries to build a dead communication artifact and fails. Please aggregate this capability with `all(child.dense_lower_supports_comm())` in `DescrptHybrid` and cover the hybrid+DPA4 case.

OutisLi (changes requested):
The new capability interface remains incomplete for composite descriptors: a supported hybrid containing DPA4 is misclassified as dense-comm capable, causing freeze to trace an implementation that always raises.

OutisLi (commented):
One maintainability issue remains in the updated DPA4 capability documentation.

deepmd/dpmodel/descriptor/dpa4.py
```diff
@@ -2278,20 +2322,27 @@ def has_message_passing(self) -> bool:
...
+        return True
+
+    def supports_edge_parallel(self) -> bool:
+        """Bridging included: multi-rank is supported for every SeZM config.
+
+        The SFPG per-node partials are completed across ranks by
+        ``_gate_partial_exchange`` (reverse-accumulate + broadcast) before
+        the gate is applied (issue #5906).
+        """
+        return True
```
OutisLi, line 2345: [P2] Remove the stale single-rank statement from this capability block

This PR deliberately separates `has_message_passing_across_ranks()` (the model **needs** a ghost exchange) from `supports_edge_parallel()` (the model **can** run under domain decomposition), and bridging now returns `True` for both. However, the `uses_graph_lower()` docstring a few lines below still says that “Bridging models still fail multi-rank fast via `has_message_passing_across_ranks`.” That is now the opposite of both the implementation and the new capability semantics, and it risks reintroducing exactly the conflation this refactor is meant to remove. Please update that sentence to reflect the `_gate_partial_exchange`-enabled multi-rank path.

deepmd/pt/model/descriptor/sezm_nn/edge_cache.py
```diff
@@ -193,16 +202,31 @@ def compute_edge_src_gate(
     log_eta = torch.zeros(
         n_nodes, dtype=edge_w.dtype, device=edge_w.device
     ).scatter_add(0, src, log_safe)
```
OutisLi, line 204: [P2] Build eta on the owner instead of reverse-accumulating source partials

The graph contract defines `src = neighbor` and `dst = center`, and LAMMPS requests a full neighbor list, so the owner of node `j` already has every edge `k -> j`. Since `w` depends only on the symmetric distance (and pair exclusion is symmetric), `eta_j = prod_k w(r_jk)` can be computed by reducing `log_safe` and `is_zero` over `dst`. Then only one forward `border_op` is needed to broadcast the completed `eta` to ghost source rows.

Reducing over `src` here makes the partials rank-incomplete by construction and forces `_gate_partial_exchange` to run `border_op_backward` followed by `border_op`. Both are MPI exchanges, and each current kernel ends with an `MPI_Barrier`, so this adds one avoidable round trip/global synchronization to every bridged DPA4 force evaluation and also requires the extra backward-op autograd machinery. Please use destination reduction plus one broadcast on the comm path (while preserving the current non-comm behavior if arbitrary non-reciprocal graphs must remain supported), or identify a supported multi-rank topology where an owner does not hold its complete destination-centered neighborhood.

OutisLi (changes requested):
The current SFPG communication sequence adds an avoidable MPI exchange and global synchronization on every force evaluation; please simplify this hot path before merging.

deepmd/pt_expt/model/make_model.py
```diff
@@ -1206,6 +1241,191 @@ def fn_spin(
                 spin,
             )
 
+        def forward_lower_graph_exportable_with_comm(
```
OutisLi: [P2] Keep the energy-only exporter out of the generic model factory

`make_model()` generates DOS, dipole, polar, property, and energy models, but this method hard-codes `do_grad_r("energy")`, `do_grad_c("energy")`, and `_translate_energy_keys()`. Consequently every non-energy model now advertises a `forward_lower_graph_exportable_with_comm` method that fails on the missing `energy` output contract if called. The current serializer's energy eligibility gate happens to hide that false API in the main path, but it leaves the common abstraction internally inconsistent and makes `hasattr`/direct programmatic use misleading.

The non-comm public wrapper is still correctly owned by `EnergyModel`, and `LinearEnergyModel` already aliases `EnergyModel.forward_lower_graph_exportable`. Please give the with-comm twin the same ownership (and alias it for linear energy), or introduce an energy-specific shared mixin, rather than putting energy-key translation on every class produced by `make_model`.

OutisLi (changes requested):
The graph with-comm exporter is energy-contract machinery and should not be attached to every model generated by the generic model factory. Please keep this capability in an energy-specific abstraction before merging.

source/lmp/tests/test_lammps_dpa4_spin_zbl_pt2.py
```diff
@@ -0,0 +1,566 @@
...
+# docstring for why these are computed live via a DeepPot subprocess call
+# rather than hardcoded or read from a sidecar file.
+expected_e = None
+expected_ae = None
+expected_f = None
+expected_fm = None
+expected_v = None
+
+
+def _cell_from_lammps_box(lmp_box: np.ndarray) -> np.ndarray:
```
OutisLi: [P2] Extract the shared native-spin LAMMPS harness

This new 566-line module shares 416 identical lines with `test_lammps_dpa4_spin_graph_pt2.py` (about 71% whole-file similarity). The duplicated surface includes `_cell_from_lammps_box`, live `DeepPot` reference generation, setup/teardown, LAMMPS construction, `force_mag` gathering, the single-rank energy/force/virial checks, and the MPI subprocess/parser/parity machinery. The actual variant-specific inputs are primarily the archive path, the close-pair geometry, and a small number of ZBL assertions.

Keeping two copies of the whole harness means any future change to the LAMMPS invocation, reference conversion, tolerances, timeout handling, or MPI output schema must be updated in parallel, and makes another stale-contract divergence likely. Please retain separate plain-native-spin and spin+ZBL test cases, but extract these shared mechanics into one helper/harness and parameterize only the archive, geometry, and variant-specific assertions.

OutisLi (changes requested):
The new spin+ZBL LAMMPS coverage should not duplicate the existing native-spin harness wholesale. Please keep the variant-specific tests independent while consolidating their shared reference, LAMMPS, and MPI mechanics before merging.

deepmd/pt_expt/model/get_model.py
```diff
@@ -164,38 +166,93 @@ def get_sezm_model(data: dict) -> EnergyModel:
...
-            models=[model.atomic_model, zbl_atomic],
-            type_map=data["type_map"],
-            weights="sum",
-            # Both exclusions belong to the composition: its children share one
-            # graph, so "excluded" must cover the analytical term too.
-            atom_exclude_types=data.get("atom_exclude_types", []),
-            pair_exclude_types=pair_exclude_types,
-        )
-        return LinearEnergyModel(atomic_model_=composed)
+        return _compose_bridging(model, data, bridging_method)
```
OutisLi, line 175: [P2] Document the capabilities this factory now supports

The `get_sezm_model()` docstring above still says that both `bridging` and `spin` are unsupported and raise `NotImplementedError`, which is now the opposite of the implementation: this branch accepts analytical bridging through `_compose_bridging()`, and the earlier branch accepts native spin by routing to `get_native_spin_model()`. What remains unsupported is virtual-atom/deepspin, LoRA, `use_compile`, and `preset_out_bias`; not spin and bridging as whole feature families.

Please update the public factory docstring with that precise support matrix. As written, it tells callers that two paths introduced or extended by this PR will fail, despite the function deliberately constructing them, and obscures the backend-interchangeability contract stated there.

OutisLi (commented):
Please update the DPA4/SeZM factory documentation to match the capabilities implemented by this PR.

source/tests/infer/gen_dpa4_spin_zbl.py
```diff
@@ -25,14 +25,14 @@
...
-set.  Edges exist only for owned centres, so eta is incomplete on every rank
-and no with-comm artifact may be exported.  ``_check_metadata`` asserts BOTH
-``has_comm_artifact is False`` AND that the nested
-``model/extra/forward_lower_with_comm.pt2`` entry is absent (mirroring
-``gen_dpa4_zbl.py``) -- an artifact appearing there would silently promise a
-multi-rank capability the model cannot honour.  Note this is the opposite of
-the UNbridged native-spin fixture (``gen_dpa4_spin.py``), which does carry
-the with-comm twin: bridging is what removes it.
+set.  Edges exist only for owned centres, so the per-node partials are
+rank-incomplete; the with-comm artifact completes them with one
```
OutisLi, line 29: [P2] Update the fixture's stale single-rank contract

This paragraph now explains how the with-comm artifact completes SFPG partials across ranks, and `_check_metadata()` below requires that artifact, but the heading at line 24 still says `Single-rank only, and this fixture PINS that limitation`. The comment above the bridging radii at line 95 likewise says they make the composition single-rank only. Both statements directly contradict this paragraph and the new two-rank LAMMPS parity coverage.

Please update both stale statements together so this fixture documents the multi-rank contract it now enforces; otherwise future maintenance can easily mistake the with-comm artifact or MPI test for a regression against the claimed fixture design.

OutisLi (commented):
Please remove the stale single-rank limitation from the spin+ZBL fixture documentation now that this fixture pins the with-comm multi-rank path.

deepmd/pt/model/descriptor/sezm.py
```diff
@@ -2280,16 +2290,66 @@ def has_message_passing(self) -> bool:
...
-        Freeze Propagation bridging is excluded: its per-node gate folds a
-        node's entire outgoing-edge set, which a single rank cannot observe for
-        ghost owners, so the edge-based with-comm artifact is not exported for
-        bridging models and multi-rank inference fails fast instead.
+        """SeZM reads ghost-neighbour features at every interaction block.
+
+        A domain-decomposed run must exchange them through ``border_op``, so
+        multi-rank inference always needs the with-comm exchange. Whether
+        multi-rank is POSSIBLE at all is :meth:`supports_edge_parallel`
+        (bridging vetoes it there).
```
OutisLi: [P2] Do not say bridging vetoes edge-parallel execution

This sentence says `supports_edge_parallel()` vetoes bridging, but the immediately following method explicitly documents `Bridging included`, returns `True` for every SeZM configuration, and relies on `_gate_partial_exchange` to complete the SFPG partials. It therefore reverses the distinction this capability split is intended to establish: bridging needs cross-rank exchange and is also edge-parallel capable.

Please remove the stale veto statement and describe the completed SFPG exchange instead. The analogous stale statement in the dpmodel twin is already tracked in the other review thread, but this PT copy remains independently contradictory.

OutisLi (commented):
Please make the PT SeZM capability documentation consistent with the bridging multi-rank behavior implemented by this PR.

doc/model/dpa4.md
```diff
@@ -573,17 +575,12 @@ inference](#multi-gpu-mpi-inference).
...
-  implemented), so a multi-rank LAMMPS run fails fast with an explicit error
-  at the first force evaluation, mirroring the dense route's single-rank
-  restriction described in [Multi-GPU (MPI)
-  inference](#multi-gpu-mpi-inference). Run native-spin models on a single
-  MPI rank (a single GPU, or CPU without `mpirun`).
-- **Spin is per local atom.** The `spin` input is `(nframes, nloc, 3)` --
-  one vector per *local* atom, not per ghost/extended atom (`nall`); there is
-  no ghost-spin exchange to populate ghost spins across a rank boundary,
-  consistent with the single-rank restriction above.
+- **Multi-rank capable.** The frozen archive embeds the with-comm artifact
```
OutisLi, line 585: [P2] Do not advertise a selectable dense freeze for plain DPA4

The preceding bullet still says a plain-energy DPA4 can freeze to either the dense or graph lower and calls `--lower-kind auto` the default. Neither matches the public `pt_expt freeze` contract: its default request is `lower_kind="nlist"`, and `freeze()` overrides every non-graph request—including an explicit `nlist`—to `graph` whenever `model_uses_graph_lower(model)` is true. Plain DPA4 is graph-capable, so users cannot select a dense artifact through this entry point.

Please describe plain DPA4 as always graph-frozen too, and explain that the default/non-graph request is overridden. The current wording gives users a choice that the CLI deliberately does not honor.

OutisLi (commented):
Please align the DPA4 freeze-route documentation with the public pt_expt freeze behavior that always selects graph for graph-capable models.

deepmd/pt_expt/utils/serialization.py
```diff
@@ -181,34 +181,17 @@ def _needs_with_comm_artifact(
     if isinstance(model, NativeSpinModelKind) and lower_kind != "graph":
```
OutisLi, line 184: [P2] Update the shared native-spin marker's stale single-rank example

This gate now deliberately excludes native spin only from the dense lower and allows the graph lower to emit a with-comm artifact. However, `NativeSpinModelKind`—the shared marker imported here to define this cross-backend distinction—still documents its motivating example as “the with-comm freeze gate: native-spin lowers are single-rank only” (`deepmd/dpmodel/model/native_spin_model.py:37-38`). That is now the opposite of the implemented contract and of the new multi-rank coverage.

Please update the marker documentation to say that native-spin graph lowers participate in the with-comm path, while only their dense lower is excluded. Leaving the stale example on the central abstraction risks reintroducing the unconditional exclusion this branch just removed.

OutisLi (commented):
Please update the shared native-spin marker documentation to match the graph with-comm contract implemented here.

deepmd/pt_expt/model/get_model.py
```diff
@@ -164,38 +166,93 @@ def get_sezm_model(data: dict) -> EnergyModel:
...
+    config with ``bridging_method`` silently dropped the bridging term
+    (backend divergence from dpmodel -- issue #5906 Task 4 audit).
+
+    Parameters
+    ----------
+    data : dict
+        The data to construct the model.
+    """
+    data = copy.deepcopy(data)
+    bridging_method = str(data.get("bridging_method", "none"))
```
OutisLi, line 269: [P1] Complete `standard` + DPA4 bridging as a public equivalent route

The non-bridged `type: "standard"` + `descriptor.type: "dpa4"` + `fitting_net.type: "dpa4_ener"` route is a useful pt_expt extension: for the same fully specified configuration I verified that it produces the same `DPA4EnergyModel`, bitwise-identical state, identical serialization, and identical energy/force/virial as `type: "dpa4"`. The canonical direct-builder bridging case is likewise identical because both paths end in `_compose_bridging`.

This new wrapper is not yet a complete extension, however. Strict `argcheck.normalize` rejects `bridging_method` under `type: "standard"`, so a valid `input.json` cannot reach this code; the new test only succeeds by calling the low-level builder directly. The route also bypasses the DPA4 builder's configuration reconciliation. For example, with descriptor-only `exclude_types=[[0, 1]]`, `type: "dpa4"` propagates the exclusion to the composition, whereas this path leaves `LinearEnergyAtomicModel.pair_exclude_types` empty and the `InnerPotential` child still evaluates that pair. On a 0.9 Å Ni-O dimer this changes the energy by about 79.97 eV and the maximum force component by about 318.48 eV/Å.

Please formally support the enhancement: make the validated `standard` + DPA4 configuration able to express bridging, restrict that interpretation to the DPA4 component combination, and share one DPA4 normalization/reconciliation owner with `get_sezm_model` so exclusions and the other DPA4 invariants cannot diverge. The regression should pass through `argcheck.normalize` and compare the resulting standard and dpa4 models, including an exclusion case.

OutisLi (changes requested):
Please make the new `standard` + DPA4 bridging route a fully supported, normalized equivalent of `type: dpa4`, rather than a direct-builder-only partial path.

source/tests/pt_expt/model/test_zbl_bridging.py
wanghan-iapcm: Fixed in [94401ad](https://github.com/deepmodeling/deepmd-kit/commit/94401ade084fbe9a89e2367314583f4b75e4ada7) — you are right, the capability flip missed both of these. Flipped `test_graph_freeze_and_deep_eval_parity` and `test_native_spin_with_bridging_graph_freeze_and_deep_eval` to `has_comm_artifact is True`, updated the adjacent comments, and additionally asserted the nested `model/extra/forward_lower_with_comm.pt2` entry so the two tests pin the artifact itself rather than only the flag. Both verified locally with the AOT compile actually running (CI unset).

deepmd/dpmodel/descriptor/make_base_descriptor.py
wanghan-iapcm: Fixed in [b61ae9a8](https://github.com/deepmodeling/deepmd-kit/commit/b61ae9a8fa23a9296fc5637e5335d51d8e75fc84). Reproduced exactly as you described before fixing:

```
uses_graph_lower False
has_mp_across_ranks True
dense_lower_supports_comm True   (dpa4 child: False)
```

`DescrptHybrid` now aggregates `dense_lower_supports_comm()` with ALL, and `supports_edge_parallel()` too since it has the same veto semantics. Covered by a new `test_hybrid_descriptor_aggregates_dense_comm` in `source/tests/common/dpmodel/test_atomic_model_capabilities.py`, which pins both branches of the ALL rule and the `DPAtomicModel` delegation.

Scope note: `deepmd/pt/model/descriptor/hybrid.py` is a separate class that does not aggregate `has_message_passing_across_ranks()` either. That is pre-existing and pt has no with-comm freeze gate, so I left it alone; pt_expt inherits the dpmodel class and picks the fix up directly.

source/tests/pt_expt/model/test_dpa4_zbl_parallel.py
wanghan-iapcm: I cannot reproduce this, and I would like your help pinning down what differs before I either fix code or change the harness — the numbers you quote are far too large to be tolerance noise, so if it is real I do not want to paper over it.

What I ran on the current head:

1. The test as written (both `gap` values, `force_mag` asserted at `rtol=atol=1e-12`): passes. 8 consecutive whole-file runs with `pytest-randomly` shuffling order also pass, so it is not order/pollution dependent.
2. Importantly, this fixture is **not** weight-deterministic: `ZBL_CONFIG` seeds the descriptor (`seed: 7`) but its `fitting_net` has no seed, so the fitting weights differ on every construction (verified by serializing two builds and diffing the first layer). Parity must therefore hold for every draw, which makes a sweep the right check.
3. So I swept 100 independent model draws x 2 gaps = 200 comparisons, recording the worst `force_mag` deviation:

```
draws: 100 fails: 0
worst force_mag abs diff: gap0.4=1.527e-16 gap1.0=3.469e-16
```

That is machine epsilon, ~13 orders of magnitude below the `3.50371e-3` you observed. `_run_self_comm` and `_run_folded` are driven directly, so the comparison is the same one the test makes.

Could you tell me which commit you ran against and how you invoked it? Two candidates I can think of: the branch has since merged `upstream/master` (which brought in #5910, relocating the pt-side ZBL injection), so a pre-merge head is one possibility; the other is a different fixture draw path than the one above. If you still see it on `4021124`, a dump of the failing `force_mag` pair plus your commit SHA would let me chase it directly.

To be explicit about what is and is not established: the sweep above is an inline investigation, not a committed test. The committed regression is the two-gap parity test at `1e-12`.

deepmd/pt_expt/model/make_model.py
wanghan-iapcm: Fixed in [5b86080](https://github.com/deepmodeling/deepmd-kit/commit/5b8608083eaac517a42cb7fa0e1f5c0bb1d4647c) by taking your first option: the method now lives on `EnergyModel` next to its non-comm twin, and `LinearEnergyModel` aliases it exactly as it aliases `forward_lower_graph_exportable`. `NativeSpinEnergyModel` is `make_native_spin_model(EnergyModel)`, so the spin route keeps it through the MRO. The API surface is now what you would expect:

```
EnergyModel True / LinearEnergyModel True / NativeSpinEnergyModel True / DOSModel False
```

The body needed no changes — it only closed over `self`, `make_fx` and `_translate_energy_keys`, all already available in `ener_model.py`. I left `_translate_energy_keys` as a module-level helper in `make_model.py`: it is a plain function rather than a method on every generated class, and `training.py` imports it from there too, so moving it would only shuffle imports around. Re-verified by re-running both with-comm freeze tests end to end (real AOTI compile), which exercise the moved method through the composition and the spin subclass.

deepmd/dpmodel/descriptor/dpa4.py
wanghan-iapcm: Fixed in [4021124](https://github.com/deepmodeling/deepmd-kit/commit/4021124336c8fce310a72cb54b688ec102c4b14e).

deepmd/pt_expt/model/get_model.py
wanghan-iapcm: Fixed in [4021124](https://github.com/deepmodeling/deepmd-kit/commit/4021124336c8fce310a72cb54b688ec102c4b14e) — replaced with the precise matrix: analytical bridging and native-scheme spin are supported (and combine), while the deepspin scheme, `lora`, `use_compile` and `preset_out_bias` still raise.

source/tests/infer/gen_dpa4_spin_zbl.py
wanghan-iapcm: Fixed in [4021124](https://github.com/deepmodeling/deepmd-kit/commit/4021124336c8fce310a72cb54b688ec102c4b14e) — the heading now reads 'Multi-rank capable, and this fixture PINS that contract', and the bridging-radii comment says they are what makes the composition need the SFPG cross-rank completion rather than what makes it single-rank.

deepmd/pt/model/descriptor/sezm.py
wanghan-iapcm: Fixed in [4021124](https://github.com/deepmodeling/deepmd-kit/commit/4021124336c8fce310a72cb54b688ec102c4b14e) — the pt copy now states that `supports_edge_parallel()` is True for every configuration including bridging, completed via `_gate_partial_exchange`.

deepmd/pt_expt/utils/serialization.py
wanghan-iapcm: Fixed in [4021124](https://github.com/deepmodeling/deepmd-kit/commit/4021124336c8fce310a72cb54b688ec102c4b14e) — the marker docstring now says native spin excludes only the DENSE lower, and that native-spin graph lowers do participate in the with-comm path.

doc/model/dpa4.md
wanghan-iapcm: Fixed in [4021124](https://github.com/deepmodeling/deepmd-kit/commit/4021124336c8fce310a72cb54b688ec102c4b14e), and you were right on both counts — I checked the CLI and it is worse than a stale default: `--lower-kind` accepts only `nlist` and `graph` (`deepmd/main.py`, `default="nlist"`), there is no `auto` choice at all, and `freeze()` overrides any non-`graph` request to `graph` with a warning whenever `model_uses_graph_lower(m)`. The bullet now says every graph-capable DPA4/SeZM model including plain energy is graph-frozen, and that a dense artifact is not selectable through this entry point.

deepmd/pt/model/descriptor/sezm_nn/edge_cache.py
wanghan-iapcm: I would like to push back on this one, because the equivalence rests on the edge set being reciprocal and our supported topologies do not guarantee that.

Your reasoning holds exactly when for every edge `j -> k` the mirror `k -> j` also exists with the same length. That is true for a LAMMPS full neighbour list built purely by cutoff. It is not true for the nlist-derived edge schemas this same code serves: `format_nlist` caps each centre at `sel`, and when `sel` binds it truncates by distance per centre, so a dense-environment atom can drop a neighbour that the sparser neighbour keeps. Measured on a deliberately inhomogeneous system (one dense cluster plus a sparse tail, `rcut=4.0`, `sel=6`, ghosts mapped back to owner images):

```
sel = 6   edges = 90
NON-RECIPROCAL ordered pairs: 12   [(0,5), (6,2), (3,6), (0,9), (7,2), (3,9)]
```

So on that path `prod_{src=j} w` and `prod_{dst=j} w` are different numbers, not two spellings of one. That leaves only bad options: reduce over `dst` everywhere and the gate is silently redefined for sel-capped inputs (pt `forward_with_edges` is exactly this path, and it is what `test_sezm_parallel_bridging_parity.py` compares against), or reduce over `dst` only when `comm_dict` is set and the comm path stops computing the same quantity as the non-comm path — which is precisely the kind of split this branch is trying to remove, and it would make the folded-vs-parallel parity test tautological rather than meaningful.

The SFPG definition is source-keyed (`eta_j = prod_{e: src_e = j} w_e`), so I would rather keep one reduction that is correct for every supported input than a faster one that is correct only under an invariant nothing asserts.

Your cost point stands on its own though, and I do not want to lose it: one extra `border_op_backward` round trip with an `MPI_Barrier` per bridged force evaluation is real. The clean version of your optimisation is to establish reciprocity as an explicit, asserted property of the graph (which the LAMMPS-fed graph genuinely has) and then take the single-broadcast path under that guarantee, keeping the general reduction otherwise. That is a larger change than this PR should carry, so I would rather file it as a follow-up than bolt a conditional fast path on here. Happy to open that issue if you agree with the framing — and if you think the reciprocity invariant already holds by construction on every path that reaches this function, say so and I will re-examine, since that would change the conclusion.

source/lmp/tests/test_lammps_dpa4_spin_zbl_pt2.py
wanghan-iapcm: Agreed on the substance — I measured it and it matches your estimate: 421 of the 566 lines are byte-identical to `test_lammps_dpa4_spin_graph_pt2.py`, and the genuinely variant-specific parts are the archive, the geometry and a handful of ZBL assertions. Two copies of the MPI subprocess/parser/parity machinery will drift.

What I do not want to do is perform that extraction blind, and right now I cannot execute either file to check it. The `.pt2` fixtures are not present in this working tree, generating them needs AOTI compiles that hit a vectorised-scatter codegen assertion in the local torch build (workable only with `cpp.simdlen=1`), the multi-rank spin parity comparisons are additionally unreliable on this box at O(1e-4) for environmental reasons unrelated to this branch, and the GPU box I would normally use to adjudicate is unreachable at the moment. Restructuring 400+ lines of MPI harness — including the subprocess output schema and teardown — with only `--collect-only` as evidence is a worse trade than the duplication, especially since it would also rewrite a pre-existing file that is not otherwise touched by this PR.

So my preference is a dedicated follow-up PR that does the extraction and is validated on a machine that can actually run both files, rather than an unvalidated refactor here. If you would rather it land in this PR and are content to let the C++/LAMMPS CI jobs be the validation, say so and I will do it — I just want that trade made explicitly rather than by accident.

deepmd/dpmodel/model/native_spin_model.py
```diff
@@ -34,8 +34,10 @@ class NativeSpinModelKind:
     ``NativeSpinEnergyModel``) are parallel products with NO subclass
     relation between them -- an ``isinstance`` against one backend's
     concrete class is silently dead in the other. Backend seams that need a
-    cross-backend family test (e.g. the with-comm freeze gate: native-spin
-    lowers are single-rank only) test against this shared marker instead.
+    cross-backend family test test against this shared marker instead. The
```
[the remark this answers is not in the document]
wanghan-iapcm: Fixed in [975e17e](https://github.com/deepmodeling/deepmd-kit/commit/975e17e8543412d8152791205890196cb91e29ba) — leftover from removing the stale parenthetical in that sentence. I kept "test" as the noun and made the verb explicit: "Backend seams that need a cross-backend family test check against this shared marker instead."

deepmd/pt_expt/model/get_model.py
wanghan-iapcm: Thanks — I reproduced both of your findings exactly, but I think the exclusion half is attributed backwards, and fixing it the way you propose would entrench the thing that actually causes the divergence. Details, then a question.

**Reachability: confirmed.** `bridging_method` is declared in `sezm_model_args()`, not `standard_model_args()`, so strict validation rejects it under `type: "standard"`:

```
ArgumentKeyError: undefined key `bridging_method` is not allowed in strict mode
```

You are right that the new test only reaches the builder directly. Worth adding: dpmodel's `get_standard_model` (`deepmd/dpmodel/model/model.py:53`) has honored `bridging_method` the same way for as long as it has existed, and is unreachable from a validated config for exactly the same reason. So the backend divergence this wrapper was added to close was a divergence between two programmatic-only paths, not between a supported route and a broken one.

**Exclusions: reproduced, same numbers, different diagnosis.** With descriptor-only `exclude_types=[[0,1]]` on a 0.9 A Ni-O dimer I measure dE = 79.97 eV and max|dF| = 318.48 eV/A, matching you. But the mechanism is not a missing reconciliation in the standard route. Since #5733 there are two exclusion owners with two different SCOPES:

- model-level `pair_exclude_types` is a graph-BUILD transform, folded into the graph before the atomic model sees it, so it removes the pair from **every child of the composition, including the analytical InnerPotential term**;
- descriptor-level `exclude_types` is applied **inside** the descriptor forward via `apply_pair_exclusion(graph, atype, self.emask)` (`deepmd/dpmodel/descriptor/dpa4.py:1514`, twins at `dpa1.py:1866` and `repformers.py:622`), so it only removes the pair from the learned features.

`edge_cache.py:279` states the split outright: *"Model-level `pair_exclude_types` is a separate, graph-BUILD transform, already folded into the incoming graph."*

What `get_sezm_model` does at `deepmd/pt_expt/model/get_model.py:148-162` (and identically pt at `deepmd/pt/model/model/__init__.py:388-403`) is copy the descriptor-scoped key up into the model-scoped one, write it back down, and raise if both are set differently. That **promotion** is what removes the pair from ZBL and yields -0.909 eV. `get_standard_model` does not promote, so the network ignores the pair while ZBL still supplies its physical repulsion: +79.056 eV.

By the one-owner rule the unpromoted behaviour is the defensible one — descriptor-scoped means descriptor-scoped — and it is also physically meaningful: hiding a pair from the network while keeping its real nuclear repulsion is a legitimate configuration, which the promotion silently forecloses. So I would rather not copy the promotion into a second builder; that spreads the scope conflation instead of removing it.

Filed #5947 to deprecate descriptor-scoped `exclude_types` in favour of the single build-seam owner. Once the promotion is gone there is nothing left to reconcile and the two builders agree by construction, which is the outcome your review is asking for by a different route. It has to be staged (deprecation warning first, pt and pt_expt together) because dropping the promotion changes energies for existing dpa4 configs that set only the descriptor key.

**Question, since it is a public-schema call rather than a bug fix.** For the reachability half I see three options: (a) delete the bridging branch from `get_standard_model` in both pt_expt and dpmodel so `type: "dpa4"` is the single bridging owner; (b) keep it programmatic, and add a guard rejecting bridging for non-DPA4 descriptors; (c) what you asked for — add the bridging arguments to `standard_model_args()` with a DPA4-only runtime restriction and a regression through `argcheck.normalize`. (c) expands a shared, backend-agnostic schema, so it pulls in cross-backend consistency coverage and is really its own PR. Which do you want? I lean (a) for this PR with (c) as a follow-up if the `standard` + DPA4 route is worth supporting properly.

wanghan-iapcm: Resolved in [eb23860](https://github.com/deepmodeling/deepmd-kit/commit/eb238607e99709aaae94d3a057e5d7b113382a38), and the outcome is closer to your request than the option list I offered — a discussion with @wanghan-iapcm found a cleaner root cause than any of (a)/(b)/(c).

The observation: a bridged model **is** a linear composition. `bridging_method` returns a `LinearEnergyModel` over `[learned, InnerPotential]`, so a builder for a non-composite model type cannot honor it without returning a different kind of model than the one requested:

```
type "standard", no bridging -> DPA4EnergyModel
type "standard", + bridging  -> LinearEnergyModel   # not what was asked for
```

So `get_standard_model` now **rejects** `bridging_method` with a message pointing at `type: "dpa4"`. Rejecting rather than ignoring preserves the property the original audit item cared about — silently dropping a bridging term yields a physically different model — while leaving exactly one bridging owner per backend. Since `argcheck` never allowed the spelling anyway, no supported configuration changes behavior.

That also answers the exclusion half without needing the promotion copied anywhere: with one owner there is nothing to reconcile between two builders. The tests in `test_get_model_bridging.py` now pin the contract in both directions — the standard route rejects, the dpa4 route composes and injects the radii — and I kept the two composition-behavior regressions (`.pt` checkpoint eval, compile-probe tolerance) intact by moving them onto the dpa4 spelling.

Two follow-ups filed, both stemming from your review:

- #5947 — remove the `descriptor.exclude_types` -> `pair_exclude_types` promotion in `get_sezm_model` (pt and pt_expt together). Your ~80 eV measurement is the evidence in that issue.
- #5948 — the underlying design: `linear_ener` is already the composition axis, but `InnerPotential` has no config-level model type, which is *why* the flag exists. Registering it lets bridging be written explicitly as a composition, at which point `pair_exclude_types` naturally belongs to the linear model and the flag becomes sugar over one normalizer.

On your reachability point specifically: you were right that the route could not be reached from a validated `input.json`, and the fix is that it should not have existed on that builder rather than that the schema should grow to reach it. If you still think `standard` + DPA4 + bridging deserves to be a supported spelling, #5948 is the place to argue for it — it would come for free there as `type: "linear_ener"`.

deepmd/pt/model/descriptor/sezm_nn/edge_cache.py
OutisLi: Your non-reciprocal counterexample is valid for the sel-capped, nlist-derived PT compatibility path, so that path must retain the source-keyed two-phase completion. It does not apply to the production pt_expt graph-lower path, however. That path uses the carry-all, sel-free NeighborGraph (all edges inside `rcut`), and DPA4's dense `call()` rejects `comm_dict` before `_run_graph`; consequently the pt_expt communication hook is reached through `call_graph`, not through a sel-truncated dense graph.

For that graph-native route, the owner has the complete destination-centred neighbourhood. Reducing `[log_eta, zero_count]` over `dst` and performing one forward `border_op` is therefore the appropriate fast path, while the generic/nlist-derived PT route can keep its current source reduction plus reverse-accumulate/broadcast.

Please either (1) add this as a clean, explicitly carry-all/reciprocal graph fast path, without changing the generic source-keyed semantics, or (2) open a focused follow-up issue that records the graph contract, routing boundary, and required benchmark/parity validation. Graph lower is the primary route going forward, so I do not want the two-exchange/two-barrier implementation to become its only long-term path. I will leave this thread unresolved until the fast path lands or the tracking issue is linked.

source/lmp/tests/test_lammps_dpa4_spin_zbl_pt2.py
OutisLi: Please perform the extraction in this PR. I do not want to merge a newly added test module with 416 of its 566 lines duplicated and rely on an unspecified follow-up to remove that debt.

The repository already has `source/lmp/tests/lammps_test_utils.py`, which owns shared LAMMPS setup and MPI-runner mechanics (`make_spin_lammps`, `run_mpi_pair_runner`). Please either extend that existing seam or add a narrowly scoped shared DPA4-native-spin harness, while keeping the plain native-spin and spin+ZBL archives, geometries, and variant-specific assertions in their respective tests. The common live-reference setup, teardown, LAMMPS construction, magnetic-force gathering, subprocess/output parsing, and common parity assertions should have one owner.

It is acceptable for the C++/LAMMPS CI jobs to provide the end-to-end runtime validation; local collection/static checks can validate the refactor's imports and parametrization. Lack of a locally reliable MPI environment is not a reason to commit two copies of the harness. Please keep this thread unresolved until the shared structure lands.

OutisLi (commented):
The PT freeze comment still describes bridged SeZM as single-rank.

`deepmd/pt/entrypoints/freeze_pt2.py:951-956` says that bridging models report `supports_edge_parallel() == False`, that Source Freeze Propagation is not rank-decomposable, and that they fall back to single-rank. That is now the opposite of the contract implemented by this PR: bridged SeZM reports `True`, `_gate_partial_exchange()` completes the SFPG partials across ranks, and the surrounding predicate therefore embeds `forward_lower_with_comm.pt2` for the `edge_vec` route. The new PT capability tests also pin that behavior.

Please update this production comment together with the capability change. Leaving the explanation inverted next to the artifact-selection predicate makes the multi-rank contract particularly easy to misread during future maintenance.

doc/model/dpa4.md
```diff
@@ -454,18 +454,20 @@ route, and `--lower-kind` selects between the pt_expt lowers only.
...
-Native spin (`scheme: "native"`) and charge/spin conditioning are *not* in
-that list: native spin reuses the `edge_vec` interface on PT and the
-NeighborGraph lower on pt_expt, and both support multi-rank. The remainder of
-this subsection describes the multi-GPU launch recipe.
+**ZBL zone bridging is multi-rank capable** (including combined with native
+spin). The Source Freeze Propagation gate folds each node's full
+*outgoing*-edge set, which no single rank observes for ghost owners; the
+with-comm artifact completes the gate's per-node `[log eta, zero count]`
+partials across ranks with one reverse-accumulate plus one forward-broadcast
+border exchange of an `(N, 2)` tensor per forward pass — negligible next to
```
OutisLi: [P2] Do not label the SFPG synchronization negligible without measurements

The `(N, 2)` payload is narrow, but its cost is not determined by bandwidth alone. This path invokes `border_op_backward` followed by `border_op`, and both C++ kernels end in an `MPI_Barrier`; because the exported force graph differentiates through both operations, their registered transpose callbacks add the corresponding communication pair to the force evaluation as well. On small systems or at higher rank counts, the added round trips/global synchronizations can dominate despite the small tensor. This is also the hot path covered by the still-open graph fast-path thread. Please remove “negligible”, or support it with benchmarks across representative system sizes and MPI rank counts.

OutisLi (commented):
The DPA4 documentation makes an unsupported performance claim about the new SFPG communication path.

deepmd/pt_expt/model/get_model.py
```diff
@@ -164,36 +170,107 @@ def get_sezm_model(data: dict) -> EnergyModel:
...
-        composed = LinearEnergyAtomicModel(
-            models=[model.atomic_model, zbl_atomic],
-            type_map=data["type_map"],
-            weights="sum",
-            # Both exclusions belong to the composition: its children share one
-            # graph, so "excluded" must cover the analytical term too.
-            atom_exclude_types=data.get("atom_exclude_types", []),
-            pair_exclude_types=pair_exclude_types,
+
+def _compose_bridging(model: Any, data: dict, bridging_method: str) -> Any:
```
OutisLi: [P2] Keep the factory return contract truthful after composition

The new composition branch makes the public annotation on `get_sezm_model()` incorrect: the plain path returns `DPA4EnergyModel`, but the bridging path returns `LinearEnergyModel`, which is not an `EnergyModel`. Typing this helper and `get_standard_model()` as `Any` only hides that mismatch and discards useful structural information from the factory API. The PT twin already uses the honest common contract, `BaseModel`. Please type `get_sezm_model()` and `get_standard_model()` against `BaseModel`, and give this helper either its concrete `LinearEnergyModel` result or the same `BaseModel` contract, instead of erasing all three paths with `Any`.

OutisLi (commented):
The refactored DPA4/SeZM factory now has an internally inconsistent return-type contract.

source/lmp/tests/test_lammps_dpa4_spin_zbl_pt2.py
wanghan-iapcm: Done in [0408b38](https://github.com/deepmodeling/deepmd-kit/commit/0408b38863ff3f0cede417e1e5d363d3552eddbb) — you were right that the missing local MPI environment was not a reason to ship two copies.

Extracted to a new `source/lmp/tests/dpa4_spin_harness.py` owning the shared mechanics: the 4-atom NiO system, the live `DeepPot` reference (`compute_expected`), LAMMPS construction (`make_lammps`), `gather_force_mag`, the timeout-bounded MPI subprocess/parser (`run_mpi_spin_runner`), the rank-parity assertion (`assert_mpi_matches_single_rank`), and the two single-rank checks (`check_single_rank_energy_force`, `check_single_rank_virial`).

```
spin_graph  605 -> 230 lines
spin_zbl    566 -> 214 lines
```

Each module keeps only its archive, its geometry (empty-rank shift vs close-pair shift) and its variant-specific assertions and docstrings.

On the seam choice: I added a narrowly scoped harness rather than extending `lammps_test_utils`, per your second option. The three pieces that differ from what is already there are DPA4-native-spin-specific — `atom_modify map yes` (the graph route needs the LAMMPS atom map, which `make_spin_lammps` deliberately does not set), the `spin_norm / hbar` `fm` convention, and a runner that bounds every invocation and SIGKILLs the whole mpirun process group on expiry (`run_mpi_pair_runner` has no timeout). Folding those into the shared module would change behaviour for unrelated tests. Happy to move it if you would rather have one file.

Validation, as you allowed: `ruff check`/`ruff format` clean, byte-compile clean, and `pytest --collect-only` yields the same 8 test IDs as before the refactor (plus the two neighbouring DPA4 modules whose prose references to the moved helper I repointed). Runtime validation is the C++/LAMMPS CI jobs. One drive-by improvement fell out of it: the graph module’s two MPI tests had inlined the assertion block that the ZBL module had already factored into a helper, so they now share it.

deepmd/pt/model/descriptor/sezm_nn/edge_cache.py
wanghan-iapcm: Tracked as #5957, per your option (2), and I have kept the generic source-keyed path unchanged in this PR.

The issue records exactly what you laid out: the carry-all/sel-free `NeighborGraph` used by the pt_expt graph lower is reciprocal, DPA4’s dense `call()` rejects `comm_dict` before `_run_graph` so the hook is only reachable via `call_graph`, and therefore the owner holds its complete destination-centred neighbourhood — making `dst` reduction plus a single forward `border_op` the right fast path there. It also carries the counterexample that keeps the nlist-derived route on the two-phase completion, so the boundary is written down rather than implied.

Requirements captured: the fast path must be explicitly labelled carry-all/reciprocal and must not change the generic semantics; a negative test that the sel-capped route still takes the two-phase completion, so the fast path cannot silently capture it; bit-for-bit parity against the current implementation including a bridging-zone pair straddling a rank boundary; and the benchmark across system sizes and rank counts.

Related: your other thread about the "negligible" wording is fixed in [2e7865c](https://github.com/deepmodeling/deepmd-kit/commit/2e7865c71) — the claim is removed rather than restated, and #5957 notes that numbers from that benchmark are what should replace it.

deepmd/pt_expt/model/get_model.py
wanghan-iapcm: Fixed in [2e7865c](https://github.com/deepmodeling/deepmd-kit/commit/2e7865c719953097003d796141c834d53955245e). `LinearEnergyModel` is indeed not an `EnergyModel` (verified: `issubclass(LinearEnergyModel, EnergyModel)` is `False`, `issubclass(..., BaseModel)` is `True`), so the annotation was wrong on the bridging path rather than merely imprecise.

`get_sezm_model()` and `get_standard_model()` are now typed against `BaseModel`, matching the PT twin, and `_compose_bridging()` declares its concrete `LinearEnergyModel` result (via `TYPE_CHECKING` to avoid an import cycle). The numpydoc `Returns` sections were updated to match, and `Any` is gone from the module.

doc/model/dpa4.md
wanghan-iapcm: Fixed in [2e7865c](https://github.com/deepmodeling/deepmd-kit/commit/2e7865c719953097003d796141c834d53955245e). I removed the claim rather than trying to support it — the exchange has not been benchmarked, and your point about the transposes is the part I had not accounted for. The text now states that both kernels end in an `MPI_Barrier`, that the force graph differentiates through them so their transposes add a matching pair of round trips to the force evaluation, and that on small systems or at high rank counts those synchronizations can dominate despite the narrow payload.

#5957 (the graph fast-path issue) records that benchmark numbers across system sizes and rank counts are what should replace the wording.

wanghan-iapcm:
@OutisLi — the PT freeze comment is fixed in [2e7865c](https://github.com/deepmodeling/deepmd-kit/commit/2e7865c719953097003d796141c834d53955245e). `deepmd/pt/entrypoints/freeze_pt2.py` no longer says bridging reports `supports_edge_parallel()=False` and falls back to single-rank; it now states that bridging models DO take the with-comm path because the SFPG per-node partials are completed across ranks by `DescrptSeZM._gate_partial_exchange`, leaving only virtual spin (the nlist interface) excluded. You were right that leaving the explanation inverted next to the artifact-selection predicate was the dangerous part.

All four points from your 2026-08-01/08-02 rounds are now addressed:

| thread | resolution |
|---|---|
| PT freeze comment | fixed, [2e7865c](https://github.com/deepmodeling/deepmd-kit/commit/2e7865c719953097003d796141c834d53955245e) |
| "negligible" SFPG claim | claim removed, [2e7865c](https://github.com/deepmodeling/deepmd-kit/commit/2e7865c719953097003d796141c834d53955245e) |
| factory return contract | typed against `BaseModel`, [2e7865c](https://github.com/deepmodeling/deepmd-kit/commit/2e7865c719953097003d796141c834d53955245e) |
| LAMMPS harness duplication | extracted, [0408b38](https://github.com/deepmodeling/deepmd-kit/commit/0408b38863ff3f0cede417e1e5d363d3552eddbb) |
| SFPG graph fast path | tracked as #5957 (your option 2) |

Two follow-ups from the earlier round remain filed rather than fixed here: #5947 (drop the `descriptor.exclude_types` promotion) and #5948 (express bridging as an explicit `linear_ener` composition, which subsumes #5947).

deepmd/pt_expt/model/get_model.py
```diff
@@ -164,36 +172,109 @@ def get_sezm_model(data: dict) -> EnergyModel:
...
+
+
+def get_standard_model(data: dict) -> BaseModel:
+    """Build a pt_expt standard model: one descriptor plus one fitting net.
+
+    ``bridging_method`` is rejected here rather than honored. Analytical
+    bridging is a COMPOSITION -- it yields a ``LinearEnergyModel`` over
+    ``[learned, InnerPotential]`` -- so a builder that accepted it would
+    return a model of a different kind than the one requested. pt_expt
+    keeps exactly one bridging owner, :func:`get_sezm_model`
```
OutisLi, line 244: [P2] Update the stale native-spin bridging ownership docstring

This new one-owner contract is contradicted by `source/tests/pt_expt/model/test_zbl_bridging.py::TestNativeSpinWithBridging`, whose docstring still says `get_standard_model` owns and returns the bridging composition. Under the code here `get_standard_model` now rejects bridging, while `get_native_spin_model` routes DPA4/SeZM to `get_sezm_model` before re-classing the result. Leaving the regression suite with the opposite ownership description makes the intended boundary internally inconsistent and invites the rejected `standard + bridging` path to be reintroduced. Please update that docstring to name `get_sezm_model` as the composition owner and describe the native-spin wrapper as re-classing that returned composition.

source/lmp/tests/test_lammps_dpa4_spin_graph_pt2.py
```diff
@@ -69,149 +71,53 @@
...
+# the x=30 slab boundary of a [0, 90] box.  With 3 even x-slabs of width
+# 30: rank 0 owns [0, 30) (all 4 atoms), rank 1 ([30, 60)) owns NOTHING but
+# receives ghosts of the two Ni atoms (3.2 and 3.9 from its lower
+# boundary), and rank 2 ([60, 90)) owns nothing but receives periodic
+# ghosts of the x < 6 O atoms (x=3.51 and x=4.27, wrapped around the box's
+# x=90/x=0 seam) -- BOTH atom-less ranks carry ghosts, so both take the
+# phantom path and none trips the genuinely-empty fail-fast.  The shifted
+# coordinates (``coord_empty_rank``) are defined below, after ``coord``.
+data_file_empty_rank = Path(__file__).parent / "data_dpa4_spin_graph_pt2_empty_rank.lmp"
+data_file_empty_rank = Path(__file__).parent / "data_dpa4_spin_graph_pt2_empty_rank.lmp"
```
OutisLi: [P3] Remove the duplicated fixture-path assignment

The refactor introduces this exact `data_file_empty_rank` assignment twice in succession. It is a runtime no-op, but it is an obvious patch artifact in a commit whose purpose is to establish a clean single-owner test harness. Please keep a single definition.

deepmd/pt_expt/model/get_model.py
wanghan-iapcm: Fixed in [285e3c6](https://github.com/deepmodeling/deepmd-kit/commit/285e3c67ce1fa22a13389a1c4e3414efde41db6f). `TestNativeSpinWithBridging`'s docstring now names `get_sezm_model` as the composition owner, notes that `get_standard_model` rejects `bridging_method` (a composition is not expressible on a non-composite model type), and describes the native-spin wrapper as re-classing the composition `get_sezm_model` returns.

Verified the description against the code rather than just rewording it: this file's `ZBL_CONFIG` carries `"type": "dpa4"`, so `get_native_spin_model`'s `model_type in ("dpa4", "sezm")` branch selects `get_sezm_model` as the backbone builder — the route the docstring now describes is the one the test actually exercises. 18 passed / 2 skipped in that file.

source/lmp/tests/test_lammps_dpa4_spin_graph_pt2.py
wanghan-iapcm: Fixed in [285e3c6](https://github.com/deepmodeling/deepmd-kit/commit/285e3c67ce1fa22a13389a1c4e3414efde41db6f) — you read it correctly, it was a patch artifact. My extraction script preserved the original comment block by line range, and that range already ended with the `data_file_empty_rank` assignment, while the rewrite template emitted it again. Single definition now; collection still yields the same 4 test IDs for that module.

OutisLi (approved):
Reviewed the latest HEAD. The previously raised correctness, performance-documentation, type-contract, and code-structure issues have been addressed.

PR #5944, opened 2026-07-31
perf: bound memory in model testing and output statistics

OutisLi:
## Summary

- evaluate `dp test` in atom-bounded chunks, with lazy on-demand decoding for LMDB inputs
- preserve run-level MAE/RMSE aggregation and detail output across chunks while consolidating backend-independent test logic under `deepmd.infer.model_test`
- build the neighbor representation declared by each atomic model during output-bias statistics, using `NeighborGraph` for graph-native models and the existing dense neighbor list for other models

## Why

Large LMDB datasets were fully decoded before frame selection, so even a small test run could require loading tens of millions of frames. Separately, graph-native atomic models were routed through a fixed-capacity dense neighbor list while calibrating output bias; because those models do not declare a finite neighbor capacity, that path could allocate tens of gigabytes and fail before fine-tuning began.

This change bounds LMDB decoding and model-evaluation memory while preserving the existing evaluation and output-statistics contracts. Ordinary `DeepmdData` inputs still materialize one test system before chunked evaluation.

## Additional correctness fixes

- preserve legacy TensorFlow spin-model dispatch and spin virial/stress reporting
- use canonical atomic tensor labels consistently across NPY and LMDB inputs
- flatten atomic parameters correctly for graph-native output statistics
- keep LMDB frame selection deterministic and honor availability subgroups during iteration
- keep detail files isolated across systems and emit a single header across chunks
- require atomic property labels when atomic metrics are requested

## Checks

- `ruff check .`
- `ruff format --check .`
- targeted common model-test and LMDB regression tests
- targeted PyTorch streaming/output-stat regression tests
- `git diff --check`

The TensorFlow atomic dipole/polar integration cases could not be run locally because this checkout has no built TensorFlow backend; they are left to CI.

## Summary by CodeRabbit

* **New Features**
  * Expanded model evaluation for energy, spin, DOS, property, dipole, and polarizability models.
  * Added chunked testing, frame limits, per-frame details, and atomic-level metrics.
* **Improvements**
  * Large LMDB test datasets now load frames lazily for improved efficiency.
  * Added clearer weighted aggregation of MAE and RMSE results.
  * Graph-capable models now use graph-based neighbor representations during inference.
* **Bug Fixes**
  * Improved handling of periodic systems, mixed atom counts, exclusions, optional outputs, missing labels, and atomic label naming.


deepmd/infer/model_test/tensor.py
```diff
@@ -0,0 +1,157 @@
...
+        sel_type = self.dp.get_sel_type()
+        sel_natoms = int(sum(sum(atype == ii) for ii in sel_type))
+
+        if self.atomic:
+            sel_mask = np.isin(atype, sel_type)
+            prediction = prediction.reshape((nframes, -1, self.ndof))[
+                :, sel_mask, :
+            ].reshape((nframes, -1))
+            reference = (
+                test_data[self.atomic_label]
```
wanghan-iapcm, line 76: Blocking. This raises `KeyError` and CI is red on it.

`DeepmdData.add()` normalises `atomic_*` keys to `atom_*` — that normalisation *is* fix #5082 (`bf3e483ca`, "Fix: dptest label mismatch"), added precisely because the label names disagreed. So `get_test()` returns `atom_dipole` and `atom_polarizability`, which is what the old `test.py` read. Here `atomic_label` is set to [`"atomic_dipole"`](https://github.com/deepmodeling/deepmd-kit/blob/5ea2b6605f0b30acb8f01a9ea33f7943af17c5c4/deepmd/infer/model_test/tensor.py#L131-L133) and [`"atomic_polarizability"`](https://github.com/deepmodeling/deepmd-kit/blob/5ea2b6605f0b30acb8f01a9ea33f7943af17c5c4/deepmd/infer/model_test/tensor.py#L147-L149), and this line looks up the un-normalised name.

From the CI run on this commit:

```
FAILED source/tests/tf/test_dp_test.py::TestDPTestDipole::test_1frame - KeyError: 'atomic_dipole'
FAILED source/tests/tf/test_dp_test.py::TestDPTestPolar::test_1frame - KeyError: 'atomic_polarizability'
deepmd/infer/model_test/tensor.py:78: KeyError
```

Both `Test Python` jobs fail. The fix is to use `atom_dipole`/`atom_polarizability`, matching what `add()` produces.

deepmd/infer/model_test/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+    ModelTester
+        A tester able to evaluate one system of that model class.
+
+    Raises
+    ------
+    RuntimeError
+        If no tester covers the model class.
+    """
+    if isinstance(dp, DeepPot):
+        tester = SpinEnerTester if dp.has_spin else EnerTester
```
wanghan-iapcm: Blocking. This silently changes the reported numbers for legacy TF spin models.

The old dispatcher used `out_put_spin = dp.get_ntypes_spin() != 0 or dp.has_spin`. This keeps only the second term. TF's `DeepEval` never overrides `get_has_spin`, so it inherits the base `return False` ([`deepmd/infer/deep_eval.py`](https://github.com/deepmodeling/deepmd-kit/blob/5ea2b6605f0b30acb8f01a9ea33f7943af17c5c4/deepmd/infer/deep_eval.py#L420-L424)), while `deepmd/tf/infer/deep_eval.py` does report a non-zero `get_ntypes_spin()` for those models. They therefore now route to `EnerTester`.

The consequence is not an error but wrong output: `mae_f`/`rmse_f` get computed over real *and* virtual spin atoms together instead of `mae_fr`/`rmse_fr` plus `mae_fm`/`rmse_fm` over the real subset, and the detail file is written as `.f.out` rather than `.fr.out`/`.fm.out`. A user comparing against a previous run sees different force errors with nothing to indicate why.

The tell that this is an oversight rather than a decision is that [`_align_spin_force_arrays`](https://github.com/deepmodeling/deepmd-kit/blob/5ea2b6605f0b30acb8f01a9ea33f7943af17c5c4/deepmd/infer/model_test/ener.py#L74-L78) still carries its `if dp.get_ntypes_spin() != 0:  # old tf support for spin` branch, which is now unreachable. Restoring the `or dp.get_ntypes_spin() != 0` term here makes that branch live again.

deepmd/dpmodel/atomic_model/base_atomic_model.py
```diff
@@ -841,31 +854,57 @@ def model_forward(
...
+                    atype,
+                    box,
+                    self.get_rcut(),
+                    pair_excl=self.pair_excl,
+                )
+                atomic_ret = self.forward_common_atomic_graph(
+                    graph,
+                    xp.reshape(atype, (-1,)),
+                    fparam=fparam,
+                    aparam=aparam,
```
wanghan-iapcm: Blocking. `aparam` is forwarded with the wrong rank on the new graph route.

Note the asymmetry on the line above: `atype` is flattened with `xp.reshape(atype, (-1,))`, but `aparam` is passed through as-is. `_compute_model_predict` hands this wrapper a rectangular `(nf, nloc, nda)` aparam, while `forward_common_atomic_graph` documents it as flat `(N, nda)` on the node axis — and the reference caller in [`make_model.py`](https://github.com/deepmodeling/deepmd-kit/blob/5ea2b6605f0b30acb8f01a9ea33f7943af17c5c4/deepmd/dpmodel/model/make_model.py#L555-L563) does `xp.reshape(ap, (nf * nloc, ap.shape[-1]))` before calling it.

Nothing downstream rescues it. `DPAtomicModel.forward_atomic_graph` only reshapes when `graph.n_local is not None`, and `build_neighbor_graph` — called four lines above — leaves `n_local` at its `None` default. So the rank-3 array reaches `GeneralFitting.call_graph`, which rejects it explicitly ("graph-route aparam must be flat (N, nda) on the node axis").

Any graph-eligible model with `numb_aparam > 0` therefore raises `ValueError` during output-bias statistics — i.e. the new path fails on exactly the models it was written for, whenever aparam is in use. `fparam` is `(nf, ndf)` on both routes so it is unaffected. Reshaping `aparam` to `(nframes * nloc, -1)` alongside the `atype` flatten should be all it needs.

deepmd/utils/data.py
```diff
@@ -338,6 +341,44 @@ def get_test(self, ntests: int = -1) -> dict:
...
+        ------
+        dict
+            One chunk of test data, keyed as :meth:`get_test`.
+        """
+        if not hasattr(self, "test_set"):
+            self._load_test_set(self.shuffle_test)
+        total = int(self.test_set["type"].shape[0])
+        if np.isfinite(numb_test):
+            total = min(total, int(numb_test))
+        step = max(1, int(chunk_atoms) // max(1, self.natoms))
```
wanghan-iapcm, line 384: Blocking, and this is the one that explains the other three.

Nothing in the suite sets `DP_TEST_CHUNK_ATOMS`, and nothing asserts that a multi-chunk run reproduces the single-shot errors or writes the same detail files. Every test uses one or two frames against a default chunk of 1M atoms, so `step` is always large enough to swallow the whole system and CI only ever walks the single-chunk path. `LmdbTestData.iter_test` has no direct coverage either.

That means the streaming behaviour this PR exists to add is the one thing the tests do not exercise, and it is why the two silent regressions above (the spin dispatch and the aparam rank) reach HEAD with green-looking local checks — combined with "Tests were not rerun in this checkout at the author's request", nothing was in a position to catch them.

What would close it: one test that runs a system twice, once with `DP_TEST_CHUNK_ATOMS` set below `natoms` so the system splits, once at the default, and asserts the reported MAE/RMSE match exactly. Given the aggregation math is exact rather than approximate, that comparison can be an equality rather than a tolerance, which makes it a sharp test. Extending it to assert the detail files match would also catch the header issue I have flagged separately.

deepmd/infer/model_test/ener.py
```diff
@@ -0,0 +1,703 @@
...
+        (
+            np.reshape(reference_energy, [-1, 1]),
+            np.reshape(prediction_energy, [-1, 1]),
+        ),
+        axis=1,
+    )
+    save_txt_file(
+        detail_path.with_suffix(".e.out"),
+        pe,
+        header=f"{system}: data_e pred_e",
```
wanghan-iapcm, line 166: Non-blocking, but easy to fix while you are in here.

`save_txt_file` does `np.savetxt(fp, data, header=header)` and opens in `"a"` mode when appending, and `np.savetxt` writes the header at the start of *every* call — it has no notion of the file already having one. This used to be called once per system; it is now called once per chunk, and every call passes the same `header=`.

So a system spanning three chunks produces:

```
# sys/foo: data_e pred_e
-1.234 -1.230
# sys/foo: data_e pred_e     <- chunk 2, mid-file
-2.345 -2.341
# sys/foo: data_e pred_e     <- chunk 3
-3.456 -3.450
```

`np.loadtxt` skips `#` lines anywhere, so the repo's own round-trip is fine and this is invisible under the 1M-atom default. But it only takes a system big enough to split — which is the case this PR is for — and any hand-rolled parser that treats line 1 as the header and the rest as data gets garbage rows. Passing `header=""` when `append` is true confines it to the chunk that creates the file. Same applies to the `.f.out`, `.fr.out`, `.v.out` and per-atom variants below.

wanghan-iapcm (changes requested):
The design here is good and I want to say that first, because most of what follows is about things lost in the move rather than the idea. I checked the aggregation math, since everything else rests on it: MAE accumulates `sum(e_i * w_i)` over `sum(w_i)` and RMSE accumulates `sum(e_i^2 * w_i)` before the root, so chunked evaluation is exactly equal to one-shot, and a merged `(error, weight)` really is reusable as a further partial — the `merge_weighted_errors` docstring's claim holds. Frame order and `numb_test` semantics are preserved too: `iter_test` serves the same prefix `get_test()` did, with `shuffle_test` still applied once inside `_load_test_set`. Of the three incidental fixes claimed, two are real and correct — spin models now report virial/stress, and `append_detail = bool(err_coll)` genuinely fixes LMDB nloc-subgroups overwriting each other's detail files.

The problem is that a 1470-line move carried three regressions across, and the feature the PR exists to add has no test to have caught them. Four blocking comments inline, one non-blocking.

The third claimed fix, "restore reachable wave-function model dispatch", is inverted rather than wrong in a way that matters: `build_tester` has no `DeepWFC` branch and `test_wfc` was deleted, so WFC testing now raises `RuntimeError`. That is fine as an outcome — the old dispatcher had no reachable `DeepWFC` branch either, so `test_wfc` was already dead code and the explicit error beats the old `UnboundLocalError` — but the body describes it as a restoration when it is a removal. Worth correcting the sentence.

Two smaller things I am not asking you to change, recorded so they are not lost. The `ModelTester` and `deepmd/infer/model_test/__init__.py` docstrings say chunking is "what makes a dataset larger than memory testable", stated across all backends; that is true only for the lazy `LmdbTestData` path, and [`data.py`'s own `iter_test` docstring](https://github.com/deepmodeling/deepmd-kit/blob/5ea2b6605f0b30acb8f01a9ea33f7943af17c5c4/deepmd/utils/data.py#L352-L356) is honest that "a set is loaded as a whole, so chunking here bounds what a consumer holds at once rather than what is read". The two contradict each other and the narrower one is right. Separately, [`LmdbTestDataNlocView.iter_test`](https://github.com/deepmodeling/deepmd-kit/blob/5ea2b6605f0b30acb8f01a9ea33f7943af17c5c4/deepmd/dpmodel/utils/lmdb_data.py) delegates on `nloc` alone while `get_test` honours `self._frame_indices`; no in-tree caller passes `frame_indices` today so it is latent, but the two accessors now disagree and the next caller gets the wrong frames silently.

One point I considered raising and decided against: routing on `uses_graph_lower()` means output bias for dpa1/dpa2 now comes from the sel-free graph builder rather than the `get_sel()`-truncated dense list, so the value changes for existing models. dpa1's own docstring already discloses that divergence and this only affects bias initialisation, so I do not think it blocks — but gating on the absence of a finite `sel`, which is the actual motivation, would keep the change to the models that need it.

deepmd/infer/model_test/tensor.py
OutisLi: Fixed in d2756cd. `DipoleTester.atomic_label` and `PolarTester.atomic_label` now use the canonical `atom_dipole` and `atom_polarizability` keys produced by `DeepmdData.add()` / `get_test()`. I also normalized raw LMDB `atomic_*` / `find_atomic_*` keys at LMDB ingestion so both input formats obey the same canonical contract.

deepmd/infer/model_test/__init__.py
OutisLi: Fixed in d2756cd. `build_tester()` now selects `SpinEnerTester` when either `dp.has_spin` or `dp.get_ntypes_spin() != 0`, restoring the legacy TF behavior. I also added a focused regression for `get_has_spin() == False` with a nonzero `get_ntypes_spin()`.

deepmd/dpmodel/atomic_model/base_atomic_model.py
OutisLi: Fixed in d2756cd. The graph output-stat route now reshapes `aparam` to `(nframes * nloc, dim_aparam)` before `forward_common_atomic_graph`, matching the reference call in `make_model.py`. A DPA4 regression now exercises output-stat graph forwarding with `numb_aparam > 0`.

deepmd/utils/data.py
OutisLi: Agreed. I added a focused PT regression in d2756cd that runs the same three-frame stress case once as a single chunk and once with `DP_TEST_CHUNK_ATOMS=192`, then compares every weighted error, the numeric detail contents, and one-header-per-file behavior. LMDB iteration also now has direct coverage for selected frame subgroups, frame limits, and required labels.

deepmd/infer/model_test/ener.py
OutisLi: Fixed in d2756cd. `save_txt_file()` now suppresses the header only when appending to an existing non-empty file, so later chunks do not insert mid-file headers while a new or empty detail file still receives its header. The multi-chunk regression asserts exactly one header in every combined detail file.

source/tests/pt/test_dp_test.py
```diff
@@ -402,6 +405,9 @@ def _prepare_virial_system(self) -> str:
         tmp_dir = tempfile.mkdtemp()
         shutil.copytree(src, tmp_dir, dirs_exist_ok=True)
         set_dir = Path(tmp_dir) / "set.000"
+        for data_file in set_dir.glob("*.npy"):
```
iProzd, line 408: Blocking: repeating every fixture array to three frames also affects the three inherited `DPTest` methods. They call `dp_test` with `numb_test=0`, so the detail file contains all three frames, but `_run_dp_test` still compares it with a single validation batch. The previous PyTorch CI therefore got 576 force rows versus 192 in all three inherited stress tests, and neither this fixture nor the helper changed in the follow-up. Please either constrain those inherited checks to one frame or compare against predictions for all three frames, while keeping `test_stress`'s explicit three-frame chunking coverage.

iProzd (changes requested):
Thanks for addressing grouped tensor detail paths and zero-selected-atom handling; both prior runtime blockers are resolved on 01fe8fdb. One blocking issue remains: the new three-frame stress fixture still affects the inherited `DPTest` methods, which run all frames but compare the detail output against a single validation batch. The prior CI failure—576 force rows versus 192 in three tests—is therefore unchanged; the follow-up commit did not touch this fixture or helper. Please fix this case and add or adjust the focused regression before merging.

source/tests/pt/test_dp_test.py
OutisLi: Fixed in b014db81c. `_run_dp_test()` now defaults to `numb_test=1`, matching the single validation batch used as its reference. This constrains the three inherited checks to one frame, while `test_stress()` still explicitly runs all three frames and preserves the chunking coverage. All four `TestDPTestStress` tests pass locally.

njzjz (approved):
## Approve

I reviewed the current head (`b014db81c`) in full, including the earlier review threads and the fresh diff, and I am approving.

### Core design is sound
- **Memory-bound chunked testing**: `ModelTester.run` walks each system via `data.iter_test(chunk_atoms=DP_TEST_CHUNK_ATOMS)`; the weighted aggregation accumulates `sum(e_i * w_i)` / `sum(e_i^2 * w_i)` before rooting, so a merged (error, weight) partial is reusable and chunked MAE/RMSE equals the one-shot result exactly.
- **LMDB**: `_select_frames`/`_read_frames` do frame selection by index arithmetic from metadata and decode only retained frames, grouping by nloc while honoring availability subgroups — this is what fixes the 38.8M-frame/54GB all-into-memory pathology with the GPU idle.
- **Output-bias stats**: graph-native atomic models now route through a carry-all `NeighborGraph` (`build_neighbor_graph`) instead of the fixed-capacity dense neighbor list (`get_sel`-sized), removing the unbounded-allocation failure mode.

### Blocking issues from earlier rounds are all resolved
- Canonical `atom_dipole` / `atom_polarizability` label keys (were `atomic_*` → KeyError) — fixed.
- Legacy TF spin dispatch restored (`build_tester` selects `SpinEnerTester` on `has_spin` *or* `get_ntypes_spin() != 0`) with a regression — fixed.
- `aparam` flattened to `(nframes*nloc, nda)` before `forward_common_atomic_graph` on the graph route — fixed.
- Repeated mid-file detail headers suppressed (header written only on a new/empty file) with a single-header-per-file multi-chunk assertion — fixed.
- Ragged tensor detail file across differing `sel_natoms` groups and zero-selected-atom handling — fixed.
- Three-frame stress fixture breaking the inherited `DPTest` methods (576 vs 192 force rows) — fixed via `_run_dp_test` defaulting to `numb_test=1`.

### Verification
- Full CI is green at head: Test Python (all 12 shards), Test C++ (all 4), CUDA/rocm builds, CodeQL, and codecov all pass.

Remaining thread comments are non-blocking (dead `siz_coll` list; a `must=True` atomic-label improvement note; CodeQL unused-`log` lint). Nothing blocks merge.

iProzd (approved):
Re-reviewed the current head after the test follow-up. Setting the inherited `dp test` checks to one frame now aligns the generated detail files with the direct single-batch model result, resolving the three-frame 576-versus-192 row mismatch. The earlier grouped tensor detail-path and zero-selected-atom issues remain fixed, the obsolete size collection was removed cleanly, and I found no new blocker.

wanghan-iapcm (approved):
All five points are addressed, and I checked each against this head rather than the replies.

- `DipoleTester` and `PolarTester` now use `atom_dipole` / `atom_polarizability`, the keys `DeepmdData.add()` actually produces. CI confirms it independently: all thirteen `Test Python` jobs pass where two were red on the `KeyError`.
- `build_tester` restores `dp.has_spin or dp.get_ntypes_spin() != 0`, so legacy TF spin models route to `SpinEnerTester` again and the `_align_spin_force_arrays` branch is reachable.
- The graph route reshapes `aparam` to `(nframes * nloc, dim_aparam)`, matching the reference call in `make_model.py`.
- `save_txt_file` suppresses the header only when appending to an existing non-empty file.

On the fourth point, the chunk regression is the one I care most about, so I checked whether it actually splits rather than trusting the environment variable. `_prepare_virial_system()` repeats `water/data/single` three times, and that system has 192 atoms, so at `DP_TEST_CHUNK_ATOMS=192` the step is `192 // 192 = 1` and the chunked run really does process three single-frame chunks against the unchunked reference. It is not a test that passes because the chunk swallowed everything.

It is also a genuinely falsifying test rather than a description of intended behaviour. Comparing the weight alongside the value (`single_err[key][1]`) matters, because an aggregation bug that scaled numerator and denominator together would survive a value-only comparison. And the one-header assertion fails outright on the previous code -- three chunks meant three headers -- so that single test pins both the chunking contract and the header fix. The DPA4 case does the same job for the aparam rank: passing a rectangular `(nf, nloc, 2)` array through `_get_forward_wrapper_func()` hit `GeneralFitting.call_graph`'s flat-node-axis rejection before this change.

Two things you did beyond what I raised are worth acknowledging: normalizing raw LMDB `atomic_*` / `find_atomic_*` keys at ingestion so both input formats obey one canonical contract addresses the root of the first issue rather than its symptom, and the direct LMDB iteration coverage for frame subgroups and limits closes a gap I had not asked about.

Approving. Thanks for taking the coverage point seriously rather than patching the three symptoms and moving on -- the missing chunked-path test was the reason all three shipped together, and it is now the thing standing in the way of a repeat.

PR #5958, opened 2026-08-04
feat(pt_expt): align the training runtime with pt

OutisLi:
## Summary

- move checkpoint layout, retention, training timing, and sharding policy into backend-independent training utilities shared by `pt` and `pt_expt`
- add `pt_expt` support for checkpoint directories and retention ratios, EMA training/checkpoints, EMA full validation, consistent training reports, and restart-safe state restoration
- support the same `zero_stage` strategies as `pt`, including DDP, ZeRO-1, and FSDP2, with collective checkpoint assembly and optimizer-state restoration
- reuse overflow-safe gradient norm reduction and defer the non-finite verdict to checkpoint boundaries so a diverged model is not saved
- make EMA checkpoint retention inherit `max_ckpt_keep` by default while preserving an explicit `ema_ckpt_keep` override

## Motivation

The `pt_expt` trainer currently lacks several operational guarantees available in `pt`: equivalent checkpoint retention and restart behavior, EMA support, distributed state sharding, stable gradient checks, and consistent progress reporting. Implementing these separately would leave two training runtimes with duplicated policies that can drift.

This PR keeps backend-specific serialization and execution in each trainer, while centralizing the policies that are independent of a backend. It also relocates the EMA, validation, and gradient helpers under `pt_expt`, which is their continuing owner as the legacy `pt` trainer is retired.

## Notable fixes

- rerunning in a directory from a longer run no longer lets stale future checkpoints evict the newly written checkpoint
- `max_ckpt_keep < 1` retains all checkpoints instead of deleting the current checkpoint
- `ckpt_keep_ratio` works when periodic saving is disabled and overrides both regular and EMA windows
- restarting from a checkpoint without optimizer state resumes the learning-rate schedule from the recorded step
- sharded checkpoints are assembled collectively, avoiding rank desynchronization at the next barrier
- regular and EMA checkpoint families are pruned independently; absent `ema_ckpt_keep` now gives both families the `max_ckpt_keep` window

## Validation

- `OMP_NUM_THREADS=1 DP_INTER_OP_PARALLELISM_THREADS=0 DP_INTRA_OP_PARALLELISM_THREADS=0 /Users/outisli/Software/miniforge3/envs/dpmd/bin/python -m pytest source/tests/common/dpmodel/test_train_checkpoint.py source/tests/common/test_argcheck_training.py -q` (`14 passed`)
- a five-step `pt_expt` training run with EMA, `save_freq=1`, and only `max_ckpt_keep=2` retained regular steps `4, 5` and EMA steps `4, 5`
- repository pre-commit checks passed

## Summary by CodeRabbit

* **New Features**
  * Added configurable checkpoint retention, restart-safe cleanup, and separate EMA checkpoints.
  * Added distributed training with multiple sharding strategies and improved checkpoint restoration.
  * Added EMA full-validation workflows with independent schedules and best-checkpoint tracking.
  * Added parameter-count reporting and improved training progress timing, averages, and ETA estimates.
  * Added safer handling of non-finite gradients.
  * Added support for checkpoint retention and EMA features across supported PyTorch backends.

* **Documentation**
  * Expanded guidance for checkpoint retention, EMA, distributed training, and inference options.

* **Bug Fixes**
  * Improved validation compatibility and handling of shared or relocated save directories.


source/tests/pt_expt/test_entrypoint.py
```diff
@@ -7,6 +7,7 @@
 from deepmd.pt_expt.entrypoints.main import (
     PTExptTrainEntrypoint,
     _ensure_pt_expt_model_suffix,
+    _ensure_stat_file_path,
```
wanghan-iapcm: Blocking. This import does not resolve, and it takes the entire test suite down with it.

`_ensure_stat_file_path` does not exist in [`deepmd/pt_expt/entrypoints/main.py`](https://github.com/deepmodeling/deepmd-kit/blob/775b68f405d820c8aba632cb705c40a6a96bbf06/deepmd/pt_expt/entrypoints/main.py) at this head -- the module defines `_ensure_pt_expt_model_suffix`, `_update_changed_model_tensors`, `_detect_lmdb_path`, `_get_neighbor_stat_data`, `_build_data_system`, `get_trainer`, `train`, `_default_output_path`, `freeze`, `change_bias` and `main`, and no `_ensure_stat_file_path`. That file is not in this PR's diff, so nothing adds it. It is also used further down at line 157 (`stat_path = _ensure_stat_file_path(str(stat_file))`), so removing just the import will not be enough.

Because pytest fails at collection rather than at test time, this does not fail one test -- it aborts the whole shard:

```
ImportError while importing test module '.../source/tests/pt_expt/test_entrypoint.py'.
E   ImportError: cannot import name '_ensure_stat_file_path' from 'deepmd.pt_expt.entrypoints.main'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
```

All twelve `Test Python` shards are red with this, `collected 19958 items` in each, and zero of them executed. That includes every test this PR adds -- `test_train_checkpoint.py`, `test_train_sharding.py`, `test_train_timing.py`, `test_training_ddp.py` -- so none of the new coverage has ever run, and neither has the pre-existing suite that would have guarded the `pt` refactor.

deepmd/dpmodel/train/checkpoint.py
```diff
@@ -0,0 +1,310 @@
...
+                continue
+            if current_step is not None and step > current_step:
+                path.unlink(missing_ok=True)
+            else:
+                retained.append((step, path))
+        retained.sort(key=lambda item: (item[0], item[1].name))
+        # The current checkpoint occupies one slot of the window when this
+        # store holds it.
+        occupied = 1 if current_step is not None else 0
+        for _, path in retained[: len(retained) + occupied - self.max_keep]:
```
wanghan-iapcm: Blocking, and this one loses user data on `pt` as well as `pt_expt`.

When `len(retained) + occupied - self.max_keep` is negative, this is not an empty slice -- Python reads a negative bound as "all but the last N", so the loop deletes the oldest checkpoints instead of deleting nothing.

I ran this class against the implementation it replaces ([`_write_checkpoint`](https://github.com/deepmodeling/deepmd-kit/blob/master/deepmd/pt/train/training.py#L2229-L2234), whose `while len(checkpoint_files) > max_ckpt_keep: checkpoint_files.pop(0)[0].unlink()` cannot over-delete by construction), same inputs, current checkpoint at the highest step:

| `max_ckpt_keep` | on disk | master keeps | this branch keeps |
|---|---|---|---|
| 5 | 4 | `100,200,300,400` | `300,400` |
| 10 | 6 | all six | `200..600` |
| 10 | 7 | all seven | `400..700` |
| 10 | 8 | all eight | `600,700,800` |
| 10 | 9 | all nine | `800,900` |

The trigger is that the checkpoint count is still *below* the retention limit, so this fires during the ramp-up phase of every run and stops once the directory fills to `max_keep`. That makes it worse the larger `max_ckpt_keep` is -- the value a cautious user picks precisely to avoid losing checkpoints -- and it is silent.

`max_keep` of 1, 2 and 3 never diverge, which is why it got through: every `prune` case in [`test_train_checkpoint.py`](https://github.com/deepmodeling/deepmd-kit/blob/775b68f405d820c8aba632cb705c40a6a96bbf06/source/tests/common/dpmodel/test_train_checkpoint.py) uses `max_keep` in {0, 1, 2}, and the two `max_ckpt_keep=7` cases only assert the configured value and never call `prune`. The window is empty for those, so the tests cannot see it.

`retained[: max(0, len(retained) + occupied - self.max_keep)]` fixes it. Worth adding a case with `max_keep=10` and nine checkpoints, since that is the shape no current test has.

wanghan-iapcm (changes requested):
The shape of this is right and I want to say that before the problems. Extracting checkpoint layout, retention, timing and sharding into `deepmd/dpmodel/train/` follows the precedent #5603 set, the three new modules import no torch so the backend-independence is real rather than nominal, and the alternative -- reimplementing the same policies inside `pt_expt` -- would have produced exactly the drift the PR body describes. Several things I went looking for turned out clean: the ZeRO collective ordering is correct (`_collect_checkpoint_states` runs `consolidate_state_dict` before the `rank != 0` return, so no rank can skip a collective), ZeRO-1 restart across a different world size still works, `pt`'s non-finite gradient handling is a pure rename with no behavioural diff, and `resolve_keep_ckpt_count` actually fixes a latent `ZeroDivisionError` on `save_freq <= 0`.

Two blocking problems, both inline, and one question.

The second one is the serious one: `CheckpointStore.prune` deletes checkpoints it is supposed to keep, on `pt` as well as `pt_expt`. I verified it by running the new class against the implementation it replaces rather than by reading, and the divergence is not subtle -- with `max_ckpt_keep=10` and nine checkpoints on disk, master keeps all nine and this branch keeps two.

I also want to flag the ordering problem this creates for review itself. Because CI never got past collection, none of the ~19958 tests ran, including all the new ones. So the checkpoint bug was not caught by the suite, and more importantly nothing else in this 2600-line diff has been exercised either -- the parts I checked by hand look right, but "looks right on inspection" is a much weaker statement than this PR deserves given it touches production `pt` training. Worth fixing the import first and letting a green run tell us what else is there before anyone reads the rest too closely.

A few smaller notes I am recording rather than asking you to act on. `format_training_message`'s `step_time` parameter and the `avg = ... s/step` field are gone, along with the `TestFormatTrainingMessageStepTime` test; those came from #5500 whose stated goal was to fold that average into the normal log line, and `TrainingTimer.format_average()` only prints once at end of run, so `pt_expt`, `jax` and `tf2` all lose the per-interval figure. `build_checkpoint_stores` gates the retention log and `store.prepare()` on `rank == 0` but every test uses the default rank, so the non-chief branch is unexercised -- the same gap I raised on #5603 for `is_chief`. `ema_ckpt_keep` moving from a hard 3 to inheriting `max_ckpt_keep` silently shrinks EMA retention for anyone who set `max_ckpt_keep` to 1 or 2; it is documented, so this is a release-note item rather than a defect. And `pt_expt` now inherits `pt`'s "delete every checkpoint numbered above the current step" rule, which is a new data-loss path for `pt_expt` users restarting from an older checkpoint in a directory holding a longer run -- pre-existing for `pt`, new for `pt_expt`, and not mentioned in `doc/train/training-advanced.md`.

source/tests/pt_expt/test_entrypoint.py
OutisLi: Fixed in c814ddd97 by removing the stale import and orphaned test instead of adding another entrypoint helper. Nested HDF5 parent creation is already owned and covered by open_stat_file, so keeping that lifecycle at the shared statistics-cache boundary avoids duplicating it. The focused entrypoint and stat-file tests now collect and pass locally.

deepmd/dpmodel/train/checkpoint.py
OutisLi: Fixed in c814ddd97. CheckpointStore now computes a non-negative overflow count before slicing, so a retention window below capacity deletes nothing. I also added the exact regression with max_keep=10 and nine checkpoints; the shared checkpoint tests pass locally, covering both pt and pt_expt.

deepmd/pt_expt/train/training.py
```diff
@@ -1872,29 +1865,92 @@ def update_finetune_bias(
...
+
+        # Optimiser -----------------------------------------------------------
+        opt_type = optimizer_params.get("type", "Adam")
+        # LambdaLR multiplies each param group's initial learning rate by the
+        # lambda value.  Warmup schedules legitimately return zero at step 0,
+        # so use the nonzero schedule base as the denominator and let the
+        # lambda initialize the optimizer to the requested warmup value.
+        initial_lr = float(self.lr_schedule.start_lr)
+        adam_betas = (
+            float(optimizer_params["adam_beta1"]),
```
wanghan-iapcm, line 1926: Blocking. This lookup now runs before the optimizer type is validated, so an unsupported type dies with `KeyError` instead of the intended message.

In the base this method checked the whitelist first ([`training.py` at c0c1f0c53](https://github.com/deepmodeling/deepmd-kit/blob/c0c1f0c53e5aaab5be8d0e510c05c0bf0eb8243a/deepmd/pt_expt/train/training.py#L1634-L1646)):

```python
opt_type = optimizer_params.get("type", "Adam")
if opt_type not in {"Adam", "AdamW", "HybridMuon"}:
    raise ValueError(f"Unsupported optimizer type: {opt_type}")
...
adam_betas = (float(optimizer_params["adam_beta1"]), ...)
```

The refactor to `if/elif/else` moved that `raise` down to line 1920, past this unconditional read. `normalize()` only fills the Adam sub-arguments for the Adam variants, so any other type -- `LKF`, which `pt` accepts, is the obvious one a user would try -- reaches line 1882 with no `adam_beta1` key and raises `KeyError: 'adam_beta1'`.

This is caught by `test_unsupported_optimizer_has_clear_error`, which is not new in this PR -- it is at line 376 of `test_training.py` in the base and passed there. It is failing now only because the stale import was suppressing collection, so the shard never got far enough to run it. It is the sole remaining red job:

```
FAILED source/tests/pt_expt/test_training.py::TestTraining::test_unsupported_optimizer_has_clear_error - KeyError: 'adam_beta1'
deepmd/pt_expt/train/training.py:1882: KeyError
```

Hoisting the check back above the parameter reads restores both the behaviour and the test:

```python
opt_type = optimizer_params.get("type", "Adam")
if opt_type not in ("Adam", "AdamW", "HybridMuon"):
    raise ValueError(f"Unsupported optimizer type: {opt_type}")
```

leaving the `else:` branch at the end as unreachable belt-and-braces, or dropping it. Worth noting this one generalises: the whole suite has only just started running on this branch, so other latent breakage may surface once the shard is green.

wanghan-iapcm (changes requested):
Both of my earlier findings are fixed, and I checked each against `c814ddd97` rather than going by the replies.

The import removal is right, including the part of your reply I had not verified when I raised it: `open_stat_file` really does own nested HDF5 parent creation ([`stat_file.py`](https://github.com/deepmodeling/deepmd-kit/blob/c814ddd9739fc176bfa7959f0cd8a9b9bbf6c3ed/deepmd/utils/stat_file.py#L193-L196) calls `target.parent.mkdir(parents=True, exist_ok=True)`), and [`test_stat_file.py`](https://github.com/deepmodeling/deepmd-kit/blob/c814ddd9739fc176bfa7959f0cd8a9b9bbf6c3ed/source/tests/common/test_stat_file.py#L55-L60) already exercises it through `tmp_path / "nested" / "stat.hdf5"`. So the deleted test was redundant rather than coverage quietly dropped, and removing the helper was the better call.

For `prune` I reconstructed the pre-fix line and ran your new `test_prune_keeps_every_checkpoint_below_the_window` against both versions instead of reading it:

```
PRE-FIX   FAIL -> deleted steps [1, 2, 3, 4, 5, 6, 7]
POST-FIX  PASS
```

It fails pre-fix for exactly the right reason, so it is a genuine regression test. The `training-advanced.md` paragraph covers the retention note I raised in the body too.

The other changes in this round check out: `_named_model_parameters` returns a list, so collecting the backups up front in `apply_shadow` is safe to re-iterate and is what makes tied weights restore correctly; `eval_env_defaults` is in scope at the new `pt/train/training.py` use site; and the `argcheck` edit is an accuracy fix rather than a narrowing, since `DP_COMPILE_INFER` and `DP_TF32_INFER` are only read under `deepmd/pt/` while `DP_AMP_INFER` is read in `deepmd/kernels/utils.py` and `pt_expt/descriptor/dpa4.py`.

One blocker left, inline. Getting collection working turned out to expose a regression this PR introduces against its own base, and it is the only red shard remaining.

Still open from my first review, as notes rather than blockers: `format_training_message` losing `step_time` and the `avg = ... s/step` field, which drops that figure for `pt_expt`, `jax` and `tf2`; and the `rank == 0` branch of `build_checkpoint_stores` being unexercised because every test uses the default rank.

deepmd/pt_expt/train/training.py
OutisLi: Fixed in a425d1e59. I moved the whitelist check back immediately after `opt_type` and before the unconditional Adam-parameter reads, then made the remaining branch the now-exhaustive `HybridMuon` path. The existing `test_unsupported_optimizer_has_clear_error` passes locally; `source/tests/common/test_argcheck_training.py` passed in the same focused run (2 passed total).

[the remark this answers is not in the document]
OutisLi: Fixed in a425d1e59. The optimizer whitelist now runs immediately after reading `opt_type`, before any Adam-specific fields, so unsupported types again raise the intended `ValueError`. The existing `test_unsupported_optimizer_has_clear_error` passes.

wanghan-iapcm (approved):
All three findings are fixed, and I checked each against `a425d1e59` rather than going by the replies.

The optimizer ordering is restored the way I suggested: the whitelist runs immediately after `opt_type`, the `HybridMuon` branch became the exhaustive `else`, and the trailing `raise` is gone. What convinces me it is a real fix rather than a plausible one is the CI transition itself -- `test_unsupported_optimizer_has_clear_error` is pre-existing and untouched, it was the sole red job on `c814ddd97`, and every check now passes. So it failed pre-fix and passes post-fix on its own merits.

The two changes in that commit I had not asked for also check out. Guarding `should_save_checkpoint` with `save_freq > 0` is logically identical to the old expression for any positive `save_freq` and only changes the `save_freq <= 0` case, which previously divided by zero -- consistent with the `resolve_keep_ckpt_count` hardening already in this PR. The `doc_max_ckpt_keep` rewrite and the `parallel-training.md` edits cover the `ema_ckpt_keep` inheritance and the ZeRO stage-2/3 constraints, which was the release-note item from my first review.

This is the first fully green run on the branch, so the suite has now actually exercised the `pt` refactor rather than aborting at collection. Nice work on the checkpoint retention regression in particular -- the `max_keep=10` case you added is exactly the shape that was missing.

PR #5960, opened 2026-08-06
perf(dpa4): batch the SO3/grid contractions over (D,F); keep use_amp through pt_expt assembly

wanghan-iapcm:
Users reported that compiled DPA4 training runs ~2x slower on `pt_expt` than on `pt`. This PR is the result of chasing that: a performance bug in how the SO3 contractions were lowered, and a correctness bug where a configured `use_amp: false` was lost while `pt_expt` assembled the model.

## Changes

**1. Weight broadcast across the node axis (`so3.py`, `lora.py`, `grid_net.py`)**

`matmul(x[..., None, :], weight[None, ...])` makes the node count `N` the matmul BATCH, so matmul broadcasts the weight to `(N, D, F, Cin, Cout)` and autograd then reduces that whole expanded gradient (`ExpandBackward0`) back to the parameter shape. At the water example's sizes a 165 K-element weight expanded to 191 M elements (~0.8 GB) per call, and the reduce was the single costliest kernel of a training step (45.6 ms, 3x per step).

The fix batches the contraction over the small `(D, F)` axes so `N` stays the GEMM ROW dimension and the weight is used in place. Micro-benchmark, fwd+bwd at the real shapes: **16.48 ms -> 1.09 ms (15x)**.

Two lookalike sites in `projection.py` are deliberately NOT changed: their operands are `requires_grad=False` buffers, so no backward reduce exists. Verified rather than assumed.

**2. The same lowering for the frame mixers (`_degree_batched_matmul`)**

Review found the `FrameContract` / `FrameExpand` mixers still on the broadcast spelling. They now share one helper, `_degree_batched_matmul`, written identically on the dpmodel side (`dpmodel/descriptor/dpa4_nn/grid_net.py`) and the pt side (`pt/model/descriptor/sezm_nn/grid_net.py`). Because it does no reshape, an empty node axis (`N == 0`) flows through unchanged instead of hitting a reshape error — pinned by a test.

**3. `use_amp` was lost during `pt_expt` model assembly (correctness)**

`use_amp` is a training-runtime policy, not model state, so it stays OUT of the portable serialization record (this is the #5963 position, and the jax deserializer actively rejects records carrying `use_amp: true`). The bug was elsewhere: `pt_expt` assembled its model by converting an already-populated dpmodel instance, and that conversion round-trips the component through `deserialize(serialize())`. Anything that is deliberately not in the portable record — `use_amp` among it — was therefore dropped, and training silently ran under bfloat16 autocast even when the input configured `use_amp: false`.

The fix is at the assembly boundary, not in the record: `pt_expt` now constructs the wrapped class directly (`auto_wrapped_class(...)` in `make_model.py`, `get_model.py`, and the bridging composition path), so a live constructor-supplied component keeps its runtime state. The rule is stated once, in the `auto_wrapped_class` docstring; the call sites reference it.

An earlier revision of this PR instead added `use_amp` to `serialize()`. That was reverted in review — it put a runtime knob into the portable record and would have broken the jax contract.

**Also removed in review: an `enable_tf32` / `DP_TF32_INFER` implementation for `pt_expt`.** It contributes nothing to the speedup measured below (the benchmark card has no TF32 silicon), and #5958 owns the `pt_expt` training-runtime alignment — including the documented position that `pt_expt` always runs at `"highest"` matmul precision. `pt_expt` therefore keeps master's warn-and-ignore behavior for `enable_tf32`.

## Benchmark

DPA4 water example (`examples/water/dpa4`), one Tesla T4, torch 2.11, fp32 (`use_amp: false`), batch size 6. Steady-state seconds per training step, obtained by differencing the wall time of a 33-step and a 3-step run of the same config, which cancels every one-time cost (import, data load, statistics, `torch.compile` / make_fx lowering). All five arms were measured in one session on the same machine; run-to-run variation is about 2-3%.

**Provenance: measured at `ae720432b`**, the head at which this PR was opened — i.e. BEFORE the review changes (change 2, the frame-mixer lowering, and change 3's move from `serialize()` to the assembly boundary). Change 1, which is where the entire speedup comes from, is unmodified since. The numbers have not been re-measured on the current head; a re-run is pending and I will post it rather than silently reuse these.

| training mode | `pt` (reference) | `pt_expt` at master | `pt_expt` at `ae720432b` | speedup vs master |
|---|---|---|---|---|
| eager | 0.891 s/step | 1.555 s/step | **0.921 s/step** | **1.69x** |
| compiled | 0.545 s/step | 1.611 s/step | **0.535 s/step** | **3.01x** |

This reproduces the reported issue at master — `pt_expt` compiled was 3.0x slower than `pt` compiled, and even slower than its own eager path, because the broadcast-weight contraction lowers to worse code under inductor than under eager cuBLAS. After the fix `pt_expt` is at parity with `pt`: eager within 3.4%, compiled within measurement noise.

## Known limitations

- **Backward numerics are covered for the frame mixers, not for the SO3 / LoRA contractions.** `test_dpa4_frame_mixers.py` compares `_degree_batched_matmul`'s weight gradient against the pt module's at rtol/atol 1e-12. The rewritten SO3 and LoRA contractions are pinned on the forward against an explicit `einsum` reference (rtol/atol 1e-12, numpy and torch namespaces); their backward is still exercised only by tracing, not compared by value.
- **The `pt` / `pt_expt` TF32 policy gap remains open.** On Ampere+ cards `pt` runs training matmuls under TF32 (`enable_tf32`, default `True`) while `pt_expt` ignores the key with a warning; the two backends are not speed-comparable there. Deferred to the #5958 training-runtime series.
- **The residual compiled gap vs `pt` is not stable across sessions.** An earlier session measured `pt_expt` compiled 10.8% slower than `pt` compiled; the benchmark above measured it 1.7% faster. Both are within a couple of run-to-run standard deviations, so I treat compiled as at parity and the earlier gap as unconfirmed.
- The history contains churn at the GridBranch router (`7518a417c` -> `01c58e665` -> `75459610a` -> `504bb2430` -> `157444204`): a matmul spelling introduced, reverted, reintroduced, and finally restored to master's line. The site is byte-identical to master in the final diff. The degenerate GEMM that profiling found there existed only on this branch, so it is not a fix — I have left the commits rather than rewriting pushed history, and would squash them on request.
- Unrelated but found while benchmarking: **torch >= 2.11 ships no Volta (CC 7.0) kernels**, and compiled training requires >= 2.11 via `check_compile_torch_version`. Compiled DPA4 training is therefore impossible on V100 with official wheels; T4 (CC 7.5) is the oldest card that works.

## Tests

- `source/tests/common/dpmodel/test_dpa4_frame_mixers.py` — `_degree_batched_matmul` vs the pt module: forward parity, the `N == 0` contract, and weight-gradient parity.
- `source/tests/common/dpmodel/test_dpa4_lora.py` — new `test_lora_so3_call_matches_einsum_contract`: `LoRASO3.call` against the explicit `einsum("ndfi,difo->ndfo")` reference with a nonzero adapter, on both the numpy and torch namespaces, for `n_focus` 1 and 2.
- `source/tests/pt_expt/model/test_get_model_dpa4.py` — `use_amp` survives model assembly (both branches), for the plain and the bridged/composed construction paths.
- `source/tests/common/dpmodel/test_descrpt_dpa4.py` — `use_amp` is absent from the portable serialization record and defaults on deserialize.
- Existing `test_grid_branch[1]`/`[2]` cover the changed SO3 contraction against the pt implementation at rtol 1e-12.
- Run locally: 434 passed / 10 skipped across the dpa4 dpmodel, pt_expt and cross-backend parity suites, plus the pt_expt model suite. CUDA-gated precision-context cases were run on a T4 (29/29).

## Test status caveat — resolved

An earlier revision of this description flagged two locally failing `pt_expt` AOTI-freeze tests (`test_zbl_bridging.py::test_native_spin_with_bridging_graph_freeze_and_deep_eval`, `test_dpa4_zbl_parallel.py::TestBridgedSpinGraphSelfComm::test_freeze_embeds_with_comm_artifact`) as unadjudicated. They are now adjudicated as **pre-existing and environmental, not caused by this branch**: a clean `upstream/master` worktree on the same machine fails both with the identical `InductorError: assert isinstance(index, CppCSEVariable) and index.is_vec` (torch 2.11 CPU-SIMD codegen bug on an `atomic_add` scatter buffer), and both tests pass on this branch with the known workaround `torch._inductor.config.cpp.simdlen = 1` (2 passed). The same bug is already documented in `source/tests/infer/gen_dpa4.py` / `gen_dpa2.py`.


OutisLi:
Please merge or rebase the latest `upstream/master` before continuing this review. PR #5963 has already been merged into master and fixes a correctness issue relevant to the contractions changed here: `FrameContract.weight` and `FrameExpand.weight` are added to the pt_expt trainable-promotion table. At the current head (`1e56cf6d`), the default `examples/water/dpa4/input.json` constructs four such modules, but all four weights are still non-trainable buffers, whereas the pt backend registers them as trainable `nn.Parameter`s. This also means the current-head assumption that these frame weights incur a trainable-weight `ExpandBackward0` reduction does not hold for pt_expt. Syncing current master should bring in the #5963 fix naturally; please re-run the relevant forward/backward and performance checks after resolving the current conflicts.

OutisLi (changes requested):
The `use_amp` fix is at the wrong abstraction boundary. `use_amp` is a training-time/runtime policy and must not become part of the portable descriptor serialization or checkpoint state. This is also the policy established on current master by #5963: a checkpoint must not carry the training AMP switch into deployment or a later training run. Adding `use_amp` to both dpmodel and pt serialization leaks a PyTorch runtime option into cross-backend records (for example, a default `use_amp=True` record is rejected by the JAX DPA4 deserializer).

The underlying pt_expt bug is real, but it happens during model assembly. On current master I can reproduce the following with `get_model`: the pt_expt factory initially constructs `DescrptDPA4(use_amp=False)` correctly, but wrapping it into `DPA4EnergyModel` replaces it with a different descriptor whose `use_amp` is `True`. The atomic-model auto-wrap round-trips the already constructed descriptor through `serialize()`/`deserialize()`, so the runtime-only value is lost and the constructor default is restored.

Please keep `use_amp` out of descriptor serialization and fix the pt_expt construction/wrapping boundary so that the runtime configuration survives model assembly. The regression test should exercise the public construction path, e.g. build via `get_model` with `descriptor.use_amp=False` and assert that `model.atomic_model.descriptor.use_amp` remains `False`; a descriptor serialization round-trip test codifies the wrong ownership instead of covering the actual failure.

deepmd/dpmodel/descriptor/dpa4_nn/grid_net.py
```diff
@@ -99,6 +99,38 @@ def _build_frame_degree_index(
...
+
+    Notes
+    -----
+    Batching over the degree axis, not over ``N``: the latter would broadcast
+    ``weight`` to ``(N, D, i, o)`` and make autograd reduce that expansion on
+    every backward.  The transposes touch only ``coeff``, which is smaller.
+    """
+    n_batch, coeff_dim, n_focus, _ = coeff.shape
+    coeff_d = xp.reshape(
+        xp.permute_dims(coeff, (1, 0, 2, 3)), (coeff_dim, n_batch * n_focus, -1)
```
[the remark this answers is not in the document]
wanghan-iapcm: Fixed in https://github.com/deepmodeling/deepmd-kit/commit/63071be846af5d1a58e520837b9657bb3b125a71 exactly as suggested: both reshapes in `_degree_batched_matmul` (the one helper behind `FrameContract` and `FrameExpand`) now use explicit channel widths taken from the weight shape, so an `N == 0` batch flows through as an empty `(0, D, F, o)` result like the previous broadcasted matmul. Regression test `test_empty_batch_passes_through` covers both mixers on the numpy and torch namespaces (verified to fail on the `-1` version). The jax namespace shares this exact code path but was not run locally.

deepmd/pt_expt/model/make_model.py
```diff
@@ -470,6 +474,55 @@ def get_min_nbor_dist(self) -> float | None:
...
+            prev_precision = torch.get_float32_matmul_precision()
+            try:
+                if self.training:
+                    precision = "high" if self.enable_tf32 else "highest"
+                else:
+                    precision = self.tf32_infer_precision
+                torch.set_float32_matmul_precision(precision)
+                yield
+            finally:
+                torch.set_float32_matmul_precision(prev_precision)
```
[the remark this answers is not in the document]
wanghan-iapcm: Skipped as stale: the code this comment targets is no longer part of this PR. The pt_expt `tf32_precision_ctx` / `enable_tf32` change was reverted from this branch (commit https://github.com/deepmodeling/deepmd-kit/commit/1e56cf6d5) and deferred to issue #5958 — the PR diff no longer touches `deepmd/pt_expt/model/make_model.py`. The only remaining `tf32_precision_ctx` is the pre-existing one in `deepmd/pt/model/model/sezm_model.py` on master, unchanged by this PR; the thread-safety consideration will be addressed together with the deferred #5958 work.

wanghan-iapcm:
@OutisLi Both points are addressed.

**Master merge (your first comment):** `upstream/master` is merged (https://github.com/deepmodeling/deepmd-kit/commit/c1792fad9), bringing in the #5963 trainable-promotion of `FrameContract.weight`/`FrameExpand.weight`. The local forward/backward batteries were re-run on the merged head (frame mixers, so3 gridnet, descriptor, native-spin, linear, get_model suites — all green); the V100 performance re-check still needs a remote run and I will attach numbers separately.

**`use_amp` boundary (your review):** agreed on all counts, fixed in https://github.com/deepmodeling/deepmd-kit/commit/fa465706e73712af04feebe047db3100188be416.
- The `use_amp` keys are reverted out of both the dpmodel and pt serialize dicts — the portable record no longer carries the runtime switch (the old round-trip test is replaced by one pinning `"use_amp" not in serialize()["config"]`).
- The real bug is fixed at the assembly boundary you identified: pt_expt `make_model` handed the raw dpmodel atomic class to the CM, so the constructed atomic model went through the auto-wrap `deserialize(serialize())` round-trip and the live descriptor was rebuilt from the portable record. `make_model` now hands the CM the auto-wrapped atomic class (`auto_wrapped_class(T_AtomicModel)`), so the atomic model is constructed directly as a torch module and the already-constructed descriptor/fitting are kept as-is — no serialize round-trip in assembly.
- The regression tests exercise the public construction path exactly as you asked: `get_model` with `descriptor.use_amp=false` asserting `model.atomic_model.descriptor.use_amp is False` (plus the default-True case and the `type: standard` route), verified to fail before the fix.

One known residual, out of this PR's scope: compositions passed as `atomic_model_=` (linear/ZBL) still convert the pre-built atomic model through the instance round-trip, so a dpa4 child's `use_amp=false` inside a composition would still reset; happy to file a follow-up issue.

OutisLi:
The latest empty-axis fix keeps an avoidable hot-path copy, and the same issue exists in the pt reference. For `F > 1`, `_degree_batched_matmul` permutes `(N,D,F,i) -> (D,N,F,i)` and then reshapes to `(D,N*F,i)`; because `N` and `F` are not collapsible in that layout, this materializes the entire coefficient tensor. `torch.einsum("ndfi,dio->ndfo", ...)` in pt `FrameContract` / `FrameExpand` lowers to essentially the same coefficient clone plus a `(D,N*F,i)` bmm.

This is avoidable by batching over `(D,F)` instead:

```python
coeff_df = xp.permute_dims(coeff, (1, 2, 0, 3))  # (D,F,N,i)
out = xp.matmul(coeff_df, weight[:, None, :, :])   # (D,F,N,o)
return xp.permute_dims(out, (2, 0, 1, 3))          # (N,D,F,o)
```

The tradeoff is expanding the much smaller weight across `F`, i.e. `D*F*i*o` elements instead of copying `N*D*F*i` coefficient elements; their ratio is `N/o`. At `(N,D,F,i,o)=(1152,9,2,96,32)`, that is 7.59 MiB versus 0.21 MiB. With torch 2.11 compiled CPU forward+backward, I measured 3.106 ms for pt einsum, 3.590 ms for the current helper, and 2.653 ms for `(D,F)` batching; at `(102510,9,2,32,32)`, 170.7/171.4 ms became 96.6 ms. These are CPU measurements, not a GPU speedup claim, but the removed coefficient materialization is deterministic and `n_focus=2` is used by shipped spin/property DPA4 examples.

Since this PR is specifically a contraction-performance cleanup, please use the `(D,F)` formulation in both dpmodel and pt `FrameContract` / `FrameExpand`, rather than retaining the copy while only replacing `-1` with explicit dimensions. The same formulation naturally handles `N == 0`; please keep the new empty-axis regression coverage and add forward/backward parity for `F > 1`.

OutisLi (changes requested):
The assembly fix still loses runtime-only configuration on the public composition paths. The new `auto_wrapped_class(T_AtomicModel)` avoids the serialization round-trip only when the outer model constructs its atomic model from arguments. `_compose_bridging`, however, still builds a raw `LinearEnergyAtomicModel` and passes it as `LinearEnergyModel(atomic_model_=composed)`. Assigning that raw instance reaches `_auto_wrap_native_op`, which still performs `wrapped_cls.deserialize(value.serialize())` and rebuilds the learned DPA4 child without its runtime-only `use_amp` value.

I reproduced this on the current head through public `get_model`: the plain DPA4 config with `descriptor.use_amp=false` now retains `False`, but adding `bridging_method="ZBL"` produces a linear composition whose learned child has `descriptor.use_amp is True`. The PT backend retains `False` for the same DPA4+ZBL config. Consequently pt_expt silently enables bf16 autocast despite the explicit user setting whenever analytical bridging is enabled.

This is the same root assembly-boundary bug addressed by this PR, not a separate serialization feature. Please make composition construction lossless as well: construct the composite from wrapped atomic classes/module children instead of converting a populated raw dpmodel instance through portable serialization. Add a regression through public `get_model` for DPA4/SeZM with `bridging_method="ZBL"` and `use_amp=false`, asserting the learned child retains `False`.

source/tests/pt_expt/model/test_get_model_dpa4.py
```diff
@@ -285,3 +285,61 @@ def test_unrelated_construction_error_propagates(self) -> None:
 
 if __name__ == "__main__":
     unittest.main()
+
+
+class TestUseAmpSurvivesAssembly(unittest.TestCase):
```
OutisLi, line 324: This test class is defined after the `if __name__ == "__main__": unittest.main()` entry point, so direct execution starts discovery before this class exists and silently skips all three new `use_amp` regression tests. I reproduced 16 tests through the file entry point, with none from `TestUseAmpSurvivesAssembly`, whereas pytest import-based discovery sees them. Please keep the test classes together and move the `unittest.main()` block to the actual end of the file.

OutisLi:
The module docstring in `source/tests/common/dpmodel/test_dpa4_frame_mixers.py` is stale after this PR. It still says that the dpmodel implementation realizes the contraction as a "broadcast batched `xp.matmul`", even though this PR specifically replaces that formulation with degree-batched multiplication. Please describe the backend-independent mathematical contract (`einsum("ndfi,dio->ndfo")`) and either omit the lowering detail or update it to the final implementation selected here. Keeping the old broadcast description makes the parity test documentation contradict the code it covers.

OutisLi:
Additional reproduction for the composition assembly request: the explicit public `linear_ener` route has the same loss even though it does not call `_compose_bridging`. With a DPA4 child configured with `descriptor.use_amp=false`, current-head pt_expt builds `model.atomic_model.models[0].descriptor.use_amp is True`; the PT backend retains `False` for the same config.

Here the round-trip occurs one level lower: `get_linear_model` constructs a raw `DPAtomicModel(descriptor, fitting, ...)`, and assigning the raw children list to the wrapped `LinearEnergyAtomicModel` makes `_try_convert_list` call `_auto_wrap_native_op` on that child, again rebuilding it through `deserialize(serialize())`. Therefore, please do not fix only the ZBL `_compose_bridging` call site. Composition construction needs one lossless rule for both the parent and learned children, with a public `linear_ener` + DPA4(`use_amp=false`) regression test in addition to the bridged case.

source/tests/pt_expt/model/test_get_model_dpa4.py
wanghan-iapcm: Fixed in https://github.com/deepmodeling/deepmd-kit/commit/8bce829481aa89a62005960b553a2f4b8f034dce: the `if __name__ == "__main__": unittest.main()` block moved to the actual end of the file, so direct execution now discovers `TestUseAmpSurvivesAssembly` too (verified: 21 tests collected either way).

wanghan-iapcm:
@OutisLi All four round-2 points are addressed in https://github.com/deepmodeling/deepmd-kit/commit/8bce829481aa89a62005960b553a2f4b8f034dce.

**Lossless composition assembly (your review + the `linear_ener` addendum):** both public routes are fixed by one rule — construct wrapped atomic classes, never convert a populated raw instance. pt_expt now binds `DPAtomicModel`/`PairTabAtomicModel`/`InnerPotentialAtomicModel` to `auto_wrapped_class(...)` at module level (also handed to the backend factory), so `get_linear_model` builds its children as live wrapped modules; `_compose_bridging` passes constructor args to `LinearEnergyModel` instead of a pre-built `atomic_model_=` raw composition, so the CM constructs the wrapped composition directly around the live learned child. Both of your reproductions are now regression tests through public `get_model` — DPA4+`bridging_method="ZBL"` and explicit `linear_ener` with a DPA4 child, each asserting the learned child keeps `descriptor.use_amp is False` — and both fail without the fix.

**(D, F) batching:** adopted exactly as proposed, in BOTH dpmodel `_degree_batched_matmul` and pt `FrameContract`/`FrameExpand` (the pt einsum is replaced by the same shared-form lowering). No reshape remains, so `N == 0` flows through naturally; the empty-axis regression stays and a new F>1 test pins forward AND backward (input + weight gradients) of both lowerings against the `einsum("ndfi,dio->ndfo")` contract at rtol 1e-12. The pt-vs-dpmodel descriptor parity battery (387 cases) passes on the new lowering. I did not re-run your timing comparison locally; the V100 numbers will come with the remote perf check.

**Stale mixer-test docstring:** rewritten to state the backend-independent einsum contract, with both backends sharing the `(D, F)`-batched lowering.

**Test placement:** `unittest.main()` moved to the true end of the file (replied inline).

OutisLi:
The PR description is stale and currently documents the implementation that was rejected during review. At the current head (`e2f54c8f`), `use_amp` is deliberately absent from portable descriptor serialization and is preserved by lossless pt_expt model assembly. However, section 2 still says that `use_amp` was added to both dpmodel and pt `serialize()` methods, and the Tests section still describes a `use_amp` serialization round-trip test. This reverses the actual ownership of the runtime policy and could lead a future maintainer to reintroduce the same abstraction error.

The Known limitations section is also outdated: the latest frame-mixer tests now compare both input and weight gradients against the einsum contract. Finally, the benchmark table predates the final `(D, F)` lowering and assembly changes; the latest author reply explicitly says that timing was not rerun after those changes. Please either rerun the benchmark on the final head or identify the exact measured commit instead of presenting it as the final-head result.

Please update the title and description together so they describe the final design: keep `use_amp` out of portable serialization, preserve runtime configuration at the pt_expt assembly boundary, state the current gradient coverage, and provide accurate benchmark provenance.

OutisLi:
The implementation now has a clean central invariant in auto_wrapped_class, but the same review-history explanation is repeated at four call sites: common.py, the wrapped atomic aliases and bridging path in get_model.py, and make_model.py. The repeated DIRECTLY / WRAPPED / NOT wording and DPA4-specific use_amp narrative reads like patch archaeology rather than stable documentation, and the copies can drift.

Please document the general invariant once in the auto_wrapped_class docstring: construct the wrapped class directly when live constructor-supplied components must retain non-serialized runtime state. Keep the call-site comments to one concise ownership line, and leave the concrete use_amp regression history in the tests and PR description. No structural code change is requested; this is about making the final implementation read as one coherent design rather than the sequence of review fixes.

deepmd/dpmodel/descriptor/dpa4_nn/so3.py
```diff
@@ -439,12 +442,16 @@ def call(self, x: Any) -> Any:
         weight_expanded = xp.take(weight, expand_index, axis=0)  # (D, Cin, F, Cout)
 
         # === Step 2. Per-focus, per-degree channel mixing ===
-        # einsum "ndfi,difo->ndfo" as a broadcast batched matmul:
-        # (N, D, F, 1, Cin) @ (1, D, F, Cin, Cout) -> (N, D, F, 1, Cout)
+        # einsum "ndfi,difo->ndfo", batched over the small (D, F) axes.
+        # Batching over the node axis N instead would broadcast the weight to
+        # (N, D, F, Cin, Cout) -- for the water example a 165K-element parameter
+        # blown up to 191M elements per call -- and autograd would then reduce
+        # that expansion back down.  It was the costliest kernel of a step.
```
OutisLi: Please keep the structural performance rationale, but remove the configuration-specific benchmark anecdote from library source. The water-specific 165K-to-191M numbers and the statement that this was the costliest kernel have no recorded shapes, backend, device, benchmark method, or commit, so they cannot be audited and will become stale as configurations and kernels change.

The durable invariant is sufficient here: batching over `(D, F)` keeps `N` as the GEMM row dimension, avoiding materialization of `N` copies of the weight and the corresponding gradient reduction. Please express that directly, for example:

```python
# Batch over (D, F) so N remains the GEMM row dimension. This avoids
# materializing N copies of the weight and the corresponding gradient reduction.
```

source/tests/common/dpmodel/test_dpa4_frame_mixers.py
```diff
@@ -211,3 +214,130 @@ def test_torch_namespace(cls) -> None:
...
+        DPFrameExpand,  # (N, D, F, C) -> (N, D, F, K*C)
+    ],
+)
+def test_empty_batch_passes_through(cls) -> None:
+    """``N == 0`` must yield an empty result, not a reshape error.
+
+    Reachable when the cross-grid leading axis is an empty graph/edge set
+    or a distributed rank owns no nodes; the degree-batched contraction
+    must keep the broadcasted matmul's empty-batch behavior (explicit
+    channel widths -- ``-1`` cannot be inferred from zero elements).
```
OutisLi: This docstring still describes the intermediate reshape-based fix: explicit channel widths and the inability to infer `-1` from zero elements. The final `(D, F)`-batched implementation contains no reshape and no inferred width, so this explanation contradicts the code the test now covers.

Please state only the stable behavior contract here: an empty node axis returns shape `(0, D, F, o)` on every supported array namespace. The empty graph/edge-set or rank-with-no-local-nodes reachability note is still useful; the deleted broadcast/reshape implementation history is not.

source/tests/common/dpmodel/test_dpa4_frame_mixers.py
```diff
@@ -211,3 +214,130 @@ def test_torch_namespace(cls) -> None:
...
+
+    from deepmd.dpmodel.descriptor.dpa4_nn.grid_net import (
+        _degree_batched_matmul,
+    )
+
+    coeff_dp = coeff.detach().clone().requires_grad_(True)
+    dp_out = _degree_batched_matmul(
+        array_api_compat.array_namespace(coeff_dp),
+        coeff_dp,
+        pt_mod.weight.index_select(0, pt_mod.degree_index).detach(),
```
OutisLi: This branch does not test the dpmodel lowering weight gradient: the expanded weight passed to `_degree_batched_matmul` is detached, and the assertions after `backward()` only inspect `coeff_dp.grad`. Consequently the module-level claim that both backends are pinned for values and gradients, and the round-2 statement that both lowerings cover input plus weight gradients, are stronger than the actual test.

I independently checked the missing case and the current implementation matches the einsum reference, with a maximum weight-gradient error of about `3.6e-15`; this is a regression-coverage gap, not evidence of a current numerical bug. Please keep a leaf copy of the original per-degree parameter requiring gradients, apply `index_select` to it, and compare that leaf weight gradient with `grad_w_mod` after the dpmodel helper backward.

deepmd/dpmodel/descriptor/dpa4_nn/lora.py
```diff
@@ -189,10 +189,12 @@ def call(self, x: Array) -> Array:
...
         expand_index = xp_asarray_nodetach(xp, self.expand_index, device=device)
         weight_expanded = xp.take(weight, expand_index, axis=0)
-        # einsum "ndfi,difo->ndfo" as a broadcast batched matmul:
-        # (N, D, F, 1, Cin) @ (1, D, F, Cin, Cout) -> (N, D, F, 1, Cout)
+        # einsum "ndfi,difo->ndfo", batched over the small (D, F) axes rather
+        # than over N, which would broadcast the weight and make autograd
+        # reduce the expansion.  LoRA twin of the so3.py contraction.
         weight_expanded = xp.permute_dims(weight_expanded, (0, 2, 1, 3))
-        out = xp.matmul(x[:, :, :, None, :], weight_expanded[None, ...])[..., 0, :]
+        out = xp.matmul(xp.permute_dims(x, (1, 2, 0, 3)), weight_expanded)
```
OutisLi, line 196: This PR changes the dpmodel `LoRASO3.call` contraction, but no existing test executes this implementation. `source/tests/common/dpmodel/test_dpa4_lora.py` only checks adapter injection and trainability, while `TestLoRASO3Adapter` exercises the separate pt implementation. A shape or algebra regression in this changed path would therefore be unobserved.

I checked the current code independently with `n_focus=2` and a nonzero `B_by_l`; its forward matches `einsum("ndfi,difo->ndfo")` exactly, so this is a coverage gap rather than evidence of a current numerical error. Please add a direct dpmodel `LoRASO3.call` regression for `n_focus=1` and `2`, set `B_by_l` nonzero so the adapter delta participates, and compare NumPy and Torch array-namespace results with the einsum contract.

deepmd/dpmodel/descriptor/dpa4_nn/so3.py
wanghan-iapcm: Done in [`81e756d3`](https://github.com/deepmodeling/deepmd-kit/pull/5960/commits/81e756d30). The comment no longer cites the water example's numbers; it states why the contraction is batched over `(D, F)` — so `N` stays the GEMM row dimension, which avoids materializing `N` copies of the weight and the matching gradient reduction on every backward. That rationale holds at any shape.

source/tests/common/dpmodel/test_dpa4_frame_mixers.py
wanghan-iapcm: Done in [`81e756d3`](https://github.com/deepmodeling/deepmd-kit/pull/5960/commits/81e756d30). The docstring now states the contract instead of naming a call site: an empty node axis yields `(0, D, F, o)` on every namespace, and it explains when that shape is reachable. It no longer goes stale if the caller moves.

source/tests/common/dpmodel/test_dpa4_frame_mixers.py
wanghan-iapcm: Valid — the test was forward-only. Fixed in [`81e756d3`](https://github.com/deepmodeling/deepmd-kit/pull/5960/commits/81e756d30): the dp branch now runs on a leaf copy of the pt module's weight (`weight_dp = pt_mod.weight.detach().clone().requires_grad_(True)`), backprops the same scalar, and compares `weight_dp.grad` against the pt module's weight gradient at rtol/atol 1e-12. So the frame-mixer lowering is now pinned on the backward by value, not only by tracing.

deepmd/dpmodel/descriptor/dpa4_nn/lora.py
wanghan-iapcm: Agreed, `LoRASO3` had no value-level coverage of the rewritten contraction. Added `test_lora_so3_call_matches_einsum_contract` in [`81e756d3`](https://github.com/deepmodeling/deepmd-kit/pull/5960/commits/81e756d30): it sets a nonzero `B_by_l` (a fresh adapter is zero-initialized, so with the default weights the test would have been vacuous), builds the reference by hand — `delta = (B @ A)^T * scaling`, folded into the weight, expanded by `expand_index`, then `einsum("ndfi,difo->ndfo")` — and compares `call()` against it on both the numpy and torch namespaces, for `n_focus` 1 and 2, at rtol/atol 1e-12.

wanghan-iapcm:
@OutisLi thanks — all six addressed, code and tests in [`81e756d3`](https://github.com/deepmodeling/deepmd-kit/pull/5960/commits/81e756d30), description rewritten just now.

On the two non-inline points:

**Stale title and description.** Retitled to *"perf(dpa4): batch the SO3/grid contractions over (D,F); keep use_amp through pt_expt assembly"* and rewrote the body to the design that is actually in the diff:
- Section 2 is now the `_degree_batched_matmul` frame-mixer lowering (one helper, written identically on the dpmodel and pt sides), including the `N == 0` property that follows from doing no reshape.
- Section 3 no longer describes adding `use_amp` to `serialize()` — that revision was reverted. It now states the final design: `use_amp` stays out of the portable record (the #5963 position, which the jax deserializer enforces), and the loss was at the `pt_expt` assembly boundary, where converting a populated dpmodel instance round-trips it through `deserialize(serialize())`. The fix constructs the wrapped class directly.
- Known limitations now state gradient coverage accurately: the frame mixers are compared on the backward by value (weight gradients, rtol/atol 1e-12); the SO3 and LoRA contractions are pinned on the forward against an explicit einsum reference, and their backward is still trace-only.
- The benchmark table is now labeled with its provenance: measured at `ae720432b`, the head at which the PR was opened, i.e. before the review changes. Change 1 — which is the entire speedup — is unmodified since. I have not re-run it on the current head; when I do I will post the new numbers rather than let these stand in for them.

**Duplicated comments.** The invariant is now stated once, in the `auto_wrapped_class` docstring, and phrased generally (live constructor-supplied components must keep non-serialized runtime state; converting a populated instance goes through `deserialize(serialize())`, which keeps only the portable record) rather than as a `use_amp` anecdote. The four call sites in `make_model.py` and `get_model.py` carry a one-line pointer to it.

Tests run locally: `test_dpa4_lora.py` 3 passed, `test_dpa4_frame_mixers.py` 18 passed, `test_get_model_dpa4.py` 21 passed, `test_descrpt_dpa4.py` 24 passed.

OutisLi (approved):

wanghan-iapcm:
Merged `upstream/master` in [`c958cd77`](https://github.com/deepmodeling/deepmd-kit/pull/5960/commits/c958cd770). One conflict, in `deepmd/pt_expt/model/get_model.py`, plus one real regression it exposed — worth a note because it changes the diff:

**The conflict.** #5964 landed on master and removed `_compose_bridging` in favour of the canonical `linear_ener` route. This branch had only edited a comment inside that function, so I took master's deletion.

**What the merge broke.** `TestUseAmpSurvivesAssembly` (this PR's own regression test) failed on the merged tree: master's new route builds the composition with `LinearEnergyModel(atomic_model_=composed)`, and `composed` came back from `get_linear_atomic_model` as a raw dpmodel instance because the factory hardcoded the dpmodel `InnerPotentialAtomicModel` / `LinearEnergyAtomicModel`. Converting that raw instance into the pt_expt wrapper round-trips it through `deserialize(serialize())`, which keeps only the portable record — so `use_amp: false` was dropped again, this time on the canonical route. Same failure mode as the original bug, at a site that did not exist when I fixed it.

**Fix.** `get_linear_atomic_model` now takes `inner_potential_model` and `linear_atomic_model`, alongside the `atomic_model` / `pairtab_model` / `zbl_model` injection it already did, defaulting to the dpmodel classes; `pt_expt` passes its wrapped ones. The composition is assembled from backend-native children, so there is no conversion to lose state in. The docstring states the obligation for any backend that wraps dpmodel classes.

Suites run locally on the merged tree: `test_get_model_dpa4.py` 22, `test_get_model_bridging.py` (pt_expt) 21, `test_zbl_bridging.py` 50, `test_bridging.py` 26, `test_dpa4_native_spin.py` 34, `test_dpa4.py` (pt_expt descriptor) 13 — all passed. PR is `MERGEABLE` again.

PR #5961, opened 2026-08-07
docs(skills): add structured DPA4 workflow support

SchrodingersCattt:
## Summary

- add a dedicated `deepmd-finetune-dpa4` skill while preserving `deepmd-finetune-dpa3`
- add progressive DPA4/SeZM and DPA4C guidance to `deepmd-train`
- document `.pt`, `.pt2`, `.pth`, and `.pb` inference artifact routing
- add model deployment, freeze-time policy, same-node export, and explicit atom-type mapping guidance to `lammps-deepmd`
- update the Agent Skills index and focused regression tests

## Design

The skills remain organized by user workflow rather than adding one umbrella DPA4 skill. Checkpoint routing is based on stored model metadata instead of filenames. DPA4/SeZM guidance covers PyTorch training and fine-tuning, LoRA, held-out evaluation, `.pt2` export, Python inference, and LAMMPS deployment. DPA4C guidance covers the public PyTorch Exportable training, freeze, compression, export-time inference policy, and LAMMPS deployment contracts.

The changed skills contain no MatMaster, Bohrium, private image, private library path, or platform-specific helper dependency.

## Validation

- `python -m pytest -q -p no:cacheprovider source/tests/common/test_agent_skills.py`: 15 passed, 1 skipped; the skipped normalization test requires a built DeePMD checkout
- `python -m black --check source/tests/common/test_agent_skills.py`
- `git diff --check`
- focused tests cross-check DPA4C export variables against `doc/model/dpa4c.md`, reject the invalid `--pt_expt` spelling, and preserve the same-node export contract


SchrodingersCattt:
Addressed the review findings in four focused commits:

- `6e27dc22`: clarify `.pt2` inference limits; retain verified `DeepPot.eval`/`dp test`, remove the unsupported metadata-only `.pt2` `eval-desc` recipe, and state that DPA4 `dp embed` supports `.pt` but not `.pt2`.
- `dc8ec953`: fix the LoRA example path and propagate selected multi-task heads through fine-tuning (`--model-branch`) and export (`--head`) without changing single-task commands.
- `92476d3d`: distinguish DPA4 `edge_vec` and NeighborGraph `.pt2` contracts, require with-comm metadata/artifact for multi-rank LAMMPS, and document device/toolchain compatibility.
- `0e24272f`: pin all online `uvx` commands consistently to the published DPA4-capable pair `deepmd-kit[gpu,torch,lmp]==3.2.0b0` and `lammps==2025.7.22.2.0`.

Validation: relative links and JSON examples checked, unsupported `.pt2` `eval-desc` command absent, `git diff --check` passes, and `skills/deepmd-finetune-dpa3/SKILL.md` still matches upstream exactly.

SchrodingersCattt:
Addressed the latest review comments in two focused commits:

- `adc86165`: qualify atomic virial by the archive's `do_atomic_virial` metadata and add `--head SELECTED_BRANCH` to the multi-task DPA4 embedding example. The wording reflects current source behavior: ordinary non-spin `dp --pt freeze` exports atomic virial by default, while specialized spin/conversion routes may not.
- `263c3408`: remove both `tee /dev/tty` pipelines so pinned `lmp -h` commands write directly to stdout in non-interactive environments.

Validation: `git diff --check`, focused assertions for the metadata/head/help changes, editor diagnostics, and confirmation that `skills/deepmd-finetune-dpa3/SKILL.md` still matches upstream.

OutisLi (changes requested):
[P1] Make required capability probes fail on a nonzero exit status

`skills/matmaster-dpa4-workflows/scripts/check_environment.py` currently treats executable discoverability as capability success. I reproduced this with a fake `dp` on `PATH` that exits 42: `--probe --require-deepmd` reports `available: true`, `returncode: 42`, and `failures: []`, then exits 0. A broken or incompatible DeePMD/LAMMPS/Bohrium runtime can therefore pass the validation gate and let automation continue packaging or submitting the job.

Please define required capability usability from the probe result when `--probe` is active (for example, discovered and `returncode == 0`) and apply that same predicate consistently to `bohr`, `dp`, and the `lmp`/`lmp_mpi` alternatives. A regression test with a discoverable executable returning nonzero would pin the contract.

OutisLi (changes requested):
[P1] Keep required case files inside the packaged case directory

`skills/matmaster-dpa4-workflows/scripts/check_environment.py` documents `--require-file` as case-relative, but currently checks `(case / name).is_file()` without enforcing that boundary. I reproduced a successful exit with `--case-dir parent/case --require-file ../model.pt` when `model.pt` exists beside the case: the report marks both `run.sh` and `../model.pt` present and leaves `failures` empty. An absolute path behaves the same way.

Bohrium packages only the directory passed as `--input_directory`, so this gate can validate a model/input that will not exist in the remote workspace. Please reject absolute paths and any resolved required path outside `case_dir` (including symlink escapes), retain normalized relative names in the report, and add regression coverage for `..` and absolute-path inputs.

OutisLi (changes requested):
The documented non-submitting validation command is not a verified Bohrium CLI interface. The current official `bohr job submit` documentation does not list either `--dry-run` or `--output`, and the available CLI reproduces this as `Error: unknown flag: --dry-run`; its submit help lists neither option. Therefore the workflow safety gate cannot run as written. Line 54 then tells the agent to remove `--dry-run` only for production, but there is no supported validation path before that production submission.

Please do not encode an unverified version-specific CLI contract. Use a version-verified, genuinely non-submitting validation interface if Bohrium provides one; otherwise keep the local job-spec/input checks as validation and leave the actual `bohr job submit` strictly behind explicit user authorization. If 2.5.17 really supports these flags, please add an authoritative reference or executable test that establishes that contract. Current CLI reference: https://bohrium-doc.dp.tech/docs/bohrctl/job/

OutisLi (changes requested):
`audit_cases.py` silently scans only the first 50,000,000 characters of each log. Any fatal signature after that boundary is ignored, so a failed long-running job can be reported as `passes_file_audit` with exit status 0.

I reproduced this with a 50,000,077-byte `run.log` in train mode whose final lines are `Traceback (most recent call last):` and `RuntimeError: failed after long training`. The script returned `passes_file_audit`, `reasons: []`, `fatal_signatures: []`, and exit code 0. The report warning about scientific acceptance does not make this safe, because this is a false result for the fatal-signature audit itself.

Please scan logs incrementally with bounded memory while accumulating fatal signatures, completion markers, and the last energy, rather than truncating the log. Reading only head plus tail would still miss failures in the middle. Please also add a regression test with a fatal signature beyond the current 50 MB boundary.

OutisLi (changes requested):
`audit_cases.py` can return success while expected cases are completely absent. The earlier `in.lmp` reproduction is one symptom: `infer_mode()` recognizes `in.lmp`, but marker-based `find_case_dirs()` does not. More fundamentally, any submitted case whose download failed, whose directory is empty, or whose files use an unlisted name has no marker and therefore disappears from `case_count` instead of being reported as missing.

I reproduced the concrete false pass with one valid training case plus a second directory containing only `in.lmp` and no log. The report contained only the valid case, returned `case_count: 1`, `passes_file_audit: 1`, and exited 0. Adding `in.lmp` to the marker tuple would fix only this filename and leave absent or empty expected cases invisible.

The workflow already requires a manifest that records every case. Please make the completeness-sensitive audit manifest-driven, or accept an explicit repeatable expected-case list, and iterate those case directories even when they contain no recognized files. Marker discovery can remain an optional convenience mode, but it must not produce a successful completeness verdict without an expected inventory. Add regression coverage for an absent case directory, an empty case directory, and an `in.lmp` case without a log alongside a valid case; all must remain visible and make the audit fail.

OutisLi (changes requested):
`make_job_spec.py` describes itself as generating and validating a Bohrium job specification, but it accepts any `--result-path` and writes it unchanged. The Bohrium `bohr job submit` contract limits automatic result destinations to `/personal` or `/share`: https://bohrium-doc.dp.tech/docs/bohrctl/job/

I reproduced this with `--result-path /tmp/ephemeral-results`; the script exited 0 and emitted that value into `job.json`. The generated specification therefore violates both the platform contract and this skill requirement that results use durable storage.

Please validate the normalized path before writing the specification. It must be `/personal`, `/share`, or a descendant of one of those directories. Use path-component containment rather than a raw string prefix so values such as `/personal-old/run` are rejected. Please cover valid durable paths plus `/tmp/result`, `/personal-old/run`, and a relative path in regression tests.

OutisLi (changes requested):
`audit_cases.py --require` does not enforce its documented case-relative boundary. `match_required()` passes the user pattern directly to `case.glob()`, so `..` and symlink traversal can satisfy a required output with a file outside the audited case.

I reproduced this with `root/checkpoint.pt` and `root/case/run.log`, where the case itself has no checkpoint. Running the audit with `--mode train --require ../checkpoint.pt` returned `passes_file_audit`, exit code 0, and recorded `../checkpoint.pt` as the successful match.

This is the same containment invariant that `check_environment.py --require-file` needs. Please avoid two local patches: introduce one shared case-path containment helper used by both scripts. It should reject absolute paths and parent traversal, and verify every resolved match remains under the resolved case directory so symlink escapes are rejected as well. Please test ordinary exact paths, globs, `..`, absolute paths, and symlink escapes.

OutisLi (changes requested):
The pinned online environment cannot provide the multi-rank DPA4 behavior documented by this PR. `skills/lammps-deepmd/SKILL.md` installs `deepmd-kit==3.2.0b0`, but the `v3.2.0b0` PT freezer hard-codes `has_comm_artifact: False` and never embeds `model/extra/forward_lower_with_comm.pt2`: https://github.com/deepmodeling/deepmd-kit/blob/v3.2.0b0/deepmd/pt/entrypoints/freeze_pt2.py#L253-L264. The C++ runtime in that same tag rejects multi-rank message-passing inference when this artifact is absent. This directly contradicts `model-deployment.md`, which says the ordinary PT freeze carries the with-comm artifact and supports multi-rank LAMMPS out of the box.

Do not replace this with another fixed DeePMD-kit release. These DPA4/PT/PT-expt capabilities are evolving on `upstream/master`, so the skill should use the latest master source and record the resolved commit SHA in the execution/provenance report. Resolve a LAMMPS version compatible with that checkout rather than forcing DeePMD-kit back to the old beta pin. If the master bootstrap needs a separate installation step instead of the current one-line `uvx` command, document that explicitly; the recommended environment must actually implement every deployment capability claimed by the skill.

OutisLi (commented):
The `Agent responsibilities` list in `skills/lammps-deepmd/SKILL.md` has lost its hierarchy. Under `Confirm the minimum simulation inputs`, only the structure/data file remains nested; the model artifact, type mapping, ensemble, and run parameters at lines 28-31 are top-level bullets, after which the numbered responsibility list resumes. The rendered document therefore splits one input contract across two unrelated list levels and reads like a patch residue.

Please indent all minimum-input bullets under the same numbered responsibility and make their punctuation/line wrapping consistent, so this section has one clear hierarchy for both human readers and agents.

OutisLi (changes requested):
[P1] Handle valid datasets without type_map.raw instead of assuming it exists

The held-out admission rule states that type.raw is always indexed into the system type_map.raw, but type_map.raw is optional in the DeePMD system format. When it is absent, DeePMD interprets type.raw directly in the model type-map order. The current rule therefore either rejects valid systems or, worse, encourages an agent to invent an element-identity mapping that the dataset does not contain; a wrong mapping can produce scientifically invalid held-out metrics while dp test still completes.

Please express the two actual contracts explicitly: when type_map.raw exists, compare/remap by element identity; when it is absent, require provenance that the dataset indices already use the candidate model type-map order (and fail closed if that cannot be established). This should remain an admission invariant rather than being inferred from filenames or index ranges.

OutisLi (changes requested):
[P1] Preserve the selected head when evaluating a multi-task checkpoint

The complete held-out command always invokes dp --pt test on selected.pt without --head. A native multi-task checkpoint cannot generally be evaluated this way: DeepEval asserts that a head must be selected unless the checkpoint happens to define a Default alias. The reference is linked as the production-admission procedure, and DPA4 training in this same skill set explicitly supports multi-task models, so the documented path fails before evaluating any frame for a valid multi-task candidate.

Please make the artifact contract explicit here: inspect model-branch for a multi-task checkpoint and pass --head SELECTED_BRANCH to every per-system dp test invocation. A frozen .pt2 produced with --head is already single-head and should not receive another branch argument.

OutisLi (changes requested):
[P1] Keep the reusable LAMMPS asset consistent with the DPA4-safe example

The updated top-level skill fixes its NVT example with atom_modify map yes, an explicit pair_coeff element mapping, and matching dump element labels, but it still points readers to assets/input.nvt.lammps as the reusable template. That asset remains on the old contract: no atom map, pair_coeff * * with implicit ordering, and no element labels. An agent following the referenced asset for DPA4 can therefore fail the required atom-map check or silently use an unaudited type mapping, even though the prose example is correct.

Please make the skill expose one canonical template rather than two divergent copies. At minimum, update the asset to the same safe mapping contract and cover the asset in the consistency test; preferably generate or verify both rendered examples from one source so future edits cannot leave this patch-style split again.

OutisLi (changes requested):
[P1] Make the advertised minimal DPA4 model section pass argcheck

The new DPA4 training reference labels its JSON as a minimal model configuration but omits model.fitting_net and then says model.type=dpa4 selects a default energy fitting network. That is not the current schema contract: the dpa4 model variant gives fitting_net.type a default only after a fitting_net object exists, while fitting_net itself is required. Normalizing the shown input fails with ArgumentKeyError: at model, key fitting_net is required. An agent copying the advertised minimal block therefore cannot start training.

Please either include an explicit fitting_net object in the minimal configuration or change the schema deliberately if omission is intended. Add a test that parses the actual fenced JSON and runs argcheck normalization; substring assertions cannot catch an invalid example.

OutisLi (changes requested):
[P1] Create the detail-output directory before the documented test command

The complete held-out example writes -d details/system.000 but never creates details/. dp test does not create parent directories: save_txt_file opens the derived .e.out/.f.out paths directly with Path.open. In a clean evaluation workspace, the production-admission command therefore evaluates the model and then fails with FileNotFoundError while trying to retain the required evidence.

Please make directory creation an explicit step before the per-system commands (and keep the existing no-overwrite check). A focused workflow test should execute the documented path from a clean temporary directory rather than only asserting that -d text is present.

OutisLi (changes requested):
[P1] Do not claim that use-pretrain-script restores the complete DPA4 model configuration

The fine-tuning skill recommends --use-pretrain-script when the architecture is unknown and says it inherits the stored model configuration except type_map. The current fine-tune builder does not do that for DPA4: _apply_pretrained_model_params copies only descriptor and fitting_net. Model-level architecture/semantics such as type, spin, bridging_method, bridging_r_inner/outer, and related specialized settings remain whatever input.json supplied. I reproduced this by applying the builder to a native-spin, ZBL-bridged pretrained config and a plain DPA4 target; the returned config contained only the copied descriptor/fitting and silently dropped spin and bridging.

This is unsafe guidance for exactly the unknown-architecture case: training may fail during state loading or proceed with a different physical model. Please either make the core option actually preserve the complete model contract (with type_map and intentional trainability/LoRA exceptions defined centrally), or narrow the skill and require every DPA4 model-level field to be inspected and reproduced explicitly. A bounded initial run is not a substitute for constructing the right model.

OutisLi (changes requested):
[P1] Make the inference precision/performance policy explicit before freezing DPA4

Every DPA4 freeze recipe runs dp --pt freeze without inspecting or recording DP_TRITON_INFER, DP_TF32_INFER, and DP_AMP_INFER. These are not ordinary runtime knobs for .pt2: the freezer reads them while constructing/tracing the model and bakes the selected kernels and precision policy into the AOTInductor archive. Existing shell exports can therefore silently change force smoothness and numerical behavior, while the default DP_TRITON_INFER=0 also leaves the production artifact on the unfused slow path. Moving freeze/run to the same node does not establish a reproducible policy by itself; at level >=2 it only makes the hardware-specific tuning relevant.

Please add one explicit pre-freeze decision shared by the training, fine-tuning, inference, and LAMMPS routes: inspect/reset these variables, choose a validated policy (levels 1/2 retain fp32 accumulation; level 3 and AMP/TF32 require task-specific validation), and record the values with the artifact provenance. The command examples should not inherit an unknown shell environment, and the production workflow should not silently bake the slow default.

SchrodingersCattt:
Addressed the latest review in fe8168c3 and dd3fb6aa (plus pre-commit auto-fix 3a871cd8): optional type_map.raw is provenance-gated; native multi-task held-out tests require the selected --head while frozen pt2 stays single-head; the LAMMPS asset now matches the documented mapping/dump contract; the minimal DPA4 JSON includes fitting_net and has an argcheck normalization test; detail directories are created collision-free and exercised from a clean temporary directory; --use-pretrain-script is limited to descriptor/fitting_net replacement; and DPA4 freeze-time Triton/TF32/AMP policy is explicit. Focused local result after the auto-fix: 13 passed, 1 skipped only because this source checkout has no built deepmd.lib; the normalization test executes where the extension is built. Full CI is still running.

SchrodingersCattt:
Added DPA4C training guidance in `783ff88d` after merging current upstream master:

- route DPA4C requests from `deepmd-train` to a dedicated progressive-disclosure reference
- state the required `dp --pt-expt` backend
- distinguish DPA4's `model.use_compile` / `model.enable_tf32` from DPA4C's `training.enable_compile` / `training.enable_tf32`
- include a complete accelerated DPA4C template, deployment commands, and a checklist
- add a regression test that pins these backend-specific paths

Validation: `source/tests/common/test_agent_skills.py` -> `14 passed, 1 skipped` (the existing normalization test skips without a built extension).

njzjz (approved):

Issue #5954, opened 2026-08-03
Record exact label availability when building LMDB datasets


OutisLi:
### Summary

LMDB generation should record exact per-frame label availability in dataset metadata. Data readers should consume that metadata instead of reading or probing frame payloads at training startup to infer label-availability signatures.

### Motivation

A training batch exposes one `find_*` value for each label. Frames with different optional-label availability therefore need either exact homogeneous grouping or conservative reduction at batching time.

Determining those groups online is problematic for large LMDB datasets:

- An exact result requires inspecting every frame before useful training work starts.
- A bounded probe avoids that startup scan for common uniform datasets, but it is inherently probabilistic.
- Rare availability signatures outside the sample may be missed. If they later enter a mixed batch, conservative reduction preserves correctness but discards valid supervision from the other frames in that batch.
- Genuinely heterogeneous datasets may still trigger an expensive full scan.

Label availability is an intrinsic property of the serialized dataset, so it should ideally be computed once when the LMDB is created rather than rediscovered by every consumer.

### Proposal

When generating an LMDB dataset:

1. Normalize label names according to the same key mapping used by readers.
2. Compute each frame's effective availability from source presence and any explicit `find_*` value.
3. Encode distinct availability bitsets as compact signature IDs.
4. Store a schema-versioned signature table and one compact signature ID per frame in LMDB metadata.
5. Store dataset-level uniformity directly when all frames share one signature.

At read time:

1. Select only labels required by the active model and loss.
2. Project the stored signatures onto those labels without decoding frame payloads.
3. Partition tracked optional labels exactly.
4. Keep mandatory-label validation during frame decoding as a corruption and contract check.
5. Fall back to the existing reader-side detection path for legacy LMDBs without this metadata.

### Implementation considerations

- Metadata should describe source facts, not model-specific policy. Decisions such as whether a missing label may use a default should remain with the consumer.
- LMDB merge and filtering utilities should preserve or rebuild the signature metadata, including remapping signature IDs where necessary.
- The metadata schema should be versioned so readers can reject malformed or incompatible indexes safely.
- A compact representation should have low overhead: typically one small integer per frame plus a small table of unique signatures.

### Expected benefits

- Exact label grouping without probabilistic misses.
- No `O(number of frames)` payload scan at training startup.
- Reusable availability information for training, statistics, validation, sampling, and inference.
- A clearer separation between immutable dataset facts and consumer policy.

### Relationship to reader-side mitigation

[...]

PR #5962, opened 2026-08-08
feat(lmdb): support mixed-size batches and lazy label availability

### LINKED ISSUE #5965: pt_expt: support correct mixed-size spin batches and ragged training
## Problem

PR #5962 adds `mix:N` batches that may contain frames with different atom counts. Spin models currently opt out of the ragged layout and therefore receive rectangular batches padded to the widest frame.

That rectangular fallback is not semantically correct today:

- The public spin-model output does not propagate the padding `mask` to `EnergySpinLoss`.
- `EnergySpinLoss` consequently uses the padded `max_nloc` as the atom count for every frame. Energy and virial normalization are wrong for shorter frames, while real-force and atomic-energy losses are diluted by padded slots.
- Native-spin `mask_mag` is derived from `atype`; phantom `atype == -1` rows should be excluded explicitly rather than relying on the last real type being non-magnetic.

A direct padded native-spin forward currently returns `atom_energy`, `energy`, `force`, `force_mag`, `mask_mag`, and `virial`, but no padding `mask`.

## Ragged support

There is no mathematical requirement for `EnergySpinLoss` to remain rectangular. Its terms are either frame-level or node-level:

- energy and virial can normalize from per-frame counts derived from `n_node`;
- coefficient-weighted atomic energy can use `segment_sum`;
- real force, magnetic force, and atomic energy can reduce directly on the flat node axis;
- `mask_mag` can remain flat with shape `(N, 1)`.

The graph-native spin lower already consumes flat `spin` and produces magnetic derivatives. The missing pieces are the public/training seams: `call_common_ragged`, native-spin `forward_ragged`, the compiled graph-training wrapper, output translation for `force_mag`/`mask_mag`/`n_node`, and a ragged-aware `EnergySpinLoss`.

## Expected behavior

1. Rectangular spin batches propagate the padding mask, exclude phantom atoms from every applicable loss term, and combine padding validity with `mask_mag`.
2. Graph-native spin models use ragged `mix:N` batches when both the model and `EnergySpinLoss` advertise support.
3. Traditional virtual-atom spin models remain rectangular until their model path has an explicit ragged contract.
4. Ragged and correctly masked rectangular calculations agree for energy, real force, magnetic force, virial, atomic energy, and coefficient-weighted energy.
5. Eager and compiled pt_expt training preserve the same layout and output semantics.

Related: #5962.

OutisLi:
## Summary

- support `mix:N` LMDB batches containing frames with different atom counts, using padded rectangular batches for dense models and a flat ragged node axis for eligible graph models
- compact phantom atoms before graph-model evaluation and make loss reductions, validation weighting, and epoch sizing use real atom counts
- resolve label availability lazily after data requirements are registered, so required, optional, defaulted, and partially available fields are handled without an eager full-dataset scan
- keep non-mixing and native-spin models on their existing rectangular public paths; the ragged regression coverage uses upstream DPA1 and avoids model-specific dependencies

## Behavioral changes

- Masked per-atom loss terms now pool all included labels across the batch instead of averaging per-frame means. This intentionally retires the bit-identical reduction guarantee from #5738/#5783 for existing `mixed_type` NPY datasets whose frames have different real atom counts: those frames are weighted by their real label counts rather than equally. Uniform-atom-count batches are unchanged. Hessian pair terms remain normalized per frame so their quadratic component count does not make large structures dominate a batch.
- Legacy LMDB files have no exact per-frame label-availability metadata. To avoid an eager O(N) startup scan, the reader uses a bounded probe and conservatively reduces per-frame `find_*` flags at collation. A missed rare signature may discard valid supervision for the affected batch, but default-filled values are never treated as real labels. Recording exact availability metadata when generating LMDB datasets is tracked in #5954.

## Testing

- all pre-commit hooks passed for the changed files
- 375 passed, 2 skipped, 1 deselected, and 13 subtests passed in the main targeted LMDB/PT/PT-expt/model suite
- 15 passed in the isolated loss-reduction and decoder-pool regression suite
- review fixes: 100 passed, 2 skipped, and 2 subtests passed in the common loss suite; 60 passed in the PT padding-loss suite; all 25 LMDB training tests passed; padded/unpadded DPA2 graph and Hessian parity tests passed

## Summary by CodeRabbit

- **New Features**
  - Added LMDB batching for frames with different atom counts, including mixed-size and ragged layouts.
  - Added ragged-batch inference and training for supported energy and spin models.
  - Added configurable data-source policies for optional labels and parameters.
  - Added safer handling of padded atoms across neighbor graphs and model outputs.

- **Bug Fixes**
  - Improved per-atom loss normalization for uneven and padded batches.
  - Prevented padded atoms from affecting neighbor searches, metrics, or losses.
  - Improved handling of missing labels and default-valued data.

- **Documentation**
  - Documented mixed-size batching, ragged data, and per-atom normalization.

Closes #5965


deepmd/pt_expt/train/training.py
```diff
@@ -2081,6 +2196,53 @@ def _raise_if_full_validation_unsupported(
...
+                not model.has_spin()
+                and model_uses_graph_lower(model)
+                and hasattr(model, "forward_ragged")
+            )
+            for data_map in data_maps:
+                data = (
+                    data_map.get(task_key) if isinstance(data_map, dict) else data_map
+                )
+                if hasattr(data, "use_ragged_batches"):
+                    data.use_ragged_batches(ragged)
```
wanghan-iapcm, line 2440: This is the root cause of the two loss findings below, and the smallest thing to change.

`ragged` is decided from the model's capabilities alone -- non-spin, graph-lower, has `forward_ragged` -- and then applied to every LMDB data system. Nothing here consults the batch-size rule or asks whether any frame in the dataset actually differs in atom count.

The consequence is that this PR is not opt-in. Before it, every LMDB run used the rectangular masked layout. After it, every dpa2/dpa3-class pt_expt LMDB run is on the flat layout, whether or not the user wrote `mix:N`. So the two problems below are not "new feature has rough edges" -- they change behaviour for configurations that exist today and did not ask for anything.

Gating this on the reader genuinely needing it -- `reader.mixed_nloc`, or whatever the equivalent predicate is once the batch rule is known -- keeps the feature and confines its blast radius to the people who opted in. It also means a uniform-frame batch keeps producing the `mask` the loss code is already written to consume, which resolves both inline findings without touching the loss at all.

deepmd/dpmodel/loss/ener.py
```diff
@@ -253,45 +253,51 @@ def call(
...
+        #
+        # ``inv``, the reciprocal real atom count of each frame, is what the
+        # extensive frame-level terms (energy, virial) divide by. ``maskf``
+        # marks the padded rows the per-atom terms must skip. A rectangular
+        # batch carries both in its mask: summing it gives the counts, and its
+        # zeros are the padding. A ragged batch pads nothing, so it states the
+        # counts alone and its per-atom terms reduce over their whole axis.
+        maskf = None
+        inv = None
+        if "n_node" in model_dict:
```
wanghan-iapcm: The ragged branch discards information the model already provides, and the loss silently changes.

`mask` is not merely "which rows are real". `_prepare_graph_nodes` builds it as `(atype >= 0) & atom_excl.build_type_exclude_mask(...)`, so it also excludes atoms removed by `atom_exclude_types`. `n_node` does not -- it is the raw per-frame node count. Taking `inv = 1/n_node` therefore over-counts the denominator whenever `atom_exclude_types` is set.

The second half is worse: because `maskf` stays `None` on this branch, every per-atom term -- `rmse_f`, `rmse_ae`, `rmse_pf` -- falls through to the unmasked `else`, which sums over the excluded atoms too. So their residuals enter the numerator as well.

Concretely: one frame, 10 atoms, `atom_exclude_types` removing 2, energy label present. Before, the energy term divided by 8 and the force MSE was the sum over the 8 kept atoms divided by 24. Now it divides by 10 and the force MSE sums all 10 over 30, with the two excluded atoms' non-zero residuals included. Same model, same data, different loss and different reported `rmse_f`, with no error.

The information is right there -- [`EnergyModel._to_public_keys`](https://github.com/deepmodeling/deepmd-kit/blob/ae387ea2a76aaf22a5cd06839d85b7c7025549db/deepmd/pt_expt/model/ener_model.py#L288-L294) copies **both** `mask` and `n_node` into the public dict. This branch just stops looking once `n_node` is present.

Either honour `mask` when the model supplies it (making this `if "n_node" in model_dict and "mask" not in model_dict`, or better, always preferring `mask`), or fix it upstream by not flipping uniform runs to ragged at all.

deepmd/dpmodel/loss/ener.py
```diff
@@ -702,21 +721,33 @@ def call(
...
                 raise NotImplementedError(
                     f"Loss type {self.loss_func} is not implemented for atom prefactor force loss."
                 )
         if self.has_gf:
+            if maskf is None and inv is not None:
+                # ``natoms`` below is one number for the whole batch, which a
+                # padded batch can honour and a concatenated one cannot: its
+                # frames differ in atom count, so ``drdq``, stored per frame
+                # against a common atom axis, has no shape to take.
+                raise NotImplementedError(
```
wanghan-iapcm, line 785: This guard is right in principle but, given the unconditional flip, it aborts runs whose frames are uniform.

The message states the precondition accurately: the generalized-force loss needs every frame in the batch to share an atom count, because `drdq` is stored against a common atom axis. The problem is that it triggers on the *layout*, not on the precondition. A dpa2/dpa3 LMDB run whose frames all have the same atom count satisfies the stated requirement perfectly -- and is now served in the concatenated layout anyway, because `_configure_batch_layout` does not consult frame sizes. It reaches this line and dies with `NotImplementedError` on the first training step, where before this PR it trained fine through the `maskf is not None` path.

So a user with `start_pref_gf != 0` who changes nothing about their config or data gets a hard failure from a PR whose feature they are not using.

The Hessian term has the same exposure without the courtesy of an error: under the same flip it moves from `masked_pair_mean` into the flat `else`, silently changing its reduction on uniform data.

Gating the layout on actual need (see the `_configure_batch_layout` comment) resolves both. If you would rather keep the flip, then this guard should test the precondition it names -- whether the frames in this batch actually differ -- rather than which layout carried them.

deepmd/dpmodel/loss/reduction.py
```diff
@@ -45,26 +92,20 @@ def masked_atom_mean(elem: Array, maskf: Array, ncomp: int) -> Array:
...
-    # to a neutral per-frame value of 0. Frames with real atoms are untouched,
-    # preserving the bit-identical guarantee.
-    has_dof = per_frame_dof > 0
-    safe_dof = xp.where(has_dof, per_frame_dof, xp.ones_like(per_frame_dof))
-    per_frame = xp.where(
-        has_dof, per_frame_sum / safe_dof, xp.zeros_like(per_frame_sum)
-    )
-    return xp.mean(per_frame)
+    total = xp.sum(elem * maskf[:, :, None])
+    total_dof = xp.sum(maskf) * ncomp
```
wanghan-iapcm: This changes loss values for data that predates this PR, and reverses a decision that was made explicitly.

The rewrite goes from mean-over-frames of (per-frame sum / per-frame dof) to a single pooled `sum / total_dof`. Those are not the same number when per-frame real atom counts differ: frame A with 10 real atoms and sum 100, frame B with 100 real atoms and sum 100, gives `(10 + 1)/2 = 5.5` before and `200/110 = 1.818` after.

The part that makes this more than a new-feature choice is reachability. The masked branch is entered whenever `mask` is in `model_pred`, and `deepmd/pt/loss/loss.py` sets `model_pred["mask"] = (atype >= 0)` when it is absent. That is pre-existing `mixed_type` infrastructure from #5738, nothing to do with LMDB. So ordinary `mixed_type` npy training whose `real_atom_types.npy` spends a different number of `-1` rows on different frames -- which is the normal case for that format -- gets different loss values and different gradients after this PR.

The previous docstring was unusually explicit that this must not happen:

> the "bit-identical for non-mixed batches" guarantee from PR #5738 is preserved (defaulting the mask to all-ones would change the reduction order at the ULP level and is deliberately NOT done here)

and #5783, which introduced that wording, was reviewed and approved on the basis that it preserved reduction order. The new docstring argues the pooled convention is better and does name the `mixed_type` npy case where the identity breaks -- so this is not hidden. But it is framed as forward-looking design rationale, never as "this used to compute X and now computes Y for datasets that already exist."

I am not asking you to revert it. Pooling by label count is defensible and probably the better convention. What I want is: say in the PR body and the release notes that the #5738/#5783 guarantee is being deliberately retired and which existing configurations change; and confirm that any fixture or golden-value test from #5738/#5783 that pinned exact numbers was re-derived rather than updated to match the new output. That distinction is the whole reason this repo treats a float mismatch as a bug signal rather than a tolerance to loosen.

deepmd/dpmodel/loss/reduction.py
```diff
@@ -45,26 +92,20 @@ def masked_atom_mean(elem: Array, maskf: Array, ncomp: int) -> Array:
...
+    # A batch of nothing but padding has no label to average over, and the
+    # ratio would be 0/0 = NaN -- poisoning the whole batch loss and, under
+    # autograd, its gradient. The division still runs on a safe denominator so
+    # that the discarded branch stays differentiable.
+    has_dof = total_dof > 0
+    safe_dof = xp.where(has_dof, total_dof, xp.ones_like(total_dof))
+    return xp.where(has_dof, total / safe_dof, xp.zeros_like(total))
 
 
 def masked_pair_mean(elem: Array, maskf: Array, ncomp: int) -> Array:
```
wanghan-iapcm, line 120: While you are here: `masked_pair_mean` was left on the old convention, and the new doctrine does not mention it.

`masked_atom_mean` just became batch-pooled on the argument -- stated in the module docstring immediately above -- that mean-of-per-frame-means "gives every frame the same weight whatever its size, which makes a label in a small frame count for more than one in a large frame". `masked_pair_mean`, used for the Hessian loss, is byte-identical to before and still computes exactly that per-frame mean.

The new "Reduction convention" section enumerates only "per-atom terms" and "frame-level terms". Pair terms are not in either list, so a reader cannot tell whether the omission is deliberate.

On the ragged batches this PR is built for, that means Hessian weights frames by atom-pair count while force, energy, dos and tensor weight by pooled label count. If keeping the old convention for pair terms is intentional, the docstring should say why; if not, it should move with the others.

source/tests/common/dpmodel/test_loss_padding.py
```diff
@@ -514,6 +534,99 @@ def _padded_atom_flat(arr_A, arr_B, ncomp):
...
+            limit_pref_ae=0.0,
+            start_pref_pf=0.0,
+            limit_pref_pf=0.0,
+            **kwargs,
+        )
+
+    @pytest.mark.parametrize("loss_func", ["mse", "mae"])
+    @pytest.mark.parametrize("intensive", [False, True])
+    @pytest.mark.parametrize("use_huber", [False, True])
+    def test_mask_and_n_node_agree(self, loss_func, intensive, use_huber):
```
wanghan-iapcm, line 567: This test cannot observe the behaviour it is named for.

`test_mask_and_n_node_agree` builds `mask = np.ones((nf, nloc))` and `n_node = np.full(nf, nloc)` with `nf, nloc = 3, 5` -- uniform counts, no padding, no raggedness. The module's own docstring says that in exactly this case "the choice between the two is unobservable". So the test asserts the two paths agree precisely where they cannot disagree, and it would pass just as happily with the mask-dropping described in the `n_node` branch comment above. It also sets `pref_f = pref_ae = pref_pf = 0`, which switches off the per-atom terms that the ragged branch actually mishandles.

A case with genuinely unequal `n_node` across frames, with the per-atom prefactors on, and with `atom_exclude_types` set on at least one variant, would distinguish them -- and would have caught the denominator issue.

Related gap: the new `compact_nodes` call in `make_model._call_common_graph` has no numeric end-to-end test with a real phantom (`atype < 0`) batch. What exists is a unit test of `compact_nodes` on a synthetic graph and a training-level smoke test asserting `isfinite(loss)`. Given the round trip is scatter-inverts-gather, a value comparison against the equivalent unpadded batch is cheap and would pin it.

wanghan-iapcm (changes requested):
Mixed-size batching is worth having and much of the machinery here is careful. Things I checked rather than assumed, and found correct: the frame-offset arithmetic in the `from_ijs` ragged rewrite, and that all four call sites were updated to pass `full((nf,), nloc)`; `compact_nodes` / `expand_node_values` being an exact round trip, since `keep_index` holds distinct positions so the `segment_sum` scatter inverts the gather; the claim that every builder already refuses an edge touching a phantom, which holds in `ase_builder`, `nv_graph_builder` and `vesin_graph_builder` alike; and the `sezm_model` rebinding, which actually *fixes* a pre-existing case where `atom_excl` was applied to raw possibly-negative `atype`. One reported problem I want to explicitly withdraw before it wastes your time: I looked hard at whether the padded generalized-force path was mis-normalised by phantom rows, and it is not -- `_finalize_atomic_ret` sets `mask` unconditionally, so that path takes the masked branch.

Four things I do want changed, and two of them are the same bug.

**The shared root cause: ragged is switched on for models that never asked for it.** [`_configure_batch_layout`](https://github.com/deepmodeling/deepmd-kit/blob/ae387ea2a76aaf22a5cd06839d85b7c7025549db/deepmd/pt_expt/train/training.py#L2199-L2248) enables the ragged layout for every LMDB data system whenever the model is non-spin, graph-lower and has `forward_ragged` -- with no reference to the batch-size rule and no reference to whether frames actually differ in size. So every pt_expt LMDB run on a dpa2/dpa3-class model changes layout, including runs with perfectly uniform frames that never opted into mixed-size batching. That is what turns two narrow new-feature concerns into regressions on existing configurations. Gating it on the reader actually needing it fixes both inline comments below at once.

**The other regression is the LMDB availability rewrite, and it is the one I would fix first.** This is not anchorable inline because the file is too large for GitHub to serve a patch, so here it is with permalinks.

PR #5839 (`e37cc0095`), which fixed #5636, deliberately did two things: an exact per-frame availability signature used to partition batches, and a hard `ValueError` in `collate_lmdb_frames` if the `find_*` flags disagreed inside a batch. This PR removes both. The scan becomes [`_probe_uniform_availability`](https://github.com/deepmodeling/deepmd-kit/blob/ae387ea2a76aaf22a5cd06839d85b7c7025549db/deepmd/dpmodel/utils/lmdb_data.py#L424-L470) over [`_AVAILABILITY_PROBE_FRAMES = 256`](https://github.com/deepmodeling/deepmd-kit/blob/ae387ea2a76aaf22a5cd06839d85b7c7025549db/deepmd/dpmodel/utils/lmdb_data.py#L95-L99) evenly spaced frames, and when the probe reports uniform, [`availability_groups`](https://github.com/deepmodeling/deepmd-kit/blob/ae387ea2a76aaf22a5cd06839d85b7c7025549db/deepmd/dpmodel/utils/lmdb_data.py#L2530-L2545) returns a single group -- so no partitioning happens for the rest of the dataset at all. The guard becomes [`_batch_find_flags`](https://github.com/deepmodeling/deepmd-kit/blob/ae387ea2a76aaf22a5cd06839d85b7c7025549db/deepmd/dpmodel/utils/lmdb_data.py#L2690-L2700), which on disagreement silently sets the flag to 0 for the whole batch. No exception, and no log line on either path.

Composed, those two changes reproduce #5636 exactly: on a dataset larger than 256 frames where the frames missing a label fall between probe points, the label is silently discarded for every frame in the batch. I am not neutral about this one -- I approved #5839 specifically because it added that guard, and the regression test that reassured me, `test_unrequested_labels_form_homogeneous_batches`, uses 8 frames, so `_evenly_spaced` returns every index and the probe is exact for it. It will keep passing while the property it was written to protect is gone. That is worse than the original bug, because the test now certifies it.

The PR's own comment concedes the probe is "a finding and not a proof" and even names caching the signature in `__metadata__` as the right fix. I would take that: keep exactness and make it cheap. If you want to keep the probe, then `collate_lmdb_frames` must keep a defensive check -- silently zeroing a label is precisely the failure mode #5839 existed to prevent.

Two smaller notes I am not blocking on. `masked_pair_mean` and the coverage gap are inline. Also, `deepmd/pt/loss/dens.py` still ignores `intensive_ener_virial` in its energy term while the adjacent `rmse_e` uses `inv_natoms**2`, so the displayed metric and the optimised quantity disagree -- that predates this PR (it is not yours to fix here), but you were rewriting those exact lines, so it is worth knowing.

CI is green, which given the above is itself informative: none of these have a test that can see them.

deepmd/pt_expt/train/training.py
OutisLi: Agreed. Commit 38cc7b9a4 moves the final policy to LmdbDataReader.use_ragged_batches: model capability is necessary, but the reader must also have mixed_nloc, which is true only for mix:N. Non-mixing graph-capable LMDB runs therefore retain the established rectangular layout. I also added a regression test for a DPA1 graph model with a fixed batch size.

deepmd/dpmodel/loss/ener.py
OutisLi: Agreed. The root cause is removed by requiring mix:N before selecting the ragged layout, so unchanged uniform LMDB runs remain rectangular and keep the generalized-force and Hessian paths they used before this PR. The generalized-force guard now explicitly identifies a genuinely ragged layout; only that unsupported representation is rejected.

deepmd/dpmodel/loss/reduction.py
OutisLi: Confirmed: pooling per-atom labels is intentional. The PR body now explicitly states that the #5738/#5783 bit-identical guarantee is retired for existing mixed_type NPY data whose frames have different real atom counts; uniform-count batches are unchanged. The affected tests were re-derived algebraically by changing the reference combination from equal frame weights to real-atom-count weights. Golden values were not merely updated and tolerances were not loosened to accept the new output.

source/tests/common/dpmodel/test_loss_padding.py
OutisLi: Agreed. The previous test could not distinguish the layouts. Commit 38cc7b9a4 replaces it with n_node = [2, 3], an actual padding slot, an excluded physical atom, and enabled force, atomic-energy, virial, and atom-prefactor terms, then compares the complete loss and metrics between rectangular and flat-ragged representations. I also converted the existing DPA2 graph and Hessian parity tests to compare padded inputs numerically against their equivalent unpadded inputs, including exact zero phantom outputs, rows, and columns.

deepmd/dpmodel/loss/reduction.py
OutisLi: Keeping masked_pair_mean per-frame is intentional. A Hessian carries a matrix whose component count grows quadratically with atom count; pooling all pair components would therefore give large structures quadratic weight in the batch. Commit 38cc7b9a4 adds pair terms as a separate documented reduction category and explains why each Hessian matrix is normalized before frames are averaged.

deepmd/dpmodel/loss/ener.py
OutisLi: Agreed. Commit 38cc7b9a4 makes the ragged loss consume the model mask whenever it is present. n_node defines only the frame segment boundaries; frame_id_from_n_node plus segment_sum(mask) computes included counts, and masked_atom_mean accepts either rectangular or flat node axes. Consequently atom_exclude_types affects both numerator and denominator exactly as on rectangular batches. The replacement test uses unequal frame sizes, real padding, and an excluded physical atom with force, atomic-energy, and atom-prefactor terms enabled.

OutisLi:
@wanghan-iapcm Thank you for the careful review. On LMDB label availability, I agree with the factual concern: a bounded probe cannot prove uniformity, and when it misses a rare signature the affected batch can lose otherwise valid supervision. I am intentionally not restoring either the eager O(N) startup scan or the batch-time ValueError in this PR. The former defeats lazy startup for large legacy LMDB datasets, while the latter makes a missed probe turn into a shuffle-dependent training abort. The current batch-level AND reduction is conservative: it may disable a disputed label for that batch, but it never treats a default-filled value as real supervision. Exactness needs to be established when the LMDB is generated, by recording a schema-versioned per-frame availability signature in metadata and using that as the reader source of truth. That work is tracked in #5954. I have made both the legacy tradeoff and the long-term exact solution explicit in the PR body. The remaining requested changes are addressed in 38cc7b9a4 and replied to individually in their threads.

iProzd (changes requested):
The backend/layout work now looks coherent, and the current head addresses the earlier ragged-model, loss-contract, mask, and Hessian issues. One blocking correctness issue remains: the bounded legacy-LMDB availability probe discussed in https://github.com/deepmodeling/deepmd-kit/pull/5962#discussion_r3750453676 can still miss a rare signature and silently demote a valid label for the entire mixed batch. I reproduced this on dc4c99a39cc406422682b095b2832fd09e0cdb71 with 512 frames: frame 1 omitted `energy`, the 256-frame probe returned one 512-frame group, and decoding frames 0 and 1 changed `find_energy` from 1 for frame 0 alone to 0 for the batch. This makes the training objective depend on batch composition without an error. Please make availability exact for tracked optional labels (metadata when available, otherwise an exact scan), or fail loudly when mixed availability reaches collation. I do not think a probabilistic correctness boundary is safe for training.

iProzd (approved):

wanghan-iapcm (approved):
All six findings from my earlier review are addressed at dc4c99a3.

- The ragged layout is now genuinely opt-in: `use_ragged_batches` ends with `ragged and self.mixed_nloc`, and `mixed_nloc` is `self._mix_rule is not None`, so only a `mix:N` rule can select the flat node axis. Existing dpa2/dpa3 LMDB runs keep the rectangular layout, which also removes the root cause of the generalized-force abort.
- The ragged loss branch consumes the model mask whenever it is present. `frame_id_from_n_node` plus `segment_sum(mask, frame_id)` give the included counts, `n_node` now only sets the frame boundaries, and every per-atom term goes through `masked_atom_mean` on both layouts, so `atom_exclude_types` affects numerator and denominator alike.
- The generalized-force guard fires only under `is_ragged`, which is the representation it actually cannot serve.
- `masked_pair_mean` is now a documented third reduction category with the quadratic-growth rationale, rather than an unexplained holdout.
- The layout-agreement test is replaced by one that can fail: `n_node=[2,3]`, a real padding slot, a model-excluded physical atom, force, atomic-energy, virial and atom-prefactor terms enabled, and a 1e-14 comparison of the total loss and every metric across the two layouts.
- The reduction tests were re-derived algebraically through a `frame_weights` parameter rather than re-baselined, so no golden value was updated and no tolerance was widened.

Two things I am accepting as documented decisions rather than defects. First, pooling retires the bit-identical guarantee from #5738/#5783 for existing mixed_type NPY data whose frames hold different real atom counts; the Behavioral changes section states this and uniform-count batches are unchanged. Second, the bounded label-availability probe may discard valid supervision for a batch on a missed rare signature, with exact metadata tracked in #5954.

Worth keeping in mind on merge: the CUDA jobs report skipping, so a PR that reworks graph-model batching and the pt_expt training loop has no GPU coverage in CI.

PR #5503, opened 2026-06-07
feat(dpa4): multiple updates for DPA4/SeZM

OutisLi:
## Summary by CodeRabbit

* **New Features**
  * GPU-accelerated O(N) neighbor-list for large periodic CUDA workloads and explicit backend selection.
  * New S2/SO(3) grid projectors and grid-network nonlinearities plus fused GPU rotation kernels for faster transforms.
  * Expanded model descriptor knobs (per-block node sizing, kmax/extra node options, per-path grid toggles) and zonal coupling support.

* **Performance**
  * Improved compile/inference lowering, shared compiled-graph caching, TF32/inference controls, and eval-time activation checkpointing.

* **Bug Fixes**
  * Hardened neighbor-list/backend validation and dispatch.

* **Documentation**
  * Major DPA4/SeZM docs and inference environment guidance.

* **Tests**
  * New and rewritten unit tests covering grids, projections, rotations, and nlist backends.

PR #5963, opened 2026-08-09
fix(dpa4): align pt_expt training and native-spin fine-tuning

OutisLi:
## Summary

- align the pt_expt DPA4 parameter set and precision behavior with the PT reference
- keep compiled training and evaluation graphs mode-specific
- make native-spin DPA4 fine-tuning from a spin-free checkpoint function-preserving and trainable

## Details

The pt_expt wrapper treated the bare NumPy weights in `FrameExpand` and `FrameContract` as buffers, so they never reached the optimizer. It also used NumPy-only operations in `ReducedEquivariantRMSNorm`, coupled inference AMP to the training switch, and reused a train-mode compiled graph during evaluation. This change promotes the missing trainable arrays, makes the norm tensor-safe, separates training and inference precision policies, and caches compiled lowers independently for train and eval mode.

Native-spin fine-tuning had two related state-transfer problems. Per-type spin gates derived from `use_spin` were persisted in checkpoints, allowing an all-zero gate from a spin-free pretraining to override the fine-tune configuration. In addition, dormant randomly initialized spin routes became active at full amplitude when a magnetic type was introduced. Configuration-derived gates are now non-persistent and archived copies are ignored when loading older checkpoints; output-bias calibration receives the native moment; and native-spin models no longer fabricate a virtual-atom scale.

DPA4 spin routes now initialize at the zero function. `prepare_finetune` resets dormant routes only when activating a spin-free checkpoint, preserves already-trained matching routes, and rejects reassignment to a different magnetic-type set. The environment-spin gate moves after the quadratic form so it retains a nonzero gradient at zero. Versioned migration preserves the function of existing checkpoints, and the reset capability is forwarded through atomic-model compositions such as ZBL bridging.

This PR contains no DPA4C-specific or LMDB changes.

## Checks

- SeZM descriptor, Triton-dispatch, model, compile, native-spin, and LoRA tests: 119 passed and 27 skipped in the full run; the two Warp-dependent tests that the sandbox could not compile were rerun outside it and passed with all 24 subtests
- DPA4 dpmodel, PT/pt_expt gradient parity, pt_expt descriptor/model, native-spin, fine-tune, and compiled dynamic-shape tests: 130 passed
- full `source/tests/pt_expt/test_training.py`: 50 passed
- pre-commit hooks and `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Native-spin models now accept per-atom spin data during prediction and compiled training.
  * Added independent training and inference controls for AMP and TF32 precision.
  * Compilation now caches separate training and evaluation graphs.
  * Native-spin fine-tuning validates pretrained model compatibility and magnetic element settings.

* **Bug Fixes**
  * Improved migration and loading of older DPA4 and SeZM checkpoints.
  * Corrected restoration of spin configuration and dormant parameters.
  * Improved compatibility with legacy exported model state.
  * Fixed spin embedding initialization and gradient behavior.


deepmd/pt_expt/model/native_spin_model.py
```diff
@@ -42,6 +42,41 @@ class NativeSpinEnergyModel(make_native_spin_model(EnergyModel)):
       ``forward_common_lower_graph_exportable``.
     """
 
+    def prepare_finetune(self, pretrained_model: BaseModel) -> None:
```
wanghan-iapcm: This is the design question. Why does this need a new interface at all?

The reset is reachable through exactly one path. `source_mask == target_mask` returns early, a source with any magnetic type raises, and a non-native-spin source raises -- so the only case that reaches `self.atomic_model.reset_spin_parameters()` is *source declares no magnetic types, target activates some*.

In that case, for any checkpoint this PR can write, the routes being reset are already zero. This same commit changed [`SpinEmbedding.__init__`](https://github.com/deepmodeling/deepmd-kit/blob/79f3d462e54896110ff2b35d7edc0c2f5a79aa01/deepmd/dpmodel/descriptor/dpa4_nn/embedding.py#L1044-L1052) so `mag_layer2.w`, `adam_spin_vec_weight` and `adam_spin_nbr_weight` are `np.zeros(...)` instead of `rng.normal(0, init_std, ...)`, and [`EnvironmentInitialEmbedding.__init__`](https://github.com/deepmodeling/deepmd-kit/blob/79f3d462e54896110ff2b35d7edc0c2f5a79aa01/deepmd/dpmodel/descriptor/dpa4_nn/embedding.py#L571-L588) so `spin_scale` is `np.zeros` instead of `np.ones`. A dormant native-spin source constructs zeros, never trains them, serializes zeros, and loads zeros. `reset_spin_parameters()` then sets zero to zero.

So the capability earns its keep only against **pre-1.2** checkpoints, whose dormant routes hold `1.0` and random values. That is a version-migration concern, and `_migrate_variables()` already runs on precisely those checkpoints and already rewrites `spin_scale`. All five spin arrays are in `_variables()` (`spin_scale`, `mag_layer1.matrix`, `mag_layer2.matrix`, `adam_spin_vec_weight`, `adam_spin_nbr_weight`), so the migration can reach them, and the source descriptor's own `spin_flags` tells it whether the routes were dormant -- the discrimination a blanket zeroing would need, since a legacy model with trained routes must keep them.

I understand the docstring's argument for asking at the model level rather than the descriptor level, given bridging compositions. That is a good reason for the *shape* of the call if it is needed. It is not a reason for needing it on the 1.2 path.

Concretely, what I would like: either move the legacy sanitization into `_migrate_variables()` and delete `prepare_finetune` / `reset_spin_parameters` along with the four no-op defaults it needs on `BaseModel`, `BaseAtomicModel`, `make_base_descriptor` and the pt mirrors -- or, if there is a 1.2-era case that genuinely needs a model-level reset, describe it and add a test with a source whose spin routes are non-zero, because no current test has one.

deepmd/pt_expt/model/native_spin_model.py
```diff
@@ -42,6 +42,41 @@ class NativeSpinEnergyModel(make_native_spin_model(EnergyModel)):
...
+        """
+        if not isinstance(pretrained_model, NativeSpinEnergyModel):
+            raise TypeError(
+                "pt_expt native-spin fine-tuning requires a native-spin pretraining"
+            )
+        source_mask = pretrained_model.spin.use_spin.tolist()
+        target_mask = self.spin.use_spin.tolist()
+        if source_mask == target_mask:
+            return
+        if any(source_mask):
```
wanghan-iapcm: Blocking on its own terms: this rejects a supported fine-tuning workflow.

The predicate compares positional boolean lists. Pretrain with `type_map = ["Fe","C"]`, `use_spin = [True, False]`, then fine-tune on a corpus that adds an element: `type_map = ["Fe","C","O"]`, `use_spin = [True, False, False]`. The magnetic species set is unchanged -- still `{Fe}` -- and the state dict transfers, but `[True,False] != [True,False,False]`, `any(source_mask)` is true, and the run dies with `RuntimeError: Changing the active magnetic-type set during fine-tuning is not supported`.

Reordering breaks the same way and is worse, because it is silent about the reason: `["Fe","C"]` -> `["C","Fe"]` leaves `pretrained_model.spin.use_spin` un-permuted, so the comparison runs against stale indices. There is no `change_type_map` override on the native-spin model in either backend -- `def change_type_map` appears on descriptors, fittings and the virtual-atom spin models, never on `native_spin_model.py` -- so nothing realigns `use_spin` when the surrounding machinery remaps everything else.

Type-map extension is first-class here, not exotic: the same finetune loop calls `change_type_map(finetune_rule.get_finetune_tmap(), ...)` on the pretrained branch immediately before this hook runs, and `get_has_new_type()` / `get_finetune_tmap()` in `deepmd/utils/finetune.py` exist to support it.

The comparison wants to be over the *set of magnetic element symbols* resolved through each side's type map, not over positional booleans. Every added test uses identical type maps on both sides, so nothing catches this.

If the design question above lands the other way and this predicate survives, it needs a case with a changed type map.

deepmd/pt_expt/model/native_spin_model.py
```diff
@@ -42,6 +42,41 @@ class NativeSpinEnergyModel(make_native_spin_model(EnergyModel)):
...
+        The reset is requested from the atomic model, not from a descriptor:
+        this class is descriptor-agnostic, and a bridging composition owns
+        several children rather than one descriptor.
+
+        Parameters
+        ----------
+        pretrained_model
+            Source model reconstructed from the fine-tuning checkpoint.
+        """
+        if not isinstance(pretrained_model, NativeSpinEnergyModel):
```
wanghan-iapcm: This rejects the workflow the PR is named for.

`prepare_finetune` is called unconditionally on every non-resuming branch, and a genuinely spin-free pretraining is a *different class*: with no `spin` section in the config, `get_model()` builds a plain `EnergyModel`, not a `NativeSpinEnergyModel`. So pretrain a general DPA4 potential, then fine-tune a native-spin model on a magnetic corpus, and the run aborts here with `TypeError` before it starts.

That is the ordinary way anyone would reach native spin -- a large non-magnetic pretraining followed by a small magnetic fine-tune -- and it is roughly what the PR body describes. It also used to work: pt_expt's selective copy fills unmatched keys from the target state (`else: new_state[key] = target_state[key]`), so a plain-`EnergyModel` source loaded fine and left the spin routes at their fresh init. With this PR's zero-init that fresh init is now exactly the zero function, which is precisely the behaviour the PR wants -- so the case that needs the least work is the one that raises.

The new test does not cover it: `TestDPA4NativeSpinFinetuneFromSpinFreePretrain` builds its "spin-free" source with `pretrain_config["model"]["spin"]["use_spin"] = [False, False]`, i.e. still a `NativeSpinEnergyModel`. A real spin-free checkpoint has no `spin` block at all.

A non-native-spin source is the case where resetting is trivially safe, since there is nothing to inherit. It should take the reset branch (or simply return), not raise.

deepmd/pt/model/atomic_model/linear_atomic_model.py
```diff
@@ -140,6 +140,11 @@ def need_sorted_nlist_for_lower(self) -> bool:
         """Returns whether the atomic model needs sorted nlist when using `forward_lower`."""
         return True
 
+    def reset_spin_parameters(self) -> None:
+        """Reset native-spin parameters in every child model."""
+        for model in self.models:
+            model.reset_spin_parameters()
```
wanghan-iapcm: This loop is unguarded, and pt has no default to fall back on.

`DPZBLLinearEnergyAtomicModel` sets `models = [dp_model, zbl_model]` with `zbl_model` a `PairTabAtomicModel`, and pt's `PairTabAtomicModel` has no `reset_spin_parameters`. The no-op defaults this PR adds live on the *dpmodel* bases -- `dpmodel/atomic_model/base_atomic_model.py` and `make_base_descriptor.py` -- but pt's `BaseAtomicModel` comes from `make_base_atomic_model(torch.Tensor)`, a different base the PR does not touch. So this line raises `AttributeError` the moment it runs against a ZBL composition.

It is unreachable today: the only caller is `SeZMNativeSpinModel.prepare_finetune`, and `self.atomic_model` there is always a `SeZMAtomicModel(DPAtomicModel)`, never a linear model. So this is latent rather than live, and I am not blocking on it by itself.

Worth noting the precedent, though: the twin capability `supports_native_spin` (#5884) was deliberately defined only on the dpmodel side, for exactly the reason that pt never composes native spin with a linear atomic model. Adding the pt forwarding loop without the pt base default takes the opposite half of that decision. Either put the no-op on `make_base_atomic_model` so the loop is safe, or drop the pt override and match #5884.

If the design question above resolves by removing the capability, this disappears with it.

source/tests/pt_expt/descriptor/test_dpa4.py
```diff
@@ -103,6 +107,42 @@ def test_consistency(self, use_env_seed, use_mapping) -> None:
...
+            with mock.patch.dict(
+                os.environ, {"DP_AMP_INFER": "1" if amp_infer else "0"}, clear=False
+            ):
+                for use_amp in (False, True):
+                    dd = make_descriptor(
+                        self.nt, self.sel_mix, self.rcut, use_amp=use_amp
+                    ).to(self.device)
+                    for training in (False, True):
+                        dd.train(training)
+                        expected = (use_amp if training else amp_infer) and (
```
wanghan-iapcm: This test cannot fail on unfixed code in CI.

`expected` is `and`-ed with `coord_ext.device.type == "cuda"`, and the gate it checks is [`_forward_blocks`](https://github.com/deepmodeling/deepmd-kit/blob/79f3d462e54896110ff2b35d7edc0c2f5a79aa01/deepmd/pt_expt/descriptor/dpa4.py#L364-L368):

```python
enabled = self.use_amp if self.training else self.use_amp_infer
if enabled and x.device.type == "cuda":
```

On CPU that `if` is false regardless of `enabled`, so `autocast` is never called; and on CPU `expected` is false for all eight combinations. The assertion therefore holds identically before and after the fix. `Test Python on CUDA` reports `skipping` on this PR, so the configuration where the test has teeth never ran.

The pt mirror does not have this problem, and the contrast is instructive -- `test_train_and_eval_amp_switches_are_independent` in `source/tests/pt/model/test_descriptor_sezm.py` passes `torch.device("cuda")` explicitly into `_compute_mode_ctx` with `torch.autocast` mocked, so its `expected = use_amp if training else amp_infer` carries no device clause and the gate is observable on CPU. The backend this PR is named for ended up with the weaker of the two tests.

Driving the pt_expt gate the same way -- assert on the decision rather than on a real forward -- would make it fail pre-fix on the shards that actually run.

deepmd/pt/train/training.py
```diff
@@ -924,6 +924,15 @@ def collect_single_finetune_params(
...
                 state_dict["_extra_state"] = self.wrapper.state_dict()["_extra_state"]
                 self.wrapper.load_state_dict(state_dict)
+                if finetune_model is not None:
+                    for model_key in self.model_keys:
+                        finetune_rule = self.finetune_links[model_key]
+                        if self.multi_task and finetune_rule.get_resuming():
+                            continue
+                        source_key = finetune_rule.get_model_branch()
+                        self.wrapper.model[model_key].prepare_finetune(
+                            pretrained_model_wrapper.model[source_key]
```
wanghan-iapcm: The open CodeQL alert lands here: `py/uninitialized-local-variable`, "Local variable 'pretrained_model_wrapper' may be used before it is initialized."

I checked the block structure and it is a false positive: the assignment at line 811 and this use both sit under identical `if finetune_model is not None:` guards at the same indentation inside the same `if "model" in state_dict:` block, so the name is always bound when this runs. (The other alert on this PR, `py/unnecessary-delete`, already reads `fixed` at this head.)

Still worth clearing, since it leaves a required check red and CodeQL will keep reporting it. Moving this hook into the existing `if finetune_model is not None:` block that already builds `pretrained_model_wrapper` would resolve the alert and remove a duplicated guard at the same time.

While you are there: pt guards with `if self.multi_task and finetune_rule.get_resuming()` while pt_expt uses a bare `if finetune_rule.get_resuming()`. They agree in practice -- `resuming` is only ever set inside the multi-task branch of `FinetuneRuleItem.build()`, so the `multi_task` conjunct is redundant -- but the two hooks arrived in the same change and reading one after the other invites the question.

wanghan-iapcm (changes requested):
A lot of this is solid and I want to name it before the objection. I checked the 1.1 -> 1.2 squaring migration algebraically rather than taking the docstring's word: `env_agg` is linear in `r_tilde`, and `D = sum_c A[c,i] A[c,j]` contracts over the coordinate axis, so it splits into `D_geometry + D_spin` with no cross terms; the old code's `a`-scaled spin channels contributed exactly `a**2 * D_spin`, so `a -> a**2` after the move is exact, and sign loss does not matter because the old contribution was already even in `a`. The AMP/TF32 train-vs-infer split is the right fix and it closes the concern njzjz raised on #5503. Extending the pt/pt_expt grad-parity test to the SO(3) and `so2_norm` branches is the right way to catch the `FrameExpand`/`FrameContract` promotion gap, and it genuinely fails without the fix, because those branches are off by default and no previous parametrization ever constructed those modules.

My objection is to the new `reset_spin_parameters` / `prepare_finetune` capability, and it is a design question rather than a list of defects -- which is why I am leading with it. Details inline, but the short version:

**The reset is a no-op for every checkpoint this PR can produce.** The same commit zero-initializes all four spin routes at construction -- `spin_scale` from `np.ones` to `np.zeros`, `mag_layer2.w` to zeros, and both `adam_spin_*_weight` from `rng.normal(0, init_std, ...)` to zeros. A dormant source built by this code holds zeros, saves zeros and loads zeros, so zeroing it again achieves nothing. The reset only does real work on a **pre-1.2** checkpoint, where dormant routes carried `spin_scale = 1.0` and random `adam_spin_*`. That is a legacy-checkpoint hazard, not a fine-tuning one.

**And the legacy path already has a hook.** `_migrate_variables()` runs on exactly those checkpoints and already rewrites `spin_scale`. I checked `_variables()` in `dpa4_nn/embedding.py`: `spin_scale`, `mag_layer1.matrix`, `mag_layer2.matrix`, `adam_spin_vec_weight` and `adam_spin_nbr_weight` are all in the serialized dict, so a migration can reach every route it would need to zero, and it can discriminate the dormant case because the source descriptor knows its own `spin_flags` -- which it must, since a legacy checkpoint with genuinely trained routes has to keep them.

I take the docstring's point that a bridging composition owns several children rather than one descriptor, so the model level is the natural place to *ask*. But that argues the plumbing is needed somewhere, not that it is needed here: the discriminating information (legacy version plus dormant flags) is available at migration time.

Why I think this is worth answering before patching anything: three of the four defects I found are *in this plumbing*, not in the fix underneath it. If the legacy sanitization moves into the migration that already exists, the new capability and all three go away together, rather than being fixed one predicate at a time. If there is a case I am missing that genuinely needs a model-level reset on a 1.2 checkpoint, say so and I will drop the objection -- but then it should have a test, because none of the current ones exercise a source whose spin routes are non-zero.

Two things independent of that question: a test that cannot fail on unfixed code, and the open CodeQL alert. Both inline.

Green otherwise -- 53 checks pass.

deepmd/pt_expt/model/native_spin_model.py
OutisLi: Agreed. I moved the legacy sanitization into the existing v1.1 -> v1.2 descriptor migration and removed the entire `prepare_finetune` / `reset_spin_parameters` interface hierarchy. For an all-false native-spin source, migration now zeros only `mag_layer2`, both `adam_spin_*` routes, and `env_seed_embedding.spin_scale`; `mag_layer1` is preserved. Active-spin sources preserve their learned routes and still square `spin_scale` to retain the represented function. This is covered in dpmodel, PT, pt_expt, and the full spin-free-native -> active fine-tuning workflow in d32b1b661.

deepmd/pt_expt/model/native_spin_model.py
OutisLi: Agreed. The compatibility check now lives at the shared `FinetuneRuleBuilder` boundary. Each side's `spin.use_spin` is normalized against its own `type_map`, then the sets of magnetic element symbols are compared. The tests cover type-map reordering, adding an inactive element, all-false native-spin source -> active target, and rejection when an already-active magnetic element set changes. DPA4 itself still does not claim end-to-end `change_type_map` support; this change fixes the compatibility predicate without overstating that capability.

deepmd/pt_expt/model/native_spin_model.py
OutisLi: I am keeping this case explicitly unsupported in this PR. Here, "spin-free pretraining" means a native-spin model whose `use_spin` mask is all false, so the source has the complete spin parameter tree and legacy migration can sanitize it. A plain `EnergyModel` has no spin parameter tree; pt_expt's unmatched-key fallback alone is not a cross-backend conversion contract, and the traditional PT transfer path would still fail on missing source keys. The shared finetune builder now rejects that ordinary-source conversion early with an actionable error. Supporting it should be a separate change that defines consistent missing-key initialization and transfer semantics for both backends.

deepmd/pt/model/atomic_model/linear_atomic_model.py
OutisLi: Agreed. This latent unsafe forwarding loop has been removed together with the `reset_spin_parameters` capability. The legacy all-false cleanup now happens during descriptor migration, so neither the PT linear atomic model nor `PairTabAtomicModel` participates in reset plumbing.

source/tests/pt_expt/descriptor/test_dpa4.py
OutisLi: Agreed. The test now drives `_forward_blocks` directly with a CUDA-device test double while mocking only the parent block execution and `torch.autocast`. Its expected value is `use_amp` in training and `DP_AMP_INFER` in evaluation, with no CPU device clause, so all policy combinations are observable on CPU and the test fails if the train/eval selection regresses. The revised test passes locally.

deepmd/pt/train/training.py
OutisLi: Resolved structurally. The model-level fine-tuning preparation hook was removed as part of moving legacy cleanup into descriptor migration, so the duplicated guarded use of `pretrained_model_wrapper` no longer exists. This also removes the pt/pt_expt resuming-guard discrepancy at that hook rather than rearranging the false-positive path.

deepmd/pt_expt/model/native_spin_model.py
```diff
@@ -42,6 +42,41 @@ class NativeSpinEnergyModel(make_native_spin_model(EnergyModel)):
...
+            raise TypeError(
+                "pt_expt native-spin fine-tuning requires a native-spin pretraining"
```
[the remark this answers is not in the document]
OutisLi: This conversion remains intentionally out of scope. The supported spin-free source is a native-spin model with an all-false mask and therefore a complete spin parameter tree. A plain `EnergyModel` lacks those parameters; pt_expt's selective fallback does not define equivalent behavior for the traditional PT path. The shared finetune builder now rejects this unsupported cross-class conversion before model construction. A separate change can add it once both backends have an explicit, tested missing-key transfer contract.

deepmd/pt_expt/model/native_spin_model.py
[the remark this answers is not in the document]
OutisLi: Fixed in d32b1b661. Native-spin compatibility is now checked in `FinetuneRuleBuilder` by resolving each mask against its own `type_map` and comparing magnetic element-name sets. The shared tests cover both reordering and extension by an inactive type, in addition to all-false activation and rejection of a changed active magnetic set.

wanghan-iapcm (approved):
All six findings from my earlier review are addressed at 3be03563, each with a test that fails on the pre-fix code:

- The `reset_spin_parameters` / `prepare_finetune` hierarchy is gone; legacy cleanup now happens in the 1.1 -> 1.2 descriptor migration, so the unsafe forwarding loop in the PT linear atomic model went with it.
- The native-spin compatibility check moved to `FinetuneRuleBuilder` and compares magnetic element sets rather than positional boolean lists, so type-map reordering and added inactive elements no longer abort.
- Conversion from a spin-free source is still unsupported, but now fails early with an actionable message instead of a `TypeError`.
- The AMP test drives `_forward_blocks` with a CUDA device double, so the train/eval policy is observable on CPU; `use_amp=False, eval, DP_AMP_INFER=1` fails on the old gate.
- The CodeQL uninitialized-local alert is resolved structurally.

I also checked the replacement mechanism, since it is new since my review. The migration is exact: the quadratic form sums over the coordinate axis only, so D splits into D_geometry + D_spin with no cross terms, an amplitude a contributed a**2 * D_spin, and the new linear gate reproduces it with s = a**2. Both dpmodel and pt carry the migration and the version bump.

Two things to keep in mind rather than block on: the migration test asserts the stored gate value, not a numeric round-trip of the forward output; and the CUDA jobs report skipping, so the autocast region itself was never executed in CI - only the selection logic was.

Issue #5824, opened 2026-07-16
feat(pt_expt): decouple DPA1 graph normalization and statistics from dense sel


OutisLi:
## Background

The pt_expt DPA1/`se_atten_v2` carry-all graph lower no longer uses `sel` to cap or pad the runtime neighbor graph. However, `sel` is still part of the descriptor's mathematical and statistical state:

- `self.nnei = sum(sel)` is used to normalize graph aggregation;
- descriptor mean/std buffers are shaped as `(ntypes, nnei, 4)`;
- input statistics build a dense `sel`-sized neighbor list and include padded zeros;
- omitting `sel` therefore still changes model construction and statistics.

Graph-native DPA1 is uncapped at runtime but is not truly `sel`-independent.

## Proposed design

Separate the three concepts currently represented by `sel`:

1. `neighbor_capacity`: required only by dense neighbor-list execution;
2. `aggregation_normalization`: serialized model semantics used by DPA1 aggregation;
3. `input_stat_policy`: defines how descriptor statistics are accumulated.

Use a versioned compatibility mapping:

- legacy configs/checkpoints map `sum(sel)` to a fixed aggregation normalizer and retain legacy padded-stat semantics;
- graph-native configs may use `neighbor_capacity=None`;
- a graph-native fixed or data-derived normalizer is computed once and serialized;
- graph-native statistics store the canonical per-type vector, such as `(ntypes, 1, 4)`, rather than duplicating it over `nnei`.

Implement statistics as streaming reductions over graph edges. Legacy padded statistics should be reproducible from sums, squared sums, real-edge counts, and the legacy slot count without materializing `(nframes, nloc, sel, 4)` tensors.

Dynamic raw-degree normalization should not become the default because it changes DPA1 semantics and may break smoothness at the cutoff. Any smooth-degree policy should be explicit and versioned separately.

## Compatibility requirements

- Existing configs and checkpoints containing `sel` must preserve their current normalization and statistics.
- Serialization/deserialization must distinguish legacy and graph-native semantics.
- PT dense execution must retain its current capacity-based behavior.
- Continuing a legacy checkpoint in pt_expt graph mode must not silently reinterpret its normalization.

## Performance requirements

- Graph-native statistics must not allocate memory proportional to `nloc * sel`.
- Runtime graph execution must remain carry-all and dynamic in the number of edges.
- The normalizer must be compile-friendly and must not introduce per-atom Python control flow.
[...]

PR #5975, opened 2026-08-17
fix(pt-expt): preserve lower semantics in backend conversion

### LINKED ISSUE #5973: [BUG][pt_expt][DPA1] Automatic graph-lower export changes predictions for dense-trained se_atten_v2 models with nonzero davg
### Bug summary

I trained a DPA1-L0 model with the standard PyTorch backend (`dp --pt train`) and converted the
  frozen `.pth` model to `.pt2` using `dp convert-backend`. The descriptor used the default
  `set_davg_zero: false` and has nonzero `davg`.

  The automatically selected graph-lower model loads and runs, but it is not numerically
  equivalent to the source model. On the same six-frame system, the force-component difference
  is:

  - MAE: `1.1381067 eV/Angstrom`
  - RMSE: `1.7619205 eV/Angstrom`
  - Maximum absolute difference: `6.7789032 eV/Angstrom`

  In contrast, a manually exported nlist/dense-lower `.pt2` model reproduces the source model
  with a force-component MAE of `5.13e-6 eV/Angstrom` and a maximum difference of `1.91e-5 eV/
  Angstrom`.

  The discrepancy is reproduced by an uncompiled CPU eager comparison between
  `forward_common_lower` and `forward_common_lower_graph`, and is already present in the
  descriptor output. As a diagnostic, setting the descriptor mean (`davg`) to zero in memory
  reduces the descriptor discrepancy to numerical noise.

  I also independently retrained four models with `set_davg_zero: true`. After automatic graph-
  lower conversion, their force-component MAE differences are `1.76e-5`–`1.84e-5 eV/Angstrom`,
  with a maximum observed difference of `1.33e-4 eV/Angstrom`.

  These results indicate that automatic conversion accepts a dense-trained `se_atten_v2` model
  with nonzero `davg`, even though its graph-lower representation is not functionally equivalent.

### DeePMD-kit Version

0.1.dev1+gbc902daf7.d20260809

### Backend and its version

PyTorch 2.11.0+cu128

Built from source

A minimal reproducer archive is attached containing:

  ```text
  reproducer/
  ├── frozen_model.pth
  ├── input.json
  ├── compare_predictions.py
  └── test_system/
      ├── type.raw
      ├── type_map.raw
      ├── nopbc
      └── set.000/
          ├── box.npy
          ├── coord.npy
          ├── energy.npy
          └── force.npy

  The test system is nonperiodic and contains six frames with seven atoms per frame.

  Relevant model settings:

  {
    "type_map": ["C", "Cl", "H", "O"],
    "descriptor": {
      "type": "se_atten_v2",
      "sel": 250,
      "rcut": 6.0,
      "attn": 128,
      "attn_layer": 0,
      "tebd_input_mode": "strip",
      "precision": "float32"
    }
  }

  set_davg_zero was not specified during training, so its effective value is false. The trained
  descriptor has nonzero davg; its largest absolute value is approximately 0.01418.

  Environment:

  DeePMD-kit: 0.1.dev1+gbc902daf7.d20260809
  Git commit: bc902daf747d93c5d7314338d287fd2dd0a7e32c
  PyTorch: 2.11.0+cu128
  Python: 3.12.12
  OS: Linux x86_64

  The build is based on the official master branch. The only local source change preserves an
  explicitly requested `CMAKE_CUDA_ARCHITECTURES=120`; no Python, model, descriptor,
  serialization, or inference code was modified.

  There is no exception or error traceback. Conversion and inference both exit successfully. This
  is a silent numerical correctness failure.

  Observed source .pth versus automatic graph-lower .pt2 difference:

  Energy MAE:                  1.3291797 eV/frame
  Energy maximum difference:  2.0507687 eV/frame
  Force-component MAE:        1.1381067 eV/Angstrom
  Force-component RMSE:       1.7619205 eV/Angstrom
  Force maximum difference:   6.7789032 eV/Angstrom

  For comparison, manually exporting the same model through nlist/dense lower gives a force-
  component MAE of 5.13e-6 eV/Angstrom and a maximum difference of 1.91e-5 eV/Angstrom.

### Steps to Reproduce

[deepmd-dpa1-graph-lower-reproducer.zip](https://github.com/user-attachments/files/31117915/deepmd-dpa1-graph-lower-reproducer.zip)

  1. Extract the attached minimal reproducer and enter its directory:

     ```bash
     unzip deepmd-dpa1-graph-lower-reproducer.zip
     cd deepmd-dpa1-graph-lower-reproducer

  2. Convert the supplied source model. dp convert-backend automatically selects graph lower:

     dp convert-backend frozen_model.pth model_graph.pt2

  3. Evaluate the source and converted models on the same six-frame system:

    dp --pt test -m frozen_model.pth -s test_system -n 32 -d pth
    dp --pt-expt test -m model_graph.pt2 -s test_system -n 32 -d graph

  4. Compare their predictions:

     python compare_predictions.py pth graph

     Expected reproduction output:

     Energy MAE difference:              1.3291797 eV/frame
     Energy maximum absolute difference: 2.0507687 eV/frame
     Force-component MAE:                1.1381067 eV/Angstrom
     Force-component RMSE:               1.7619205 eV/Angstrom
     Force-component maximum difference: 6.7789032 eV/Angstrom

  No retraining, LAMMPS, or Kokkos is required. The same underlying dense-versus-graph
  discrepancy was also reproduced independently in uncompiled CPU eager execution.

_No response_

OutisLi:
## Summary

- preserve the source artifact's concrete lower-input semantics during backend conversion instead of resolving `lower_kind="auto"`
- expose `lower_input_kind` from `.pte` and `.pt2` metadata so exported artifacts retain their lower across subsequent conversions
- reject graph-to-dense-only conversions rather than silently changing the model function
- document that graph semantics require a graph-native training and freeze workflow

## Root cause

`dp convert-backend` always passed `lower_kind="auto"` to the pt_expt serializer. A dense-trained DPA1 model was therefore reinterpreted as graph-native whenever the reconstructed target model advertised graph support. Dense padding contributes `-davg/dstd` when `davg` is nonzero, while the graph lower contains no padding edges, so the generated artifact represented a different function.

## Verification

- targeted conversion and metadata tests: 9 passed
- PTE serialization round-trip: 1 passed
- real nonzero-`davg` `.pth` to `.pt2` conversion selected `lower_input_kind=nlist`
- source versus converted artifact: energy delta 0, force max delta 8.882e-16, virial max delta 6.661e-16
- Ruff, diff checks, and all pre-commit hooks passed

Closes #5973

Related to #5862 and #5824.

## Summary by CodeRabbit

* **New Features**
  * Backend conversions preserve supported lower-input representations, including dense, graph, neighbor-list, and canonical formats.
  * Model exports across supported frameworks now include lower-input representation metadata.
  * Legacy files without metadata use compatible automatic defaults.
  * Unsupported target representations are rejected to prevent silent semantic changes.
  * Schema-neutral model containers retain lower-input metadata for later conversion.

* **Documentation**
  * Added guidance on conversion behavior, preservation rules, and supported representations.

* **Tests**
  * Expanded coverage for metadata preservation, legacy artifacts, serialization formats, and conversion validation.


deepmd/entrypoints/convert_backend.py
```diff
@@ -40,8 +50,24 @@ def convert_backend(
 
     sig = inspect.signature(out_hook)
     hook_kwargs: dict[str, Any] = {}
+    # Existing backend serializers describe dense models and predate explicit
+    # lower metadata. Graph-capable artifact serializers override this default
+    # from their archive metadata.
+    lower_input_kind = data.get("lower_input_kind", "nlist")
```
wanghan-iapcm: Blocking. Defaulting to `"nlist"` for every source that does not carry `lower_input_kind` regresses two cases.

**(a) The automatic graph selection is lost for all non-pt_expt sources.** Only pt_expt `.pt2`/`.pte` populate this key, so `.pth`, `.pb`, `.dp`, jax and paddle sources are now hard-pinned to `"nlist"`. The `"auto"` path did not consult artifact metadata at all -- `_resolve_lower_kind` deserializes `data["model"]` and asks `model_uses_graph_lower(model) and _supports_graph_export(model)`, i.e. a property of the model that is available regardless of source format. That is why `.pth`/`.dp` -> `.pt2` of a graph-eligible model produced a graph or `dpa1_canonical` artifact, which is what 2aee81cec (#5758) added it for. The docstring's premise that "Formats without explicit lower metadata are dense neighbor-list models" does not hold for `.pth`/`.dp`.

The genuinely broken case in #5973 is narrower: a *dense-trained* model whose reconstructed dpmodel happens to advertise graph support. For a graph-native descriptor (DPA4C has no dense lower at all -- `disable_graph_lower()` raises) the old resolution was the only correct answer, and it is now unreachable.

**(b) Native-spin models can no longer be converted to `.pt2` at all.** `deepmd/dpmodel/model/native_spin_model.py:261` stamps `type="native_spin"`, and `deepmd/pt_expt/utils/serialization.py:1344` raises unconditionally when `data["model"]["type"] == "native_spin"` and `lower_kind != "graph"`. `_resolve_lower_kind` short-circuits only on `"auto"`, so the new concrete `"nlist"` passes straight through into that guard. `dp convert-backend model.dp model.pt2` for a model trained with `spin.scheme == "native"` now aborts with `ValueError: native-spin models implement only the NeighborGraph lower`, where it previously succeeded. Native spin is graph-lower-only by construction, so `"nlist"` is never a valid default for it.

Suggestion: keep the model-derived resolution as the fallback when the source dict carries no `lower_input_kind` (i.e. pass `"auto"` through in that case) and use the metadata value only when it is actually present. That preserves the source's semantics where the source states them, and keeps the model-property answer where it does not.

source/tests/test_convert_backend.py
```diff
@@ -46,5 +46,65 @@ def detect_backend(path: str) -> type[InputBackend] | type[OutputBackend]:
 
     convert_backend(INPUT="model.input", OUTPUT="model.output")
 
-    assert captured["lower_kind"] == "auto"
+    assert captured["lower_kind"] == "nlist"
```
wanghan-iapcm: Blocking. The test that would prevent this bug from coming back has not been committed.

The bug's essence is "a dense-trained model came out of the conversion as graph-native". Reproducing it requires the intersection of three conditions: going through `convert_backend`, a model with nonzero `davg` whose reconstructed dpmodel advertises graph support, and a `.pt2` target. Only then does `_resolve_lower_kind` run and pick `"graph"`.

The old test crossed the first condition only, and because it used stub backends `_resolve_lower_kind` never executed -- which is also why it ended up asserting `lower_kind == "auto"`, encoding the bug as the expected behaviour. All three tests here keep that same shape: hand-written `InputBackend`/`OutputBackend` classes plus `monkeypatch.setattr(Backend, "detect_backend_by_model", ...)`, so the real `_resolve_lower_kind`, `deserialize_to_file` and export path still never run. The two metadata tests monkeypatch `torch.export.load` for the same reason. They do fail on pre-fix code, so they are not vacuous -- but what they pin down is the kwarg plumbing, not the semantics.

The PR description reports exactly the right check ("real nonzero-`davg` `.pth` to `.pt2` conversion selected `lower_input_kind=nlist`", energy delta 0). Please turn that into a committed test: build a model with nonzero `davg`, run the real `convert_backend` to `.pt2`, and assert the resulting `metadata.json` records `nlist` (asserting the energy/force agreement as well would be stronger). Pre-fix that assertion sees `graph` and fails, which is what makes it a regression test rather than a restatement of the new code.

Two smaller coverage notes while you are in here, not blocking on their own: the no-op path of the new `elif` (a `"nlist"` source going to an output backend whose `deserialize_hook` genuinely has no `lower_kind` -- the ordinary tf/pt/dpmodel case) is never exercised, and neither is the `_LOWER_INPUT_KINDS` rejection branch.

wanghan-iapcm (changes requested):
Thanks for tracking this down -- the diagnosis in #5973 is right, and reading the source artifact's own `lower_input_kind` instead of re-deriving it at the target is the correct direction. Two things need work before this can go in; both are inline.

The short version: the new default is applied to sources that never carried the field, and that is a different decision from the one the bug required. `_resolve_lower_kind` answered a question about the *model* (`model_uses_graph_lower` + `_supports_graph_export`), which is available from any source format, so pinning every non-pt_expt source to `"nlist"` removes a correct answer along with the incorrect one.

I also checked the rejection branch for graph -> `.dp`/`.pth`/`.pb` and concluded it is right as written: those backends only implement the padded dense lower, so allowing that conversion would be the same silent change of function this PR is fixing. No change requested there.

deepmd/entrypoints/convert_backend.py
OutisLi: Fixed in d3920903. `convert_backend` now distinguishes absent metadata from an explicit source ABI: an absent `lower_input_kind` is passed as `auto` only to targets that support `lower_kind`, while an explicit value is forwarded unchanged. Dense-only targets accept absent metadata or `nlist` and reject an explicit non-dense ABI. Source serializers now report their actual contract across standard PyTorch, TensorFlow, TensorFlow 2, and JAX; PT2/PTE continue to preserve archive metadata. Standard PyTorch delegates this to `model.export_lower_input_kind()`, whose default is `nlist` while model-specific graph-native overrides remain authoritative. Tests cover missing-to-auto, every explicit kind, dense-only no-op/rejection, JAX/TF2 metadata, virtual-spin ABI, and cross-backend round trips.

source/tests/test_convert_backend.py
OutisLi: Fixed in d3920903 with a real end-to-end regression. `test_convert_regular_pt_dpa1_preserves_dense_semantics` builds DPA1 data with nonzero `davg`, writes a real `.pth`, invokes the real `convert_backend` path to `.pt2`, asserts `metadata.json` records `lower_input_kind=nlist`, and compares actual energy and force at `rtol=atol=1e-10`. I also added the requested dense-only no-op coverage and target-owned invalid-kind rejection, plus PT2/PTE legacy metadata fallback cases. The focused conversion/PT2 suite passes: 47 passed.

deepmd/pt/utils/serialization.py
```diff
@@ -53,6 +53,7 @@ def serialize_from_file(model_file: str) -> dict:
         "pt_version": str(torch.__version__),
         "model": model_dict,
         "model_def_script": model_def_script,
+        "lower_input_kind": model.export_lower_input_kind(),
```
wanghan-iapcm, line 56: Blocking. This makes `dp convert-backend` fail outright for every SeZM/DPA4 PyTorch model.

`SeZMModel.export_lower_input_kind()` returns `"edge_vec"` (`deepmd/pt/model/model/sezm_model.py:3171-3183`), and the class is registered as `"SeZM"`/`"sezm"`/`"DPA4"`/`"dpa4"` at lines 677-680, with `SeZMNativeSpinModel` inheriting it. So this line now stamps `lower_input_kind="edge_vec"` on the interchange dict, `convert_backend` forwards an explicit value verbatim, and `deepmd/pt_expt/utils/serialization.py:1429` rejects it -- `_SUPPORTED_LOWER_INPUT_KINDS` at lines 55-57 is `{"nlist", "graph", "dpa1_canonical", "dpa4c_canonical"}` and does not contain `"edge_vec"`. Targeting a backend that does not take `lower_kind` fails one step earlier, in `convert_backend` itself.

Before this PR the same conversion worked: `"auto"` went into `_resolve_lower_kind`, which deserializes the model (`deepmd/pt_expt/model/dpa4_model.py:35-41` registers `"sezm"`/`"SeZM"`/`"dpa4"`/`"DPA4"`, so the pt dict is deserializable there) and returns `"graph"` or `"dpa4c_canonical"` -- both supported. This is deterministic, not an edge case.

It also contradicts this PR's own documentation. `doc/backend.md` gains the sentence "model families with a graph-native deployment ABI report their corresponding kind" -- `edge_vec` is that kind, for that family, and it is now the only family that cannot be converted at all.

The same trade-off has a milder second face worth deciding deliberately: because the base default is `"nlist"`, a plain `.pth` can no longer be promoted to `graph`/`dpa1_canonical`/`dpa4c_canonical` on conversion, and there is no `--lower-kind` on the CLI to ask for it. Pinning dense is the correct default given #5973, but the promotion path disappears entirely rather than becoming opt-in.

Two directions, either is fine: add `"edge_vec"` to the kinds pt_expt accepts and map it onto the schema it already implements (`deepmd/pt/entrypoints/freeze_pt2.py` and `deepmd/pt_expt/infer/deep_eval.py` both already branch on it, so the vocabulary gap is only in the new validator), or have the pt serializer report nothing for models whose ABI the target cannot consume, so they fall back to `auto` as before.

Worth noting how this got through: the new tests do parametrize `"edge_vec"` through `convert_backend`, but only against an inline stub `OutputBackend` that records the kwarg instead of calling the real pt_expt deserializer, so the failure is invisible in CI. That is the same stub-shaped gap as the one we just closed on the conversion side -- whichever fix you pick, it needs a case that reaches the real target.

wanghan-iapcm (changes requested):
Both of my earlier points are genuinely addressed -- thanks for the rewrite, the "source declares its contract, absent metadata falls back to `auto`" split is the right shape.

Verified against HEAD rather than taking the replies at face value:

- Absent metadata now reaches `auto`, so `.dp` sources get the model-derived resolution back, and a native-spin `.dp` -> `.pt2` no longer trips the `native_spin` guard.
- `test_convert_regular_pt_dpa1_preserves_dense_semantics` is a real regression test: it forces `davg` to a nonzero value, writes an actual `.pth`, runs the real `convert_backend`, reads the real `metadata.json` out of the produced archive, and compares energy and force at `rtol=atol=1e-10`. Nothing is monkeypatched, it is not skipped, and the main CI matrix collects it. Pre-fix it would record `graph` and fail on the metadata assertion. That is exactly what was missing.

One new problem came in with the rewrite, inline below: making the PyTorch source declare a fixed ABI breaks the one model family that already had a non-dense one.

Non-blocking, for a follow-up if you prefer: the dpmodel backend is now the only one that does not declare a contract (`deepmd/dpmodel/utils/serialization.py` is untouched, and `load_dp_model` returns whatever is stored). The consequence is that the same `.dp` behaves differently depending on when it was produced -- one written before this change has no key and resolves through `auto`, while one written afterwards by a `pth -> dp` conversion carries `"nlist"` and is pinned dense. Since tf, tf2 and jax all now state `"nlist"` explicitly, `.dp` reads as an oversight rather than a decision.

On CI: the checks are green, but `Test C++ on CUDA` and `Test Python on CUDA` both come back `skipped` at job level for this head (only the aggregate `Pass testing on CUDA` gate is green), so the CUDA-side canonical operators are not covered by anything that ran here.

deepmd/pt/utils/serialization.py
OutisLi: Fixed in 4ae8ce98. The PT serializer continues to report SeZM/DPA4’s actual source ABI (`edge_vec`); the pt_expt deserializer now owns a data-driven source-to-target mapping and materializes that ABI as its native NeighborGraph `graph` schema. The emitted artifact therefore records `lower_input_kind="graph"`, matching the schema it actually accepts, while the existing graph-eligibility checks remain in force.

I also added a regression that builds a real PT DPA4 `.pt` checkpoint, verifies that the source reports `edge_vec`, runs the public `convert_backend` path into a real `.pte` target (including `torch.export` save/load), and verifies that the target reports `graph`. The regression passes locally; the related metadata/validation tests (11) and generic conversion tests (9) also pass, as do all pre-commit hooks.

deepmd/pt_expt/utils/serialization.py
```diff
@@ -52,6 +52,18 @@
...
+# Backend conversion supplies the source artifact's lower ABI. Each accepted
+# source kind maps to the concrete schema emitted by pt_expt. PT SeZM's
+# ``edge_vec`` lower and pt_expt's ``graph`` lower carry the same directed-edge
+# model semantics, while the target materializes its native NeighborGraph ABI.
+_LOWER_INPUT_KIND_TARGETS = {
+    "nlist": "nlist",
+    "graph": "graph",
+    "dpa1_canonical": "dpa1_canonical",
+    "dpa4c_canonical": "dpa4c_canonical",
+    "edge_vec": "graph",
```
wanghan-iapcm: Blocking, though it should be a small fix. Mapping `edge_vec` to `graph` unconditionally turns a conversion that works today into a hard error for SeZM/DPA4 models that are not graph-lower eligible.

`SeZMModel.export_lower_input_kind()` returns `"edge_vec"` unconditionally — it never consults eligibility. `SeZMPropertyModel` subclasses `SeZMModel` and does not override it, so a property-fitting model reports `"edge_vec"` too. But `model_uses_graph_lower` returns `False` when `"energy" not in model.atomic_output_def().keys()`, which is exactly that case, so the innermost gate raises `ValueError: lower_kind='graph' requested but the model is not graph-lower eligible`.

Before this commit the same model converted fine: `convert_backend` passed `"auto"`, and `_resolve_lower_kind` pre-checked `model_uses_graph_lower` and fell back to `"nlist"`. Now the source's unconditional `"edge_vec"` is forwarded explicitly and skips that pre-check. So `dp convert-backend sezm_property.pt out.pt2` regresses from working to failing, and the same applies to any other SeZM/DPA4 configuration that turns graph eligibility off — compression, `set_davg_zero=False`, `use_three_body`, `disable_graph_lower()`.

The new test only exercises an energy-fitting DPA4 config, so nothing catches it.

The cleanest fix is probably to make the mapping conditional rather than a static table entry: resolve `edge_vec` to `"graph"` only when `model_uses_graph_lower(model)` holds, and fall back to `"nlist"` otherwise — i.e. put back the pre-check that `"auto"` used to perform. Alternatively, make `SeZMModel.export_lower_input_kind()` itself eligibility-aware, which would keep the table honest as a pure lookup. Either way a property-model case in `test_convert_pt_dpa4_maps_edge_vec_to_graph`'s neighbourhood would pin it.

deepmd/entrypoints/convert_backend.py
```diff
@@ -40,8 +48,18 @@ def convert_backend(
 
     sig = inspect.signature(out_hook)
     hook_kwargs: dict[str, Any] = {}
+    lower_input_kind = data.get("lower_input_kind")
```
wanghan-iapcm, line 51: The ABI declaration is only half-rolled-out, so the result now depends on which file format the weights arrived in.

pt, tf, tf2 and jax all declare a kind, but `deepmd/dpmodel/utils/serialization.py` (`save_dp_model`/`load_dp_model`) and `deepmd/pd/utils/serialization.py` do not, so `data.get(...)` is `None` for a `.dp` source and this sends `"auto"`. `_resolve_lower_kind` then picks `graph`/`dpa1_canonical`/`dpa4c_canonical` for an eligible model — where the very same weights arriving as a `.pth` would have been pinned to `"nlist"`.

So the same model converts to a different artifact depending on whether it came through `.dp` or `.pth`, which is the class of surprise this PR is otherwise removing. `doc/backend.md` frames the no-metadata path as a legacy/backward-compatibility fallback, but `.dp` is not legacy — it is a current format that will simply never carry the field unless it is added.

Adding the declaration to the dpmodel serializer would close it. Paddle is less urgent, since its `serialize_from_file` raises `NotImplementedError` today, but it is the same gap.

Related, lower stakes: the rejection just below fires for any target whose hook lacks a `lower_kind` parameter, which includes `.dp`. Unlike `.pb`/`.pth`, a `.dp` bakes no execution schema at all — it is a weights container, and whoever re-exports it later picks their own lower. Blocking `dp convert-backend sezm.pt out.dp` therefore protects nothing while losing a working path. Worth narrowing the guard to targets that actually materialise a lower.

deepmd/pt_expt/utils/serialization.py
```diff
@@ -1393,14 +1435,17 @@ def deserialize_to_file(
...
+        compatible source ABIs are mapped to the target's native schema while
+        preserving their execution semantics. A graph lower always preserves
+        the fused inference operators (``DP_CUDA_INFER >= 2``) and the
+        per-atom virial.
         The selected schema is recorded as ``lower_input_kind`` in
         ``metadata.json``.
     """
-    lower_kind = _resolve_lower_kind(model_file, data, lower_kind)
+    lower_kind = _resolve_target_lower_kind(model_file, data, lower_kind)
     if data["model"].get("type") == "native_spin" and lower_kind not in (
```
wanghan-iapcm: Minor, but this guard can now be tripped by a value the caller never supplied, and its message then misdirects them.

`_resolve_lower_kind` special-cases `type == "spin_ener"` but not `"native_spin"`, so a native-spin model arriving with `"auto"` — a `.dp` source, for instance — falls through to the generic eligibility branch and can resolve to `"nlist"` or, when `canonical_model_eligible` holds, `"dpa1_canonical"`. Neither is in this guard's allowed set, so it raises and tells the user to pass `lower_kind='graph'` when they passed `"auto"`, or nothing at all through `dp convert-backend`.

The mismatch predates this PR, but `_resolve_target_lower_kind` is new and sits directly in front of the guard, which makes it the natural place to either extend the `native_spin` special case into `_resolve_lower_kind` or reword the message to distinguish a caller-supplied kind from an internally-resolved one.

wanghan-iapcm (changes requested):
All three of my earlier points are genuinely resolved, and I checked each against the code rather than the replies.

The `auto` fallback restores the model-derived resolution for sources that declare nothing. The dense-semantics regression is real end-to-end — nonzero `davg`, a real `.pth`, the real `convert_backend`, the real `metadata.json`, plus an energy/force comparison at 1e-10. And I spent a while on `edge_vec -> graph`, since it is the only row in the table whose key and value differ: it holds, and for a better reason than the comment gives. `deserialize_to_file` re-deserializes the model from the weights, builds pt_expt's own NeighborGraph sample inputs and re-traces through `torch.export`, and `meta["lower_input_kind"]` is written after that trace, so the recorded ABI cannot drift from the compiled signature. That matters, because the two ABIs are not interchangeable at all — a six-tuple starting with float `coord` versus a ten-tuple with no `coord`, `atype` in slot 0, and four CSR arrays. A relabel would have been silently catastrophic; a re-trace is sound. I also confirmed the innermost gate re-asserts both `_supports_graph_export` and `model_uses_graph_lower`, so nothing here bypasses the eligibility checks.

What holds this up is one regression that came in with the last commit, inline below, plus two things worth settling in the same pass. None of them is in the original problem area — the fix for #5973 itself looks right to me.

For completeness, things I checked that turned out fine and that nobody needs to spend time on: a graph lower written into a `.pte` is a supported artifact, not a broken one (the `.pte` path is the Python `torch.export.load` consumer; the C++ AOTI runtime only handles `.pt2`, and `doc/backend.md` already says compiled `.pte` retains its recorded kind). The new top-level `lower_input_kind` key does not trip the serialization-version rule, since the interchange dict carries no top-level `@version` — only the nested `data["model"]` is versioned. And the base-class `export_lower_input_kind` default closes a latent `AttributeError` for non-SeZM models reaching the freeze path rather than creating a new risk.

Separately, and not something to fix here: `dp freeze` in `deepmd/pt_expt/entrypoints/main.py` force-overrides `lower_kind` to `"graph"` for any structurally graph-capable model, warning only about a "~1e-4" difference. That gate looks at structural eligibility and never at `davg`, which is precisely the variable #5973 showed can drive the two lowers apart by far more than 1e-4. So after this PR, `dp freeze` and `dp convert-backend` apply opposite policies to the same weights, and I think it is freeze that is now wrong. Worth a separate issue rather than widening this one.

deepmd/pt_expt/utils/serialization.py
OutisLi: Fixed in 3166cd95. The static `edge_vec -> graph` row is gone. `_select_graph_lower_kind` now deserializes the target model and chooses `graph` only when both `model_uses_graph_lower(model)` and `_supports_graph_export(model)` hold; otherwise `edge_vec` materializes as `nlist`. The regression uses a real pt_expt `PropertyModel` with a DPA4 descriptor, stores `edge_vec` provenance, runs the public `.dp -> .pte` conversion, and verifies that the target metadata is `nlist`; the real energy DPA4 path still verifies `graph`.

I also reproduced the specific PT `SeZMProperty` wrapper before this fix: pt_expt rejects its wire type at `BaseModel.deserialize` (`Unknown model type: SeZMProperty`) before lower selection, and the same was true before 4ae8ce98. This commit therefore does not claim to add that previously unsupported wrapper conversion; it fixes the unsafe mapping for target-supported graph-ineligible models.

deepmd/entrypoints/convert_backend.py
OutisLi: Fixed in 3166cd95. `Backend.preserves_lower_input_kind` now expresses whether a target without a `lower_kind` argument is a schema-neutral container. `DPModelBackend` declares that capability, so `convert_backend` no longer rejects a non-dense source ABI when writing `.dp`/`.yaml`/`.yml`.

The DPModel conversion serializer reports `auto` only when the file contains no concrete provenance and preserves any stored `lower_input_kind` unchanged. `auto` is deliberate: a DPModel file binds no executable input signature, so stamping every native file as `nlist` would be false for graph-only families. A real regression now converts PT DPA4 `.pt -> .dp -> .pte`, verifies that `edge_vec` survives the container, and verifies that the executable target materializes `graph`.

deepmd/pt_expt/utils/serialization.py
OutisLi: Fixed in 3166cd95. Native-spin validation now lives in `_resolve_target_lower_kind`, where both the caller-supplied value and the resolved target schema are still available. An unbound `.dp -> .pte` conversion now says that automatic native-spin selection requires a `.pt2` target because no dense lower exists; it no longer presents internally resolved `nlist` as a user choice. Automatic `.pt2` selection is separately verified to resolve to `graph`, while an explicitly requested invalid lower retains the explicit-contract error.

wanghan-iapcm (approved):
Approving. All three points from my last round are resolved in 3166cd95, checked against the code rather than the replies.

The static `edge_vec -> graph` row is gone. `_resolve_target_lower_kind` now routes `edge_vec` through `_select_graph_lower_kind`, which deserializes the target model and returns `graph` only when both `model_uses_graph_lower` and `_supports_graph_export` hold, falling back to `nlist` otherwise. That restores the pre-check `auto` used to perform, which was the regression. The new `test_edge_vec_uses_dense_lower_for_non_energy_target` crosses exactly the cell the previous test missed, an `edge_vec` source with a graph-ineligible target: on 4ae8ce98 the static table maps it to `graph` and the innermost gate raises, so the test fails there and passes now. That is a real regression test.

On the `.dp` declaration, I accept the argument. A `.dp` written by a conversion keeps the concrete kind it arrived with, so `.pth -> .dp -> .pt2` now carries `nlist` end to end, and only a natively authored `.dp` with no provenance reports `auto`, which is the legacy fallback by construction. `preserves_lower_input_kind` on `DPModelBackend` also reopens `pt -> .dp`, which the previous guard had closed for no benefit. The schema-neutral test fails on the old guard with "Cannot preserve lower_input_kind 'graph'", so it too proves the change.

The native-spin guard now lives where both the caller's value and the resolved kind are visible, and the two `auto` messages describe the actual constraint instead of presenting an internally resolved `nlist` as a user choice. The `.pte` regression pins the message.

Also noted: 11be89dc gives the argcheck fixture the `numb_steps` that dargs 0.5.1's root-level `extra_check` now requires. That is the same failure currently showing on other open PRs, so CI here is green on its own merits, and master will pick up the fix with this merge.

PR #5984, opened 2026-08-22
feat(data): support extxyz training and validation datasets

qchempku2017:
## Summary

Add first-class support for using labeled `.xyz` and `.extxyz` files directly
as DeePMD-kit training and validation datasets.

Users can now write:

```json
{
  "training": {
    "training_data": {
      "systems": [
        "data/train_part_1.extxyz",
        "data/train_part_2.xyz"
      ],
      "batch_size": "auto"
    },
    "validation_data": {
      "systems": ["data/validation.extxyz"],
      "batch_size": "auto"
    }
  }
}
```

No separate conversion command or modification of the input JSON is required.

## Implementation
- Make dpdata>=1.1.0 a core dependency.
- Detect explicitly listed .xyz and .extxyz files during shared input normalization.
- Read frames through dpdata's existing extxyz implementation.
- Materialize the data as standard DeePMD NumPy systems before existing DeepmdDataSystem and backend-specific loaders are initialized.
- Normalize a deep copy of the parsed configuration, leaving the original configuration and input file unchanged.
- Use the same normalized paths for neighbor statistics and training-data initialization.

This insertion point covers TensorFlow, PyTorch, JAX, PaddlePaddle, PyTorch Exportable, and TensorFlow 2 without adding separate extxyz parsers to their data loaders.

## Labels and units

The conversion supports:
- atomic species and positions;
- periodic cells and all-periodic/all-nonperiodic PBC;
- total energy aliases supported by dpdata;
- both force and forces per-atom property names;
- direct virial and virials fields;
- ASE-style stress and stresses fields.

Stress is converted using:
```python
virial = -volume * stress
```

Six-component stress uses ASE Voigt order:
`xx yy zz yz xz xy`

Nine-component stress is interpreted in row-major tensor order. Unit metadata supported by dpdata is converted to DeePMD's eV/angstrom conventions.
Missing energy or forces produces a descriptive frame-specific error. Virial or usable stress is required when the configured loss has a nonzero virial prefactor.
Partially periodic inputs such as `pbc="T T F"` are rejected because traditional DeePMD NumPy systems cannot represent per-axis PBC.

## Heterogeneous inputs
Frames are grouped deterministically in first-occurrence order by:
- chemical composition and atom count;
- periodic versus nonperiodic behavior;
- available label set.

Compatible atom layouts are canonicalized through dpdata while preserving coordinate/force correspondence.
If one extxyz file expands into multiple internal systems, configurations using list-valued batch_size, explicit sys_probs, or indexed auto_prob blocks are rejected because their per-input mapping would be ambiguous.
Lists may mix extxyz files with existing DeePMD system directories. Mixing extxyz and LMDB in the same list is rejected because the existing LMDB path uses scalar-path semantics.

## Caching
Converted data is stored in a persistent DeePMD NumPy cache under the platform temporary directory, configurable through DEEPMD_EXTXYZ_CACHE.
The cache key includes:
- canonical source path;
- complete source-file SHA-256;
- conversion settings;
- converter schema version;
- dpdata version.
Cache publication uses a per-key lock and atomic directory rename. The source hash is checked again after conversion so a file modified during conversion is not published under a stale key. rglob_patterns continues to apply only to DeePMD directory discovery. Arbitrary .xyz files beneath a parent directory are not discovered implicitly, while explicitly listed extxyz files are not filtered by those patterns.

## Compatibility
Existing behavior remains unchanged for:
-DeePMD NumPy systems;
- raw systems and their existing conversion workflow;
- HDF5 systems;
- scalar LMDB inputs;
- directories containing type.raw;
- parent-directory recursive discovery.
ASE remains optional and is not required for extxyz training.

## Tests
Added focused coverage for:
- periodic and nonperiodic structures;
- energy, positions, forces, cells, and PBC round trips;
- direct virial labels;
- six- and nine-component ASE stress conversion;
- stress sign, unit, and component ordering;
- energy and force unit conversion;
- relative and absolute paths;
- multiple input files and deterministic ordering;
- training and validation initialization;
- heterogeneous frame splitting;
- atom-layout changes without coordinate/force corruption;
- missing energy, force, and virial errors;
- coordinate-only XYZ rejection;
- partially periodic PBC rejection;
- cache reuse and source-content invalidation;
- configuration and input-file immutability;
- NumPy, raw, HDF5, LMDB, parent-directory, and rglob_patterns compatibility.

Validation performed:
- 33 focused and legacy regression tests passed.
- ruff check . passed.
- Python compilation checks passed.
- A one-step PyTorch CPU training and validation smoke test completed from extxyz inputs and reported energy, force, and virial RMSE values for both datasets.

The native editable build was not completed on the test machine because its Windows environment lacks nmake and a C++ compiler. The PyTorch smoke test used the Python fallback with custom C++ operators disabled.

## Summary by CodeRabbit

- **New Features**
  - Training and validation now accept labeled `.xyz` and `.extxyz` files directly.
  - Extended-XYZ data is automatically validated, converted, cached, and reused.
  - Supports mixed datasets, heterogeneous-system splitting, energies, forces, cells, virials, and stresses.
  - Provides clear validation for missing labels, unsupported periodicity, and ambiguous configurations.

- **Documentation**
  - Added configuration examples and guidance for supported labels, units, caching, and file discovery.
  - Clarified native training formats and extended-XYZ conversion behavior.

- **Tests**
  - Added comprehensive coverage for conversion, caching, validation, and end-to-end training workflows.

### DIFF
--- a/deepmd/utils/argcheck.py
+++ b/deepmd/utils/argcheck.py
@@ -5360,12 +5360,12 @@ def training_data_args() -> list[
     doc_systems = (
         "The data systems for training. "
         "This key can be a list or a str. "
-        "When provided as a string, it can be a system directory path (containing 'type.raw') or a parent directory path to recursively search for all system subdirectories. "
-        "When provided as a list, each string item in the list is processed the same way as individual string inputs, i.e., each path can be a system directory or a parent directory to recursively search for all system subdirectories."
+        "Each value can be a system directory path (containing 'type.raw'), a parent directory path to recursively search for system subdirectories, or an explicitly named labeled '.xyz' or '.extxyz' file. "
+        "Extended-XYZ files are read with dpdata and transparently cached as DeePMD NumPy systems. Every frame must contain species, positions, total energy, and atomic forces; virial or ASE-style stress is also required when virial loss is enabled. "
+        "Files containing heterogeneous atom counts or compositions are split into fixed-shape systems in first-occurrence order. Partially periodic PBC cannot be represented by traditional DeePMD NumPy systems and is rejected. If such a file expands into multiple systems, list-valued batch_size, sys_probs, and indexed auto_prob blocks are rejected as ambiguous. "
+        "Lists may contain multiple extended-XYZ files and may mix them with existing DeePMD system directories."
     )
-    doc_patterns = (
-        "The customized patterns used in `rglob` to collect all training systems. "
-    )
+    doc_patterns = "The customized patterns used in `rglob` to collect DeePMD system directories. Explicit '.xyz' and '.extxyz' files are not discovered or filtered by these patterns. "
     doc_batch_size = f'This key can be \n\n\
 - list: the length of which is the same as the {link_sys}. The batch size of each system is given by the elements of the list.\n\n\
 - int: all {link_sys} use the same batch size.\n\n\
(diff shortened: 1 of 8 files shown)


njzjz (changes requested):
Thanks for your contribution. However, several months ago, we decided that the conversion should:

1. accept any format supported by dpdata (see https://github.com/deepmodeling/deepmd-kit/issues/5237), rather than only extxyz; and
2. always use deepmd/lmdb as the target format, given its superior performance.

qchempku2017:
> Thanks for your contribution. However, several months ago, we decided that the conversion should:
> 
> 1. accept any format supported by dpdata (see #5237), rather than only extxyz; and
> 2. always use deepmd/lmdb as the target format, given its superior performance.

Indeed a better decision. I hope such functionality will be released with the next formal version of deepmd-kit.

PR #6003, opened 2026-08-26
fix(dpmodel): preserve nopbc batch semantics

### LINKED ISSUE #6002: [BUG] pt_expt drops nopbc metadata but retains the zero box placeholder in normalize_batch
## Bug summary

For a valid non-periodic DeepMD system containing `nopbc` and no `box.npy`,
`DeepmdData` creates an all-zero `box` placeholder and
`DeepmdDataSystem` adds an empty `default_mesh` to encode non-periodicity.

`deepmd.dpmodel.utils.batch.normalize_batch()` then drops `default_mesh` as
metadata while retaining the all-zero `box`. The `pt_expt` training path calls
this function and passes the retained box to the model. Downstream DPA4 /
neighbor-graph code treats every non-`None` box as periodic and attempts to
normalize coordinates with the singular zero cell.

The non-periodic semantic is therefore lost between `DeepmdDataSystem` and the
model. Depending on the execution/compilation path, this can cause a singular
matrix failure, non-finite intermediates, or silently incorrect training.

## DeePMD-kit Version

`v3.2.0`, official tag and current `master` commit
`8cfd46e37448feca361ff3427a6c2b514ce5a2bb`.

The reproducing conda package reports `deepmd-kit 3.2.0`. Its
`deepmd/dpmodel/utils/batch.py` and `deepmd/utils/data.py` are byte-identical to
the files at the official v3.2.0 tag.

## Backend and its version

- backend: `pt_expt` (PyTorch Experimental)
- PyTorch: `2.12.0`
- Python: `3.12.14`
- NumPy: `2.5.2`
- array-api-compat: `1.15.0`
- CUDA runtime packages: `13.3`
- NVIDIA driver: `580.95.05`

The minimal reproducer reaches the faulty shared batch/region code on CPU and
does not require a GPU.

## How did you download the software?

Conda (`conda-forge` build
`deepmd-kit 3.2.0 cuda130py312h3eba875_mpi_openmpi_4`).

## Input Files, Running Commands, Error Log, etc.

Save the following as `reproduce_nopbc_zero_box.py`:

```python
#!/usr/bin/env python3
"""Minimal reproducer for lost nopbc semantics in normalize_batch()."""

from __future__ import annotations

import tempfile
from pathlib import Path

import numpy as np

from deepmd.dpmodel.utils.batch import normalize_batch, split_batch
from deepmd.dpmodel.utils.region import normalize_coord
from deepmd.utils.data_system import DeepmdDataSystem


with tempfile.TemporaryDirectory() as tmpdir:
    system = Path(tmpdir) / "nopbc"
    set_dir = system / "set.000"
    set_dir.mkdir(parents=True)

    # A valid non-periodic H2 system: nopbc is present and box.npy is absent.
    (system / "type.raw").write_text("0 0\n")
    (system / "type_map.raw").write_text("H\n")
    (system / "nopbc").touch()
    np.save(
        set_dir / "coord.npy",
        np.array([[0.0, 0.0, 0.0, 0.74, 0.0, 0.0]], dtype=np.float64),
    )

    data_system = DeepmdDataSystem(
        [str(system)],
        batch_size=1,
        test_size=1,
        rcut=6.0,
        type_map=["H"],
        trn_all_set=True,
        shuffle_test=False,
    )
    raw = data_system.get_batch()
    normalized = normalize_batch(raw)
    inputs, _ = split_batch(normalized)

    print("pbc:", data_system.data_systems[0].pbc)
    print("raw default_mesh size:", raw["default_mesh"].size)
    print("raw box:", raw["box"])
    print("normalized has default_mesh:", "default_mesh" in normalized)
    print("normalized box is None:", inputs["box"] is None)
    print("normalized box all zero:", np.allclose(inputs["box"], 0.0))

    try:
        normalize_coord(
            inputs["coord"].reshape(1, 2, 3),
            inputs["box"].reshape(1, 3, 3),
        )
    except Exception as exc:
        print("normalize_coord exception:", type(exc).__name__, str(exc))
```

Run it with:

```bash
python reproduce_nopbc_zero_box.py
```

Observed output:

```text
pbc: False
raw default_mesh size: 0
raw box: [[0. 0. 0. 0. 0. 0. 0. 0. 0.]]
normalized has default_mesh: False
normalized box is None: False
normalized box all zero: True
normalize_coord exception: LinAlgError Singular matrix
```

Expected behavior:

```text
normalized box is None: True
```

No periodic coordinate normalization should be attempted for this system.

## Steps to Reproduce

1. Create a standard DeepMD NPY system with `type.raw`, `type_map.raw`,
   `set.000/coord.npy`, and an empty `nopbc` marker, but no `box.npy`.
2. Load one batch using `DeepmdDataSystem`.
3. Observe that the raw batch has `default_mesh.size == 0` and an all-zero
   `box` placeholder.
4. Pass the batch through `normalize_batch()` and `split_batch()`.
5. Observe that `default_mesh` has been dropped while `box` remains an
   all-zero array rather than `None`.
6. Passing that box to `normalize_coord()` attempts to invert the singular
   zero cell.

The script performs all six steps in a temporary directory and does not use
external data.

## Code path

1. `deepmd/utils/data.py`
   - registers `box` with `must=self.pbc` and default `0.0`;
   - when `nopbc` is present and `box.npy` is absent, `_load_data()` creates an
     all-zero placeholder with `np.full()`.
2. `deepmd/common.py` and `deepmd/utils/data_system.py`
   - `make_default_mesh(False, False)` returns an empty array;
   - `DeepmdDataSystem.get_batch()` adds it as `default_mesh`.
3. `deepmd/dpmodel/utils/batch.py`
   - `_DROP_KEYS` contains `default_mesh`;
   - `normalize_batch()` drops it but retains `box` as a model input.
4. `deepmd/pt_expt/train/training.py`
   - calls `normalize_batch(data_sys.get_batch())`, then `split_batch()`.
5. `deepmd/dpmodel/utils/neighbor_graph/builder.py`
   - treats `box is not None` as periodic and calls `normalize_coord()`.
6. `deepmd/dpmodel/utils/region.py`
   - calls `xp.linalg.inv(cell)`, which is invalid for the zero cell.

There is consistent handling elsewhere in the same repository:
`dpa_adapt/finetuner.py` converts `nopbc` or all-zero boxes to `None`.

## Suggested fix and regression coverage

Before dropping `default_mesh`, `normalize_batch()` could translate its encoded
PBC state into the canonical model input:

```python
default_mesh = batch.get("default_mesh")
is_nonperiodic = default_mesh is not None and np.size(default_mesh) in (0, 1)
...
if is_nonperiodic:
    out["box"] = None
```

The sizes `(0, 1)` cover standard and mixed-type non-periodic systems according
to `make_default_mesh()`; periodic systems use sizes `(6, 7)`.

Suggested regression tests should cover standard and mixed-type non-periodic
batches, verify `box is None` after normalization, and verify that periodic
boxes remain unchanged.

## Further information

In a four-replica DPA4 fine-tuning workflow using only valid `nopbc` systems,
the unconverted zero box caused force errors to degrade immediately from the
first training step. A local compatibility shim that sets `box=None` when the
raw `default_mesh` is empty restored normal training in controlled runs; model,
data, optimizer, seed, and learning-rate settings were otherwise held fixed.

GitHub issue searches for `normalize_batch default_mesh`, `zero box nopbc`,
`non-periodic box pt_expt`, and `nopbc DPA4 training` found no existing report
in `deepmodeling/deepmd-kit` as of 2026-08-26.

hcustc:
## Summary

- translate the legacy non-periodic `default_mesh` encodings (sizes 0 and 1) to the canonical model input `box=None`
- preserve periodic boxes for both standard-type and mixed-type batches
- add regression coverage for the batch boundary and a real temporary `nopbc` NPY system

## Root cause

`DeepmdDataSystem` represents a valid non-periodic system with an empty or single-element `default_mesh` and an all-zero `box` placeholder. `normalize_batch()` dropped the mesh metadata but retained the placeholder, so downstream code treated the non-`None` box as periodic and attempted to invert a singular cell.

The fix performs the semantic conversion before the metadata is discarded, without mutating the input batch.

Fixes #6002.

## Testing

- `.venv/bin/python -m pytest source/tests/common/test_batch_nopbc.py source/tests/common/test_batch_charge_state.py -q` (10 passed, 10 subtests passed)
- `.venv/bin/ruff check .`
- `.venv/bin/ruff format --check .`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of non-periodic systems during batch normalization.
  * Placeholder box data is now omitted from model inputs for non-periodic systems.
  * Periodic systems continue to retain their box data.
  * Non-periodic data without box information no longer receives an unintended box.

* **Tests**
  * Added coverage for periodic and non-periodic batch normalization, including systems with missing or placeholder mesh and box metadata.


deepmd/dpmodel/utils/batch.py
```diff
@@ -37,6 +37,8 @@ def normalize_batch(batch: dict[str, Any]) -> dict[str, Any]:
     * ``"type"`` is renamed to ``"atype"`` (int64).
     * ``"natoms_vec"`` (1-D) is tiled to 2-D ``[nframes, 2+ntypes]``
       and stored as ``"natoms"``.
+    * Non-periodic ``default_mesh`` encodings convert the zero ``box``
+      placeholder to ``None``.
```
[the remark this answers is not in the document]
hcustc: Updated the docstring to describe the actual semantic conversion: non-periodic `default_mesh` encodings canonicalize the model input to `box=None`, without implying that `normalize_batch()` inspects the box values. Thanks for catching this.

deepmd/dpmodel/utils/batch.py
```diff
@@ -56,6 +58,8 @@ def normalize_batch(batch: dict[str, Any]) -> dict[str, Any]:
         Normalized batch dict (new dict; the input is not mutated).
     """
     out: dict[str, Any] = {}
+    default_mesh = batch.get("default_mesh")
+    is_nonperiodic = default_mesh is not None and np.size(default_mesh) in (0, 1)
```
wanghan-iapcm: The name claims more than the expression computes, and it cost me real time.

`is_nonperiodic` reads as a fact about the batch. What it actually decides is narrower: *the `DeepmdDataSystem` mesh encoding says non-periodic*. For any batch without a `default_mesh` it is `False` — and `False` there does not mean "periodic", it means "this encoding did not speak". LMDB batches are the concrete case: `decode_lmdb_batch` emits no `default_mesh`, so the flag is always `False` for them.

That is fine, and I checked why: a non-periodic LMDB dataset has no box in any frame, and `decode_lmdb_batch` drops a field that is absent from any frame of the batch rather than back-filling it, so the batch simply carries no `box` key and the model correctly gets none. No placeholder, nothing to translate. But nothing at this line says so, and reading `is_nonperiodic == False` for an LMDB batch invites exactly the wrong inference — I went and built a bug report out of it before checking the decode path.

`mesh_says_nonperiodic`, or a one-line comment noting that batches without a mesh carry no box placeholder either, would close that off. Worth doing precisely because the value is load-bearing for a correctness fix.

The choice of `default_mesh` over the alternatives is right, incidentally, and worth keeping: `pbc` is decided solely by the `nopbc` marker (`deepmd/utils/data.py:1135-1139`), independently of whether `box.npy` exists, so a system can legitimately be non-periodic *and* carry a real box. A `find_box` test or an all-zero-box test would misjudge that case; `default_mesh` gets it right, since `DeepmdDataSystem.default_mesh` comes straight from `make_default_mesh(self.data_systems[ii].pbc, ...)`.

source/tests/common/test_batch_nopbc.py
```diff
@@ -0,0 +1,90 @@
...
+    @staticmethod
+    def _batch(default_mesh_size: int, box: np.ndarray) -> dict:
+        return {
+            "coord": np.zeros((1, 2, 3)),
+            "type": np.zeros((1, 2), dtype=np.int32),
+            "box": box,
+            "default_mesh": np.zeros(default_mesh_size, dtype=np.int32),
+        }
+
+    def test_nonperiodic_box_is_none(self) -> None:
```
wanghan-iapcm, line 33: Every batch these tests build carries a `"default_mesh"` key, so the `default_mesh is not None` short-circuit is never taken, and the `"box" in out` guard's false side is untested too. Both are reachable — an LMDB batch takes the first one on every step of training — so a case each would be cheap and would document the intended no-op.

Separately, `mixed_type` is unpacked from the tuple but only ever used as a `subTest` label; `_batch()` takes no such argument and the batch contents do not vary with it. What the loop actually covers is "both non-periodic mesh sizes map to `None`", which is worth covering, but the parametrization reads as though mixed-type behaviour were being exercised. Either drop the variable or make `_batch()` vary with it.

`TestDeepmdDataSystemNopbc` is a good test — it builds a real nopbc NPY system, runs the real `get_batch()` through `normalize_batch`/`split_batch`, and fails on pre-fix code because `inputs["box"]` is the zeros array rather than `None`.

wanghan-iapcm (commented):
The diagnosis is right and the fix is correct. The layer is right too, which is worth recording: doing this at the producer instead — `if not self.pbc: b_data["box"] = None` in `DeepmdDataSystem.get_batch_standard`/`get_batch_mixed` — would break TensorFlow and JAX training on nopbc systems, since `get_batch()` is shared with backends that require `box` to be an array. `normalize_batch` is the first point downstream of that contract.

Two notes, neither blocking.

deepmd/dpmodel/utils/batch.py
hcustc: Fixed in 4399883b. The predicate is now named mesh_says_nonperiodic, so a false value no longer reads as proof that a mesh-less batch is periodic. I kept default_mesh as the source of truth and added the mesh-less no-op coverage in the sibling test update.

source/tests/common/test_batch_nopbc.py
hcustc: Fixed in 4399883b. The unit tests now cover both no-op branches explicitly: a batch without default_mesh preserves its box, and a non-periodic mesh without a box does not add one. The mesh-size loops now label mesh_size directly, and the non-periodic unit case uses a real nonzero box so it pins the mesh semantics independently of the zero-box NPY integration case. Local validation: ruff format ., ruff check ., and the two focused test files (12 passed, 10 subtests passed).

wanghan-iapcm (approved):
Both earlier threads are addressed at this head: the predicate reads as what it is, and the two no-op branches (no `default_mesh`, non-periodic mesh without a `box`) each have a test. I re-ran the new test file against the pre-fix `normalize_batch` (2 failures, `box` comes back as the zeros array) and against this head (5 passed), and the merge commit brings nothing unrelated into the diff.

I also traced every consumer of `normalize_batch` output (tf2 trainer, pt_expt trainer, stat and env-mat-stat code): all of them already accept `box=None`.

One informational note, not a request: a data system that mixes periodic and non-periodic systems will now hand the model `box` as an array on some steps and `None` on others. That is the correct signal, but it is a new argument-type alternation, so `tf.function` and `torch.compile` may retrace on each switch. Worth remembering if someone reports slow mixed-set training later.

CI on this head has just been started; approving on the code.

PR #6009, opened 2026-09-05
feat: add named model presets for DPA4 and DPA4C

OutisLi:
## Summary

- add `model.preset`: `"<family>-<grade>-<version>"` (for example `dpa4-nano-v20260901` or `dpa4c-neo-v20260901`) fills in `type`, the 118-element `type_map`, `descriptor` and `fitting_net` from a released architecture; entries written next to the preset take precedence, `type` and `type_map` as a whole and `descriptor` / `fitting_net` key by key, so run-specific options (`use_amp`, `seed`, `sel`, charge and spin conditioning) are written alongside
- presets are expanded right after the input is loaded, before multi-task sharing, fine-tuning rules and argument checking, and inside `update_deepmd_input` for the other input consumers; the `preset` key is removed, so `out.json` records the expanded model and an unexpanded preset fails the strict argument check
- ship DPA4 `v20260820` (nano to pro) and `v20260901` (nano to ultra) and DPA4C `v20260901` (nano to plus); the table in `deepmd/utils/model_preset.py` lists shared options once, then per-grade knobs and per-version changes
- multi-task: a `preset` next to `model_dict` is the base of every branch and of the `shared_dict` entries the branches reference as `descriptor` or `fitting_net`; entries written next to `model_dict` are branch defaults; `examples/water/dpa4/input_multitask_preset.json` is the shared-descriptor multitask example written with a preset
- correct the DPA4C example and docs to the released Neo grade (channels 64, lmax 2, radial_modes 0, fitting width 256), drop per-grade parameter and benchmark tables from the user docs, and add `examples/water/dpa4/input_preset.json`

## Validation

- `python -m pytest source/tests/common/test_model_preset.py source/tests/common/test_examples.py source/tests/common/test_doc_train_input.py source/tests/common/test_compat_optimizer.py source/tests/common/test_dpmodel_train.py -q` — 45 passed, 218 subtests passed
- every preset normalizes to the same model as its release `input.json` (key-by-key comparison after `normalize`, runtime keys excluded)
- 20-step training runs: multi-task with the new `input_multitask_preset.json` (shared descriptor and case-embedded fitting from the preset) and with a top-level `descriptor` override next to the preset; `dp --pt train` with `dpa4-nano-v20260901` and `dpa4-neo-v20260820`, `--finetune` from the resulting checkpoint, multi-task with a top-level and a branch-level preset, `dp --pt-expt train` with `dpa4c-nano-v20260901` and with the corrected DPA4C example
- `ruff check` / `ruff format --check` clean; all pre-commit hooks pass

## Summary by CodeRabbit

- **New Features**
  - Added named DPA4 and DPA4C model presets with case-insensitive selection for single-task and multitask configurations.
  - Presets now expand automatically during training and input conversion, with explicit settings taking precedence.
- **Documentation**
  - Documented preset naming, available grades, expansion behavior, and override rules.
  - Added DPA4 preset training examples and updated DPA4C examples.
- **Tests**
  - Added coverage for preset expansion, overrides, multitask models, validation, and example configurations.

### DIFF
--- a/deepmd/dpmodel/train/entrypoint.py
+++ b/deepmd/dpmodel/train/entrypoint.py
@@ -27,6 +27,9 @@ from deepmd.utils.argcheck import (
 from deepmd.utils.compat import (
     update_deepmd_input,
 )
+from deepmd.utils.model_preset import (
+    expand_model_preset,
+)
 
 log = logging.getLogger(__name__)
 
@@ -59,6 +62,9 @@ class AbstractTrainEntrypoint(ABC):
         log.info("Configuration path: %s", options.input_file)
         options = self.prepare_options(options)
         config = self.load_config(options.input_file)
+        if "model" in config:
+            # Version-0 inputs gain their model section only in update_input.
+            config["model"] = expand_model_preset(config["model"])
         self.validate_options(config, options)
 
         config = self.preprocess_config(config, options)
--- a/deepmd/pd/entrypoints/main.py
+++ b/deepmd/pd/entrypoints/main.py
@@ -70,6 +70,9 @@ from deepmd.utils.data_system import (
     get_data,
     process_systems,
 )
+from deepmd.utils.model_preset import (
+    expand_model_preset,
+)
 from deepmd.utils.path import (
     DPPath,
 )
@@ -255,6 +258,7 @@ def train(
     if LOCAL_RANK == 0:
         SummaryPrinter()()
     config = j_loader(input_file)
+    config["model"] = expand_model_preset(config["model"])
     # ensure suffix, as in the command line help, we say "path prefix of checkpoint files"
     if init_model is not None and not init_model.endswith(".pd"):
         init_model += ".pd"
--- a/deepmd/pt/entrypoints/main.py
+++ b/deepmd/pt/entrypoints/main.py
@@ -91,6 +91,9 @@ from deepmd.utils.data_system import (
     get_data,
     process_systems,
 )
+from deepmd.utils.model_preset import (
+    expand_model_preset,
+)
 from deepmd.utils.stat_file import (
     StatFileSpec,
 )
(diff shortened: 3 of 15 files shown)


OutisLi:
Pushed `472bef96`, addressing every actionable finding from this round:

- Copilot (module docstring): `type`/`type_map` are whole-region replacements, `descriptor`/`fitting_net` merge key by key — the docstring now states both rules precisely.
- Copilot (non-mapping `model_dict`): a `model_dict` (or a branch) that is not a mapping is left untouched for the argument check to reject.
- CodeRabbit (`399`, inherited branches): the top-level `type`/`type_map`/`descriptor`/`fitting_net` written next to `model_dict` are now merged into every branch before that branch's own preset expands.
- CodeRabbit (`343`, key-wise merge on a malformed `type_map`): the key-wise merge is now restricted to `descriptor`/`fitting_net`; a `type_map` given as a mapping is treated as a whole-value override instead of raising `TypeError` before the argument check runs.
- Codex P2 (shared-role scan): the scan and the branch expansion now read the same branches merged with the top-level defaults, so a `descriptor`/`fitting_net` reference a branch inherits only through the top-level default is still recognized as referenced.
- Codex P2 (leftover top-level keys on `pt_expt`/`jax`/`tf2`): the four regions are removed from the top level once distributed to every branch, since the shared dpmodel-based multi-task preprocessing does not cascade them on its own (unlike the PyTorch backend). Verified with an actual `dp --pt-expt train` run of a DPA4C multi-task config with a top-level preset and a top-level `descriptor` override.
- Codex P2 (alias collision): an explicit override using a legacy key alias (e.g. `so2_layers`) is now canonicalized against the preset's own component type before merging, so it replaces the preset's canonical key (`mixing_layers`) instead of sitting next to it and failing strict validation.

Added a regression test for each. Full suite: 45 passed; every preset still normalizes to the same model as its reference input.

njzjz (approved):

PR #6011, opened 2026-09-08
feat(pt,pd): exact full-scan output statistics via streaming QR

anyangml:
## Summary

Output statistics (energy bias / std) are fitted from `data_stat_nbatch` batches per system. Elements too rare to appear in that sample get no support in the least-squares design matrix, so their bias collapses to the min-norm zero, and they are also missing from `observed_type`. On large mixed-type datasets this shows up as a large deviation for a handful of rare elements.

This adds `model.data_stat_full` (bool, default `false`, pt + pd), which computes bias, std and observed types from **every** training frame instead.

## How

`ReduStatAccumulator` (`deepmd/utils/out_stat.py`) folds frames into a running QR factor of the augmented design matrix `[natoms | output_redu | 1]`. Since `R.T @ R == A.T @ A`, the least-squares solution and the residual std are identical to a fit over all frames, while memory stays at `O((ntypes + ndim + 1) ** 2)` regardless of the number of frames. Blocks are buffered so one QR amortizes over many small batches.

`assigned_bias` (preset out bias), `intensive` and `atom_exclude_types` are all column-linear operations, so they apply exactly to the compressed factor rather than to the frames.

Wiring: the trainer attaches a `ReduStatScanner` to the stat sampler as the `redu_stat_scanner` attribute; `_make_wrapped_sampler` forwards it and `_compute_output_stats_global` picks it up, so no signature churn through the atomic-model hierarchy. The scan rebuilds each loader without its sampler, so a `DistributedSampler` or `WeightedRandomSampler` cannot hide frames from the chief rank.

Falls back to the sampled path (with a warning) for delta bias (`model_forward`) and `stats_distinguish_types=False`, neither of which can be expressed on the compressed factor. Descriptor input statistics still use `data_stat_nbatch`.

## Effect

Reproduction with one element present in a single late frame, sampling the first 20 frames:

```
true       [18.2172 82.5263 54.0435 68.9262 62.7498]
sampled    [18.2174 82.5263 54.0436 68.9259  0.    ]   <- rare element collapses
full scan  [18.2172 82.5263 54.0435 68.9261 62.75  ]
```

## Tests

`source/tests/common/test_out_stat.py` gains 15 cases checking the streaming accumulator against `compute_stats_from_redu` for the `intensive`, `assigned_bias` and `type_mask` combinations (rtol 1e-9), plus repeated compression with batch size 1, shape preservation and the empty-accumulator error. Existing global/atomic output-stat tests pass unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01ByhbNb5V8vZAGqY88H3wF6

## Summary by CodeRabbit

* **New Features**
  * Added an optional full-dataset statistics mode that processes every training frame.
  * Added streaming statistics processing to reduce memory use during full-data scans.
  * Added support for full-dataset statistics in PyTorch and PaddlePaddle configurations.
  * Improved observed element-type detection using complete atom-count data when available.
  * Full-scan statistics are cached and reused when valid, with safeguards for interrupted updates.
  * Batch-based input statistics remain unchanged when full-dataset output statistics are enabled.

* **Tests**
  * Added coverage for streaming accumulation, caching, intensive statistics, assigned biases, and type filtering.

### DIFF
--- a/deepmd/pt/utils/stat.py
+++ b/deepmd/pt/utils/stat.py
@@ -41,8 +41,10 @@ from deepmd.utils.path import (
     DPPath,
 )
 from deepmd.utils.stat_file import (
+    load_output_stat_full_scan,
     load_paired_items,
     replace_paired_items,
+    save_output_stat_full_scan,
 )
 
 log = logging.getLogger(__name__)
@@ -286,6 +288,11 @@ def scan_redu_stats(
     -------
     ReduScanResult
         The accumulators, the per-type atom counts and the frame count.
+
+    Notes
+    -----
+    Statistics initialization runs on the chief process only, so one scan of
+    the training set is performed per run, not per rank.
     """
     stats: dict[str, ReduStatAccumulator] = {}
     natoms_total = np.zeros(ntypes, dtype=np.int64)
@@ -574,8 +581,39 @@ def compute_output_stats(
     assert isinstance(keys, list)
     requested_keys = list(keys)
 
+    # a full scan cannot replace the sampled path for delta bias or type-blind
+    # statistics, so resolve it before the cache is consulted
+    redu_scanner = get_redu_stat_scanner(merged)
+    if redu_scanner is not None and (
+        model_forward is not None or not stats_distinguish_types
+    ):
+        log.warning(
+            "Falling back to sampled output statistics: a full scan supports "
+            "neither delta bias nor stats_distinguish_types=False."
+        )
+        redu_scanner = None
+
     # try to restore the bias from stat file
     bias_atom_e, std_atom_e = _restore_from_file(stat_file_path, keys)
+    if (
+        bias_atom_e is not None
+        and redu_scanner is not None
+        and not load_output_stat_full_scan(stat_file_path)
+    ):
+        # the cache holds sampled values, which is what data_stat_full replaces
+        if getattr(stat_file_path, "mode", None) == "r":
+            log.warning(
+                "`data_stat_full` is set, but the read-only statistics cache "
+                "holds output statistics estimated from sampled batches; they "
+                "are used as they are."
+            )
+        else:
+            log.info(
+                "Recomputing output statistics: the cache holds values "
+                "estimated from sampled batches, which `data_stat_full` "
+                "replaces."
+            )
+            bias_atom_e, std_atom_e = None, None
 
     # failed to restore the bias from stat file. compute
     if bias_atom_e is None:
@@ -673,7 +711,7 @@ def compute_output_stats(
             stats_distinguish_types,
             intensive,
             model_pred_g,
-            get_redu_stat_scanner(merged),
+            redu_scanner,
         )
         bias_atom_a, std_atom_a = _compute_output_stats_atomic(
             sampled,
@@ -709,6 +747,7 @@ def compute_output_stats(
                 bias_atom_e,
                 std_atom_e,
             )
+            save_output_stat_full_scan(stat_file_path, redu_scanner is not None)
 
     bias_atom_e = {kk: to_torch_tensor(vv) for kk, vv in bias_atom_e.items()}
     std_atom_e = {kk: to_torch_tensor(vv) for kk, vv in std_atom_e.items()}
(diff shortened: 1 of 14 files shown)


deepmd/pt/utils/stat.py
```diff
@@ -222,6 +233,103 @@ def make_stat_input(
...
+    Returns
+    -------
+    ReduScanResult
+        The accumulators, the per-type atom counts and the frame count.
+    """
+    stats: dict[str, ReduStatAccumulator] = {}
+    natoms_total = np.zeros(ntypes, dtype=np.int64)
+    nframes = 0
+    log.info(
+        "Scanning all frames of %d systems for output statistics", len(dataloaders)
```
iProzd, line 301: Every rank runs this scan in full — the stats are not rank-guarded and not broadcast, so under DDP the startup I/O is N times a pass over the whole training set.

Correct as it stands, since that is what makes the ranks agree without a collective; just worth a note in the docs, or a rank-0 scan plus a broadcast of the `R` factors.

deepmd/utils/argcheck.py
```diff
@@ -3398,6 +3398,13 @@ def model_args(
     doc_type_map = "A list of strings. Give the name to each type of atoms. It is noted that the number of atom type of training system must be less than 128 in a GPU environment. If not given, type.raw in each system should use the same type indexes, and type_map.raw will take no effect."
     doc_data_stat_nbatch = "The model determines the normalization from the statistics of the data. This key specifies the number of `frames` in each `system` used for statistics."
     doc_data_stat_protect = "Protect parameter for atomic energy regression."
+    doc_data_stat_full = (
```
iProzd, line 3401: `compute_output_stats` returns early on a cached `stat_file`, which records nothing about how the values were produced — so setting `data_stat_full: true` and rerunning against an existing stat file silently reuses the old sampled bias, which is the value the flag exists to replace.

The two open Copilot threads about `type_mask` coming from `sampled[0]` can be closed: `atom_exclude_types` is written to every sample from one model-level source in `base_atomic_model.py:250-253`, so it is uniform by construction.

iProzd (commented):
Reviewed the full diff against `28b7d068` at head `36b955a4`. No blocking findings.

I checked the exactness claim independently: `ReduStatAccumulator` matches `compute_stats_from_redu` to ~1e-13 over ten cases, including a rank-deficient design, `assigned_bias`, `intensive`, and a 1e6 column-scale imbalance. Two minor observations inline.

deepmd/pt/utils/stat.py
anyangml: The scan is rank-guarded, just not at this call site.

`initialize_statistics` — the only path that reaches `ReduStatScanner.scan` — is wrapped in `Trainer._run_stat_on_chief`, which forwards to `run_stat_on_chief(action, is_chief=self.rank == 0, ...)`. That helper executes `action()` under `if is_chief:`; peers only take part in the failure broadcast. So it is one pass over the training set per run, not per rank, and the ranks agree because the chief's values reach them the same way the sampled ones always have.

The other two consumers of `get_sample_func` are guarded as well: `update_finetune_bias` goes through the same `_run_stat_on_chief`, and `change_bias_after_training` sits behind `self.rank == 0 or dist.get_rank() == 0`. Both reach `change_out_bias`, whose default `change-by-statistic` mode passes `model_forward` — which the scan declines anyway, falling back to the sampled path.

It is precisely because only the chief scans that `_full_pass_loader` rebuilds each loader without its sampler: under a `DistributedSampler` the chief would otherwise see only `1/world_size` of the frames and the exactness claim would not hold.

Your other comment was a real gap and is fixed: the cache now records whether its output statistics came from a full scan, so `data_stat_full` rejects a sampled cache and recomputes instead of silently reusing it.

Reading `scan_redu_stats` on its own does leave the rank question open, so I have added a Notes section to its docstring recording that statistics initialization runs on the chief only.

iProzd (approved):
Both points from my earlier review are resolved — you're right that `initialize_statistics` runs under `_run_stat_on_chief`, so the scan is once per run, not per rank. The stat-file mode marker and its tests cover the cached-sampled-values case.

PR #6015, opened 2026-09-11
fix(pt): pass PACKED and Wigner strides in the flash-bwd tuning sweep

iProzd:
## Summary

- Root cause: #6001 changed the two flash-attention backward kernels (`_flash_bwd_kernel`: new `PACKED` constexpr; `_flash_bwd_block_kernel`: 20 → 26 positional args) and updated the launcher `_launch_backward`, but not the two raw launches in the tuning sweep `sweep_flash_bwd`, which omit `PACKED` and the six Wigner strides.
- Trigger: the sweep runs only in `dp --pt freeze` at `DP_TRITON_INFER >= 2` (the #6001 default) when the GPU's built-in table lacks the family. #6001 filled the new families only for the RTX PRO 6000, so on other GPUs (e.g. H20) every default DPA4 freeze fails: `TypeError: dynamic_func() missing 1 required positional argument: 'PACKED'`.
- Fix: derive `packed` and the strides with `_rotation_strides`, as `_launch_backward` does; no kernel, table or policy change. Test: a GPU regression test runs the real `flash_bwd` sweep (fails on master, passes with the fix).

## Validation (H20)

- New test: fails before the fix, passes after.
- `test_descriptor_sezm_triton.py`: 37 passed.
- Default-level `dp --pt freeze` (`examples/water/dpa4`): all six tuning groups complete.
- `test_sezm_export.py` + `test_dpa4_export.py`: 45 passed; ruff and pre-commit clean.

## Follow-ups

- The freeze then fails in the with-comm sidecar export (FakeTensorMode mismatch, another #6001 regression; `nloc` constraint violation at level 0): separate PR.
- The H20 built-in table lacks the four #6001 tuning families, so every level-2 freeze re-sweeps: separate change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved flash-attention backward configuration sweeps to correctly handle packed layouts and Wigner strides.
  - Ensured both edge and block backward kernels receive the required layout metadata during tuning.

- **Tests**
  - Added GPU coverage for real backward-sweep execution and configuration registration.
  - Validated that selected configurations come from the expected candidate sets.

### DIFF
--- a/deepmd/pt_expt/kernels/triton/sezm/sweep_tile_configs.py
+++ b/deepmd/pt_expt/kernels/triton/sezm/sweep_tile_configs.py
@@ -152,6 +152,7 @@ from deepmd.pt_expt.kernels.triton.sezm.flash_atten import (
     _flash_bwd_block_kernel,
     _flash_bwd_kernel,
     _flash_bwd_op,
+    _rotation_strides,
 )
 from deepmd.pt_expt.kernels.triton.sezm.so2_stack_fp16x3 import (
     _mixing_stack_fp16x3_bwd_op,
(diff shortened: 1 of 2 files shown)


OutisLi (approved):
The sweep launch arguments match the current backward-kernel signatures and the production launcher. The added regression test exercises the real sweep and configuration registration.

The with-comm freeze failure mentioned in the follow-up notes is addressed in #6016.

PR #6016, opened 2026-09-11
feat(dpa4): add fixed radial presets and improve export correctness

OutisLi:
Add fixed radial-basis options and single-envelope DPA4 configurations, with preset-aware JSON Schema validation and completion. Edge-free SeZM frames follow the same descriptor computation as frames with neighbors, and CUDA export metadata matches runtime output layouts.

## Changes

- Support `bessel/fix` and `gaussian/fix` in DPA4 and DPA4C. The fixed variants retain the checkpoint parameter names and shapes while excluding the basis frequencies or centers from optimization. Add the `v20260911` model presets.
- Deliberately retire the unreleased `dpa4-*-v20260901` presets without aliases; inputs using them must select a retained preset explicitly. DPA4 retains `v20260820` and `v20260911`; DPA4C retains `v20260901` and `v20260911`.
- Accept an integer DPA4 `env_exp` to apply only the message-passing cutoff envelope. Keep PT, dpmodel, PT-expt and the fused CUDA radial operator aligned, including gradients.
- Let descriptors declare the radial-input matrices that HybridMuon routes to AdamW. Both trainers collect these declarations without requiring parameter renames or input-file options. Model, atomic-model and hybrid-descriptor compositions propagate the declarations with their indexed parameter paths.
- Run the radial embedding, environment seed and interaction blocks for edge-free frames, preserving the isolated-atom cutoff limit. Handle empty edge sets in the Triton radial-mixing reference, grid branches and Cartesian tensor products.
- Avoid caching trace-bound FakeTensors in the CUDA radial and Wigner helpers. Express the scalar SO(3) readout as a weighted product/reduction so symbolic export does not require an einsum contraction-path search.
- Match the grid-product and geometric-embedding fake output layouts to the contiguous CUDA allocations. This prevents AOTInductor from using incorrect strides for non-contiguous operands; regression tests check forward and backward operator metadata against actual CUDA execution.

- Generate preset-aware JSON schemas from the existing argument definitions and preset registry. Editors accept partial overrides, suggest preset names and model-specific fields, and retain type checking. Multi-task schemas recognize inherited and per-branch presets.

## Validation

- Review regressions cover 12 real PT/PT-expt optimizer-routing cases, including nested compositions and spin wrappers, and compare radial-input updates with `torch.optim.AdamW`. Related CPU suites pass in two runs (141 passed with 5 subtests, and 122 passed); 31 GPU tests pass across empty-edge/parity and fp32 acceleration suites.
- Targeted CPU suites for presets, radial-basis serialization, PT/dpmodel/PT-expt parity, force-loss gradients, AdamW routing, spin TorchScript and scalar readout: 104 passed.
- CUDA validation after a regular editable CUDA build: 20 passed, 33 subtests passed on RTX PRO 6000 Blackwell with PyTorch 2.13.0+cu130. Covers grid/GIE numerical contracts, non-contiguous input opchecks, CUDA/Triton descriptor and coordinate-gradient parity, and export dispatch.
- Single- and double-envelope CUDA AOTI packages both compile, load, execute and match the exported graph's outputs. The non-contiguous layout regressions fail before the fake-layout fix and pass afterwards.
- An edge-free frame completes forward and force-loss backward with `DP_TRITON_TRAIN=1` and `DP_CUDA_TRAIN=1`, matching the reference path.
- JSON Schema, preset and argument regression suites: 110 passed. The official VS Code JSON language service validates the reported input without diagnostics, catches invalid override types, and offers the correct preset/descriptor/fitting completions for single- and multi-task inputs.
- Ruff lint, formatting checks and `git diff --check` pass.

## Summary by CodeRabbit

- **New Features**
  - Added support for single or separate radial-basis and message-passing envelopes.
  - Added `bessel/fix` and `gaussian/fix` options to keep basis parameters fixed during training.
  - Added updated DPA4 and DPA4C model presets.
  - HybridMuon can route selected matrix parameters through AdamW.
  - Model presets are now recognized in generated JSON Schema.

- **Bug Fixes**
  - Improved edge-free frame handling, zero-distance calculations, and accelerated/traced execution.
  - Preserved compatibility across serialization and backend implementations.

- **Documentation**
  - Updated descriptor, basis, envelope, preset, and schema configuration guidance.


deepmd/utils/argcheck.py
```diff
@@ -6454,13 +6460,17 @@
     str
         JSON schema.
     """
+    from deepmd.utils.json_schema import (
+        with_model_presets,
+    )
```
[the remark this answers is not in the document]
OutisLi: Fixed in 1a17b1583f0678d4cc53bc41bbc5e1af3b61235a. Declarative preset tables now live in model_preset_data.py, which depends only on the standard library. json_schema imports the data module, while model_preset retains the existing public re-exports and expansion logic. This breaks both reported cycles structurally without adding delayed-import workarounds. Import-graph and public-re-export checks pass, along with the schema/preset/doc-generator regression suite.

deepmd/pt_expt/kernels/cuda/dpa4/edge_radial.py
```diff
@@ -198,18 +206,23 @@ def __init__(self, envelope: Any, basis: Any) -> None:
...
-        self._basis_type = BESSEL if basis.basis_type == "bessel" else GAUSSIAN
+        self._basis_type = BESSEL if basis.basis_family == "bessel" else GAUSSIAN
         self._env = series_coefficients(envelope.p)
-        self._rbf = series_coefficients(basis.envelope.p)
+        self._rbf = series_coefficients(basis.exponent)
         self._series: tuple[torch.Tensor, torch.Tensor] | None = None
 
     def series(self, device: torch.device) -> tuple[torch.Tensor, torch.Tensor]:
         """The two Horner series on the compute device."""
         if self._series is None or self._series[0].device != device:
```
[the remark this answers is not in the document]
OutisLi: Addressed in 1a17b1583f0678d4cc53bc41bbc5e1af3b61235a with a cache-then-fullgraph-compile regression. The reproducer exposed two different causes: Dynamo traced the lazy fake-op registration despite the functools cache, and converting the Gaussian coefficient tensor to float inside the traced call produced a SymFloat for a native float argument. Registration and immutable scalar binding now happen in EdgeRadialCuda.__init__; the bound call invokes the registered operator directly. Dynamo correctly lifts the cached series tensors, so no extra FakeTensor branch or cache duplication is needed. The regression passes and checks that the original eager cache is retained; the wider CUDA suite passes as well (15 tests and 10 subtests).

OutisLi:
Review and CI fixes are in 1a17b1583f0678d4cc53bc41bbc5e1af3b61235a.

- Python group 7: CI uses dargs 0.5.1, whose repeated-object schema uses additionalProperties instead of the older items layout. Preset schema generation now receives an independent, non-repeating model schema and constructs model_dict directly, avoiding dependence on either repeated-object representation. Reproduced and verified with dargs 0.5.1.
- Python groups 8 and 11: eight parity cases still supplied the removed n_radial keyword to build_edge_cache. The stale test-helper argument is removed. Schema/preset/doc generation and the affected parity selection pass together: 116 passed.
- Fixed radial bases: dpmodel preserves trainable=False for /fix; backend promotion consumes that shared metadata. PT-expt keeps frozen named Parameters and no longer needs separate suffix-specific freezing passes. Original/deserialized fixed bases are checked in JAX and TF2; related JAX/PT-expt/gradient checks pass (30 tests), as do the actual TF2 checks (5 tests).
- CUDA radial compilation: op registration and immutable scalar conversion occur at binding time. Eager-cache-then-fullgraph-compile passes without a new hot-path FakeTensor guard.
- One-term envelopes: Python dispatch and native validation accept edge order 1 and basis orders 0/1. Normal dispatch previously fell back for these configurations. Forward/coordinate-gradient parity covers Bessel and Gaussian bases with orders 0/1/7, including zero distance. This also exposed and fixed the Bessel removable singularity using phi(0)=frequency and phi'(0)=0. The rebuilt CUDA suite passes: 15 tests and 10 subtests.
- Both CodeQL import cycles are broken by moving declarative preset data to a dependency-free module; existing public preset imports remain available.
- The single-task HybridMuon .values() finding is not applicable: ModelWrapper.model is always a ModuleDict, including the Default single-task entry. Single-task and multi-task optimizer smoke checks both pass. The legacy apply_envelope compatibility suggestion is intentionally outside this PR's supported checkpoint scope; no legacy deserializer branch is added. Basis names are already validated by RadialBasis construction, so the suggested duplicate validation is not added.
- The C++ job failed before compilation: archive.ubuntu.com was unreachable, leaving build-essential without an installation candidate. No source change is indicated by that log; this push starts a fresh CI run.

All local pre-commit hooks passed. New GitHub CI results are pending.

OutisLi:
The remaining Python group 11 failures are addressed in 7afe4f6f1fe5ad1120a9a79c587deda801304035. Both failures were in the binding-selection unit test: it mocked op_available=True on the CPU-only build but still performed real native registration. Moving registration to binding initialization exposed this incomplete mock; a CUDA-enabled local installation concealed it. The test now mocks registration alongside availability, retaining all fp32/fp64 selection assertions. Production kernel and dispatch code are unchanged.

Reproduced the exact missing-operator error in an isolated process without loading the local native library: before, 2 failed / 2 passed; after, 4 passed. The normal CPU run passes 10 tests (7 GPU-only skips), and the actual CUDA regression passes 15 tests (2 cuTile cases deselected). All pre-commit hooks pass. The Test CUDA label is being triggered for this commit to run the repository's full GPU CI.

OutisLi:
The DPA4 v20260901 presets are deprecated and removed in favor of v20260911. Given the short interval between these versions, we do not expect downstream adoption of these DPA4 presets and do not consider a compatibility layer necessary. To keep the catalog clean, their definitions are deleted outright, without deprecated aliases or fallback mappings.

DPA4 retains v20260820 and v20260911; DPA4C retains both v20260901 and v20260911. All 24 retained configurations are unchanged. The v20260911 sizes are Nano, Mini, Neo, Air, Plus, Pro, Max and Ultra for DPA4, and Nano, Mini, Neo, Air and Plus for DPA4C.

The model documentation uses compact version/available-size tables, without parameter-by-parameter release descriptions. The DPA4C water example selects dpa4c-nano-v20260911 and supplies only run-specific overrides, so no explicit Neo widths override the preset. Generated JSON schemas and editor completions follow the retained catalog.

Validation: 96 preset/schema/doc-generator/example tests and 219 subtests passed. The retained preset definitions match the original catalog exactly; both documentation tables match the registry. The official VS Code JSON language service validates the example and offers 24 preset completions.

deepmd/dpmodel/descriptor/dpa4_nn/radial.py
```diff
@@ -588,7 +619,6 @@ def deserialize(cls, data: dict[str, Any]) -> RadialBasis:
             n_radial=int(config["n_radial"]),
             basis_type=str(config.get("basis_type", "bessel")),
             exponent=int(config.get("exponent", 7)),
```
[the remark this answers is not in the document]
OutisLi: This repeats the legacy serialized-dictionary compatibility question already discussed in this PR. The numerical difference for a dictionary containing `apply_envelope: false` is real, but compatibility with those old descriptor dictionaries and `.pte` / `.pt2` artifacts is explicitly outside the agreed scope.

I checked the `.pt` checkpoint loading paths: they reconstruct the model from `model_params` and then load its `state_dict`, rather than feeding that old radial config through `RadialBasis.deserialize`. The DPA4C constructor still selects the raw radial basis (`exponent=0`). Following the agreed compatibility boundary, no legacy flag migration is added here.

OutisLi:
Addressed in 0ad6b492e261ee18fbfaba583a70b88893133000.

- Empty-edge grid paths use explicit channel/frame dimensions in PT and dpmodel/PT-expt, retaining the full isolated-atom block computation. Matching reshape fixes cover the accelerated activation adapters and first-/second-order references.
- Expanded validation exposed two pre-existing fp64 precision losses: the segmented-softmax caller downcast fp64 inputs to fp32, and the standalone gated-activation reference also used unconditional fp32 casts. Softmax now binds only for fp32 compute; the activation reference preserves fp64. The fp32 GPU kernels are unchanged. The original strict GPU training-gradient comparison and magnetic-force finite differences pass without relaxed tolerances.
- The embedding compile test uses the shared runtime allowlist (PyTorch 2.11/2.12/2.13). The fused-value-path test constructs its dense reference independently of ambient acceleration gates.

Final local validation on PyTorch 2.13.0+cu130: 14 PT compile/embedding tests; 80 PT-expt export/model tests (including real CPU/CUDA AOTI packages); 134 kernel/training-path tests; 37 PT-expt trainer/descriptor/accelerated tests; 122 CPU grid/descriptor regressions; 8 GPU empty-edge regressions; and 17 precision derivative/binding checks all passed. The 30-test PT export suite also passed. Sixteen symbolic grid checks each reuse one compiled graph for edge counts 7, 9, 11, 17, 32 and 396; zero edges use one additional graph and preserve output/gradient parity. CUDA/Triton-enabled runs use all four inference/training gates set to 1, with separate gate-off controls. All pre-commit hooks passed.

deepmd/pt/model/descriptor/sezm.py
```diff
@@ -1421,26 +1428,25 @@ def forward(
...
+        # frame holds other edges.
         with nvtx_range("blocks"):
             x = x.to(dtype=self.dtype)  # (N, D, 1, C)
             if force_embedding is not None:
                 x = x + force_embedding.to(dtype=self.dtype)
-            if self.blocks and edge_cache.src.numel() > 0:
+            if self.blocks:
                 edge_cache = edge_cache_to_dtype(edge_cache, self.dtype)
                 with self._compute_mode_ctx(extended_coord.device):
                     x = self._forward_blocks(x, edge_cache, rad_feat_per_block)
```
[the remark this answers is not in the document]
OutisLi: Fixed the shape invariant in the grid transformations. Frame/channel splitting and packing use explicit dimensions, so all interaction blocks still run for isolated atoms. The same changes are present in PT and the dpmodel implementation used by PT-expt. CUDA/Triton-enabled checks also exposed the same zero-element reshape ambiguity in the gated-activation adapters and the SO2/radial first- and second-order references; those use explicit trailing dimensions as well. No empty-edge model bypass or shape-dependent fallback is added.

Regression coverage includes all three reported configurations, the cutoff limit, empty NeighborGraph versus fully masked neighbors, and separate inference-force / force-training second-order paths. A symbolic Inductor check of 16 PT/PT-expt grid configurations confirms one compiled graph across edge counts 7, 9, 11, 17, 32 and 396, including an edge count equal to an internal coefficient dimension; an additional graph for the zero-sized axis also reproduces eager outputs and gradients.

deepmd/dpmodel/descriptor/dpa4_nn/radial.py
```diff
@@ -588,7 +619,6 @@ def deserialize(cls, data: dict[str, Any]) -> RadialBasis:
...
-            apply_envelope=bool(config.get("apply_envelope", True)),
             precision=precision,
```
wanghan-iapcm, line 622: This is a silent behaviour change for existing serialized DPA4C models. On master, `DescrptDPA4C` built its basis with `exponent=5, apply_envelope=False`, so every serialized dict (including the v3.2.0 release) contains `{"exponent": 5, "apply_envelope": false}`. This `deserialize` now ignores the key and constructs `C3CutoffEnvelope(exponent=5)`, and `call()` multiplies it in. `@version` is still 1 and `check_version_compatibility(version, 1, 1)` is unchanged, so nothing raises.

Reproduced with a `RadialBasis(rcut=6, n_radial=4, exponent=5, apply_envelope=False)` serialized on master and reloaded here, evaluated at r = 5.9:

```
master : [ 8.87e-03 -1.77e-02  2.65e-02 -3.52e-02]
this PR: [ 4.54e-08 -9.07e-08  1.36e-07 -1.80e-07]
```

Please bump `@version` to 2 and, when `version == 1`, map `config.get("apply_envelope", True) is False` to `exponent=0` before constructing the object. Same treatment in any other backend that reads this dict.

deepmd/pt/model/model/spin_model.py
```diff
@@ -375,6 +375,10 @@ def expand_aparam(aparam: torch.Tensor, nloc: int) -> torch.Tensor:
         return aparam
 
     @torch.jit.export
+    def adam_route_patterns(self) -> list[str]:
```
wanghan-iapcm, line 377: The new method landed between the existing `@torch.jit.export` and `get_type_map`, so the decorator now applies to `adam_route_patterns` and `get_type_map` lost it. Scripting still works (I checked with an se_e2_a spin model: `torch.jit.script` succeeds and `get_type_map` is still callable because the exported `get_ntypes` reaches it), so this is not a functional break, but the intent is clearly the other way round: nothing in TorchScript needs `adam_route_patterns`, and `get_type_map` is what the C++ side calls by name. Please move the decorator back. Note that the spin TorchScript test in `test_sezm_spin_model.py` mocks `torch.jit.script`, so it would not have caught a real scripting failure here.

wanghan-iapcm (changes requested):
Thanks, the CUDA/export fixes and the fixed-basis variants look right, and the CI matrix is green. One blocking point on serialization, two small non-blocking ones inline.

**Blocking:** `RadialBasis` drops `apply_envelope` from its serialized dict without a `@version` bump or a legacy mapping (see inline). Released v3.2.0 DPA4C models carry `exponent: 5, apply_envelope: false`, and at this HEAD they reload with the envelope applied, silently.

Non-blocking, in the body since the lines are spread over three files: the new `@pytest.mark.parametrize` lines for `env_exp` (`source/tests/pt_expt/descriptor/test_dpa4_accelerated.py`, `source/tests/pt_expt/model/test_dpa4_export.py`) and `options` (`test_empty_edge_grid_paths` in `source/tests/pt_expt/descriptor/test_dpa4.py`) are missing the trailing comment that the pt_expt tests use to explain the parameter.

deepmd/dpmodel/descriptor/dpa4_nn/radial.py
OutisLi: The numerical difference for a legacy component dictionary is real, but this is the same compatibility scenario excluded in the [earlier clarification](https://github.com/deepmodeling/deepmd-kit/pull/6016#discussion_r3995446974) and the [PR summary](https://github.com/deepmodeling/deepmd-kit/pull/6016#issuecomment-5636907733). That scope remains intentional.

The supported `.pt` checkpoint loading paths reconstruct the model from `model_params` and load its `state_dict`; they do not load an old radial dictionary from the checkpoint through this deserializer. The DPA4C constructor continues to select the raw radial basis with `exponent=0`.

This PR does not provide compatibility for legacy descriptor/component dictionaries or old `.pte` / `.pt2` artifacts. No legacy `apply_envelope` migration or serialization-version change is added for that excluded path. A failure in the supported `.pt` checkpoint loading path would be a distinct issue to fix.

deepmd/dpmodel/descriptor/dpa4_nn/radial.py
```diff
@@ -588,7 +619,6 @@ def deserialize(cls, data: dict[str, Any]) -> RadialBasis:
             n_radial=int(config["n_radial"]),
```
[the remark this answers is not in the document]
OutisLi: This repeats the legacy-dictionary compatibility scenario addressed in the [previous reply](https://github.com/deepmodeling/deepmd-kit/pull/6016#discussion_r3995446974). The numerical observation is valid, but compatibility with old descriptor/component dictionaries and `.pte` / `.pt2` artifacts is outside this PR's intended scope.

The supported `.pt` checkpoint path rebuilds from `model_params` and loads `state_dict`; DPA4C still constructs a raw radial basis with `exponent=0`. That loading path does not consume the legacy radial dictionary used in this reproduction. The stated compatibility boundary is unchanged, so no legacy flag migration or format-version scaffolding is added.

deepmd/pt/model/model/spin_model.py
OutisLi: Fixed in 476d8683c3714bb3402c525fcf674f66e341c882. `@torch.jit.export` is attached to `SpinModel.get_type_map`, while `adam_route_patterns` remains an ordinary Python optimizer-routing method. A real TorchScript script/save/load regression checks the restored type-map APIs and that optimizer routing is not exported. All 33 tests across the energy spin, virtual-type spin, and SeZM spin suites pass; the normal pre-commit hooks pass as well.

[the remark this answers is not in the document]
OutisLi: Fixed in 476d8683c3714bb3402c525fcf674f66e341c882. `@torch.jit.export` is attached to `SpinModel.get_type_map`, while `adam_route_patterns` remains an ordinary Python optimizer-routing method. A real TorchScript script/save/load regression checks the restored type-map APIs and that optimizer routing is not exported. All 33 tests across the energy spin, virtual-type spin, and SeZM spin suites pass; the normal pre-commit hooks pass as well.

deepmd/pt/train/training.py
```diff
@@ -1106,6 +1109,9 @@ def update_finetune_bias(
                     "enable_gram": bool(self.opt_param.get("enable_gram")),
                     "flash_muon": bool(self.opt_param.get("flash_muon")),
                     "magma_muon": bool(self.opt_param.get("magma_muon")),
+                    "adam_patterns": adam_route_patterns(
+                        self._get_inner_module().model.values()
+                    ),
```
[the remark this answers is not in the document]
OutisLi: This path is not live. The PT trainer imports `JIT` from `deepmd.pt.utils.env`, where it is a constant `False` and is never set to `True` anywhere in the tree. The `if JIT:` script/save branch therefore never runs during HybridMuon setup, and `adam_route_patterns()` is collected from the eager modules.

The spin regression is about the freeze/export ScriptModule after `jit.save`/`jit.load`: optimizer routing is intentionally not `@torch.jit.export`ed, and the restored inference artifact must not expose it. That is a separate contract from trainer grouping. `pt_expt` also collects the patterns from eager `self.models` before any scripting.

`use_compile` (`torch.compile`) is unrelated and does not drop the Python method. If trainer JIT is ever enabled, the patterns should be stored before `torch.jit.script`; that is not a current training path, so no grouping change or JIT+HybridMuon trainer regression is added.

deepmd/dpmodel/descriptor/dpa4_nn/radial.py
wanghan-iapcm, line 626: Reopening this. The `.pt` state_dict path is not the only way in: `DescrptDPA4C.deserialize`, `.dp` save/load and `dp convert-backend` all reach this function, and released v3.2.0 DPA4C dicts carry `exponent: 5, apply_envelope: false`.

Serializing `RadialBasis(rcut=6, n_radial=4, exponent=5, apply_envelope=False, precision="float64")` on the base commit and reloading it with this code, at r=5.9:

```
base: [[ 0.0088705  -0.01771669  0.02651432 -0.03523927]]
head: [[ 4.54e-08 -9.07e-08  1.36e-07 -1.80e-07]]
```

`@version` is still 1 and `check_version_compatibility(version, 1, 1)` is unchanged, so nothing raises; the model just evaluates to something else. Two lines fix it: bump `@version` to 2, and for version 1 map `apply_envelope is False` to `exponent=0`. A test that feeds a version-1 dict with `apply_envelope: false` through `deserialize` and checks `envelope is None` would keep it from regressing.

deepmd/pt_expt/descriptor/dpa4.py
```diff
@@ -159,14 +159,13 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
 
 def _promote_trainable(module: torch.nn.Module, names: tuple[str, ...]) -> None:
     """Re-register the given float buffers of *module* as Parameters."""
-    if not getattr(module, "trainable", True):
-        return
+    trainable = bool(getattr(module, "trainable", True))
```
wanghan-iapcm, line 162: The `if not getattr(module, "trainable", True): return` guard that #5522 put here is gone, so a frozen module's buffer is now promoted to `nn.Parameter(buf, requires_grad=False)` instead of staying a buffer. It then shows up in `named_parameters()`, and `training.py` builds the optimizer with `optimizer_class(self.wrapper.parameters(), ...)` unfiltered, so frozen tensors land in `param_groups` and, under FSDP2, in the sharded set. `gaussian/fix` in the new DPA4C preset makes this the default path.

The new `test_fixed_basis_types_freeze_the_basis` asserts the frozen entry appears in `named_parameters()`, so I take it this is deliberate (identical parameter names between `/fix` and trainable variants). If so, a one-line comment here saying why would help, and the optimizer param list should probably filter on `requires_grad`. Non-blocking.

wanghan-iapcm (dismissed):
Thanks for the fixes: the `@torch.jit.export` placement is correct now, and the new zero-edge tests do fail on the base code (`cannot reshape tensor of 0 elements into shape [0, 7, 1, -1]`) and pass here, so the export fix is genuinely covered.

**Blocking (reopened inline):** `RadialBasis.deserialize` still reloads released DPA4C dicts with the envelope applied, without a version bump. Numbers inline.

**Non-blocking, inline:** the `_promote_trainable` guard change and the ZBL/linear `adam_route_patterns` no-op.

**Non-blocking, here since the lines are outside the diff hunks:**

- `descrpt_dpa4c_args`' `basis_type` doc at https://github.com/deepmodeling/deepmd-kit/blob/aaa15b314ce8e35fc5cac1de29dda1a68088c79a/deepmd/utils/argcheck.py#L492-L499 still reads "`bessel` or `gaussian`", but the new `dpa4c-*-v20260911` preset ships `basis_type: "gaussian/fix"`. Only `descrpt_se_zm_args`' doc was updated, so `train-input.rst` will document the default preset's own value as unsupported.
- `adam_route_patterns` resolves off `atomic_model.descriptor` in https://github.com/deepmodeling/deepmd-kit/blob/aaa15b314ce8e35fc5cac1de29dda1a68088c79a/deepmd/dpmodel/model/dp_model.py#L22-L41, but `LinearEnergyAtomicModel` (hence `DPZBLModel`) keeps its sub-models in `.models`, so DPA4 under ZBL gets an empty pattern list and trains without the AdamW routing. `hybrid_muon.py`'s `getattr(model, "adam_route_patterns", list)()` means there is no error either. Either walk `atomic_model.models` or decline explicitly with a log line.
- The new `@pytest.mark.parametrize` lines are still missing the trailing explanatory comment used across the pt_expt tests: `test_dpa4.py` (`options`), `test_dpa4_accelerated.py` (`env_exp`, both places), `test_dpa4_export.py` (`env_exp`).

deepmd/dpmodel/descriptor/dpa4_nn/radial.py
OutisLi: You are right that `.dp` loading and `dp convert-backend` can reach this deserializer. The distinction here is between released software that can produce that representation and published model artifacts that we need to continue supporting: we have not published DPA4C `.dp` models using this legacy representation.

As clarified in the [earlier reply](https://github.com/deepmodeling/deepmd-kit/pull/6016#discussion_r3995446974), compatibility with those old intermediate descriptor dictionaries is outside this PR's intended scope. The `.pt` checkpoint loading path rebuilds from `model_params` and loads `state_dict`; it does not consume the old radial dictionary from a `.dp` file. DPA4C still constructs its raw radial basis with `exponent=0`.

Someone who independently exported an old `.dp` file could encounter the difference you describe, but supporting that old representation is not part of the compatibility commitment here. We therefore do not plan to add an `apply_envelope` migration or format-version handling for it. This is an explicit scope decision, not a claim that the deserialization entry points do not exist or that `.dp` has the same portability constraints as `.pt2`.

wanghan-iapcm (commented):
Thanks, the `RadialBasis` `@version` 1 to 2 bump with the legacy `apply_envelope: false` to `exponent = 0` mapping is exactly right. I reproduced it against a dict serialized by the v3.2.0 code: exact agreement on both bases (`max|ref - got| = 0.0`), and both new tests fail on the pre-fix sources. Threads 1 and 2 are closed from my side, and I have dismissed the earlier changes-requested review.

One factual note on the scope argument in https://github.com/deepmodeling/deepmd-kit/pull/6016#discussion_r4042977929, recorded so the next `@version` bump does not re-run it: the risk was never limited to published `.dp` artefacts. `serialize_from_file()` followed by `BaseModel.deserialize(data["model"])` is how `--finetune` and `dp compress` consume a `.pt2` on the pt_expt backend (https://github.com/deepmodeling/deepmd-kit/blob/aa41c0b663f613c392dc47a235970eb9b89918ed/deepmd/pt_expt/train/training.py#L2087-L2089 and https://github.com/deepmodeling/deepmd-kit/blob/aa41c0b663f613c392dc47a235970eb9b89918ed/deepmd/pt_expt/entrypoints/compress.py#L45-L46, both on master as well). v3.2.0 shipped `dpa4c` registered with `apply_envelope=False` and a `serialize_hook` that reads `model.json` out of a `.pt2`, so any DPA4C `.pt2` a user trained on v3.2.0 carries the legacy dict and would have been re-read through this path with a silently different envelope. Models users trained with released software count. Moot now that the mapping is in.

Remaining, none of it on the serialization question: one correctness gap in the new optimizer routing and the preset-version removal (inline), plus two items from my 09-17 pass that are still open:

- `descrpt_dpa4c_args` still has `doc="DPA4 radial basis type: \`bessel\` or \`gaussian\`."` (https://github.com/deepmodeling/deepmd-kit/blob/aa41c0b663f613c392dc47a235970eb9b89918ed/deepmd/utils/argcheck.py#L498). This PR makes dpa4c's `RadialBasis` accept `bessel/fix` and `gaussian/fix`, and the new `v20260911` preset ships a `/fix` value, so `train-input.rst` and the generated JSON schema advertise the wrong set for dpa4c. The parallel doc in `descrpt_se_zm_args` was updated; this one was missed.
- The trailing comments on the new `@pytest.mark.parametrize` lines are still missing on six lines (inline in `test_dpa4.py`).

Related to the zero-edge sweep, not a regression: with `edge_cartesian=True` a zero-edge frame still raises `ValueError: cannot reshape array of size 0 into shape (0,newaxis)` from https://github.com/deepmodeling/deepmd-kit/blob/aa41c0b663f613c392dc47a235970eb9b89918ed/deepmd/dpmodel/descriptor/dpa4_nn/cartesian.py#L511, the same `-1`-on-a-zero-leading-axis idiom this PR rewrote in `grid_net.py` and `activation.py`. It fails on master too, but it is the one sibling the sweep missed, and `test_empty_edge_grid_paths` does not parametrize `edge_cartesian`.

I would hold off merging until the CUDA jobs report: `Test C++ on CUDA` and `Test Python on CUDA` are still pending at this head, and 235 of the accelerated tests skip on CPU, so the AOTInductor stride and layout fixes that are the other half of this PR are currently verified nowhere.

deepmd/pt/model/model/model.py
```diff
@@ -128,3 +128,23 @@ def get_min_nbor_dist(self) -> float | None:
...
+        -------
+        list[str]
+            Substrings of parameter names, composed from the tensors the
+            descriptor declares through its own ``adam_route_patterns``: rows
+            of these matrices that correspond to rarely visited inputs receive
+            almost no gradient, and Adam moves each row with its own gradient
+            history, whereas Muon's orthogonalized update moves every row of a
+            matrix at the same rate.
+        """
+        descriptor = getattr(getattr(self, "atomic_model", None), "descriptor", None)
```
wanghan-iapcm: `adam_route_patterns()` resolves `getattr(getattr(self, "atomic_model", None), "descriptor", None)`, but `LinearEnergyAtomicModel` / `DPZBLLinearEnergyAtomicModel` hold their sub-models in `self.models` and have no `.descriptor`. For a ZBL-bridged or linear-composite DPA4 the chain resolves to `None` and this returns `[]`, so the radial-input matrices silently stay on the Muon path: the exact routing this PR adds is a no-op there, with no error and no symptom except worse convergence. `SpinModel` got an explicit delegate at `spin_model.py:377`; the linear/ZBL wrapper did not. The same applies to a DPA4 nested inside a `hybrid` descriptor, where the hard-coded `f"descriptor.{p}"` prefix would not match `...descriptor.descriptor_list.0....`.

Both fall out of one root cause: the capability is probed by name with a default instead of being declared. A concrete default `adam_route_patterns() -> []` on `make_base_descriptor` / `make_base_model`, overridden by the composite wrappers to union over `self.models` with the right prefix, fixes all three and restores the contract #5967 established, where a typo raises instead of degrading to `[]`. Please add a test with a non-`DPModelCommon` model; nothing currently covers this path. Same shape in `deepmd/dpmodel/model/dp_model.py:36`.

deepmd/utils/model_preset_data.py
```diff
@@ -0,0 +1,255 @@
...
+            "radial_so2_rank": 4,
+            "n_focus": 3,
+            "message_node_so3": False,
+            "ffn_so3_grid": False,
+            "grid_branch": 0,
+        },
+    },
+}
+# Descriptor options and available grades for each version.
+_DPA4_VERSIONS: dict[str, dict[str, Any]] = {
```
wanghan-iapcm, line 138: `_DPA4_VERSIONS` replaces `"v20260901"` with `"v20260911"` rather than adding it, while `_DPA4C_VERSIONS` below keeps `v20260901` and adds `v20260911`. `dpa4-*-v20260901` is on master today (`deepmd/utils/model_preset.py:206`, from the merged #6009), so after this PR any input using `dpa4-<grade>-v20260901` fails with `Unknown model preset`. Verified by execution: it resolves on the merge base and raises at this head, while `dpa4c-mini-v20260901` still resolves in both.

This also drops the sentence "Existing versions are never edited." from the module docstring, which is the policy the change violates. If retiring `v20260901` is deliberate, please keep the entry and deprecate it, or say so explicitly in the PR description and keep the policy line honest. Non-blocking since #6009 is in no release tag yet.

source/tests/pt_expt/descriptor/test_dpa4.py
```diff
@@ -56,6 +59,56 @@ def setup_method(self) -> None:
         TestCaseSingleFrameWithNlist.setUp(self)
         self.device = env.DEVICE
 
+    @pytest.mark.parametrize(
```
wanghan-iapcm, line 62: Trailing comment explaining the parameter is still missing here on `"options"` (62) and `"training"` (70), and in `test_dpa4_accelerated.py` on lines 89, 93, 244 and 245. Lines 112-113 and 408 in this same file have them, so this is an oversight. Raised on 09-17, only partly done.

deepmd/utils/model_preset_data.py
OutisLi: The retirement is deliberate, as noted in the earlier [preset decision](https://github.com/deepmodeling/deepmd-kit/pull/6016#issuecomment-5643228942). I have made it explicit in the PR description: the unreleased `dpa4-*-v20260901` names are removed without aliases, and inputs using them must select a retained preset explicitly. DPA4 keeps `v20260820` and `v20260911`; DPA4C keeps `v20260901` and `v20260911`. The removed DPA4 entries are not restored.

source/tests/pt_expt/descriptor/test_dpa4.py
OutisLi: Completed all six parameter comments in 97924bf09190e72b62f732f7569d862cbbdf46bc: edge-local configurations, inference versus force training, dtype/binding eligibility, single versus double envelopes, inference backend, and default versus single envelope. The empty-edge test also includes Cartesian lmax=1/2 cases. Related CPU tests and all 31 GPU empty-edge/parity and acceleration tests pass; normal pre-commit hooks pass.

OutisLi:
Addressed the follow-up review in 97924bf09190e72b62f732f7569d862cbbdf46bc.

- AdamW routing is a declared capability on the shared descriptor, atomic-model and model bases. Wrappers delegate explicitly and prefix the real child paths (`atomic_model`, `models.<index>`, `descrpt_list.<index>` and `backbone_model`); the collector no longer silently accepts a missing method. PT and dpmodel/PT-expt implementations agree. The regression covers real plain, hybrid, nested-linear, ZBL, virtual-spin and ordinary-descriptor models in both backends, including PT's non-`DPModelCommon` linear model. All 12 cases pass; routed matrix updates match `torch.optim.AdamW`, while an unrelated matrix stays on Muon. These methods only run during optimizer setup.
- Cartesian edge contractions use their fixed nine Cartesian components, avoiding the ambiguous `(0, -1)` reshape without bypassing isolated-atom computation. The existing empty-edge suite now also covers `lmax=1/2`, graph-versus-masked-neighbor/NumPy parity, inference force backward and force-training double backward.
- DPA4C `basis_type` documentation includes both `/fix` forms. All six reported parametrization comments are present.
- The deliberate removal of unreleased DPA4 `v20260901` presets is explicit in the PR description; DPA4C `v20260901` remains available.

Validation: related CPU suites passed in two runs (141 passed, 18 skipped and 5 subtests; 122 passed). The final 12-case routing regression also passed. GPU validation passed all 14 empty-edge/parity tests and all 17 fp32 acceleration tests. Normal pre-commit hooks, Ruff checks and `git diff --check` passed. CUDA CI is being triggered on the new head.

njzjz (approved):

source/tests/pt_expt/descriptor/test_dpa4_train_paths.py
```diff
@@ -119,8 +121,9 @@ def test_cuda_train_gate_accepts_shared_truthy_values(monkeypatch, value: str) -
     ids=("training", "inference"),
 )
 @pytest.mark.parametrize("enabled", [0, 1])
+@pytest.mark.parametrize("precision", ["float32", "float64"])
```
wanghan-iapcm, line 124: Missing the trailing comment explaining the parameter, the same point as the six lines fixed in this commit. `precision` here selects the dtype the gated stage binds under.

source/tests/pt_expt/model/test_dpa4_export.py
```diff
@@ -168,7 +168,10 @@ def test_dpa4_fp32_cpu_export_runs_without_cuda_only_ops(monkeypatch) -> None:
 
 
 @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is required")
-def test_dpa4_fp32_cuda_export_runs_with_fast_ops(monkeypatch) -> None:
+@pytest.mark.parametrize("env_exp", [None, 5])
```
wanghan-iapcm, line 171: Same as above: this line is added by this PR and has no trailing comment. `env_exp` is default double vs single envelope.

source/tests/pt_expt/model/test_dpa4_export.py
```diff
@@ -294,10 +298,15 @@ def test_dpa4_triton_force_assembly_survives_cpu_trace(monkeypatch) -> None:
         pytest.param(torch.device("cuda"), id="cuda"),
     ],
 )
+@pytest.mark.parametrize(
+    "env_exp",
+    [pytest.param(None, id="double_envelope"), pytest.param(5, id="single_envelope")],
+)
```
wanghan-iapcm, line 304: And this block. The `target_device` block just above it is in the same state. A trailing comment after the closing paren, as in `test_dpa4_interop.py`, is the pattern used elsewhere in this PR.

wanghan-iapcm (commented):
Re-reviewed on head `97924bf09190e72b62f732f7569d862cbbdf46bc`. My three open threads are resolved, verified against the file at head rather than against the replies:

- AdamW routing on composite models: the `atomic_model.descriptor` lookup is gone; routing is now a composed prefix chain with concrete `[]` defaults at all three roots, and I confirmed the attribute names it builds (`models` ModuleList, `descrpt_list` ModuleList, `backbone_model`) really exist.
- `dpa4-*-v20260901`: I checked independently — `git tag --contains 97cd2a33f` is empty and the newest tag `v3.2.0` predates #6009, so the name never shipped in a release. Accepted as a master-only retirement.
- The six parametrize trailing comments I named are done.

Both fixes in this commit have regression tests that I ran against pre-fix code and against head:

- `test_adam_routing` — 5 failed / 7 passed pre-fix (`hybrid`, `linear`, `zbl` layouts, `assert set() == {...}`), 12 passed at head.
- `test_empty_edge_paths` — 4 failed / 6 passed pre-fix (`RuntimeError: cannot reshape tensor of 0 elements into shape [0, -1] ... ambiguous`, exactly the two `edge_cartesian` parametrizations x both `training` values), 10 passed at head.

CI: 40 checks, the only non-pass entries are `deploy_pypi_index` and `upload_pypi`, both release-publication jobs that skip on every PR. `Test Python on CUDA` and `Test C++ on CUDA` both pass, so the GPU paths did execute.

One suggestion inline on the new parametrize lines; nothing blocking.

PR #6019, opened 2026-09-12
feat: add the Uni-Mol v1 backbone and its self-supervised pretraining

iProzd:
# PR title

feat: add the Uni-Mol v1 backbone and its self-supervised pretraining

# PR description

Uni-Mol is a molecular representation model: a transformer over all atom pairs
in which geometry enters only through pairwise distances. It was pretrained on
about 209 million RDKit conformers with three self-supervised objectives and no
energies or forces at all. This adds a port of Uni-Mol v1 that is faithful
enough to load the released weights and reproduce the published objective.

Two things motivate it. Uni-Mol's data and objectives become available to
multi-task training alongside DFT-labelled data, which is what makes a
controlled comparison between the two kinds of supervision possible at all.
And molecular property work gains a pretrained backbone with a large user base
behind it.

## Scope

Uni-Mol is **not** a potential energy surface model. It attends over every atom
pair with no cut-off and no smooth envelope, so it is not extensive, it does not
support periodic boundaries, and its forces are neither smooth nor conserved.
The descriptor rejects frames carrying periodic images, declares itself
unavailable for edge-parallel and communication paths, and is not offered for
molecular dynamics or frozen deployment.

Nothing existing changes behaviour. Every new component is reachable only by
name from a configuration, `gelu` and `gelu_tf` keep their current meaning, the
new data hook defaults to off, and no new dependency is added: PyTorch is
imported lazily and only to read a checkpoint file, and RDKit only when the
offline converter is asked for two-dimensional conformers.

## What is here

- **Backbone** (`deepmd/dpmodel/descriptor/unimol.py`, `unimol_nn/`): the
  15-layer pre-layer-norm encoder, self-attention that returns its pre-softmax
  logits so the pair representation accumulates across layers, the Gaussian
  distance basis with per-element-pair affine parameters, and both norm
  regularisers.
- **Heads and objective** (`fitting/unimol_pretrain.py`, `loss/unimol.py`):
  element prediction, coordinate denoising through the pair channel, pairwise
  distance prediction, with upstream's weights of 1, 5, 10, 0.01 and 0.01.
- **Data** (`dpmodel/utils/unimol_transform.py`, `utils/unimol_data.py`): the
  masking and noise pipeline as plain per-frame functions, a per-frame
  transform hook on the LMDB reader, and a streaming converter for the
  upstream dataset.
- **Weights** (`utils/unimol_checkpoint.py`): imports the released
  `mol_pre_all_h_220816` and `mol_pre_no_h_220816` checkpoints.
- **Exact GELU**: `gelu_erf` registered in every backend activation table.
  Uni-Mol uses the error-function form; deepmd's `gelu` is the tanh
  approximation, which differs by up to 4.7e-4 per element.
- **Training**: the objective declares the transform it needs and the trainer
  installs it on that task's datasets, next to where it already registers the
  label requirements. Supervised losses declare nothing and their data path is
  untouched.
- Model, atomic model, argcheck entries, PyTorch-Exportable wrappers,
  documentation and an example configuration.

A run from the shipped example trains: `dp --pt-expt train` reports all five
terms on both the training and validation curves and writes checkpoints.

## How closely it matches upstream

Every component is checked against tensors dumped from upstream Uni-Mol
(commit `90f52c4`) running unmodified on the same molecules. The golden archive
ships with the tests and the header of `source/tests/common/dpmodel/test_unimol.py`
says how to regenerate it.

| What | Agreement | Limited by |
| --- | --- | --- |
| data transforms: tokens, targets, edge types, both coordinate arrays | bitwise | nothing |
| data transforms: distance matrix | 1.9e-6 absolute | fp32 rounding of scipy vs a sqrt of squares |
| encoder, fed upstream's own attention bias, 15 layers | 1.2e-15 relative | fp64 rounding |
| three heads | 5e-16 relative | fp64 rounding |
| whole objective on the released weights | 3.4e-7 relative | the three fp32 choices below |

Bitwise agreement on the transforms is the part worth pausing on: it means the
random stream itself is reproduced, down to which atoms are masked and what
noise each one receives, not merely that the statistics match.

The remaining 3.4e-7 is upstream's own use of fp32 in three places, not an
implementation difference:

1. the Gaussian basis is evaluated in fp32, reproduced by default and
   switchable with `single_precision_basis`;
2. the distance matrix is precomputed in fp32 by upstream's data pipeline,
   while the descriptor computes distances inside the model, which is more
   accurate and is what gradients flow through; `single_precision_distance`
   reproduces upstream's numbers instead, which is what the released-weight check
   uses to reach 3.7e-7;
3. `log_softmax` and both norm regularisers are evaluated in fp32, reproduced.

Training trajectories cannot be matched exactly in any case: upstream
pretrained a pure fp16 model with fused kernels and its own Adam variant.

## Scope of the checks

Beyond the parity tests, three whole-path checks were run, and each found real
defects that component tests had not:

- Driving the model through the PyTorch-Exportable backend found arrays built
  without a device, which land on the host while the batch is on the
  accelerator.
- A short training run found the same class of defect in the loss and the
  fitting, and confirmed the objective actually falls: 8.48 to 3.02 over 60
  steps, with every term decreasing.
- Running `dp --pt-expt train` from a configuration file found five gaps
  between a file and the first step: the trainer's loss factory did not know
  the objective; the fitting lacked the accessors the atomic model calls on
  any fitting; the transform ran after the reader had already checked for the
  labels it was about to produce; the converter wrote a zero cell, which the
  neighbour-list builder inverted; and the example addressed its LMDB dataset
  with a list rather than a string.

## What an adversarial review of this branch found

The branch was reviewed before submission by independent passes over upstream
fidelity, interface compliance, edge cases, test quality and reviewability,
with every finding put to a separate attempt at refutation. Thirty-two survived
and are fixed here. The ones worth knowing about:

- The token embedding and all four Gaussian basis tables never received a
  gradient. They were bare arrays, which this backend turns into buffers, and
  then, once they were parameters, the array-API wrapper that placed them on
  the device copied them out of the autograd graph. Inference and checkpoint
  parity were unaffected, which is why the parity tests stayed green
  throughout.
- Every module was handed the same seed, so all fifteen encoder blocks started
  bitwise identical whenever a seed was set.
- The corruption was frozen at the first epoch, so every pass masked each
  molecule identically.
- Cropping inside the per-frame transform left a frame inconsistent with the
  batch layout, which is settled before the transform runs.

The tests that should have caught the first two could not fail: the gradient
test accepted any parameter with a gradient, which the three heads alone
satisfied. Those tests are strengthened rather than merely repaired.

## Decisions a reviewer may want to question

- **Real atoms are identified from the neighbour list, not from `atype`.** By
  the time a descriptor is called, virtual atoms have been clamped to type 0
  and are indistinguishable from a real first element. Frames with fewer than
  two real atoms are rejected, since that inference is ambiguous for them, and
  the converter drops such molecules.
- **A second entry point returns token-resolution output.** The descriptor
  five-tuple cannot carry the two virtual tokens, the pair channel or the norm
  regularisers that the heads read, so the atomic model overrides one method
  rather than any component being forked.
- **The two regularisers are broadcast over the local atoms.** They are frame
  scalars, but only per-atom variables survive the atomic-output machinery; the
  loss averages them back with the real-atom mask, which returns the original
  value exactly.
- **The distance head keeps the virtual columns** and is padded to
  `max_atoms + 2`, because upstream's objective counts them, and a static shape
  is what the output definition needs.
- **The distance target is derived in the loss** rather than stored. Storing it
  would cost O(natoms^2) per frame, which is impractical at 209 million
  conformers.
- **The legacy `numpy.random` interface is used deliberately** in the
  transforms, with a noqa and a reason on every call: upstream seeds the global
  legacy generator, and a `Generator` would draw a different stream.
- **The loss base class gains an optional `frame_transform`.** A
  self-supervised objective has to corrupt its input as the data is read, and
  this is the smallest way to say so without the trainer special-casing a
  particular loss. Supervised losses inherit the default and are unaffected.

## Third-party code

The ported code follows Uni-Mol (commit `90f52c4`) and the Uni-Core modules it
builds on (commit `ace6fae`), both MIT licensed, Copyright (c) DP Technology.
Parts of Uni-Core derive in turn from fairseq, Copyright (c) Facebook, Inc. and
its affiliates, also MIT licensed. Each ported file carries its provenance in
the header, naming the upstream file and commit for every class.

## Tests

All of it runs under the repository's own gate: `pre-commit` passes on every
changed file.

`source/tests/common/dpmodel/test_unimol.py`,
`source/tests/common/dpmodel/test_unimol_data.py` and
`source/tests/pt_expt/model/test_unimol.py`: 28 tests covering the transforms,
the encoder, the basis, the heads, the descriptor, the objective, the
registered model path, a training run driven from a configuration, agreement
between the array-API and PyTorch-Exportable implementations, gradient flow,
dropout behaviour, serialization round trips, the guards, the data conversion
and the reader hook.

Tolerances have stated causes rather than being tuned until they pass. One test
exists only to measure the fp32 basis gap between NumPy and Torch, so that the
looser bounds elsewhere have a number behind them.

## Not in this PR

Training the objective on DPA descriptors in multi-task, which needs a
coordinate head over the equivariant features and a new pair readout, and the
removal of the unrelated dead `denoise` code, which is a separate cleanup.

## Review round three

Three files changed since the last review, in response to @wanghan-iapcm's
second pass.

**`deepmd/dpmodel/array_api.py`** — `xp_erf` now has a TensorFlow branch.
`xp_erf` is introduced by this pull request, along with the exact-erf GELU it
serves, so this is a defect this pull request introduced rather than a
pre-existing one being tidied up. Every namespace other than JAX and torch fell
through to a NumPy round-trip; for an `ndtensorflow` array that is wrong twice
over. Under `tf.function` the conversion is refused outright, because
`__array__` raises on a graph tensor. In eager mode it succeeds while detaching
the `erf` factor from the tape, so the exact GELU differentiates as though it
were `Phi(x)` alone — silently, and for every backend user of `gelu_erf` rather
than only Uni-Mol.

Removing the branch again makes the point: the gradient test fails with the
derivative collapsed onto `Phi(x)`, the graph-mode test fails, and the
forward-value test still passes. A test that compares values could not have
found this.

**`deepmd/dpmodel/descriptor/unimol.py`** — the docstring for the dropout rates
said they are inert here and applied by the PyTorch-Exportable wrapper. They are
applied by the shared encoder, for torch arrays in training mode; inference is
the identity, and training on another array namespace raises
`NotImplementedError`.

**`source/tests/consistent/test_activation.py`** — three tf2 cases: the value
comparison's gradient counterpart, a `tf.function` trace, and an anchor on the
namespace name.

Two notes on those tests. The gradient probe uses an even number of points so
the grid straddles zero without landing on it, because `relu` and `relu6` have
no derivative there: autodiff reports the subgradient 0 while a central
difference reports 0.5, and neither is wrong. And the namespace anchor is
deliberately not gated on `INSTALLED_TF2` — `find_spec` does not import the
module, so it runs everywhere, including the ordinary runs where the tf2 cases
skip. A guard that skips alongside the thing it guards would protect nothing.

### When the tf2 cases run

They are gated on `INSTALLED_TF2`, which requires `DEEPMD_TEST_TF2=1`. As far as
I can see nothing in the repository sets that variable — `test_python.yml` sets
`DP_TEST_TF2_ONLY` for the dedicated `source/tests/tf2` job — so on a normal run
they skip rather than fail. I ran them locally with `DEEPMD_TEST_TF2=1` and
TensorFlow pinned to the CPU (TF 2.21's bundled kernels do not match this
machine's driver): 36 passed. Flagging it because a test that always skips
protects nothing, and I would rather say so than leave the impression that this
path is covered. Whether to wire these into CI is the maintainers' call.

The TF1 backend is not reachable from this change: `xp_erf` has one caller in
the tree, `deepmd/dpmodel/utils/network.py:361`, and TF1's own `gelu_erf` in
`deepmd/tf/common.py` calls `tf.math.erf` directly without going through it.

## Summary by CodeRabbit

* **New Features**
  * Added Uni-Mol v1 molecular pretraining support across supported backends, including descriptors, models, fitting, loss functions, and data transforms.
  * Added tools to convert Uni-Mol LMDB datasets and import released Uni-Mol checkpoints.
  * Added exact erf-based GELU activation support.
  * Added the `adam_eps` optimizer option for PyTorch Exportable training.

* **Bug Fixes**
  * Improved dataset conversion recovery when replacing an existing dataset.
  * Made Uni-Mol data corruption streams reproducible and independent across datasets.

* **Documentation**
  * Added Uni-Mol usage documentation and a complete pretraining example configuration.


deepmd/dpmodel/descriptor/unimol.py
```diff
@@ -0,0 +1,520 @@
...
+                "the unimol descriptor needs at least two real atoms per frame; "
+                "single-atom frames cannot be told apart from padding"
+            )
+
+        real = xp.astype(real_mask, coord.dtype)
+        # BOS and EOS sit at the centroid of the real atoms, which keeps the
+        # sequence translation invariant. Upstream centres the coordinates in
+        # its data pipeline and then places both at the origin, so the two agree
+        # whenever the data went through that transform.
+        if self.virtual_token_position == "centroid":
```
wanghan-iapcm, line 363: With the default `virtual_token_position="centroid"` (argcheck default, and the shipped `examples/unimol/pretrain/input.json` does not override it), `build_tokens` places BOS/EOS at the centroid of the coordinates it receives, which during pretraining are the corrupted coordinates. `_clean_distances` in `deepmd/dpmodel/loss/unimol.py` (L95-102) puts the virtual tokens at the origin, which is the centroid of the clean coordinates only. On a 6-atom frame with one noised atom the BOS position came out as roughly (-0.07, -0.03, -0.16) Å while the clean centroid is 0, so the two BOS/EOS columns of every masked row's distance target are regressed against the wrong label by an amount of the order of the coordinate noise.

The tests only run with `"origin"`, so the default configuration is not covered. Either compute the target with the same rule as the descriptor, or force `"origin"` on the pretraining path and cover the default in a test.

deepmd/dpmodel/utils/unimol_transform.py
```diff
@@ -0,0 +1,580 @@
...
+            *specials,
+            *(
+                i
+                for i, sym in enumerate(vocabulary)
+                if i not in specials and sym not in type_index
+            ),
+        ]
+        # Identifies this transform's draw sequence within a process, so that
+        # the training and the validation set do not share one.
+        self.stream = int(np.random.SeedSequence().generate_state(1)[0])
```
wanghan-iapcm: `np.random.SeedSequence()` with no argument draws OS entropy, and `_next_epoch(self.stream, self.seed)` mixes this into every per-frame draw. So two runs with the same `seed` and `DP_LMDB_NUM_WORKERS=0` produce different masks (verified: two processes, seed=1, same frame, different selections). The argcheck doc for the seed says "A run is reproducible from it only when one process decodes the data", which this contradicts. Deriving `stream` from `seed` plus a train/validation discriminator would restore that.

deepmd/utils/unimol_data.py
```diff
@@ -0,0 +1,282 @@
...
+    env.close()
+    if frame_idx == 0:
+        shutil.rmtree(staging, ignore_errors=True)
+        raise ValueError(
+            f"{src} produced no usable frame, so there is nothing to train on; "
+            f"{skipped} record(s) were skipped for holding one atom or an "
+            "element outside the type_map"
+        )
+    if os.path.exists(dst):
+        shutil.rmtree(dst)
```
wanghan-iapcm: `rmtree(dst)` followed by `os.rename(staging, dst)` is not atomic: an interruption between the two lines removes the previous dataset and leaves nothing in its place, which is exactly what the comment above the staging directory says this code is meant to avoid. Renaming `dst` aside first (`dst` -> `dst.bak`, `staging` -> `dst`, then remove `dst.bak`) keeps a valid dataset on disk at every step.

wanghan-iapcm (changes requested):
Thanks, this is a careful port: the encoder matches upstream to 1e-11, the loss terms and weights line up, and the transform hook is cleanly isolated from the existing data path. Three blocking points inline (distance target under the default `virtual_token_position`, non-reproducible corruption seed, non-atomic dataset replacement), and three non-blocking notes below.

Non-blocking:

- `deepmd/dpmodel/descriptor/unimol.py` L90-91 documents `max_seq_len` as "kept for configuration compatibility", but `get_rcut()` (L223-225) derives the reported cutoff from it, so it is not inert. Either the docstring or the dependency should change.
- `deepmd/dpmodel/loss/unimol.py` L76-77 `_frame_scalar` divides by `xp.sum(weights)` with no guard; a frame with zero real atoms gives NaN. `_smooth_l1` and `_masked_nll` in the same file already guard the empty case, so this is just for consistency.
- The trainer builds a fresh transform per dataset, so the validation set is re-corrupted on every pass and the validation loss is not comparable across epochs. Worth one sentence in `doc/model/unimol.md`.

deepmd/dpmodel/loss/unimol.py
```diff
@@ -0,0 +1,333 @@
...
+        """Build Uni-Mol's corruption, which also produces the labels.
+
+        A fresh object each time: it carries the counter that stands in for the
+        epoch, and the training and validation sets must not share one.
+        """
+        from deepmd.dpmodel.utils.unimol_transform import (
+            make_unimol_data_transform,
+        )
+
+        return make_unimol_data_transform(
```
wanghan-iapcm, line 293: This factory never passes `stream=`, and `deepmd/pt_expt/train/training.py` (https://github.com/deepmodeling/deepmd-kit/blob/3b48dd1edde77dbe11c4dc1ddbe5bc668b83bab2/deepmd/pt_expt/train/training.py#L1882-L1903) calls it once for the training set and once for the validation set. Both transforms therefore hash `(data_seed, "default")` to the same `self.stream`, and `_next_epoch` keys its cache by `(stream, pid)`, so the two datasets share one generator. Checked at this head: `train.stream == valid.stream` is `True`, and one validation draw changes the next training corruption. That is the situation the docstring here and the comment in `training.py` say must not happen, and it is a regression from the previous head, where the OS-entropy stream kept them apart. Multi-task models with several Uni-Mol tasks collide the same way.

The label has to come from the caller: give `frame_transform` a `stream` argument that the trainer fills with something like `f"{model_key}/training"` and `f"{model_key}/validation"`, and add a test that goes through `UniMolLoss.frame_transform` and the trainer wiring rather than through `make_unimol_data_transform` directly; `test_training_and_validation_do_not_share_a_stream` passes distinct labels by hand, which no production caller does, so it stays green with this bug in place.

deepmd/utils/unimol_data.py
```diff
@@ -0,0 +1,297 @@
...
+            f"{skipped} record(s) were skipped for holding one atom or an "
+            "element outside the type_map"
+        )
+    # Move the old dataset aside rather than deleting it first: between a
+    # delete and a rename there is a moment with nothing at `dst`, which is the
+    # loss this staging directory exists to prevent. A rename is atomic, so
+    # every instant has either the old dataset or the new one in place.
+    previous = f"{dst}.replaced"
+    if os.path.exists(previous):
+        shutil.rmtree(previous)
```
wanghan-iapcm: Two things about the recovery path. If the previous run was interrupted between `os.rename(dst, previous)` and `os.rename(staging, dst)`, the only copy of the dataset is at `dst.replaced` and `dst` does not exist; this run's first action is to `rmtree` it, before the new conversion has succeeded. If the new conversion then fails as well, both copies are gone. When `previous` exists and `dst` does not, rename `previous` back to `dst` instead of deleting it. Separately, between the two renames there is still an instant with nothing at `dst` (njzjz-bot's open point), so the comment above overstates what the sequence guarantees; it should say that no data is lost, not that `dst` is valid at every instant.

None of this is covered: every `convert_unimol_lmdb` call in `source/tests/common/dpmodel/test_unimol_data.py` converts into a fresh `dst`, so the `had_previous` branch, the rollback in the `except`, and the stranded-`.replaced` case have no test.

wanghan-iapcm (changes requested):
The three blocking points from the previous round are addressed (virtual-token position, seed-derived stream, rename-aside in the converter), and the new tests pass at this head. Two problems remain, one of them introduced by the stream change; details inline.

Non-blocking, from the same change to `_next_epoch` (https://github.com/deepmodeling/deepmd-kit/blob/3b48dd1edde77dbe11c4dc1ddbe5bc668b83bab2/deepmd/dpmodel/utils/unimol_transform.py#L318-L343):

- The generator entropy is now `[seed, stream]` only, so every decoder worker (`DP_LMDB_NUM_WORKERS>1`) seeds an identical generator and draws the same epoch sequence at each call; frames still differ through `index`, but the per-worker decorrelation the old `pid` entropy gave is gone. The docstring should state that the reproducibility guarantee holds for a single decoding process and that with workers the draws are shared across them.
- `_EPOCH_STREAMS` is never reset, so "one process decoding a run twice draws the same sequence" holds only for the first run in an interpreter; the new test has to clear the table by hand to simulate a fresh process. Either key the table by a per-transform token, or expose a reset the trainer calls when it installs the transforms.

deepmd/dpmodel/array_api.py
```diff
@@ -516,6 +516,34 @@ def xp_sigmoid(x: Array) -> Array:
...
+        return torch.special.erf(x)
+    from scipy.special import (
+        erf,
+    )
+
+    if array_api_compat.is_numpy_array(x):
+        return erf(x)
+    # array-api-strict and friends: round-trip through NumPy.
+    xp = array_api_compat.array_namespace(x)
+    return xp.asarray(erf(np.asarray(x)), dtype=x.dtype)
```
wanghan-iapcm, line 544: `xp_erf` has native branches only for jax and torch. A tf2 array (`deepmd._vendors.ndtensorflow`) falls through to this line and round-trips through NumPy, which breaks in both modes:

- under `tf.function`: `TypeError: cannot convert a TensorFlow graph tensor to a NumPy array`, so tf2 cannot train or freeze with `gelu_erf`;
- eager with `GradientTape`: the erf factor is detached from the tape, so the gradient is silently wrong. Reproduced at this head: d/dx of the exact GELU at `[0.5, 1.0, -0.3]` comes out as `[0.6915, 0.8413, 0.3821]`, which is Φ(x) alone; the correct values are `[0.8675, 1.0833, 0.2677]`.

`gelu_erf` is added to `VALID_ACTIVATION`, so it is accepted for every descriptor and fitting on every backend, not only Uni-Mol. `source/tests/consistent/test_activation.py` picks it up automatically but compares forward values eagerly only, which is why it passes. A `tf.math.erf` branch for the ndtensorflow namespace, plus a gradient check in that test, would close it.

deepmd/dpmodel/descriptor/unimol.py
```diff
@@ -0,0 +1,526 @@
...
+        Number of attention heads, which is also the width of the pair channel.
+    max_atoms : int
+        Largest molecule accepted, which fixes ``sel``.
+    max_seq_len : int
+        Upstream's sequence guard. Nothing in the forward pass consults it, but
+        :meth:`get_rcut` reports a radius derived from it, so it is not inert.
+    activation_function : str
+        Activation of the blocks and heads. Uni-Mol uses the exact GELU.
+    dropout, emb_dropout, attention_dropout, activation_dropout : float
+        Dropout rates. They are inert in this backend-agnostic implementation
```
wanghan-iapcm, line 96: Both halves of this sentence are false at this head: `unimol_nn/encoder.py:63-79` defines `dropout()`, which calls `torch.nn.functional.dropout` for torch arrays and raises `NotImplementedError` otherwise, and it is called at encoder.py:333, 420, 427 and 428; `deepmd/pt_expt/descriptor/unimol.py` contains no dropout code. Please describe what actually happens (applied here when the array is a torch tensor and the module is in training mode; unsupported for other array namespaces).

wanghan-iapcm (changes requested):
Third pass, on 92ef3ef1.

Both blockers from the previous round are fixed, and the fixes are tested for real: I restored the pre-fix tree (3b48dd1e) and ran the new tests against it. `test_the_trainer_gives_each_dataset_its_own_corruption` fails there (both datasets draw the same stream) and passes at HEAD; `test_a_stranded_backup_survives_a_failed_publish` fails there ("the only copy of the dataset was destroyed") and passes at HEAD. Thanks for routing `stream` through the trainer and for the restore-instead-of-delete publish.

Two new items are inline. One of them (`gelu_erf` on the tf2 backend) is a wrong-gradient bug in code this PR adds to every backend, so I am keeping the request-changes state for that alone.

Non-blocking, take or leave:

- A negative `data_seed` crashes at the first decoded frame: `_next_epoch` feeds `[seed, stream]` to `np.random.default_rng`, which rejects negative entropy, and only the stream label is laundered through `_stream_entropy`. `argcheck` sets no lower bound. https://github.com/deepmodeling/deepmd-kit/blob/92ef3ef1a164035ac7b041041e3c6d9f982497f4/deepmd/dpmodel/utils/unimol_transform.py#L332-L346
- The `None` side of `mask_token_head` / `coord_head` / `dist_head` is exercised by no test (six `is not None` branches). https://github.com/deepmodeling/deepmd-kit/blob/92ef3ef1a164035ac7b041041e3c6d9f982497f4/deepmd/dpmodel/fitting/unimol_pretrain.py#L108-L133
- The new `adam_eps` is not plumbed into the HybridMuon path: the Adam/AdamW branch passes `eps=`, the HybridMuon branch forwards `adam_betas` only, and the optimizer's Adam-style update hard-codes `ADAM_EPS = 1e-20`. https://github.com/deepmodeling/deepmd-kit/blob/92ef3ef1a164035ac7b041041e3c6d9f982497f4/deepmd/pt_expt/train/training.py#L2300-L2314

CI: 54 of 58 checks passed, the 4 others are label-gated skips; the Python and C++ suites ran.

deepmd/dpmodel/loss/unimol.py
```diff
@@ -0,0 +1,320 @@
...
+        self,
+        learning_rate: float,
+        natoms: int,
+        model_dict: dict[str, Array],
+        label_dict: dict[str, Array],
+        mae: bool = False,
+    ) -> tuple[Array, dict[str, Array]]:
+        """Evaluate the five terms and their weighted sum."""
+        del learning_rate, natoms, mae
+        mask = model_dict.get("mask")
```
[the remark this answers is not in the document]
iProzd: `mask` does reach the loss on this head. `BaseAtomicModel._finalize_atomic_ret`
sets it at `deepmd/dpmodel/atomic_model/base_atomic_model.py:640`, on the line
after it zeroes every per-atom output at the padded rows
(`base_atomic_model.py:638`), so the fitting does not declare it. It is present
in `translated_output_def()` and in what `forward` and `forward_lower` return.

The consequence the comment describes — the norm terms averaging over padding —
is now covered by `test_the_norm_regularisers_ignore_the_padding` in
`source/tests/pt_expt/model/test_unimol.py`. It builds a batch whose short frames
are padded with `atype = -1`, then asserts that the padded rows are zero, that
the masked mean equals the scalar every real row carries, and that the unmasked
mean equals that scalar diluted by exactly the real-atom fraction
(`unmasked == masked * n_real / nloc`). On the fixture that is 0.9686 against
0.6239 over 59 rows of which 38 are real.

`_token_mask_from_atoms(None, ...)` is not reached from the shipped path either:
the frame transform emits `unimol_token_mask`, and the loss prefers that label
when it is present.

deepmd/dpmodel/descriptor/unimol.py
```diff
@@ -0,0 +1,520 @@
...
+    def call(
+        self,
+        coord_ext: Array,
+        atype_ext: Array,
+        nlist: Array,
+        mapping: Array | None = None,
+        fparam: Array | None = None,
+        comm_dict: dict | None = None,
+        charge_spin: Array | None = None,
+    ) -> tuple[Array, None, None, None, None]:
```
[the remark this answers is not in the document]
iProzd: The alert is accurate: this signature does narrow the one it overrides.
`NativeOP.call` is declared `call(self, *args, **kwargs)`
(`deepmd/dpmodel/common.py:114` on master), and every concrete implementation
in this tree narrows it to the arguments it actually takes — 35 such
overrides across 17 files under `deepmd/dpmodel/` on master today, none of
them introduced by this pull request.

Matching the base signature here would make this one call site inconsistent
with the rest of the tree and would remove the argument names and types that
make the narrow form useful, so this is left as it is. That is a choice, not
a claim that the alert is wrong.

If the project would rather these alerts were silenced, that is a repository-
wide change and I am happy to do it as a separate pull request.

deepmd/dpmodel/fitting/unimol_pretrain.py
```diff
@@ -0,0 +1,363 @@
...
+                pad = xp.zeros(
+                    (nf, nloc, width - dist.shape[-1]),
+                    dtype=dist.dtype,
+                    device=array_api_compat.device(dist),
+                )
+                dist = xp.concat([dist, pad], axis=-1)
+            out["pair_dist"] = dist
+        return out
+
+    def call(self, descriptor: Array, atype: Array, **kwargs) -> dict[str, Array]:  # noqa: ANN003
```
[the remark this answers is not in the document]
iProzd: The alert is accurate: this signature does narrow the one it overrides.
`NativeOP.call` is declared `call(self, *args, **kwargs)`
(`deepmd/dpmodel/common.py:114` on master), and every concrete implementation
in this tree narrows it to the arguments it actually takes — 35 such
overrides across 17 files under `deepmd/dpmodel/` on master today, none of
them introduced by this pull request.

Matching the base signature here would make this one call site inconsistent
with the rest of the tree and would remove the argument names and types that
make the narrow form useful, so this is left as it is. That is a choice, not
a claim that the alert is wrong.

If the project would rather these alerts were silenced, that is a repository-
wide change and I am happy to do it as a separate pull request.

deepmd/pt_expt/model/unimol_pretrain_model.py
```diff
@@ -0,0 +1,97 @@
...
+    def forward(
+        self,
+        coord: torch.Tensor,
+        atype: torch.Tensor,
+        box: torch.Tensor | None = None,
+        fparam: torch.Tensor | None = None,
+        aparam: torch.Tensor | None = None,
+        do_atomic_virial: bool = False,
+        charge_spin: torch.Tensor | None = None,
+    ) -> dict[str, torch.Tensor]:
```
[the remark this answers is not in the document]
iProzd: The alert is accurate: `forward` and `forward_lower` here take more arguments
than the bases they override.

This is the shape every model wrapper in `deepmd/pt_expt/model/` has — 44
`forward`/`forward_lower`/`forward_lower_exportable` definitions across 11
files on master, all narrowing `torch.nn.Module.forward` to the arguments the
model actually takes. Following that convention is deliberate, so it is left
as it is rather than changed for this one model. Again, a choice rather than
a disagreement with the alert.

If the convention should change, that is a repository-wide decision and I
would be glad to follow it.

deepmd/pt_expt/model/unimol_pretrain_model.py
```diff
@@ -0,0 +1,97 @@
...
+        extended_coord: torch.Tensor,
+        extended_atype: torch.Tensor,
+        nlist: torch.Tensor,
+        mapping: torch.Tensor | None = None,
+        fparam: torch.Tensor | None = None,
+        aparam: torch.Tensor | None = None,
+        do_atomic_virial: bool = False,
+        comm_dict: dict[str, torch.Tensor] | None = None,
+        charge_spin: torch.Tensor | None = None,
+    ) -> dict[str, torch.Tensor]:
```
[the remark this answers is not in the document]
iProzd: The alert is accurate: `forward` and `forward_lower` here take more arguments
than the bases they override.

This is the shape every model wrapper in `deepmd/pt_expt/model/` has — 44
`forward`/`forward_lower`/`forward_lower_exportable` definitions across 11
files on master, all narrowing `torch.nn.Module.forward` to the arguments the
model actually takes. Following that convention is deliberate, so it is left
as it is rather than changed for this one model. Again, a choice rather than
a disagreement with the alert.

If the convention should change, that is a repository-wide decision and I
would be glad to follow it.

deepmd/utils/unimol_data.py
```diff
@@ -0,0 +1,252 @@
...
+    env = lmdb.open(
+        path, subdir=False, readonly=True, lock=False, readahead=False, meminit=False
+    )
+    try:
+        with env.begin() as txn:
+            cursor = txn.cursor()
+            for _, value in cursor:
+                # The upstream records are pickles written by the dataset
+                # authors; only convert files you trust.
+                yield pickle.loads(value)
```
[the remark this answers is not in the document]
iProzd: The risk is real and the classification is right: `pickle.loads` on a file
named by the caller will execute whatever that file contains.

The input format is not ours to choose. Uni-Mol's released datasets are
pickled Python objects inside LMDB, and this converter exists to read exactly
those artifacts; there is no alternative encoding for the files it must
consume. What is in place is an explicit trust boundary rather than a
mitigation: the function carries a `.. warning::` saying to convert only
files from a source you trust (`deepmd/utils/unimol_data.py:56`), the same
point is in `doc/model/unimol.md:114`, and the deepmd dataset the converter
writes is msgpack, which does not carry the risk forward.

So this is a decision to document the boundary rather than to narrow it, not
a position that the finding is invalid. A restricted unpickler limited to the
types these records hold would be a genuine narrowing; I left it out because
it turns a documented boundary into an implied guarantee, and there is no
publisher-side digest for these files to verify against. If the maintainers
would prefer the allowlist, or a checksum option, I will add it.

wanghan-iapcm (approved):
Approving at a309c884. The three threads from my earlier reviews are addressed at this head, checked against the head tree rather than against the replies.

- `xp_erf` now has a native TensorFlow branch (`deepmd/dpmodel/array_api.py`), returning `tf.math.erf` on the underlying tensor re-wrapped in the same namespace, ahead of the SciPy round-trip. Two new cases in `source/tests/consistent/test_activation.py` cover the gradient through `tf.GradientTape` and traceability under `tf.function`. I restored `array_api.py` from 92ef3ef1a with the tests at head: exactly those two fail for `gelu_erf` and the forward-value case still passes, so the regression test does expose the defect.
- The dropout docstring on `DescrptUniMol` now matches `unimol_nn/encoder.py::dropout`.
- The `virtual_token_position` guard, the removal of `SeedSequence()` entropy, the staged `.partial`/`.replaced` conversion and the per-split `stream=` all stand.

I also fetched upstream at the pinned commit 90f52c4 to check the distance-loss column mask: `cal_dist_loss` selects columns with `src_tokens.ne(padding_idx)`, a position mask with the diagonal included, which is what this port does. The comment in `loss/unimol.py` is accurate.

Three notes that do not block, all about sync points or documentation rather than the code itself:

- `virtual_token_position` defaults to `"centroid"` in argcheck (`deepmd/utils/argcheck.py:2879`), but `UniMolPretrainAtomicModel` refuses anything other than `"origin"` (`deepmd/dpmodel/atomic_model/unimol_atomic_model.py:52-61`), and neither the argcheck doc nor `doc/model/unimol.md` says so. A user who omits the key, as the docs allow, gets the `ValueError` from the guard. Either make `"origin"` the default or state the requirement in the doc string; the test added in 4f43a3327 pins the current mismatch, so it would need updating with the change.
- `examples/unimol/pretrain/input.json` is not listed in the `input_files` tuple of `source/tests/common/test_examples.py`, so the only example exercising the ~300 new argcheck lines is never run through `normalize()`. It does pass when run by hand.
- `doc/credits.rst` has one citation block per descriptor family and gains none for Uni-Mol, and `CITATIONS.bib` is untouched, although the code is a port of an externally published, MIT-licensed project.

One more, smaller: `gelu_erf` is not in `ACTIVATION_TO_FUNCTYPE` in `deepmd/utils/tabulate_math.py` or the parallel table in `deepmd/tf/utils/tabulate.py`, so `dp compress` on a net using it fails with a generic unknown-activation error. It fails loudly, so this is only about the message — but it is worth a comment recording that compression is deliberately unsupported, because mapping it onto functype 2 would silently pick up the tanh derivatives.

CI note: both CUDA jobs report `skipping` at this head, so nothing here ran on GPU. The tf2 shard does select the new activation cases (`-k tf2` with `DP_TEST_TF2_ONLY=1`), so those two did execute.

njzjz (approved):

PR #6022, opened 2026-09-14
feat(stat): isolated-atom energy reference through preset_out_bias and vacuum_ref

OutisLi:
`preset_out_bias` documents that the bias of an assigned atom type is set to the preset value. In `change-by-statistic` mode (fine-tuning, `dp change-bias`, `change_bias_after_training`) the preset was fed to the least-squares fit as if it were a shift, so an assigned type ended at `pretrained_bias + preset`, the residual fitted for the other types subtracted `natoms * preset` instead of `natoms * (preset - pretrained_bias)`, and every further call accumulated the preset again. The four copies of the mode branch (the pt, pd and dpmodel atomic models and the pt model wrapper) shared the defect.

This PR makes the preset the bias of an assigned output in every mode, adds bundled tables of isolated-atom energies so that a table is named instead of typed, and adds the fitting option `vacuum_ref`, which references the network output of every atom to the output the same network gives an isolated atom of its type. With both options the energy of an atom without neighbors is exactly its preset bias, whatever the network parameters are, so the dissociation limit of an energy model is pinned to the isolated-atom energies of the reference calculation.

### `preset_out_bias`

- An output with at least one assigned type is fixed by the preset alone. Every element that occurs in the training data must be assigned (a missing element is an error), the assigned types take the preset value in both `set-by-statistic` and `change-by-statistic`, a type absent from the data gets a zero bias at initialization and keeps its stored bias at fine-tuning, and no statistics are computed for, read from or written to the cache for such an output; its output std keeps its stored value. The types excluded by `atom_exclude_types` (the virtual types of a spin model, for example) need no preset, and elements outside the `type_map` are ignored. Outputs without a preset are fitted as before.
- Four forms are accepted: a dict keyed by element symbol, the name of a bundled table, the path of a JSON file holding such a dict, and a list with one entry per type of the `type_map` (`null` leaves a type unassigned; the form for tensor outputs). A table is resolved once when the input is processed, in single-task configurations, in every branch of a multi-task configuration and for a preset written next to `model_dict`, and its values are stored in the model, so a trained model depends neither on the file nor on the bundled data.
- Bundled tables live in `deepmd/utils/preset_out_bias_tables.json`, one entry per name with its `source` and its `values` keyed by element symbol, so further tables can be added without code changes. The entries are the isolated-atom reference energies of the UMA training tasks published with fairchem (`configs/uma/training_release/element_refs/iso_atom_elem_refs.yaml`, MIT license), in eV: `omat24` (89 elements), `omol25` (83, neutral atoms), `omc25` (94), `odac25` (94) and `oc20` (97). A bundled name takes precedence over a file of the same name.
- All preset handling lives in `deepmd.utils.preset_out_bias` (normalization, table resolution, remapping, validation, per-type rows), shared by the pt, pd and dpmodel atomic models; the per-backend `change_out_bias` bodies reduce to the shared helper plus the fit of the remaining outputs, and `_store_out_stat` writes bias rows without touching the std. `change_type_map` remaps the preset together with the stored bias; an output the model does not produce, a preset on a fitting whose statistics do not distinguish atom types, and an assigned preset on a dipole model are rejected.
- The dpmodel factory (pt_expt, jax) and the pt linear and ZBL builders forward the option; the pt_expt DPA4/SeZM builder accepts it.

### `vacuum_ref` (fitting option, default off)

- The atomic energy becomes `E_i = bias(t_i) + f(x_i; c_i) - f(x_vac(t_i); c_i)`, where `x_vac(t)` is the descriptor of an isolated atom of type `t` computed by the same descriptor with the current parameters and `c_i` is the atom's own conditioning (frame parameters, atomic parameters, case embedding). Without frame or atomic parameters the reference is one row per type; otherwise the reference rows follow the atoms. The reference is evaluated at every training step, so it follows the parameters; forces and virials are unchanged.
- The reference atom is the neutral atom in its ground state: with charge/spin conditioning it carries zero charge and the ground-state multiplicity of the element, and with native spin a spin vector of one Bohr magneton per unpaired electron (`deepmd.utils.vacuum_reference`, from the electron-configuration table); charged, excited or differently magnetized isolated atoms keep the deviation the network learns. The condition tables are built once per type map at the atomic model (pt: buffers of `SeZMAtomicModel`; dpmodel: numpy attributes of `DPAtomicModel`, non-persistent buffers on pt_expt).
- Routes: on pt_expt every graph-native model (DPA1, DPA2, DPA4, DPA4C) carries one isolated node per type through the same forward as single-node frames appended to the flat node axis (`append_isolated_frames`); on the PyTorch backend the SeZM edge route appends the reference nodes in `forward_with_edges`. The dense route evaluates single-atom frames.
- Freezing removes the reference atoms from every exported model: the vacuum descriptor of every type is evaluated once on the export object and folded into the fitting bias, or, for a fitting with frame or atomic parameters whose reference varies between atoms, stored in the fitting as a deployment constant (`vacuum_table`, kept out of checkpoints and serialization) from which the fitting evaluates its references. The pt_expt export folds inside `_trace_and_export_impl`, so `dp freeze`, `.pte`/`.pt2` conversion and `change-bias` on frozen artifacts all resolve the reference on the object they export, and the archive keeps the live model. A SeZM checkpoint in `dens` mode is rejected by the freeze: the DeNS head serves training alone and is not exported.
- The fused pt_expt fitting operator and the fused energy/force route take the reference through the per-type bias (`graph_fitting(fit, descriptor, atype, atom_bias)`); a fitting whose reference varies between atoms is served by the autograd route.
- Scope: `vacuum_ref` is available for DPA4/SeZM on the PyTorch backend and for the graph-native models on pt_expt. The TensorFlow and Paddle fittings reject a serialized `vacuum_ref: true` model with `NotImplementedError` instead of ignoring the option. `atom_ener` is unchanged on every backend; a fitting rejects the two options together.
- Documentation: the section "Isolated-atom energy reference" of `doc/model/train-energy.md`, the argcheck entries `preset_out_bias` and `vacuum_ref`, and the examples `examples/water/dpa4/input_e0.json` (single-task) and `examples/water/dpa4/input_multitask_e0.json` (per-branch presets).

## Breaking changes

- Reusing output-statistics caches requires a per-output preset record matching the current constraints. Existing caches without that record are recomputed in update mode; read-only users must regenerate the output statistics once with `stat_file_mode: update`. This cache validation change does not alter model or checkpoint serialization.
- A partial preset is no longer fitted around: if an output assigns any type, every element observed in the data must be assigned, otherwise the statistics raise. Previously the unassigned observed types were fitted with least squares around the preset.
- With `set-by-statistic`, the types of an assigned output that do not occur in the data get a zero bias instead of a fitted one, and the output std of such an output keeps its initial value; preset outputs never read or write the output-statistics cache.
- `DescrptSeZM.forward_with_edges` takes `vacuum_conditions: dict | None` (the reference-atom inputs) and returns `(descriptor, latent, vacuum)`; `deepmd.pt_expt.kernels.graph_fitting.graph_fitting` takes the per-type bias explicitly.
- The fitting serialization of every backend includes `vacuum_ref`, and its `@version` is bumped (general fitting 4 to 5, polarizability 5 to 6, property 6 to 7, population 4 to 5); older dictionaries without the key load with `vacuum_ref: false`, and the TensorFlow and Paddle fittings emit `false` and reject `true` behind the version check. `compute_stats_do_not_distinguish_types` loses its unused `assigned_bias` parameter.

## Validation

- Preset semantics: `source/tests/common/test_preset_out_bias.py` (forms, bundled tables, resolution in single-task and multi-task configurations, rows, excluded and unknown types), the pt, dpmodel and pd `test_atomic_model_global_stat.py` (`set` → `change` → `change` with an assigned output, the rejection of an unassigned observed element, the output std under a full preset, a spin model with a dict preset, presets that need no data, `change_type_map`), the jax end-to-end test and the argcheck example inputs.
- Reference identity `E_i = bias + f(x_i) - f(x_vac)` and "an isolated atom gives exactly its bias", in float64 to 1e-10: `source/tests/pt/model/test_fitting_vacuum_ref.py` (the full `(fparam, aparam, case_embd, aparam_as_mask) × mixed_types` product, `torch.jit.script`, `default_fparam`, fold and stored table, state-dict round trip, `change_type_map`), `source/tests/common/dpmodel/test_fitting_invar_fitting.py`, `test_fitting_call_graph.py`, `test_vacuum_ref_model.py` (se_e2_a dense and DPA1 graph routes), `source/tests/pt_expt/fitting/test_dpa4_ener.py`. `source/tests/tf/test_fitting_vacuum_ref.py` checks that the TensorFlow fittings reject a serialized `vacuum_ref` and load a dictionary of the previous version; the Paddle, dpmodel and PyTorch serialization tests do the same.
- Reference rows through the descriptors: `source/tests/pt/model/test_sezm_vacuum_ref.py` (plain, charge/spin conditioning with native spin, DeNS, forces unchanged, fold, fused training kernels under AMP), `source/tests/pt/model/test_sezm_parallel.py` (reference rows under the LAMMPS-style communication path), `source/tests/pt_expt/model/test_dpa4_vacuum_ref.py` (graph route with charge/spin and native spin, a padded node axis, dense and graph routes agree to 1e-12, `.pt2` and `.pte` freezes with and without frame parameters, the archive keeps the live model), `source/tests/pt_expt/model/test_fused_vacuum_ref.py` (fused energy/force route of DPA1 and compressed DPA4C equals autograd exactly, autograd fallback with frame parameters), `source/tests/pt/model/test_sezm_vacuum_freeze.py` (`.pt2` freezes on CPU and CUDA targets, with and without frame parameters, a `dens` checkpoint is rejected).
- End to end: the two example inputs train (single-task, multi-task, two-process DDP); a model frozen with `dp --pt freeze` gives an isolated O and H exactly their preset energies. Paddle is not installed here and the TensorFlow operators are not built; the pd and tf tests mirror pt and are verified by CI.

## Summary by CodeRabbit

* **New Features**
  * Added flexible per-element output-bias presets, including bundled tables, JSON files, and element-mapped values.
  * Added isolated-atom (`vacuum_ref`) energy references for supported energy models, including folding references during export.
  * Added vacuum-reference support for DPA4/SeZM and native-spin workflows.
  * Added support for varying graph frame sizes and isolated reference frames.

* **Bug Fixes**
  * Preset biases now validate observed types and remain fixed during statistical adjustments.
  * State-dependent statistics no longer incorrectly reuse cached results.
  * Improved model serialization and export consistency for bias and vacuum-reference settings.


deepmd/dpmodel/atomic_model/base_atomic_model.py
```diff
@@ -803,40 +811,47 @@ def change_out_bias(
...
         else:
             raise RuntimeError("Unknown bias_adjust_mode mode: " + bias_adjust_mode)
+        out_bias, out_std = compute_output_stats(
+            sample_merged,
+            self.get_ntypes(),
+            keys=self.bias_keys,
+            stat_file_path=stat_file_path,
+            model_forward=forward,
+            rcond=self.rcond,
+            preset_bias=preset_bias,
```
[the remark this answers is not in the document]
OutisLi: Fixed in 687b5107d958abd2c1cf82c4714aae9f94fa1c2c across the dpmodel, PyTorch, and Paddle output-statistics implementations.

Calls with `preset_bias` or `model_forward` now bypass both cache restoration and cache writes. This prevents an old absolute bias from being used as a residual, prevents a changed preset from being ignored, and prevents state-dependent results from contaminating the ordinary label-only cache. Label-only cache reuse is retained.

The regression compares cached and uncached calls while changing the preset or predictor, verifies that state-dependent calls do not create an output cache, and verifies the original label-only cache is unchanged and reusable without sampling. Local dpmodel/PyTorch runs passed; Paddle is not installed locally, so its mirrored implementation was linted and byte-compiled and remains subject to CI.

deepmd/pt/model/atomic_model/base_atomic_model.py
```diff
@@ -622,41 +635,56 @@ def change_out_bias(
...
         else:
             raise RuntimeError("Unknown bias_adjust_mode mode: " + bias_adjust_mode)
+        out_bias, out_std = compute_output_stats(
+            sample_merged,
+            self.get_ntypes(),
+            keys=self.bias_keys,
+            stat_file_path=stat_file_path,
+            model_forward=forward,
+            rcond=self.rcond,
+            preset_bias=preset_bias,
```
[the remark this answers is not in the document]
OutisLi: Fixed in 687b5107d958abd2c1cf82c4714aae9f94fa1c2c across the dpmodel, PyTorch, and Paddle output-statistics implementations.

Calls with `preset_bias` or `model_forward` now bypass both cache restoration and cache writes. This prevents an old absolute bias from being used as a residual, prevents a changed preset from being ignored, and prevents state-dependent results from contaminating the ordinary label-only cache. Label-only cache reuse is retained.

The regression compares cached and uncached calls while changing the preset or predictor, verifies that state-dependent calls do not create an output cache, and verifies the original label-only cache is unchanged and reusable without sampling. Local dpmodel/PyTorch runs passed; Paddle is not installed locally, so its mirrored implementation was linted and byte-compiled and remains subject to CI.

deepmd/dpmodel/atomic_model/base_atomic_model.py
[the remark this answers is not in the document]
OutisLi: Fixed in 687b5107d958abd2c1cf82c4714aae9f94fa1c2c across the dpmodel, PyTorch, and Paddle output-statistics implementations.

Calls with `preset_bias` or `model_forward` now bypass both cache restoration and cache writes. This prevents an old absolute bias from being used as a residual, prevents a changed preset from being ignored, and prevents state-dependent results from contaminating the ordinary label-only cache. Label-only cache reuse is retained.

The regression compares cached and uncached calls while changing the preset or predictor, verifies that state-dependent calls do not create an output cache, and verifies the original label-only cache is unchanged and reusable without sampling. Local dpmodel/PyTorch runs passed; Paddle is not installed locally, so its mirrored implementation was linted and byte-compiled and remains subject to CI.

deepmd/pd/model/atomic_model/base_atomic_model.py
```diff
@@ -505,34 +517,43 @@ def change_out_bias(
...
         else:
             raise RuntimeError("Unknown bias_adjust_mode mode: " + bias_adjust_mode)
+        out_bias, out_std = compute_output_stats(
+            sample_merged,
+            self.get_ntypes(),
+            keys=self.bias_keys,
+            stat_file_path=stat_file_path,
+            model_forward=forward,
+            rcond=self.rcond,
+            preset_bias=preset_bias,
```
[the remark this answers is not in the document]
OutisLi: Fixed in 687b5107d958abd2c1cf82c4714aae9f94fa1c2c across the dpmodel, PyTorch, and Paddle output-statistics implementations.

Calls with `preset_bias` or `model_forward` now bypass both cache restoration and cache writes. This prevents an old absolute bias from being used as a residual, prevents a changed preset from being ignored, and prevents state-dependent results from contaminating the ordinary label-only cache. Label-only cache reuse is retained.

The regression compares cached and uncached calls while changing the preset or predictor, verifies that state-dependent calls do not create an output cache, and verifies the original label-only cache is unchanged and reusable without sampling. Local dpmodel/PyTorch runs passed; Paddle is not installed locally, so its mirrored implementation was linted and byte-compiled and remains subject to CI.

OutisLi:
Addressed the review findings in 687b5107d958abd2c1cf82c4714aae9f94fa1c2c:

- Preset-dependent and model-residual output statistics neither read nor write ordinary output-stat caches. Plain label-only statistics still reuse their cache; descriptor/input-stat caching is unaffected.
- Preset normalization rejects all non-finite values, including positive and negative infinity.
- Assigned bias rows are pinned before residual-model prediction. This prevents non-finite stored rows from polluting the predictions and the fitted residual; for finite stored values it is algebraically equivalent to the constrained shift fit. Unassigned rows are unchanged.
- Dipole models explicitly reject assigned presets because their output path does not apply a bias. The misleading dipole example was removed; no dipole forward physics was changed.
- The shared bridging builder forwards the learned configuration's preset to the composition that computes output statistics. Existing PT-expt DPA4 preset rejection and bridging-family restrictions are retained.

Validation:

- 215 CPU statistics/cache/model-construction tests passed (28 optional-backend skips), plus 34 CPU fine-tuning/change-bias/export tests.
- 57 GPU PT/PT-expt tests passed.
- An independent subagent review found no blocking correctness, simplicity, or backend-consistency issues. It independently checked 13 cases, including NaN/+Inf/-Inf checkpoint round trips in dpmodel, PT-expt and JAX, finite-bias equivalence, and the full prediction of a bridged model.
- Normal pre-commit hooks passed. Paddle is not installed in the local PR environment; its source was linted and byte-compiled, but not executed locally.

CUDA CI is being requested for the new commit, along with another review from njzjz.

deepmd/pt/model/task/invar_fitting.py
```diff
@@ -143,13 +140,13 @@ def serialize(self) -> dict:
...
         data["type"] = "invar"
         data["dim_out"] = self.dim_out
-        data["atom_ener"] = self.atom_ener
         return data
 
     @classmethod
     def deserialize(cls, data: dict) -> "GeneralFitting":
         data = data.copy()
         check_version_compatibility(data.pop("@version", 1), 4, 1)
+        data.pop("atom_ener", None)
```
wanghan-iapcm: This drops `atom_ener` on load without any replacement. TensorFlow (https://github.com/deepmodeling/deepmd-kit/blob/8ef385e90eeb98a2f437785ce2b08098e90bffea/deepmd/tf/fit/ener.py#L946-L952) and Paddle still serialize it, so a tf/pd model that relies on it, which is the DPRc workflow documented in `doc/model/dprc.md`, converts to pt/dpmodel with different energies and no warning. The same gap exists for configurations: `atom_ener` passes argcheck (the backend label is documentation only) and then fails in the constructor with a bare `TypeError`. This is also what fails 20 cross-backend consistency tests on this head (`consistent/fitting/test_ener.py`, `consistent/model/test_{ener,dos,dpa1}.py`), because the serialized dicts of tf/pd and pt/dpmodel no longer have the same keys.

Please provide a migration path: translate `atom_ener` to `preset_out_bias` plus `vacuum_ref` (in the config converter and in `deserialize`, with a deprecation warning), or reject it explicitly with a message that names the replacement. Whichever you choose, the serialized key set has to agree across backends again. Note the two obstacles: the new rule rejects partial lists, and `vacuum_ref` is only available for DPA4/SeZM on the PyTorch backend, so a plain-fitting `atom_ener` config has nowhere to go unless one of those is relaxed.

deepmd/dpmodel/fitting/general_fitting.py
```diff
@@ -577,6 +589,7 @@ def serialize(self) -> dict:
             "precision": self.precision,
             "mixed_types": self.mixed_types,
             "exclude_types": self.exclude_types,
+            "vacuum_ref": self.vacuum_ref,
```
wanghan-iapcm, line 598: `vacuum_ref` is a new serialized field, and `@version` stays at 4. Every previous addition to this dict bumped it (2 in #3803, 3 in #4417, 4 in #4888), and the repository rule is to bump the version and add the backward-compatibility branch in `deserialize` in the same commit. As it stands an older release accepts version 4 and then fails in `cls(**data)` with an unexpected keyword. On this head it also fails 6 cases of `consistent/model/test_frozen.py`, whose frozen fixtures do not carry the key. Please bump the version, default the field when it is absent in `deserialize`, and apply the same in the pt, pd and jax fittings; the tf/pd `data.pop("vacuum_ref", False)` should live behind that version check rather than beside it.

deepmd/dpmodel/fitting/general_fitting.py
```diff
@@ -887,5 +1103,14 @@ def call_graph(
         ap1 = None if aparam is None else xp.reshape(aparam, (n, 1, aparam.shape[-1]))
         # fparam: dense API expects (nf, nfp); here nf'=N single-atom frames, so the
         # node-level (N, nfp) IS the per-(pseudo)frame param -- tiled over nloc'=1.
-        ret = self.__call__(d1, a1, gr=g1, g2=g2, h2=h2, fparam=fparam, aparam=ap1)
+        ret = self.__call__(
```
wanghan-iapcm, line 1133: The graph route forwards `vacuum_descriptor` here for every fitting, but only the energy fittings accept it: `DipoleFitting.call` and `PolarFitting.call` do not, and the dense route above guards the keyword with `needs_vacuum_descriptor()`. On this head that fails `consistent/model/test_dipole.py`, `consistent/model/test_polar.py` and both non-energy cases of `pt_expt/model/test_dos_graph.py` with `TypeError: DipoleFitting.call() got an unexpected keyword argument 'vacuum_descriptor'`. Apply the same guard here, or give the base fitting a no-op default for the keyword.

source/tests/pd/model/test_get_model.py
```diff
@@ -73,15 +72,38 @@ def test_model_attr_energy_float(self):
...
+                    [1.0],
+                    [3.0],
+                ]
+            },
+        )
+
+    def test_model_attr_energy_unknown_element(self):
+        model_params = copy.deepcopy(model_se_e2_a)
+        model_params["preset_out_bias"] = {"energy": {"C": 3.0}}
+        with self.assertRaises(ValueError):
```
wanghan-iapcm: This asserts `ValueError` for a key outside the `type_map`, but the shared normalizer ignores such keys (`entries = [spec.get(name) for name in type_map]` in `deepmd/utils/preset_out_bias.py`), the description says they are ignored, and the pt sibling was renamed `test_model_attr_energy_unknown_element_ignored` to assert exactly that. CI confirms it: `ValueError not raised`. Please align this test with the pt one.

deepmd/dpmodel/atomic_model/dp_atomic_model.py
```diff
@@ -334,9 +382,141 @@ def forward_atomic(
...
+                [
+                    xp.broadcast_to(
+                        xp.astype(xp.reshape(charge_spin, (-1, 2)), dtype), (nf, 2)
+                    ),
+                    xp.zeros((1, 2), dtype=dtype, device=device),
+                    xp.asarray(conditions["charge_spin"], dtype=dtype, device=device),
+                ],
+                axis=0,
+            )
+        if spin is not None:
```
wanghan-iapcm: The ground-state spin rows for the reference atoms are appended only when the caller passed `spin`. `vacuum_conditions()` builds the per-type spin table whenever native spin is enabled, and the dense route (https://github.com/deepmodeling/deepmd-kit/blob/8ef385e90eeb98a2f437785ce2b08098e90bffea/deepmd/dpmodel/atomic_model/dp_atomic_model.py#L415-L427) and the pt SeZM path (https://github.com/deepmodeling/deepmd-kit/blob/8ef385e90eeb98a2f437785ce2b08098e90bffea/deepmd/pt/model/atomic_model/sezm_atomic_model.py#L171-L206) always apply it, so with native-spin embedding enabled and `spin` omitted the two routes evaluate the reference atoms under different conditions and the isolated-atom identity no longer holds on the graph route. No test covers `spin=None` here. The reference rows should get their spin regardless of whether the real atoms carried one (zero rows for the real atoms if needed), and a test with `spin=None` on a native-spin model should compare the two routes.

wanghan-iapcm (changes requested):
The bug fix in the first commit is real and the new tests prove it: on master, `test_preset_bias` had the accumulated value hard-coded as its expectation and never called change-by-statistic twice; the new step-3 identity and `test_preset_pinned_in_both_modes` fail on the pre-fix logic (5 of 14 in the pt file, with the bias doubled to `[-20, -27.2]`) and pass at this head. The two features on top need more work before this can go in. The inline comments carry the blocking points; the rest is here.

CI on this head is red: 12 checks fail, 33 tests, all of which trace to the points below.

- 16 cases in `source/tests/consistent/fitting/test_ener.py` and the tf cases of `consistent/model/test_{ener,dos,dpa1}.py` (20 in total) fail on `assert_equal(data1, data2)` with key `atom_ener`: tf and pd still serialize it, pt and dpmodel no longer do. See the inline comment on `invar_fitting.py`.
- 6 cases in `consistent/model/test_frozen.py` fail on key `vacuum_ref`: the frozen fixtures were written without it and the live pt/pt_expt fittings now emit it. See the inline comment on `general_fitting.py`.
- `consistent/model/test_{dipole,polar}.py` and `pt_expt/model/test_dos_graph.py[dipole, polar]` fail with `TypeError: DipoleFitting.call() got an unexpected keyword argument 'vacuum_descriptor'`. See the inline comment on `general_fitting.py` at the graph call.
- `pd/model/test_get_model.py::test_model_attr_energy_unknown_element` fails with `ValueError not raised`. See the inline comment on that test.
- Both cases of the new `source/tests/jax/test_preset_out_bias.py` fail with `preset_out_bias['energy'] does not assign the elements ['O'] that occur in the data`: the PR's own new test uses a partial preset, which the PR's own new rule rejects. Either the test or the rule needs to change.

Non-blocking:

- Partial presets now raise, while the TensorFlow fitting keeps the least-squares-around-preset path (https://github.com/deepmodeling/deepmd-kit/blob/8ef385e90eeb98a2f437785ce2b08098e90bffea/deepmd/tf/fit/ener.py#L330-L342), so the backends disagree on the same input, and `doc/model/dprc.md` still shows `atom_ener: [null, null, 0.0, ...]` under a heading that claims PyTorch support. The declared breaking change is fine; the divergence and the doc should be stated or fixed.
- The only cross-backend consistency case for `atom_ener` (`ener_fitting_case(atom_ener=[-12345.6, None])`) was deleted from `consistent/fitting/test_ener.py` and no `vacuum_ref=True` case replaces it; the consistency framework no longer covers this path at all.
- The description says freezing removes the reference atoms from every exported model. The `.pt2` path and the pt_expt export fold (https://github.com/deepmodeling/deepmd-kit/blob/8ef385e90eeb98a2f437785ce2b08098e90bffea/deepmd/pt/entrypoints/freeze_pt2.py#L989-L994, https://github.com/deepmodeling/deepmd-kit/blob/8ef385e90eeb98a2f437785ce2b08098e90bffea/deepmd/pt_expt/utils/serialization.py#L1831-L1836), but the TorchScript `.pth` freeze in `deepmd/pt/entrypoints/main.py` never calls `fold_vacuum_reference`, so a `.pth` keeps the live reference computation. That is safe, but the text should say which exports fold.
- Fine-tuning after a fold is safe as far as I can see: folding runs only on freshly built export objects, checkpoints are never folded, and the deployment table is not persisted. Worth one sentence in the docs, since the folded fitting has `vacuum_ref=False` and a reader will wonder.
- `argcheck` labels `vacuum_ref` as pt_expt-only under `fitting_ener` while the dpmodel and jax paths implement it; `doc_preset_out_bias` says both that `null` leaves a type unassigned and that every element in the data must be assigned; the `forward_with_edges` docstring in `sezm.py` omits `spin`.
- The `preset_bias`/`assigned_bias` machinery in `{pt,pd,dpmodel}/utils/stat.py` and `utils/out_stat.py` has no caller left outside TensorFlow; two implementations of preset semantics now coexist.

source/tests/pd/model/test_get_model.py
```diff
@@ -73,15 +72,38 @@ def test_model_attr_energy_float(self):
...
+                    None,
+                    [1.0],
+                    [3.0],
+                ]
+            },
+        )
+
+    def test_model_attr_energy_unknown_element(self):
+        model_params = copy.deepcopy(model_se_e2_a)
+        model_params["preset_out_bias"] = {"energy": {"C": 3.0}}
```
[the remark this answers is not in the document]
OutisLi: Fixed in 55bd073e8: the Paddle test now mirrors the PyTorch one and asserts that an element outside `type_map` is ignored while the known element is normalized.

deepmd/pt/model/model/sezm_model.py
```diff
@@ -1565,17 +1566,21 @@ def core_compute(
                 spin=spin,
                 comm_dict=comm_dict,
                 nloc=nloc,
+                vacuum_conditions=self.atomic_model.vacuum_conditions()
+                if fitting_net.needs_vacuum_descriptor()
```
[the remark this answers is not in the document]
OutisLi: Fixed in 55bd073e8. `_sezm_structure_key` now includes `fitting.needs_vacuum_descriptor()`, so a compiled graph traced with reference rows is never shared with a folded model, and `SeZMModel.fold_vacuum_reference` drops the per-instance compile slots of both heads after delegating the fold. `test_fold_drops_the_compiled_graphs` covers it.

deepmd/dpmodel/atomic_model/dp_atomic_model.py
```diff
@@ -334,9 +382,141 @@ def forward_atomic(
...
+            spin = xp.concat(
+                [
+                    xp.reshape(spin, (-1, 3)),
+                    xp.asarray(conditions["spin"], dtype=spin.dtype, device=device),
+                ],
```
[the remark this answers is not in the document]
OutisLi: Checked against the supported flows: a native-spin (`use_spin`) DPA4/SeZM model always receives `spin`. In training the spin label is mandatory for such models (or zero-filled when `allow_missing_label` is set), and the inference evaluators pass it, so the reference rows are always appended next to real rows that carry spin and get the ground-state spin of their type through `vacuum_conditions["spin"]`. A call without `spin` on a native-spin descriptor is outside the supported interface, so the descriptor keeps mirroring the real rows instead of inventing spin values for them.

deepmd/pt/model/task/invar_fitting.py
OutisLi: Thanks. Addressed in 55bd073e8: `atom_ener` is kept on the PyTorch and dpmodel fittings exactly as on master (`InvarFitting`/`EnergyFittingNet`, argcheck default `[]`, serialization), so no migration is needed and the serialized key set agrees across backends again: every backend emits `atom_ener` as before plus `vacuum_ref`. A fitting rejects the two options together with a `ValueError`, and the argcheck/user docs say so. The upstream `consistent/fitting/test_ener.py` cases with `atom_ener` are restored unchanged.

deepmd/dpmodel/fitting/general_fitting.py
OutisLi: Done in 142e79b46, following the convention of #3803, #4417 and #4888: the general fitting `@version` is 5 on dpmodel, pt and pd (JAX inherits the dpmodel serialization), the subclasses that pin their own version follow (polarizability 6, property 7, population 5), and the TensorFlow fittings are bumped the same way. A dictionary of an earlier version has no `vacuum_ref` key and loads with the constructor default `False`, as `dim_case_embd` and `default_fparam` did; a dictionary of the new version is rejected by older releases at the version check. The TensorFlow and Paddle guards now read `version >= 5 and data.pop("vacuum_ref")` after `check_version_compatibility`; on Paddle the check and the guard both live in `InvarFitting.deserialize`. One correction: the `test_frozen` failures on 8ef385e were the `atom_ener`/`vacuum_ref` key mismatch between the backends, not the version; they passed once the key sets agreed, before the bump.

deepmd/dpmodel/fitting/general_fitting.py
OutisLi: Fixed in 55bd073e8: `call_graph` passes `vacuum_descriptor` only when a table is given, the same guard as the dense route, so dipole and polar fittings take the graph route again (`test_dos_graph.py` non-energy cases pass).

source/tests/pd/model/test_get_model.py
OutisLi: Fixed in 55bd073e8: the Paddle test now mirrors the PyTorch one (`test_model_attr_energy_unknown_element_ignored`).

deepmd/dpmodel/atomic_model/dp_atomic_model.py
OutisLi: Checked against the supported flows and left unchanged. A native-spin (`use_spin`) model always receives `spin`: the spin label is a mandatory data requirement in training (or zero-filled under `allow_missing_label`), the spin evaluators pass it at inference, and it is an input of the exported graph. The dense route you compare against has no `spin` input at all, so it is not a route for native-spin models; the routes that serve them (the pt SeZM path and the pt_expt graph route) always carry spin, and their reference rows always get the ground-state spin from `vacuum_conditions()["spin"]`. A `spin=None` call on a native-spin descriptor is outside the interface, so no default is synthesized for it.

deepmd/dpmodel/fitting/general_fitting.py
```diff
@@ -802,6 +1033,8 @@ def _call_common(
             outs = self.nets[()](xx)
             if xx_zeros is not None:
                 outs -= self.nets[()](xx_zeros)
```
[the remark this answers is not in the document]
OutisLi: Not changed here. The mixed-type branch is upstream master's `atom_ener` code restored verbatim (this PR removes nothing from it and adds nothing to it), and its documented contract for `mixed_types` is a single-element list (`remove_vaccum_contribution`: "For `mixed_types` provide `[True]`"), so a partial list there is outside the contract rather than a regression of this PR. Changing that semantics would alter existing PyTorch models trained with it and belongs in its own issue.

deepmd/pt/model/task/sezm_ener.py
```diff
@@ -699,88 +706,56 @@ def _forward_case_film(
...
-        outs = torch.zeros(
-            (nf, nloc, net_dim_out),
-            dtype=self.prec,
-            device=descriptor.device,
-        )
-        results = {}
-
         fitting = self.filter_layers.networks[0]
+        results = {}
         atom_property = fitting(xx, self.case_embd)
```
wanghan-iapcm, line 732: This branch no longer applies `atom_ener`. On master the case-FiLM path built `xx_zeros` and subtracted `fitting(xx_zeros, self.case_embd)` when `remove_vaccum_contribution` was set (https://github.com/deepmodeling/deepmd-kit/blob/master/deepmd/pt/model/task/sezm_ener.py#L708-L770); the previous head removed it, and this round restored the subtraction only in the base `_forward_common` (https://github.com/deepmodeling/deepmd-kit/blob/c1c339e3f8502ea26e648a38bd2c2b1c98a3df98/deepmd/pt/model/task/fitting.py#L995-L1063), which this method bypasses. With `case_film_embd=True` a user-set `atom_ener` is therefore silently a no-op, and `fitting_sezm_ener` in argcheck still offers `atom_ener` (https://github.com/deepmodeling/deepmd-kit/blob/c1c339e3f8502ea26e648a38bd2c2b1c98a3df98/deepmd/utils/argcheck.py#L2960-L2966), so the combination is reachable from a config. Please reinstate the zero-input subtraction here (through the same `conditioning_columns` so the fparam/aparam/case order matches), and add a test with `case_film_embd=True` and a non-empty `atom_ener` that checks an isolated atom gives the preset value.

deepmd/pt/model/atomic_model/sezm_atomic_model.py
```diff
@@ -115,6 +127,95 @@ def __init__(
...
+    def fold_vacuum_reference(self) -> None:
+        """Fold the vacuum reference of the fitting so the forward carries no reference atoms.
+
+        The vacuum descriptor of every type is evaluated once with the current
+        parameters and handed to the fitting, which folds the reference into
+        its bias or stores the table (see its ``fold_vacuum_reference``).
+        """
+        fitting = self.fitting_net
+        if fitting.needs_vacuum_descriptor():
+            fitting.fold_vacuum_reference(self.vacuum_descriptor())
```
wanghan-iapcm, line 227: Only `fitting_net` is folded here. The DeNS head has its own `vacuum_ref` and `needs_vacuum_descriptor()` (https://github.com/deepmodeling/deepmd-kit/blob/c1c339e3f8502ea26e648a38bd2c2b1c98a3df98/deepmd/pt/model/descriptor/sezm_nn/dens.py#L580-L588), so when both heads reference the vacuum the DeNS head keeps asking for reference atoms after the fold, while `SeZMModel.fold_vacuum_reference` (https://github.com/deepmodeling/deepmd-kit/blob/c1c339e3f8502ea26e648a38bd2c2b1c98a3df98/deepmd/pt/model/model/sezm_model.py#L3259-L3266) drops the dens compiled graph as if it had been folded. Please fold `dens_fitting_net` here as well (or skip the dens drop when it was not folded) and cover it with a test that freezes a DeNS model with `vacuum_ref` on both heads.

wanghan-iapcm (changes requested):
The five points from the previous round are addressed at this head: `atom_ener` is back on the PyTorch and dpmodel fittings line for line as on master, with the two options rejected together; the fitting `@version` is 5 in every backend with the subclass offsets kept and the upper bounds updated; the graph route only forwards the vacuum descriptor when it has one; the Paddle test matches the PyTorch one; and the native-spin argument holds, since the inference entry raises without `spin` and the native-spin call makes it mandatory. The jax test now assigns every element, the consistency cases are back, and the SeZM compile cache is keyed on the reference state and dropped on fold. I ran the invar-fitting, global-stat and vacuum-ref test files at this head locally: 38 of 38 pass.

Two new problems remain, both inline. Non-blocking:

- The TensorFlow guards that reject a serialized `vacuum_ref: true` (https://github.com/deepmodeling/deepmd-kit/blob/c1c339e3f8502ea26e648a38bd2c2b1c98a3df98/deepmd/tf/fit/ener.py#L905-L910 and the dos, dipole and polar counterparts) and the Paddle guard have no test, and nothing loads a version-4 dictionary through the version-5 deserializer; one small test per backend would close both branches.
- `argcheck` still labels `vacuum_ref` under `fitting_ener` as pt_expt-only (https://github.com/deepmodeling/deepmd-kit/blob/c1c339e3f8502ea26e648a38bd2c2b1c98a3df98/deepmd/utils/argcheck.py#L2869) while the PyTorch fitting implements it; the entry under `fitting_sezm_ener` carries no label at all.

CI on this head has 14 checks still pending as I write this, including the Python test shards, so I cannot call it green yet.

deepmd/pt/model/task/sezm_ener.py
OutisLi: Fixed in 0fcb0eca8: `_forward_case_film` builds `xx_zeros` beside the atoms through the same `conditioning_columns` and subtracts `fitting(xx_zeros, self.case_embd)` when `remove_vaccum_contribution` is set, as the base forward does; the DeNS energy head inherits it. `test_atom_ener_subtracts_the_zero_descriptor_output` checks, with and without case FiLM, that an atom with a zero descriptor (the isolated atom of the `atom_ener` convention) gives exactly its bias and that a general descriptor gives `f(x) - f(0) + bias`.

deepmd/pt/model/atomic_model/sezm_atomic_model.py
OutisLi: Resolved differently in 0fcb0eca8. The DeNS head serves training alone: `dens` emits a direct-force tensor that no inference consumer takes, and both exportable lowers already raise for a `dens` model. So the DeNS head is not folded; instead `freeze_sezm_to_pt2` rejects a checkpoint in `dens` mode right after loading it, before any fold, and `SeZMModel.fold_vacuum_reference` drops only the energy head's compiled graphs, while the DeNS head keeps its reference and its graph. The docstrings of both folds say so. `test_dens_checkpoint_is_rejected` saves a `dens` checkpoint with `vacuum_ref` on both heads and checks that the freeze raises. The pt_expt backend already rejects a `dens` checkpoint at deserialization.

OutisLi:
Round 3 (0fcb0eca8, fdd413761), on top of a merge with the current master:

- Inline threads: the case-FiLM branch of the SeZM fitting applies `atom_ener` again, and the SeZM freeze rejects a checkpoint in `dens` mode instead of folding the DeNS head, which serves training alone (details in the threads).
- Non-blocking points of the last review: `source/tests/tf/test_fitting_vacuum_ref.py` checks that the four TensorFlow fittings reject a serialized `vacuum_ref: true` and load a dictionary of the previous version; the Paddle `InvarFitting` test does the same, and the dpmodel and PyTorch serialization tests load a version-4 dictionary. `vacuum_ref` under `dpa4_ener` is labeled `(Supported Backend: PyTorch, PyTorch Exportable)`. The label under `ener` stays `PyTorch Exportable`: the dpmodel atomic model evaluates the isolated-atom descriptor for any descriptor, while the PyTorch `DPAtomicModel` does not, so a plain `ener` fitting with `vacuum_ref` on the PyTorch backend raises `vacuum_ref requires the vacuum descriptor of every atom type` at its first forward.
- On the njzjz-bot review: the mixed-type `atom_ener` behaviour is master's and is unchanged here (CodeRabbit withdrew the same finding), and the DeNS head is not exportable, so the freeze rejects it rather than folding it.
- The TensorFlow and Paddle tests are verified by CI only; neither backend is built locally.

njzjz (approved):

deepmd/pt/model/task/invar_fitting.py
```diff
@@ -103,13 +108,23 @@ def __init__(
         seed: int | list[int] | None = None,
         exclude_types: list[int] = [],
         atom_ener: list[torch.Tensor | None] | None = None,
+        vacuum_ref: bool = False,
```
wanghan-iapcm, line 111: **Blocking.** `vacuum_ref` is accepted and serialized here, but the pt backend has nothing to feed it. `deepmd/pt/model/atomic_model/dp_atomic_model.py` has no `vacuum` reference at all, and `resolve_vacuum_ref()` is called only from the dpmodel `DPAtomicModel` and from `pt/model/atomic_model/sezm_atomic_model.py`. So a plain pt `ener` model built with `vacuum_ref: true` keeps `needs_vacuum_descriptor() == True` forever and `GeneralFitting.vacuum_input` raises `ValueError: vacuum_ref requires the vacuum descriptor of every atom type` on every forward. Reproduced with a plain `se_e2_a` + `ener` model: it constructs cleanly and fails at the first forward. The same happens when a version-5 dict with `vacuum_ref: true` is deserialized into pt, where pd and TF both give a clear `NotImplementedError`. pt is the only backend that neither supports the option nor rejects it.

Please either reject `vacuum_ref=True` at construction or deserialize time for atomic models that cannot supply a vacuum descriptor, or route pt `DPAtomicModel` through `resolve_vacuum_ref` as well, and add a test for that combination: `test_fitting_vacuum_ref.py` and `test_sezm_vacuum_ref.py` only go through the fitting class or the SeZM atomic model.

deepmd/dpmodel/utils/stat.py
```diff
@@ -307,6 +289,11 @@ def compute_output_stats(
     assert isinstance(keys, list)
     requested_keys = list(keys)
 
+    # Presets and model residuals depend on state not recorded in the cache.
+    # Neither reuse nor persist these values as ordinary label statistics.
+    if preset_bias is not None or model_forward is not None:
+        stat_file_path = None
```
wanghan-iapcm, line 272: **Blocking.** `stat_file_path = None` when `preset_bias is not None` silently disables the cross-backend stat file for TensorFlow. `deepmd/tf/model/ener.py` (untouched by this PR) passes both `stat_file_path` and `preset_bias={"energy": self.fitting.atom_ener_v}` whenever `atom_ener` is configured, and #5551 routed TF through `compute_output_stats` precisely so it would write and restore that shared file; the comment saying so is still there. With this line a TF deck that sets `atom_ener` neither reads nor writes the energy bias from the stat file, with no warning, so #5551 becomes a no-op for exactly those decks. The line is new in this PR (the merge-base has no `stat_file_path = None` in any of the three `stat.py` files); the same line sits at `deepmd/pt/utils/stat.py:581` and `deepmd/pd/utils/stat.py:393`.

Suggest narrowing the condition to the case it is meant to guard (`model_forward is not None`, the delta-bias path) and handling presets by not consulting the cache for the preset keys only, rather than nulling the path for the whole call.

wanghan-iapcm (changes requested):
All seven threads from the earlier rounds are addressed at this head: `atom_ener` is back on pt and dpmodel and mutually exclusive with `vacuum_ref`, `@version` is 5 in every backend, the graph route only forwards the vacuum descriptor to fittings that accept it, the Paddle test matches, the case-FiLM branch subtracts the zero-descriptor output again, and the DeNS case is handled by rejecting `dens` checkpoints at freeze time. The core preset fix is correct and consistent across the four copies, and I reconstructed the pre-fix behaviour on the merge-base with the same model and sample: the second `change-by-statistic` call doubles the preset (`[-10, -13.6]` to `[-20, -27.2]`), and `test_preset_pinned_in_both_modes` fails there and passes here. CI is green apart from the label-gated CUDA jobs.

Two things introduced in this PR still need fixing before merge; both inline, both narrow, neither touches the core fix.

**Non-blocking, here since they span several files:**

- `_convert_preset_out_bias_to_array` was removed from `deepmd/pt/model/model/__init__.py` (merge-base https://github.com/deepmodeling/deepmd-kit/blob/46fdc3e8b0c92524e85dad48ee00a8555a0a7595/deepmd/pt/model/model/__init__.py#L405) without a construction-time replacement, so a preset list of the wrong length no longer errors when the model is built, only later at stat time.
- `conditioning_columns` (https://github.com/deepmodeling/deepmd-kit/blob/afaba8758035ea2475d10381938d3e9ba3778811/deepmd/pt/model/task/fitting.py#L760) now requires `aparam.numel() == nf * nloc * numb_aparam` with `nloc` taken from the descriptor, where `_forward_common` used to re-derive `nloc` from `aparam`. That is a silent contract tightening with no test on either side of it.
- `data.pop("vacuum_ref")` without a default in `deepmd/pd/model/task/invar_fitting.py` (https://github.com/deepmodeling/deepmd-kit/blob/afaba8758035ea2475d10381938d3e9ba3778811/deepmd/pd/model/task/invar_fitting.py#L151) and the four TF fittings, against the file-local `data.pop(k, None)` convention; a hand-edited version-5 dict raises `KeyError` instead of taking the default.
- `entries = [spec.get(name) for name in type_map]` (https://github.com/deepmodeling/deepmd-kit/blob/afaba8758035ea2475d10381938d3e9ba3778811/deepmd/utils/preset_out_bias.py#L165) silently maps an element missing from a bundled table to `None`; if that element never occurs in the data its bias ends up zero with no message. A warning naming the element would save someone a long debugging session.
- The change that makes a partial `preset_out_bias` (`[null, 0., 1.]`) a hard error is documented and deliberate, but it is a user-visible behaviour change and belongs in the release notes.

deepmd/pt/model/task/invar_fitting.py
OutisLi: Fixed in 58e8c1404b7eeba4d9e54c5056ddfa0620114c72. The check belongs at the atomic-model boundary, since the fitting is valid when its owning model can supply the reference. PT `DPAtomicModel` declares that it does not support `vacuum_ref` and rejects the option during construction with an actionable `NotImplementedError`; `SeZMAtomicModel` explicitly supports it and keeps its existing reference computation. Deserialization goes through the same check. There is no added forward-path check.

The regression builds a real ordinary `se_e2_a` model and exercises both `get_model` and `BaseModel.deserialize`, with and without a preset. All four cases reproduce the missing rejection before the fix and pass afterwards. The supported SeZM vacuum-reference and TorchScript suites also pass.

deepmd/dpmodel/utils/stat.py
OutisLi: Fixed in 58e8c1404b7eeba4d9e54c5056ddfa0620114c72. The dpmodel, PT and Paddle implementations only disable the statistics cache when `model_forward is not None`. Absolute constrained statistics, including the `preset_bias` argument that TensorFlow passes for `atom_ener`, retain their save/load behavior.

The model-level `preset_out_bias` path already handles assigned output keys separately and removes only those keys before calling `compute_output_stats`; no whole-call cache suppression is needed for them.

The cross-backend regression writes and restores constrained absolute statistics without resampling, computes two different model-dependent shifts without reusing or overwriting that cache, and verifies that the shifts create no cache files. It failed before the fix and passes for dpmodel and PT. Paddle is not installed locally, and the TF native backend is not built; the shared function called by TF is covered directly.

OutisLi:
Published 58e8c1404b7eeba4d9e54c5056ddfa0620114c72 for the two blocking findings:

- Ordinary PT atomic models reject unsupported `vacuum_ref` during construction/deserialization. The supported DPA4/SeZM path is unchanged, and the check adds no forward-path overhead.
- Only model-dependent delta statistics bypass the output-statistics cache. Absolute constrained statistics used by TF `atom_ener` can be saved/restored; model-level `preset_out_bias` outputs remain excluded individually from statistical fitting.

For the non-blocking observations: wrong-length preset lists already fail in `BaseAtomicModel.__init__ -> normalize_preset_out_bias` (verified through actual model construction). Valid flat and per-atom `aparam` layouts still agree; an atom-count mismatch already failed at concatenation in the old implementation, and the explicit check diagnoses it earlier. Added a regression with `vacuum_ref` both off and on. The serializers emit the version-5 field, so no extra compatibility path is added for a manually incomplete version-5 record. Unobserved types intentionally remain unassigned, while an observed type missing its preset raises. The partial-preset behavior change is already explicit in the PR description's Breaking changes section.

Validation: the final related CPU run passed 91 tests and 11 subtests (26 skipped for unavailable backends/device conditions); the preceding GPU validation passed 31 tests and 6 subtests, including AMP and CUDA/Triton training. Normal pre-commit hooks, Ruff and `git diff --check` pass. TF native integration and Paddle execution remain for CI because those backends are not built/installed in the local PR environment.

source/tests/consistent/utils/test_stat.py
```diff
@@ -49,6 +59,71 @@
...
+        np.testing.assert_allclose(
+            as_numpy(actual["energy"]), as_numpy(expected["energy"])
+        )
+        fresh_dir = tmp_path / f"delta-{value}"
+        fresh_dir.mkdir()
+        fresh_path = DPPath(str(fresh_dir), "a")
+        compute(sampled, NTYPES, ["energy"], stat_file_path=fresh_path, **kwargs)
+        assert not list(fresh_dir.iterdir())
+
+    restored, _ = compute(no_sampling, NTYPES, ["energy"], stat_file_path=path)
```
wanghan-iapcm: This final assertion is the part I would like changed. `original` was computed with `preset_bias={"energy": [None, [0.25]]}` a few lines above, and this call passes no preset at all, yet asserts it gets that same constrained bias back out of the stat file. That turns "a cached bias silently carries a preset the caller did not ask for" into a tested contract, and the next person who tries to fix the caching hole will have to delete this assertion to do it.

The cache records no fingerprint of the preset, so the two directions are symmetric: a preset-constrained bias is served to a caller with no preset, and a caller with a new preset is served the old cached value without the preset ever being consulted. Please either drop this assertion or invert it, and add the case the guard actually protects: `preset_bias` set together with a populated `stat_file_path`, asserting the preset wins over the cache. That combination is the one no test on either side of `687b5107d`/`58e8c1404` has ever crossed, which is why the over-broad guard reached review in the first place.

deepmd/dpmodel/utils/stat.py
```diff
@@ -307,6 +289,11 @@ def compute_output_stats(
     assert isinstance(keys, list)
     requested_keys = list(keys)
 
+    # Model residuals depend on parameters not recorded in the statistics cache.
+    # Neither reuse nor persist them as absolute output statistics.
+    if model_forward is not None:
```
wanghan-iapcm, line 271: Root cause for the test comment on `source/tests/consistent/utils/test_stat.py:121`. Narrowing this to `model_forward` does fix what I raised — TensorFlow gets its cross-backend stat file back — but all `assigned_bias` / `make_preset_out_bias` handling sits inside the `if bias_atom_e is None:` cache-miss branch below, so on a cache hit `preset_bias` is never read. `deepmd/tf/model/ener.py:251-260` is now the only caller that still passes `preset_bias`, and it passes a live `stat_file_path` alongside it, so a changed `atom_ener` is silently overridden by a stale cache.

Two ways out, either is fine: disable the cache again when `preset_bias is not None` (and then TF loses the shared file, which is what I objected to — so probably not this), or record the preset in the saved statistics and re-apply/validate it after `_restore_from_file`. Whichever you pick, a regression test covering `preset_bias` + an existing stat file would close the gap. Note the same hole exists on master, so this is not a new defect — it is the one the earlier in-branch commit had closed and this commit reopens.

deepmd/dpmodel/atomic_model/dp_atomic_model.py
```diff
@@ -334,9 +386,141 @@ def forward_atomic(
...
+                [
+                    xp.broadcast_to(
+                        xp.astype(xp.reshape(charge_spin, (-1, 2)), dtype), (nf, 2)
+                    ),
+                    xp.zeros((1, 2), dtype=dtype, device=device),
+                    xp.asarray(conditions["charge_spin"], dtype=dtype, device=device),
+                ],
+                axis=0,
+            )
+        if spin is not None:
```
wanghan-iapcm: This branch is guarded on `spin is not None`, but `conditions["spin"]` (line 518) is only populated by `vacuum_conditions()` when the model supports native spin (lines 188-189). `_compute_model_predict` forwards `spin` taken from the sampled data, not from model capability, so a dataset carrying a `spin` field with a non-native-spin `vacuum_ref` model raises a bare `KeyError` from inside the statistics pass. Either guard on `"spin" in conditions` (and ignore the data field with a warning) or raise a message that names the mismatch.

wanghan-iapcm (changes requested):
Re-reviewed at `58e8c140`. Both of my blocking threads from 09-18 are handled at HEAD, and both fixes carry a regression test that I reconstructed and confirmed fails on pre-fix code (3 failures for the stat fix, 1 for the `vacuum_ref` rejection) and passes at HEAD. CI is genuinely green: 57 of 59 checks pass, the two `skipping` ones are the PyPI release jobs, and both CUDA jobs actually ran.

**Thread `deepmd/pt/model/task/invar_fitting.py:111` — resolved.** `DPAtomicModel._supports_vacuum_ref` is checked in `__init__`, which both `get_standard_model` and `deserialize` (via `super().deserialize` -> `cls(**data)`) reach, so every construction path is covered, and the new test exercises both entries with and without a preset. `BaseFitting` declares `vacuum_ref: bool = False`, so fittings outside `GeneralFitting` (e.g. `EnergyFittingNetDirect`) do not break — I checked that by constructing a `direct_force_ener` model at HEAD. `pt_expt` correctly wraps the dpmodel atomic model and is unaffected. `doc/model/train-energy.md` and the two `argcheck` `supported_backends` annotations agree with the guard.

**Thread `deepmd/dpmodel/utils/stat.py:295` — fixed literally, but it re-opens the converse hole.** The narrowed guard does give TensorFlow its cross-backend stat file back, and double-application on the second `change-by-statistic` pass is still prevented (structurally now: `change_out_bias` strips preset keys into Step 1 and never passes `preset_bias` down). But `_restore_from_file` records no fingerprint of the preset, and all `assigned_bias` handling lives inside the cache-miss branch, so on a cache hit the preset is not consulted at all. TensorFlow is now the only caller that still passes `preset_bias` (https://github.com/deepmodeling/deepmd-kit/blob/58e8c1404b7eeba4d9e54c5056ddfa0620114c72/deepmd/tf/model/ener.py#L250-L261), and it passes a live `stat_file_path` with it — so changing `atom_ener` while a stat file exists silently keeps the old bias. This is the exact finding that coderabbit and Copilot raised earlier in this PR and that `687b5107d` had closed. It also exists on master, so the reversal restores master behaviour rather than creating something new; what I would like changed is the new test assertion that locks it in (inline below).

One more point is inline on `dp_atomic_model.py` (the spin branch of `append_vacuum_frames`). Sub-threshold notes, no obligation to act:

- `resolve_vacuum_ref` (https://github.com/deepmodeling/deepmd-kit/blob/58e8c1404b7eeba4d9e54c5056ddfa0620114c72/deepmd/utils/vacuum_reference.py#L122-L126) switches `vacuum_ref` off in place when `preset_out_bias` does not assign that output, with no log line, and `serialize()` then records `false` — so a `true` in `input.json` disappears from the checkpoint silently. Every other unsupported combination in this PR raises. A `log.warning` would be enough.
- A few of the new `pt_expt` test files have bare `@pytest.mark.parametrize` lines without the trailing comment the repository convention asks for: `source/tests/pt_expt/fitting/test_dpa4_ener.py:125`, `source/tests/pt_expt/model/test_dpa4_vacuum_ref.py:273`, `source/tests/pt_expt/model/test_dpa4c_graph_lower.py:242,518,538,672,727`. Worth a pre-commit rule rather than a review round.

Checked and found clean: serialization `@version` bumps and their deserialize branches across dpmodel/pt/pd/tf (GeneralFitting 5, polar 6, property 7, population 5, tf ener/dos/dipole 5, tf polar 6); `atom_ener` parity and the `xx_zeros` subtraction in both dpmodel and pt; `remap_preset_out_bias` index handling; `_convert_preset_out_bias_to_array` has no remaining importer anywhere in the tree; array-API rules in the new dpmodel code; numpydoc sections on the new public helpers; both branches of the new `vacuum_ref` boolean have tests.

Maintainability, for the record rather than as a request. Understanding cost goes up (+7k lines, two new cross-backend concepts: the preset normalisation chain and the train-time/export-time dual form of `vacuum_ref`), partly offset by the preset expansion logic collapsing from one copy per backend into `deepmd/utils/preset_out_bias.py`. Change cost: adding a backend now touches one preset call site instead of a per-backend copy, but the `vacuum_ref` side is net-new at five places (pt capability flag, dpmodel `append_vacuum_frames`, pt_expt node-axis append, tf/pd deserialize rejections). Verification cost is flat: most new behaviour has its own tests, but the TF `atom_ener` + existing stat file path, where the inline finding lives, has none.

My reading of the hardest piece, so you can correct me if it is wrong: `BaseAtomicModel.change_out_bias` now splits outputs into two groups. Outputs named in `preset_out_bias` skip statistics entirely and are written from `preset_out_bias_rows` (with `keep_unassigned=True` under `change-by-statistic`, zeros under `set-by-statistic`); only the remaining outputs go through `compute_output_stats`, and under `change-by-statistic` what is fitted there is the residual against the current model, not an absolute bias. That is what stops the second `change-by-statistic` pass from stacking the preset again, and it is also why `preset_bias` no longer has any caller in dpmodel/pt/pd and only TF still reaches the cache-hit hole.

deepmd/dpmodel/utils/stat.py
OutisLi: Fixed in e25ea1e2f4562a671c3088717f906ed268ac3e9c. The shared statistics-cache reader/writer now stores the normalized preset constraints beside each output's bias and standard deviation. A cache is reused only when those constraints match, including an explicit unconstrained record. Adding, changing or removing a preset causes a fresh fit; matching configurations still use the shared cache, including TensorFlow's `atom_ener` path. PT, dpmodel/TF and Paddle use the same cache implementation.

Simply overwriting the assigned rows is insufficient: in a two-type reproduction changing the second type's preset from 1 to 2 changes the correct bias from [3.8, 1] to [3, 2], so the free row must be refitted too. Bias, standard deviation and preset metadata are saved in one recoverable transaction for both HDF5 and directory caches.

A cache without a preset record cannot establish whether its stored values were constrained. Update mode recomputes it; read-only mode reports the missing/mismatched record and requires a one-time update-mode regeneration. Model and checkpoint serialization are unaffected by this cache change. Regressions cover changed/removed presets, unchanged-cache reuse, output scoping, read-only checks, missing metadata and interrupted metadata writes.

OutisLi:
Published e25ea1e2f4562a671c3088717f906ed268ac3e9c for this review round.

- Cache reuse now validates the saved per-output preset constraints. A changed or removed preset refits the complete output instead of returning stale constrained values or only overwriting fixed rows. The formerly incorrect assertion is replaced with populated-cache transition regressions. One shared implementation serves PT, dpmodel/TF and Paddle, with bias/std/preset written together through the existing recoverable storage transaction.
- Reference spin rows are appended only for active native-spin conditioning. Unused spin on the statistics/atomic graph path no longer raises `KeyError`; the full magnetic-force interface remains unchanged.
- All seven reported parametrization comments are present. The intentional multi-task behavior that disables `vacuum_ref` for an output not fixed by a preset is unchanged; no warning is added.

Output-statistics caches without preset metadata need one update-mode recomputation, since their previous constraints cannot be inferred safely. Read-only use reports missing or mismatched metadata. This affects the statistics cache, not model/checkpoint serialization.

Validation: the broader statistics run passed 116 tests and 5 subtests; the cache/vacuum/selected compiled-graph run passed 91 tests; GPU vacuum, fitting, AMP and training-kernel checks passed 29 tests. The final pre-commit regression rerun passed 62 tests (25 skipped for unavailable backend conditions). Normal pre-commit hooks, Ruff and `git diff --check` pass. TensorFlow native integration and Paddle execution remain for CI because these backends are unavailable in the local PR environment.

wanghan-iapcm (approved):
Re-reviewed at 10a3b47f46a41ea7c6cb2ca8eb484d8154b4346e. All three blocking points from my previous review are resolved.

1. `source/tests/consistent/utils/test_stat.py` — the final assertion no longer freezes the cache hole into a contract. The test now loops over removing, adding and changing a constraint (including two spellings of the same value and an explicitly unconstrained preset) and asserts, for each, that the cached call equals the freshly computed one. The "preset set + populated stat file" case is covered.

2. Preset fingerprint. `deepmd/utils/stat_file.py` now writes a `preset_bias_<key>` record beside each output and reuses the cache only when it matches, including an explicit empty record for an unconstrained output. It is used by all three statistics implementations (`deepmd/dpmodel/utils/stat.py:275`, `deepmd/pt/utils/stat.py:563`, `deepmd/pd/utils/stat.py:366`), and the call site I originally flagged — `deepmd/tf/model/ener.py:251-261`, which passes `preset_bias` and `stat_file_path` together — is now protected. Normalization to a `(ntypes, -1)` float64 array makes equivalent spellings compare equal. I reproduced the pre-fix failures: with the 58e8c140 sources and the head tests, `source/tests/common/test_stat_file.py` gives 7 failed / 21 passed, `source/tests/pt/test_stat_file_mode.py` 4 failed / 14 passed, and the cache test in `source/tests/consistent/utils/test_stat.py` 3 failed; all pass at head.

3. `append_vacuum_frames`. The guard is now `self.add_spin_ebd and spin is not None`, which is the same flag `init_vacuum_conditions` uses to populate `conditions["spin"]`, so the bare `KeyError` is gone. Reproduced: with the one-line revert, `test_non_spin_model_ignores_unused_spin` fails with `KeyError: 'spin'` at `dp_atomic_model.py:518`; it passes at head.

Two points that remain, neither blocking, left here as notes for a follow-up:

- `GeneralFitting.call_graph` forwards `vacuum_descriptor=` to `self.__call__` for every fitting, but `DipoleFitting.call` and `PolarFitting.call` do not declare that parameter. `DPDipoleAtomicModel` rejects an assigned preset, so dipole is covered; `DPPolarAtomicModel` has no such guard, so a polarizability model with `vacuum_ref: true` and an assigned `polarizability` preset reaches `PolarFitting.call(vacuum_descriptor=...)` and raises `TypeError`. Either reject `vacuum_ref` on the tensor fittings or guard the keyword the way the dense route does.

- `vacuum_ref` is a plain `Argument` in `fitting_ener()`, so `vacuum_ref: true` in a TensorFlow or Paddle `input.json` passes argcheck and is then swallowed by `**kwargs` in `deepmd/tf/fit/ener.py`. The `NotImplementedError` guards you added fire only when deserializing an already-serialized model, so a user training on those backends silently gets a model without the reference. A construction-time rejection would close this.

The breaking changes are declared in the PR description, which is what I wanted to see. CI is green, including "Test Python on CUDA" and "Test C++ on CUDA".

PR #6027, opened 2026-09-15
fix(pt): fall back to single-rank .pt2 when the optional with-comm artifact export fails

Fillianore:
## Problem

`dp --pt freeze` for a DPA4/SeZM checkpoint aborts during the export of the **optional** parallel with-comm artifact, producing no model at all — only an unloadable partial `.pt2` that misses `model/extra/metadata.json` (loading it raises `ValueError: Invalid .pt2 file ... missing 'model/extra/metadata.json'`).

Environment where this reproduces reliably:

- deepmd-kit master @ 3a6ca02 (the two flash-bwd sweep bugs are already fixed by #6015; the sweep itself completes)
- torch 2.12.1+cu130 (which pins triton 3.7.1)
- CUDA GPU (NVIDIA RTX PRO 5000 Blackwell)

## Root cause

Inside `_export_with_comm_artifact`, `aoti_compile_and_package` → `aot_export_module` → `detect_fake_mode(flat_args)` raises:

```
AssertionError: fake mode (<FakeTensorMode at 0x...>) from fake tensor input 96
doesn't match mode (<FakeTensorMode at 0x...>) from fake tensor input 161
```

The flattened inputs mix fake tensors from two different `FakeTensorMode` instances: one allocated by the main lower-graph `make_fx` trace (`forward_common_lower_exportable`, run earlier in the same freeze), one by `torch.export`'s `make_fake_inputs` for the with-comm export. This is a torch 2.12.1 behavior change; there is no deepmd-side flag to avoid the with-comm export, since `with_comm` is unconditionally true for the SeZM/DPA4 edge contract.

## The fix

The with-comm artifact is optional by design: the archive format and the loader already handle its absence, and the writer already guards `if with_comm_bytes is not None`. So a failed export should degrade instead of aborting:

- catch the exception around `_export_with_comm_artifact`,
- log a `WARNING` explaining the archive will be single-rank only,
- report `has_comm_artifact=False` in the metadata honestly (`with_comm and with_comm_bytes is not None`).

`dp --pt freeze` then completes and produces a valid single-rank `.pt2`. Multi-rank inference keeps requiring the with-comm artifact, which remains unavailable under torch 2.12 until the FakeTensorMode interaction is addressed on the torch side (happy to open a separate issue with the full traceback if useful).

## Validation

Finetuned DPA4/SeZM checkpoint (STO bulk DFT data, 219-frame test set), torch 2.12.1+cu130, with this patch:

- `dp --pt freeze -c model.ckpt.pt -o frozen` exits 0 and logs `DEEPMD WARNING Parallel with-comm artifact export failed (...); the frozen .pt2 will support single-rank inference only (has_comm_artifact=false)`
- the frozen archive loads and its predictions match the eager checkpoint:

| metric | frozen .pt2 (patched freeze) | checkpoint baseline |
|---|---|---|
| Energy MAE/Natoms | 2.620273e-03 eV | 2.620222e-03 eV |
| Force MAE | 6.365506e-03 eV/Å | 6.365509e-03 eV/Å |
| Virial MAE/Natoms | 2.177323e-03 eV | 2.177272e-03 eV |

Per-frame/per-component comparison against the checkpoint outputs: max abs deviation 4.2e-05 (virial), 1.2e-05 (forces), 9.4e-06 (energies) — AOTInductor float32 compilation noise.

## Summary by CodeRabbit

* **Bug Fixes**
  * Freezing now continues when exporting a parallel-inference artifact fails.
  * The resulting archive is marked appropriately and remains usable for single-rank inference.
  * A warning is recorded to help identify the failed parallel artifact export.


Fillianore:
Thanks for the reviews!

**Regression coverage added in 0024148** (`test_with_comm_export_failure_falls_back_to_single_rank` in `source/tests/pt/model/test_sezm_export.py`, mirroring the existing success-path `test_freeze_embeds_with_comm_artifact`):

- forces `_export_with_comm_artifact` to raise `RuntimeError` through the `freeze_sezm_to_pt2` caller,
- asserts the freeze completes and a `WARNING` announcing the single-rank fallback is logged (proves the with-comm branch was actually entered, not skipped),
- asserts `model/extra/metadata.json` is present, `has_comm_artifact` is `false`, and `model/extra/forward_lower_with_comm.pt2` is absent,
- loads the resulting archive with `aoti_load_package` and runs it single-rank, checking finite `energy_redu` / `energy_derv_r` outputs.

Validated locally on torch 2.12.1+cu130: the new test passes (46s, real CPU AOTInductor compile of the tiny SeZM model — only the with-comm export is mocked) and the sibling success-path test still passes.

Regarding the `action_required` GitHub Actions state: this is the first PR from this fork, so the workflows need maintainer approval to run; I have nothing pending on my side. Please approve the workflow runs if the CI evidence is needed before merge.

PR #6032, opened 2026-09-17
fix(logging): deduplicate distributed training logs

OutisLi:
## Summary

- Add backend-neutral rank-aware logging for torchrun-compatible launchers. Console INFO messages default to once per node; explicit scopes support global or per-rank events, and errors remain visible from every rank.
- Retain emitted records in separate per-rank log files and preserve the parent's logging policy in spawned LMDB scan workers. Workers reopen ordinary log files in append mode without taking ownership of collective MPI-IO handlers.
- Apply the reporting policy to PT and PT-expt startup, parameter counts, sharding, and schedule/checkpoint summaries without changing checkpoint ownership or distributed participation. Keep compile start/finish events visible per rank and report compiled-graph reuse at DEBUG level.

## Validation

- 58 related tests passed across logging, training schedules, checkpoint handling, and common/PT-expt entrypoints, including a real spawned LMDB scan worker.
- Four independent CPU subprocesses with launcher-style rank metadata verified node/global/all console scopes, rank labels, and per-rank file retention.
- All pre-commit hooks passed on the changed files.

Real multi-host/GPU training and MPI-IO integration were not exercised locally.

## Summary by CodeRabbit

- **New Features**
  - Distributed training logs now provide clearer, rank-aware output across nodes and processes.
  - Training workers preserve the configured logging behavior, including per-rank log files where applicable.
  - Build summaries now show total distributed processes and processes running on the current node.

- **Bug Fixes**
  - Reduced duplicate or missing messages during distributed training.
  - Standardized logging for checkpoints, scheduling, training, compilation, and data processing across nodes.
  - Improved visibility of parameter counts, sharding details, and fine-tuning progress.


OutisLi:
Review and CI fixes are in 078cc7cb1c2740e47292267188651266b815d34f.

- PT and PT-expt precompile-start messages use `rank_scope="all"`, matching completion events. This addresses the inline comment and the corresponding outside-diff PT-expt observation.
- The default data-size-based task-probability message uses `is_node_main_process(rank)`, so each node's local chief reports it. Sampling probabilities and checkpoint ownership are unchanged; the existing node-summary regression covers this path.
- Python group 10 failed in two LMDB full-validation tests while serializing a process-pool argument (`TextIOWrapper`). The former worker configuration copied live formatter/filter objects. A formatter holding a terminal stream reproduces that failure locally. Worker transfer now contains only DeePMD-owned declarative settings, and parent/worker handlers use the same factory. External logging integrations and collective MPI-IO handlers remain in the parent. The real spawned-LMDB regression includes a stream-holding logging integration and still checks rank-file output.

Local validation: 206 tests and 18 subtests passed, including both CI-failing validation cases. MPI master/workers file-policy probes and both precompile-log probes also passed with mocked distributed coordination. All normal pre-commit hooks passed. The complete GitHub CI matrix has not yet been rerun to completion.

njzjz (approved):

PR #6036, opened 2026-09-21
fix(pt): import e3nn lazily so it stays an optional dependency

iProzd:
> **Correction to the original description.** It said `e3nn` is declared only in the `dpa-adapt`
> extra. That was wrong: `deepmd` assembles extras through a custom build backend, and
> `backend/find_pytorch.py` also declares `e3nn>=0.5.9` under the `torch` extra. I had checked only
> `pyproject.toml`, where none of that appears. The description below is rewritten around what is
> actually broken, and a follow-up commit fixes the same wrong statement where it had been repeated
> in a docstring and a test docstring. The behaviour of the change is unaffected.

`deepmd/pt/model/descriptor/sezm_nn/projection.py` imports `e3nn` at module scope, and that module is
reached unconditionally from `deepmd/pt/model/descriptor/__init__.py`. `e3nn` ships with the `torch`
and `dpa-adapt` extras, so most installs have it — but not the PyTorch CPU install that
`doc/install/easy-install.md` documents, which deliberately uses no extra so that pip does not pull a
CUDA build of torch:

```bash
pip install torch --index-url https://download.pytorch.org/whl/cpu
pip install deepmd-kit
```

On that documented path:

```python
>>> import deepmd.pt.model.model
ModuleNotFoundError: No module named 'e3nn'
```

The whole PyTorch backend is unusable there, whether or not the user goes anywhere near a SeZM
descriptor, and the traceback points at a grid-projection helper rather than at a missing package.

## Change

Import `e3nn.o3` inside the two projection builders that need it, and raise an `ImportError` naming
both the package and an extra that ships it when it is absent.

Both builders need it, so switching grid backends is not a way to avoid the dependency and the message
does not suggest it:

| builder | e3nn symbols |
|---|---|
| `_build_e3nn_projection_mats` | `ToS2Grid`, `FromS2Grid` |
| `_build_lebedev_projection_mats` | `spherical_harmonics` |

Import timing and the error text are the only changes. No default, no numerical behaviour and no
public API is touched.

## Tests

`source/tests/pt/test_optional_e3nn_import.py` blocks `e3nn` with a meta-path hook in a subprocess
(already-imported modules cannot be un-imported in-process) and checks that
`deepmd.pt.model.model` and `deepmd.pt.entrypoints.main` both import, and that building a projector
without `e3nn` reports the package and the extra. Without the fix all three fail; with it all three
pass.

`source/tests/pt/model/test_descriptor_sezm_grid_projection.py` is unchanged by this PR and still
passes (18 tests, 179 subtests), so the e3nn path itself is unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * The PyTorch backend can now be imported without installing the optional `e3nn` package.
  * SeZM projector functionality provides clear installation guidance when `e3nn` is unavailable.
  * Missing dependencies required by an installed `e3nn` package now report their original error details instead of being misidentified as a missing `e3nn` installation.

* **Tests**
  * Added coverage for optional `e3nn` imports, installation guidance, and dependency error handling.


wanghan-iapcm (approved):
Reviewed at 73656da4. The premise holds: the documented PyTorch CPU install path (torch from the CPU index, then a bare `pip install deepmd-kit`) carries no extra and therefore no e3nn, and on the merge-base `import deepmd.pt.model.model` fails there. `projection.py` was the only module-scope e3nn import left under `deepmd/`. I ran the new subprocess tests with only `projection.py` reverted to the merge-base: 4 failed with the original `ModuleNotFoundError`; at head all 4 pass, and both branches of the helper (package absent, package present but a dependency missing) are exercised. CI: 12 Python shards and 4 C++ builds passed.

njzjz (approved):

